PR #13, opened 2015-08-26
fix an infinite loop caused by unclosed quotation tag

### LINKED ISSUE #12: Unclosed string in variable editor causes a freeze
Normally, typing a `"` causes its closing quote to appear. But if you delete one, the browser locks up. For example:

image

asiandrummer:
Self descriptive. It'll stay in the loop hoping to find `"`(34), only to fail.
Resolves #12.


asiandrummer:
screen shot 2015-08-26 at 12 19 12 pm

leebyron:
Yisssss

PR #14, opened 2015-08-26
Let people use the beta of React 0.14 as a peer dependency

steveluscher:


leebyron:
Does "0.13 - 0.14" work?

zpao:
You've got some crazy version strings going on, combining `~` with `.x`. If you use `^0.13 || ^0.14.0-beta1` that will take all of 0.13 and all of 0.14 >= 0.14.0-beta1 (you might actually want beta3, your call). That will also pick up stable versions of 0.14 when they are out.

```
> require('semver').satisfies('0.14.1', '^0.13 || ^0.14.0-beta1')
true
```

http://semver.npmjs.com/ is also a good tool for visualizing published versions.

zpao:
And for reference, it seems like prereleases don't get caught in a bunch of ranges, eg 0.13 - 0.14 won't allow betas

```
> require('semver').satisfies('0.14.0-beta3', '0.13 - 0.14')
false
```

steveluscher:
Thanks! Updated.

fson:
This would be nice to get in the next release, as it would also enable using GraphiQL with the relay-starter-kit more easily.

leebyron:
Whoa, I didn't know about that semver visualization, thanks @zpao it's super helpful to know about.

PR #15, opened 2015-08-26
Documentation Explorer for GraphiQL

asiandrummer:
Uses a provided GraphQLSchema instance to build up a documentation explorer.
- Types/fields/arguments defined are displayed with definitions
- Navigation stack to go back to previous definition looked up


asiandrummer:
Some screenshots:
screen shot 2015-08-26 at 2 35 53 pm
screen shot 2015-08-26 at 2 37 29 pm
screen shot 2015-08-26 at 2 36 04 pm

devknoll:
Took a quick glance, doesn't seem like descriptions for enum values or fields are rendered? (Or, less urgently, deprecation notices/reasons)

Also, GraphQL spec says:
> A GraphQL server may return the description field using Markdown syntax. Therefore it is recommended that any tool that displays description use a Markdown renderer

Any reason we can't run the descriptions through marked? :smile:

asiandrummer:
> Took a quick glance, doesn't seem like descriptions for enum values or fields are rendered? (Or, less urgently, deprecation notices/reasons)

Descriptions for fields are rendered, but not for enum values. Fixing.

> Any reason we can't run the descriptions through marked? :smile:

I've been completely forgetful about this. Attaching a markdown renderer... Thanks for pointing it out.

devknoll:
Hooray :smile:

KyleAMathews:
Very nice!

src/components/DocExplorer.js
```diff
@@ -7,38 +7,92 @@
...
+      }
+    />;
+    return this.marked(text);
+  }
+
+  _getTypeLink(type) {
+    return type.ofType ?
+      <span>
+        [
+          <a className="doc-type">{type.ofType.name}</a>
```
fson: This assumes every wrapper type is a list. There's also the NonNull wrapper and the combinations of these two.
E.g. if your schema has

```js
fields: {
  foo: { type: new GraphQLNonNull(GraphQLString) },
  bar: { type: new GraphQLList(GraphQLString) },
  baz: { type: new GraphQLNonNull(new GraphQLList(GraphQLString)) },
}
```
this would render:
```
foo : [String]
bar : [String]
baz : [String]
```
The expected result would be:
```
foo : String!
bar : [String]
baz : [String]!
```

css/doc-explorer.css
```diff
@@ -0,0 +1,103 @@
...
+}
+
+#graphiql-container .doc-explorer-contents {
+  position: relative;
+  height: 100%;
+  background-color: #ffffff;
+  border-left: 1px solid #d6d6d6;
+  border-top: 1px solid #d6d6d6;
+  padding: 14px 15px;
+  overflow: hidden;
```
fson: It seems that `overflow: auto` would be a better choice here. If you have a big schema it overflows the screen and can't be scrolled.

fson:
I've played with this using our test schema and it looks very promising. I'm sure this will be a loved feature of GraphiQL!

src/components/DocExplorer.js
```diff
@@ -56,48 +110,48 @@ export class DocExplorer extends React.Component {
     );
   }
 
-  // TODO: This part could be optionalized
-  _generateStartPage(schema) {
-    var queryType = schema.getQueryType();
-    var mutationType = schema.getMutationType();
-    var directives = schema.getDirectives();
+  _renderTypeValues(type) {
+    var _renderMarkdown = this._renderMarkdown;
```
fson: When `_renderMarkdown` is called below, it throws `TypeError: Cannot read property 'marked' of undefined`. Maybe this line should be `this._renderMarkdown.bind(this)`?

src/components/DocExplorer.js
```diff
@@ -7,38 +7,92 @@
...
+    this.marked = Marked;
+  }
+
+  _renderMarkdown(text) {
+    return <span
+      dangerouslySetInnerHTML={
+        {__html: this.marked(text)}
+      }
+    />;
+    return this.marked(text);
```
fson: This line is unreachable.

src/components/DocExplorer.js
asiandrummer: Ah, another ignorance on my part with NonNulls. Fixing.

asiandrummer:
screen shot 2015-09-04 at 2 49 45 pm

leebyron:
graphiql-docs

KyleAMathews:
Cool! Looking forward to using this!

PR #19, opened 2015-08-28
Use port 8081 so example works with relay-starter-kit out of the box

LegNeato:
https://github.com/relayjs/relay-starter-kit binds the graphql server to port `8080` and serves it at `/graphql`. This means we can't use the example out of the box.

Changing the port to 8081 means that relay-starter-kit works with the precompiled GraphiQL example with zero changes.


LegNeato:
Hmmm, the graphiql example appears to be using a hard-coded test schema. I wonder if we can have them play nicer. Will think about this some more.

PR #20, opened 2015-08-29
Use port 8081 if 8080 is taken so example works with relay-starter-kit

LegNeato:
This makes it so we try to bind to `8080`. If it fails, we assume that they are using [relay-starter-kit](https://github.com/relayjs/relay-starter-kit) and instead bind to `8081`. A note is output in case something else is using `8080`.

This lets us use the Test/canned data when `8080` is free (current behavior) and defer to relay-starter-kit for its data when it has already started up on `8080`.

Test Plan
-------------

1. Start relay-starter-kit, see it works.
2. Stop relay-starter-kit.
3. Start GraphiQL example, see it uses Test data.
4. Stop  GraphiQL example.
5. Start relay-starter-kit, see it works.
6. Start GraphiQL example, see it uses relay-starter-kit data.

Goes hand-in-hand with https://github.com/relayjs/relay-starter-kit/pull/20, which is needed to do a query from `localhost:8081` (GraphiQL example)  to `localhost:8080` (relay-starter-kit data)

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,8 +1,21 @@
 Example GraphiQL Install
 ========================
 
-1. Run `npm install` in root directory
+Without [relay-starter-kit](https://github.com/relayjs/relay-starter-kit)
+-------------------------------------------------------------------------
+1. Run `npm install` in the GraphiQL repository root directory
 2. Navigate to this directory (example) in Terminal
 3. `npm install`
 4. `npm start`
 5. Open your browser to [http://localhost:8080/]()
+
+
+With [relay-starter-kit](https://github.com/relayjs/relay-starter-kit)
+----------------------------------------------------------------------
+1. Run `npm start` in [relay-starter-kit](https://github.com/relayjs/relay-starter-kit)
+2. Open another Terminal window.
+3. Run `npm install` in the GraphiQL repository root directory
+4. Navigate to this directory (example) in Terminal
+5. `npm install`
+6. `npm start`
+6. Open your browser to [http://localhost:8081/]()
--- a/example/index.html
+++ b/example/index.html
@@ -71,7 +71,7 @@
 
       // Defines a GraphQL fetcher using the fetch API.
       function graphQLFetcher(graphQLParams) {
-        return fetch(window.location.origin + '/graphql', {
+        return fetch('http://localhost:8080/graphql', {
           method: 'post',
           headers: { 'Content-Type': 'application/json' },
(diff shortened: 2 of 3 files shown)


fson:
This assumes that port `8080` is taken by relay-starter-kit. For example, I had to run relay-starter-kit on another port because I have a database running in the background that takes `8080`.

Maybe it would be simpler to add GraphiQL to the relay-starter-kit instead and not the other way around? That would mean the starter kit would have a HTML page with the correct configuration to load GraphiQL and you wouldn't need to install/run anything else, just `npm start` the starter kit.

LegNeato:
I think this is not perfect but materially better than what we have now.

* If `8080` is taken by your database, the GraphiQL example currently doesn't work anyway.
* If `8080` is taken by your database, relay-starter-kit doesn't work.
* After this diff there is a message saying what we assume relay-starter-kit is on `8080` if in use and telling the user to free up `8080` if that is not the case.
 * More advanced users who need `8080` for something else can then of course edit the code or stop using the example.
* There are docs explaining steps both with and without relay-starter-kit in the README. It would be hard to miss because you generally need to see the README to start the GraphiQL example.

This PR isn't optimizing for people who know what they are doing and starting a new project...it is optimizing to get new people up and running and poking at the ecosystem with pre-cooked bits and having those bits play well together.

I thought it was super weird that I coudn't easily use GraphiQL to play around with react-starter-kit's responses and data models to see how everything fit together.

I guess we could have one project expose an endpoint and have the other query, but that seems a bit complicated for simple demoware.

fson:
@LegNeato I agree these projects should play better together, and optimizing the set up is a great idea. Maybe I was being unclear, but wouldn't the experience with relay-starter-kit be even smoother, if GraphiQL was included in the relay-starter-kit? I think it would be quite easy to add there as a devDependency an serve using the development server included.

fson:
@LegNeato I took a stab adding GraphiQL in the starter kit here: https://github.com/relayjs/relay-starter-kit/pull/22

LegNeato:
Ah, sweet...I don't think I understood what you were getting at. :D. That does look better to me!

leebyron:
Yeah, I think @fson's approach is correct. Relay starter kit should just include a graphiql install, rather than this graphiql example which is designed to be as simple as possible having a condition to work with relay starter kit.

I'm happy to adjust the port to something more likely to be free though.

LegNeato:
Going to close this out in favor of relayjs/relay-starter-kit#22

PR #21, opened 2015-08-29
Update jsdom to allow development using iojs

fson:
Use a newer version of jsdom, so GraphiQL can be developed with iojs like Relay, Jest etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "eslint-plugin-react": "3.2.3",
     "flow-bin": "0.14.0",
     "graphql": "0.4.2",
-    "jsdom": "3.0.0",
+    "jsdom": "6.2.0",
     "mocha": "2.2.5",
     "react": "0.13.3",
     "uglify-js": "^2.4.24",


leebyron:
cc @asiandrummer who I believe is using this version on purpose to facilitate testing.

asiandrummer:
This is deliberately set because jsdom with versions higher than 4.0 will require iojs instead of nodejs.

fson:
@asiandrummer Since Relay has also moved to the latest version of Jest, which requires iojs, would it make sense to standardise the development enviroment on iojs? Currently I need to switch between the old version of node for this project and iojs for Relay. It seems like this would be an easy upgrade path to node v4 once the node&iojs merge has finalized, because iojs v3 quite close to it already.

asiandrummer:
@fson I can't think of any reasons not to. @leebyron may have other opinions though - could you comment?

leebyron:
We don't want to require iojs/node4 just yet since node v0.12 is still widely used and we don't want to burden potential users/contributors

leebyron:
Is there a way to intelligently use different versions of jsdom depending on the version of node being used? Surely this isn't the first time jsdom has faced this issue.

asiandrummer:
jsdom's suggestion is to use 3.x until the merger happens; https://github.com/tmpvar/jsdom/blob/master/Changelog.md#400 for reference.

Could we host json-lint in another place? That'll allow us to remove jsdom dependency (as well as some others such as mocha + chai) since jsdom is only used in mocha-bootload.js file (and in running mocha tests).

leebyron:
I think we will want to increase test coverage over time, and testing stuff that uses code mirror probably will continue to require jsdom

fson:
Thanks for explaining the reasoning behind using node v0.12 @leebyron. Since it seems that it's not possible to support both v0.12 and v4.0 in the same code base when depending on `jsdom` and you don't want to make the switch yet, I'll close this issue for now and keep using v0.12 for GraphiQL.

leebyron:
FYI - Just bumped up jsdom's dependency as we're trying to move all FB's development to use node v4.

fson:
:+1:

PR #26, opened 2015-08-31
Fix license field in package.json

luisobo:
As per npm docs https://docs.npmjs.com/files/package.json#license


leebyron:
Thanks!

PR #27, opened 2015-09-05
Add PropTypes to components.

skevy:
Thought this might be helpful during development :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
   ],
   "babel": {
     "optional": [
-      "runtime"
+      "runtime",
+      "es7.classProperties"
     ],
     "loose": [
       "es6.classes",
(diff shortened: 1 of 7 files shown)


resources/build.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 rm -rf dist/ && mkdir -p dist/ &&
-babel src --ignore __tests__ --out-dir dist/ &&
+babel src --ignore __tests__ --out-dir dist/ --stage=0 &&
```
asiandrummer: Why was --stage=0 needed?

skevy: Class properties.

leebyron: Can we adjust it to only add that one feature so it's more explicit? stage=0 adds a lot of random stuff.

skevy: Sure! Not sure how I missed the babel settings in package.json. I will add `es7.classProperties` explictly.

skevy:
@leebyron i think that'll do it.

leebyron:
Dope

PR #28, opened 2015-09-05
Add the ability to set a prefix for data stored in localStorage

skevy:
I needed this on my electron wrapper around GraphiQL...but I think it could be useful to others if they're embedding GraphiQL in an application that uses localStorage already.


asiandrummer:
Good one - thanks!

leebyron:
Would it be a better API to be able to provide a `storage` instance as a prop? That would allow switching up local/session storage, mocking storage for tests, or doing any number of other things including this.

leebyron:
Prefixing the names used is a great idea none the less

PR #30, opened 2015-09-09
Fix GraphiQL component trying to use undefined properties

fson:
- `TypeError: Cannot read property 'storage' of undefined`
- `TypeError: this.storageGet is not a function`


asiandrummer:
whoaa silly ones

PR #35, opened 2015-09-24
make doc explorer scrolls down

vedmalex:
Make it overflowable, in the cases when there is too many fields they hidden.


leebyron:
Thanks for the report. Unfortunately this patch causes the doc explorer header to scroll away, so a different approach will be necessary.

vedmalex:
Thank you!

PR #38, opened 2015-10-01
Switch React.findDOMNode to ReactDOM.findDOMNode

asiandrummer:
Integrating changes from facebook/react#4702.


asiandrummer:
Oops, this change will be in React v0.14. I'll reopen this when that happens.

PR #44, opened 2015-10-21
Add `subscription` operation support

skevy:
Adds `subscription` operation support that landed in graphql-js 0.4.8.

Do not merge until https://github.com/graphql/codemirror-graphql/pull/2 is merged and dependency is updated here.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -10,7 +10,7 @@
     "express": "^4.12.4",
     "express-graphql": "^0.3.0",
     "graphiql": "../",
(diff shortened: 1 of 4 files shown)


skevy:
@leebyron any thoughts on getting this merged? I know people in the community are playing with subscriptions, and it could be helpful to be able to test with GraphiQL (since no Relay support exists yet).

leebyron:
Thanks for the reminder, this is good stuff.

PR #47, opened 2015-11-03
Update README example to use ReactDOM instead of React for render

BobertForever:
`React.render` is deprecated in favor of `ReactDOM.render`, this reflects that in the tiny example in the README


BobertForever:
Whoops, #41 already does this.

Closing

PR #51, opened 2015-11-05
Allow React 0.14.x in peer dependencies

gre:
If you try to use latest React 0.14.2 with graphiql, it won't work. This PR fix it


leebyron:
Thanks!

PR #52, opened 2015-11-06
Update package.json to rely on React ^0.14.0

saikat:
React has been upgraded to 0.14.2, so graphiql is now breaking my NPM install because it is relying on 0.14.0 exactly.  Is there a reason for it to rely exactly on 0.14.0?

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,8 @@
   },
   "peerDependencies": {
     "graphql": "^0.4.7",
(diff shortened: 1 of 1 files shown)


leebyron:
Just fixed by #51 

This was just a mistake. We usually rely on specific versions only, not ranges, to remove breakages from version changing, but of course that should not apply for peerDependencies.

PR #56, opened 2015-11-10
fixed to execute a version of query with auto-completed leafs

asiandrummer:
Simply return the result query from filling leaves of the query, and either use that or the query from this.state to fetch results.


leebyron:
Land ho!

PR #61, opened 2015-11-25
Loading spinner when running editor queries and fetching the schema

asiandrummer:
A CSS-driven loading spinner when running editor queries and fetching the schema, in case they take a while. Shown in a result window and documentation explorer (since it needs a schema to be rendered).

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -106,6 +106,7 @@ html, body {
 }
 
 #graphiql-container .resultWrap {
+  position: relative;
   display: -webkit-flex;
   display:         flex;
   -webkit-flex-direction: column;
--- a/css/loading.css
+++ b/css/loading.css
@@ -1,0 +1,46 @@
+#graphiql-container .spinner-container {
+  position: absolute;
+  top: 50%;
+  height: 36px;
+  width: 36px;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 10;
+}
+
+#graphiql-container .spinner {
+  vertical-align: middle;
+  display: inline-block;
+  height: 24px;
+  width: 24px;
+  position: absolute;
+  -webkit-animation: rotation .6s infinite linear;
+  -moz-animation: rotation .6s infinite linear;
+  -o-animation: rotation .6s infinite linear;
+  animation: rotation .6s infinite linear;
+  border-left: 6px solid rgba(150, 150, 150, .15);
+  border-right: 6px solid rgba(150, 150, 150, .15);
+  border-bottom: 6px solid rgba(150, 150, 150, .15);
+  border-top: 6px solid rgba(150, 150, 150, .8);
+  border-radius: 100%;
+}
+
+@-webkit-keyframes rotation {
+  from { -webkit-transform: rotate(0deg); }
+  to { -webkit-transform: rotate(359deg); }
+}
+
+@-moz-keyframes rotation {
+  from { -moz-transform: rotate(0deg); }
+  to { -moz-transform: rotate(359deg); }
+}
+
+@-o-keyframes rotation {
+  from { -o-transform: rotate(0deg); }
+  to { -o-transform: rotate(359deg); }
+}
+
+@keyframes rotation {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(359deg); }
+}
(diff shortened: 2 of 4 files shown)


src/components/GraphiQL.js
```diff
@@ -199,6 +200,10 @@ export class GraphiQL extends React.Component {
 
   componentDidMount() {
     if (!this.state.schema) {
+      this.setState({
+        showSpinner: true
```
leebyron: We should only show a spinner when there is an user-produced query to the server underway, not for the automated fetching of the schema. GraphiQL is still usable without a schema, just typeahead and linters are not yet enabled, that's fine.

This implementation also has a race condition where we could issue a query before the schema returns. The schema would return and turn off the spinner, but the query itself hasn't returned yet.

src/components/GraphiQL.js
```diff
@@ -171,6 +171,7 @@ export class GraphiQL extends React.Component {
       variableEditorHeight: this._storageGet('variableEditorHeight') || 200,
       docsOpen: false,
       docsWidth: this._storageGet('docExplorerWidth') || 350,
+      showSpinner: false,
```
leebyron: I'd recommend calling this `isWaitingForResponse` since thats the semantic value, and the spinner is just the visualization of this value.

src/components/GraphiQL.js
```diff
@@ -297,6 +308,13 @@ export class GraphiQL extends React.Component {
               </div>
             </div>
             <div className="resultWrap">
+              {(() => {
+                return this.state.showSpinner && (
+                  <div className="spinner-container">
+                    <div className="spinner" />
+                  </div>
+                );
+              })()}
```
leebyron: why the IIFE? couldn't you also write:

```
{this.state.showSpinner &&
  <div className="spinner-container">
    <div className="spinner" />
  </div>}
```

src/components/GraphiQL.js
```diff
@@ -297,6 +308,13 @@ export class GraphiQL extends React.Component {
...
               <ResultViewer ref="result" value={this.state.response} />
```
leebyron: Perhaps the ResultViewer should be grayed out or something while another query is underway and the spinner is placed over it.

src/components/ResultViewer.js
```diff
@@ -67,6 +67,8 @@ export class ResultViewer extends React.Component {
   }
 
   render() {
-    return <div className="result-window" />;
+    return (
```
leebyron: unnecessary change

leebyron:
You should also squash your commits together. You can do that with `rebase -i`, and then `push -f` to update the PR

asiandrummer:
screen shot 2015-11-25 at 3 01 34 pm

css/loading.css
```diff
@@ -0,0 +1,62 @@
+#graphiql-container .spinner-container {
+  position: absolute;
+  top: 20%;
+  height: 36px;
+  width: 36px;
+  left: 50%;
+  transform: translate(-50%,-50%);
```
leebyron: nit: space after the comma

css/loading.css
```diff
@@ -0,0 +1,62 @@
...
+  border-left: 6px solid rgba(150, 150, 150, .15);
+  border-right: 6px solid rgba(150, 150, 150, .15);
+  border-bottom: 6px solid rgba(150, 150, 150, .15);
+  border-top: 6px solid rgba(150, 150, 150, .8);
+  border-radius: 100%;
+}
+
+@-webkit-keyframes rotation {
+  from {
+  -webkit-transform: rotate(0deg);
```
leebyron: nit: indent inside the {} or one-line it. I think it may be nicer to read:

```css
from { -crap-transform: rotate(0deg); }
to { -crap-transform: rotate(359deg); }
```

src/components/GraphiQL.js
```diff
@@ -329,11 +338,18 @@ export class GraphiQL extends React.Component {
...
+      this.setState({
+        isWaitingForResponse: false,
+        response: error && error.stack || error
+      });
     });
   }
 
   _runEditorQuery = () => {
+    this.setState({
+      isWaitingForResponse: true,
```
leebyron: this should also set `response: null`

css/app.css
```diff
@@ -115,6 +116,10 @@ html, body {
   border-left: solid 1px #e0e0e0;
 }
 
+#graphiql-container .resultWrap.isWaitingForResponse .CodeMirror {
+  background: rgb(200, 200, 200);
```
leebyron: I don't think you need this, the spinner looks nicer on a white background IMHO

css/loading.css
```diff
@@ -0,0 +1,62 @@
+#graphiql-container .spinner-container {
+  position: absolute;
+  top: 20%;
```
leebyron: 50% would be nicer to vertically center

PR #64, opened 2015-11-28
Fixes #48: add a prettify query button

### LINKED ISSUE #48: "prettify" query
my common usecase is to paste a query that was generated by Relay and see what it does.

However, the Relay queries are pretty obfuscated, it would be great to have a button that prettify the query textarea.

I was playing around with this code:
```js
function blockIndent(editor, from, to) {
  editor.operation(function() {
    for (var i = from; i < to; ++i)
      editor.indentLine(i, "smart");
  });
}
```

but it just indents, it does not do the "return to new line" work.

gre:
addresses #48

it adds a **`{}`** prettify button. this is implemented using `graphql` parse & print function: `print(parse(this.state.query))`

(pretty useful when you paste an obfuscated relay query and want to debug it)

screen shot 2015-11-28 at 11 14 35

on hover:

screen shot 2015-11-28 at 11 14 42

after clicking:

screen shot 2015-11-28 at 11 14 57

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -215,6 +215,21 @@ html, body {
   padding-left: 3px;
 }
 
+#graphiql-container .tool-button {
+  cursor: pointer;
+  text-decoration: none;
+  font-family: sans-serif;
+  color: #333;
+  padding: 2px 4px;
+  margin: 0 4px;
+}
+#graphiql-container .tool-button:active {
+  color: #999;
+}
+#graphiql-container .tool-button.error {
+  color: #c00;
+}
+
 #graphiql-container .execute-button {
   background: -webkit-linear-gradient(#fdfdfd, #d2d3d6);
   background:         linear-gradient(#fdfdfd, #d2d3d6);
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -8,10 +8,13 @@
 
 import React, { PropTypes } from 'react';
 import ReactDOM from 'react-dom';
+import { print } from 'graphql/language/printer';
+import { parse } from 'graphql/language/parser';
 import { GraphQLSchema } from 'graphql/type';
 import { buildClientSchema } from 'graphql/utilities';
 import find from 'graphql/jsutils/find';
 import { ExecuteButton } from './ExecuteButton';
+import { PrettifyButton } from './PrettifyButton';
 import { QueryEditor } from './QueryEditor';
 import { VariableEditor } from './VariableEditor';
 import { ResultViewer } from './ResultViewer';
@@ -261,6 +264,7 @@ export class GraphiQL extends React.Component {
             <div className="topBar">
               {logo}
               <ExecuteButton onClick={this._runEditorQuery} />
+              <PrettifyButton onClick={this._prettifyQuery} />
               {toolbar}
             </div>
             {!this.state.docsOpen &&
@@ -347,6 +351,12 @@ export class GraphiQL extends React.Component {
         this.setState({ response: JSON.stringify(result, null, 2) });
       }
     });
+  }
+
+  _prettifyQuery = () => {
+    const query = print(parse(this.state.query));
+    const editor = this.refs.queryEditor.getCodeMirror();
+    editor.setValue(query);
   }
 
   _onEditQuery = value => {
--- a/src/components/PrettifyButton.js
+++ b/src/components/PrettifyButton.js
@@ -1,0 +1,50 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE-examples file in the root directory of this source tree.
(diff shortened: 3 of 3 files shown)


skevy:
Awesome!!

leebyron:
Interesting! I think we may want to improve the button affordance, it's not clear that it is clickable. One significant downside of this prettify technique is that it eats comments. Obviously not a big deal for cleaning up obfuscated queries that have no comments, but something to consider.

gre:
I'm not sure how we can preserve comments. Sounds like something to fix on the graphql parser/printer side.

I have intentionally made the `{}` button looking secondary (less important than the  ▶one), but actually I would be happy to implement any better UI if you have an idea on how to make it look more clickable (it already reacts on hover as you can see).
I feel that if using a circle or a square border will look weird next to the circle ▶button.


I've also inspired from the Web Console:
screen shot 2015-11-30 at 20 57 10

vladar:
That would be so useful with Relay

src/components/PrettifyButton.js
```diff
@@ -0,0 +1,38 @@
...
+export class PrettifyButton extends React.Component {
+  static propTypes = {
+    onClick: PropTypes.func
+  }
+
+  onClick = e => {
+    const { onClick } = this.props;
+    if (onClick) {
+      e.preventDefault();
+      onClick();
```
asiandrummer: nit: maybe just `this.props.onClick()`

src/components/GraphiQL.js
```diff
@@ -349,6 +353,12 @@ export class GraphiQL extends React.Component {
     });
   }
 
+  _prettifyQuery = () => {
+    const query = print(parse(this.state.query));
```
asiandrummer: This could throw if the query isn't complete/has validation errors I think.

css/app.css
```diff
@@ -215,6 +215,18 @@ html, body {
...
+#graphiql-container .tool-button {
+  cursor: pointer;
+  text-decoration: none;
+  font-family: sans-serif;
+  color: #ccc;
+  padding: 2px 4px;
+  margin: 0 4px;
+}
+#graphiql-container .tool-button:hover {
+  color: #333;
```
asiandrummer: Following up on GraphiQL.js@357, maybe the prettify button could change color if it's possible without throwing errors, not on hover?
Also, nit: `.prettify-button`

src/components/PrettifyButton.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE-examples file in the root directory of this source tree.
+ */
+
+import React, { PropTypes } from 'react';
+
+
+/**
+ * PrettifyButton
```
asiandrummer: maybe nit, but I'd love a bit more documentation on this component.

asiandrummer:
Sorry it took a while to get to this @gre. I think this is a nice feature addition - although we have a known issue with comments being deleted, I think this will be very useful for folks working with reduced/minimized query string, especially in Relay land.

css/app.css
gre: I think :hover need to stay as it's a feedback that the button is clickable. that said, I can make the button changing color when there is an error?

gre: I intentionally make the button lighter when it's not hover so it doesn't pollute too much the UI . I'm happy to implement any other design

skevy:
I'm so excited to have this in GraphiQL, @gre! :)

gre:
me too :)

css/app.css
asiandrummer: Maybe more of a personal preference, but a lighter button feels like the query's not possible to be prettified to me (like a disabled button). Instead of :hover how about :active selector for the clickable feedback, just like .execute-button?

asiandrummer:
@gre Could we squash the commits into one commit?

css/app.css
gre: let me see if I can do something about this

asiandrummer: :+1:

gre:
@asiandrummer are you sure I should squash my commit and lose the history? is this really the convention here? (merge should hide the local branch history no?)

src/components/PrettifyButton.js
gre: done

src/components/GraphiQL.js
gre: this is addressed in the prettifybutton code now :)

asiandrummer:
@gre I was thinking we could only have a single or a couple commit related to the feature, for the readability of the commit history reasons. I was especially thinking we could include (or amend the message of) the 2nd commit if that's okay.

Also I think merge actually shows the local branch history (https://github.com/graphql/graphiql/pull/15), but I might be missing something :p

gre:
this is squashed now

asiandrummer:
image
Citizen Hyo thanks you! I'm really excited for this feature.

PR #66, opened 2015-11-29
Migrate to babel6

craigbilner:
Mirrored same loose options and only added minimum plugins needed to run build

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,55 @@
+{
+  "presets": [
+    "react"
+  ],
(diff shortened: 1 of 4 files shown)


leebyron:
Thanks for queuing this up! We're waiting on some fixes from eslint and istanbul to support babel6 and then will upgrade the whole graphql fleet.

leebyron:
Merged in 042fa126a65edf74ea6e0c1bbf3eefa2ea8e0b43

PR #70, opened 2015-12-10
Typeahead in VariableEditor

asiandrummer:
Powered by CodeMirror. Simply search for all Variable (http://facebook.github.io/graphql/#Variable), de-duplicate/sort, and serve them up to the typeahead. Typeahead is trigger on typing letters/backspace/quotes (")/and underscore character.

Still have some UI improvements/testing to finish - just opening a PR for early reviews.

### DIFF
--- a/src/codemirror/hint/json-hint.js
+++ b/src/codemirror/hint/json-hint.js
@@ -1,0 +1,128 @@
+/**
+ *  Copyright (c) 2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE-examples file in the root directory of this source tree.
+ */
+
+import CodeMirror from 'codemirror';
+
+CodeMirror.registerHelper('hint', 'json', (editor, options) => {
+  var cur = editor.getCursor();
+  var token = editor.getTokenAt(cur);
+  var tokenStart = token.type === null ? token.end :
+    /\w/.test(token.string[0]) ? token.start :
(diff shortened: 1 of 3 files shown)


src/codemirror/hint/json-hint.js
```diff
@@ -0,0 +1,128 @@
...
+CodeMirror.registerHelper('hint', 'json', (editor, options) => {
+  var cur = editor.getCursor();
+  var token = editor.getTokenAt(cur);
+  var tokenStart = token.type === null ? token.end :
+    /\w/.test(token.string[0]) ? token.start :
+    token.start + 1;
+
+  // Match all variableDefinitions, even if they're not declared on the top.
+  // De-duplicate before filtering/sorting the list
+  let hintList = options.query.match(/\$\s*[_A-Za-z][_0-9A-Za-z]*/g);
```
leebyron: I think this is probably not what we want - this will cause false-positive matches for something like "I'm a string with a $thing in it" and it also lacks the information we want to do a good type-ahead: the relevant input schema. Just having a list of variables only lets us typeahead names, but doesn't let us typeahead or lint values like allowed enum values or input objects with deeper fields.

leebyron:
I think this ultimately belongs in codemirror-graphql. That would allow us to share the lexicalDistance function, for example. The full typeahead we want will also leverage a lot of what we've already built for hinting and linting graphql.

asiandrummer:
I agree - closing and moving this over to codemirror-graphql.

PR #73, opened 2015-12-17
Fixes #69: Show correct description text for each enum value.

### LINKED ISSUE #69: Enum documentation shows wrong descriptions
the overall enum description shows up in the documentation viewer for the enum values even if the enum values have a valid description.

ryanspradlin:
Fixes: #69 

The `DocExplorer` component was incorrectly showing the type description text instead of the individual enum value description text for each enum value. (`type.description` was being passed instead of the correct `value.description`).


asiandrummer:
whoa good catch - thanks!

PR #80, opened 2016-01-16
Search in Doc Explorer

asiandrummer:
Search functionality in Documentation Explorer.
It's pretty basic with searching/getting results/clicking links - I'll incrementally add in more details in upcoming PRs.

EDIT: added no-search-result case and minimum character length

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -56,7 +56,10 @@ export class DocExplorer extends React.Component {
   constructor() {
     super();
 
-    this.state = { navStack: [] };
+    this.state = {
+      navStack: [],
+      searchValue: null
+    };
   }
 
   _onToggleBtnClick = () => {
@@ -64,17 +67,25 @@ export class DocExplorer extends React.Component {
   }
 
   _onNavBackClick = () => {
-    this.setState({ navStack: this.state.navStack.slice(0, -1) });
+    this.setState({
+      navStack: this.state.navStack.slice(0, -1),
+      searchValue: null
+    });
   }
 
   _onClickTypeOrField = typeOrField => {
     this.showDoc(typeOrField);
   }
 
+  _onSearch = event => {
+    this.setState({ searchValue: event.target.value });
+  }
+
   shouldComponentUpdate(nextProps, nextState) {
     return (
       this.props.schema !== nextProps.schema ||
-      this.state.navStack !== nextState.navStack
+      this.state.navStack !== nextState.navStack ||
+      this.state.searchValue !== nextState.searchValue
     );
   }
 
@@ -103,12 +114,22 @@ export class DocExplorer extends React.Component {
           field={typeOrField}
           onClickType={this._onClickTypeOrField}
         />;
+    } else if (this.state.searchValue) {
+      title = 'Search Results';
+      content =
+        <SearchDoc
+          searchValue={this.state.searchValue}
+          schema={schema}
+          onClickType={this._onClickTypeOrField}
+          onClickField={this._onClickTypeOrField}
+        />;
     } else if (schema) {
       title = 'Documentation Explorer';
       content =
         <SchemaDoc
           schema={schema}
           onClickType={this._onClickTypeOrField}
+          onSearch={this._onSearch}
         />;
     }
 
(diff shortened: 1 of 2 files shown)


src/components/DocExplorer.js
```diff
@@ -45,8 +45,8 @@ export class DocExplorer extends React.Component {
   // Public API
   showDoc(typeOrField) {
     var navStack = this.state.navStack;
-    var isCurrentlyShown =
-      navStack.length > 0 && navStack[navStack.length - 1] === typeOrField;
+    var isCurrentlyShown = navStack.length > 0 &&
```
asiandrummer: This is an unnecessary change - reverting.

asiandrummer:
screen shot 2016-01-15 at 4 17 58 pm
screen shot 2016-01-15 at 4 17 50 pm

asiandrummer:
screen shot 2016-01-15 at 4 43 39 pm
screen shot 2016-01-15 at 4 39 30 pm

src/components/DocExplorer.js
```diff
@@ -56,7 +56,10 @@ export class DocExplorer extends React.Component {
   constructor() {
     super();
 
-    this.state = { navStack: [] };
+    this.state = {
+      navStack: [],
+      searchValue: null
```
leebyron: I think searchValue should probably be part of the nav stack rather than it's own thing, so that you can search, click on something, click back and be where you were before.

This could be done in a separate PR though.

src/components/DocExplorer.js
```diff
@@ -103,12 +114,25 @@ export class DocExplorer extends React.Component {
...
+      title = 'Search Results';
+      content = this.state.searchValue.length > 2 ?
+        <SearchDoc
+          searchValue={this.state.searchValue}
+          schema={schema}
+          onClickType={this._onClickTypeOrField}
+          onClickField={this._onClickTypeOrField}
+        /> :
+        <span className="doc-alert-text">
+          Minimum search length is 3 characters.
```
leebyron: I don't think we need to explicitly show this on the screen. Also, in a separate PR later we may want to remove this restriction so you could search for things like "id" but perhaps just change the search algorithm so it's not overwhelmed by < 3 ch searches.

src/components/DocExplorer.js
```diff
@@ -141,6 +165,15 @@ export class DocExplorer extends React.Component {
           </div>
         </div>
         <div className="doc-explorer-contents">
+          {!typeOrField &&
+            <label className="search-box-outer">
+              <input className="search-box-input"
+                onKeyUp={event => this._onSearch(event)}
+                type="text"
+                name="SearchBox"
```
leebyron: I don't think you need `name=` here

src/components/DocExplorer.js
```diff
@@ -148,6 +181,142 @@ export class DocExplorer extends React.Component {
...
+  render() {
+    const searchValue = this.props.searchValue;
+    const schema = this.props.schema;
+    const onClickType = this.props.onClickType;
+    const onClickField = this.props.onClickField;
+
+    const typeMap = schema.getTypeMap();
+
+    // Compose a search index for all possible arguments from every types.
+    let possibleArguments = [];
```
leebyron: This isn't used anywhere below, maybe we should remove it for this PR.

Also, we don't want to compute this index on every render since it's likely to be expensive for larger schema. We only want to compute an index once per schema.

src/components/DocExplorer.js
```diff
@@ -148,6 +181,142 @@ export class DocExplorer extends React.Component {
...
+    let matchedArgs = [];
+
+    Object.keys(typeMap).forEach(typeName => {
+      const type = typeMap[typeName];
+      let matchedOn = [];
+      if (typeName.indexOf(searchValue) !== -1) {
+        matchedOn.push('Type Name');
+      }
+
+      if (type.description && type.description.indexOf(searchValue) !== -1) {
```
leebyron: I would remove this for this initial PR - I'm concerned that searching over descriptions will be expensive for larger schema, and perhaps we'll want to do a specialized index for this. Let's come back to it in a separate PR once the main search functionality is working.

src/components/DocExplorer.js
```diff
@@ -148,6 +181,142 @@ export class DocExplorer extends React.Component {
...
+                  onClick={event => onClickField(field, type, event)}>
+                  {field.name}
+                </a>
+                {' on '}
+                <TypeLink type={type} onClick={onClickType} />
+              </div>
+            );
+          }
+
+          if (field.args && field.args.length) {
```
leebyron: probably want `else if` here so the same field doesnt appear twice if an argument also matches

src/components/DocExplorer.js
```diff
@@ -148,6 +181,142 @@ export class DocExplorer extends React.Component {
...
+                {' on '}
+                <TypeLink type={type} onClick={onClickType} />
+              </div>
+            ));
+          }
+        });
+      }
+    });
+
+    if (matchedTypes.length === 0 &&
```
leebyron: style nit:

```js
if (matchedTypes.length === 0 &&
    matchedFields.length === 0 &&
    matchedArgs.length === 0) {
  return ...
}

File: src/components/DocExplorer.js [hunk 9]
```diff
@@ -148,6 +181,142 @@ export class DocExplorer extends React.Component {
...
+                <TypeLink type={type} onClick={onClickType} />
+              </div>
+            );
+          }
+
+          if (field.args && field.args.length) {
+            let matches = field.args.filter(arg => {
+              return arg.name.indexOf(searchValue) !== -1;
+            });
+            Array.prototype.push.apply(matchedArgs, matches.map(arg =>
```
leebyron: Since you're rendering fields here, you might want to add them to `matchedFields`. Also, rather than rendering the same field multiple times for each matched arg, you should just render all the matched args within this one field, moving the `matches.map` inside the `( )` selection below.

leebyron:
This is awesome! There are some comments inline, but this is pretty close to landing as an initial version of search.

src/components/DocExplorer.js
asiandrummer: Exactly; I would like the search action to be a part of navStack - something like:
navStack = [
  testType: { ... },
  graphiql_search: { value: ... },
];

asiandrummer:
Responded to comments:
- removed ch restriction for now
- got rid of the argument indexing, turns out we didn't need it after all
- removed specific match types and included wholesome within `search results`
- a field with multiple argument matches

asiandrummer:
screen shot 2016-01-21 at 2 56 40 pm

src/components/DocExplorer.js
leebyron: A related note about navStack for the future: it would be cool if scroll position were also maintained as part of the back stack. Today if you click from a long scrolling view to another long scrolling view your scroll position is maintained when it should probably be reset to "0", but then when going "back" the original scroll position should be restored. It's a fairly small detail, but it will help make navigating larger schema easier on the mind.

src/components/DocExplorer.js
```diff
@@ -103,12 +114,21 @@ export class DocExplorer extends React.Component {
           field={typeOrField}
           onClickType={this._onClickTypeOrField}
         />;
+    } else if (this.state.searchValue) {
+      title = 'Search Results';
+      content = <SearchDoc
```
leebyron: nit: use the same styling here as in the else block below, line break and indent.

src/components/DocExplorer.js
```diff
@@ -148,6 +176,111 @@ export class DocExplorer extends React.Component {
   }
 }
 
+// Render Search Results
+class SearchDoc extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    return (this.props.schema !== nextProps.schema) ||
+      (this.props.searchValue !== nextProps.searchValue);
```
leebyron: nit: no need for the wrapping () on each of these terms.

src/components/DocExplorer.js
```diff
@@ -148,6 +176,111 @@ export class DocExplorer extends React.Component {
...
+    const searchValue = this.props.searchValue;
+    const schema = this.props.schema;
+    const onClickType = this.props.onClickType;
+    const onClickField = this.props.onClickField;
+
+    const typeMap = schema.getTypeMap();
+
+    let matchedTypes = [];
+    let matchedFields = [];
+    let matchedArgs = [];
```
leebyron: this is no longer being used, rm

src/components/DocExplorer.js
```diff
@@ -148,6 +176,111 @@ export class DocExplorer extends React.Component {
...
+              <div className="doc-category-item">
+                <a className="field-name"
+                  onClick={event => onClickField(field, type, event)}>
+                  {field.name}
+                </a>
+                {'('}
+                <span>
+                  {matches.map(arg =>
+                    <span className="arg" key={arg.name}>
+                        <span className="arg-name">{arg.name}</span>
```
leebyron: nit: indented 2sp too far

src/components/DocExplorer.js
```diff
@@ -148,6 +176,111 @@ export class DocExplorer extends React.Component {
...
+                  onClick={event => onClickField(field, type, event)}>
+                  {field.name}
+                </a>
+                {' on '}
+                <TypeLink type={type} onClick={onClickType} />
+              </div>
+            );
+          } else if (field.args && field.args.length) {
+            let matches = field.args.filter(arg => {
+              return arg.name.indexOf(searchValue) !== -1;
```
leebyron: you can make this a directly returned expression instead of a function block

```
const matches = field.args.filter(arg => arg.name.indexOf(searchValue) !== -1);
```

leebyron:
This looks great. I have a few other comments in there. One overarching minor thing: it's nice to use `const` instead of `let` as the default for defining variables, and use `let` as the exception when you need to reassign it in the future.

Once you've covered all the comments and updated it, merge away!

src/components/DocExplorer.js
```diff
@@ -148,6 +176,111 @@ export class DocExplorer extends React.Component {
...
+            });
+            matchedFields.push(
```
asiandrummer: This will allow search results to match arguments/render fields every time. Fixed:
```
if (matches.length > 0) {
```

asiandrummer:
Thanks @leebyron!
Next diffs will include navStack integration and a newer format in Doc Explorer.

Issue #74, opened 2015-12-19
HTML embedded in description field is not escaped


runeh:
Steps to reproduce

- Define a graphql type with a description field that contains markup. For example `<input type="checkbox">`
- Inspect the type in docs panel in graphiql

Expected behaviour

- The string `<input type="checkbox">` is shown in docs panel

Actual behaviour

- A checkbox is shown in the docs panel

![screenshot 2015-12-19 02 39 57](https://cloud.githubusercontent.com/assets/64223/11910571/f7040820-a5f9-11e5-8875-e2faf28c4b62.png)

As far as I can tell, it's just a matter of adding `{sanitize: true}` as a config option when calling the `marked` function.

asiandrummer:
Thanks for the report - I'm checking this out.

asiandrummer:
Added sanitizing option to DocExplorer

lelong71:
It was kind of nice to be able to put a link in description.

leebyron:
Yeah this is good. Just as a heads up for anyone who lands here with questions:

If you'd like to include some example code in a description, use backticks: `` to do so.

```
`<input type="checkbox">`
```

Becomes:

`<input type="checkbox">`

That just removes any ambiguity and ensures all markdown renderers will include your example code correctly and not accidentally render HTML or aggressively sanitize and remove it.

Also, markdown has syntax for including the types of things we want to encourage in description documentation like `*italics*`, `**bold**`, `# Headers`, and `[links][http://url.com]`

PR #84, opened 2016-01-26
Escape HTML embedded description by sanitizing output

asiandrummer:
addresses #74


leebyron:
This is the right call. I think it's reasonable that we limit display of markdown to only non-html markdown.

Issue #24, opened 2015-08-31
Cmd-Space key combination conflicts in OS X


dittos:
In OS X, multilingual users use Cmd-Space key combination to switch between input methods. And Ctrl-Space is mapped to Spotlight shortcut key.

It would be better if some other key combination is added, like Option-Space. (Originally I was going to send a pull request, but didn't come up with the key combination that feels right :disappointed:)

lqez:
:+1:

eincs:
:+1:

leebyron:
I have no preference here since Cmd-Space is my habit from TextMate and Sublime. Those of you affected by this, what key combo do you use in other IDEs to trigger typeaheads/autocomplete?

charypar:
Would shift-space work as an alternative for people? I have both cmd-space and ctrl-space mapped in the OS

leebyron:
I'm just curious what people use for auto-complete in IDEs when these keys are already mapped.

asiandrummer:
Maybe we could optionalize the key mapping as well, unless there is a set of mappings popular to many people across. Shortcut key mapping as a prop in GraphiQL should be good enough.

leebyron:
I'd rather not optionalize it since that could be confusing for what works in one version of GraphiQL to not work elsewhere. Also, optionalizing by GraphiQL installation will just end up benefiting the author of the GraphiQL install rather than it's users.

Instead I think it makes sense to just have multiple key-combos that open the typeahead - basically any combo that's not already used for text editing is a viable candidate. That way if you have one of our multiple key-combos mapped to something else in your OS or browser that it can take precedence but there are always alternatives to use. This also means that we don't need to come to any consensus here, we can just add a couple reasonable ones.

How about all cmd+space, alt+space, opt+space and shift+space trigger typeahead? Are any of these commonly used by text inputs in any OS that would be frustrating to use? Like triggering an alternative keyboard or something?

jamesroseman:
I'm a fan of this solution as well. The multiple options for universality feels right, and I agree with @leebyron when it comes to benefiting the author over benefiting the users. That said, Cmd + Space has been working fine for me.

asiandrummer:
>  That way if you have one of our multiple key-combos mapped to something else in your OS or browser that it can take precedence but there are always alternatives to use.
I thought exactly the opposite was the case - sorry for the misunderstanding. Also agree with multiple options.

PR #85, opened 2016-01-27
Enable extra key mappings for typeahead

asiandrummer:
addresses #24. Added `Alt-Space` and `Shift-Space` as alternatives.


leebyron:
#winning

PR #86, opened 2016-01-28
Integrate search results into Documentation Explorer navigation stack

asiandrummer:


asiandrummer:
screen shot 2016-01-27 at 5 32 08 pm
This happens now.

src/components/DocExplorer.js
```diff
@@ -78,7 +83,15 @@ export class DocExplorer extends React.Component {
   }
 
   _onSearch = event => {
-    this.setState({ searchValue: event.target.value });
+    if (event.target.value === '') {
```
leebyron: Is there any way this event could be triggered twice where both trigger this condition? If so then the nav will go backwards twice. For example, are there keys that can be pressed such that when the keyup event below is fired, we accidentally interpret it as a "nav back" event?

src/components/DocExplorer.js
```diff
@@ -43,23 +43,31 @@ export class DocExplorer extends React.Component {
   }
 
   // Public API
-  showDoc(typeOrField) {
-    var navStack = this.state.navStack;
-    var isCurrentlyShown =
-      navStack.length > 0 && navStack[navStack.length - 1] === typeOrField;
-    if (!isCurrentlyShown) {
-      navStack = navStack.concat([ typeOrField ]);
+  showDoc(navItem) {
```
leebyron: This API gets less clear as a result of this change, perhaps there's a better way to do this. Search should probably enter the nav stack via a different API to preserve this

src/components/DocExplorer.js
```diff
@@ -43,23 +43,31 @@ export class DocExplorer extends React.Component {
...
+    // Creating a deep copy in order to trigger state change when lastEntry is
+    // modified.
+    var navStack = this.state.navStack.slice(0);
+    let lastEntry = navStack.length > 0 && navStack[navStack.length - 1];
+    const isCurrentlyShown = lastEntry === navItem;
+
+    // First check if a navItem argument and the last entry from navStack are
+    // both Search Results - if so, push back the last entry with searchValue
+    // updated.
+    if (navItem.name === 'Search Results' &&
```
leebyron: For example, this seems a bit suspect. This creates a coupling to this particular string. Also, as search expands in features, for example to search within a particular type, it's unlikely that we'll want to always have this title.

leebyron:
This is pretty cool, I'm excited to have this.

asiandrummer:
@leebyron I made some changes based on your suggestions as to how to work with the navStack. Introducing another API to integrate Search results into the navStack clarified things up a quite bit.

leebyron:
Most dope

PR #91, opened 2016-02-14
Display the toolbar's scrollbar only when needed

BenoitZugmeyer:
Everybody seems to be using an OS with invisible scrollbars, but at least on Linux, `overflow: scroll` makes the scrollbars always visible.

screenshot_20160214_184528

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -69,7 +69,7 @@
 }
 
 #graphiql-container .toolbar {
-  overflow-x: scroll;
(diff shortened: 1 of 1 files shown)


asiandrummer:
Whoa, I didn't know that's the case. I want to see what happens to the window (container) height when the Toolbar does/doesn't overflow, and make sure things don't break. I'll merge this when I confirm.

Thanks for the catch!

PR #92, opened 2016-02-14
Fix example build script execution

BenoitZugmeyer:
On Linux, `npm run build` failed:

```
$ npm run build

> @ build /home/.../graphiql/example
> . build.sh

sh: line 0: .: build.sh: file not found

...
```


example/package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "start": "node dist/server.js",
     "prestart": "npm run build",
-    "build": ". build.sh"
+    "build": "./build.sh"
```
asiandrummer: I think this should be:
  . ./build.sh

I didn't know what `.` does and had to look up; for a better clarity, I propose:
  source ./build.sh

PR #93, opened 2016-02-15
Client can set different editor options

Bjvanminnen:
I find myself wanting to change the CodeMirror editor options whenever I use GraphiQL (auto closing brackets make me angry). I'm not sure I love this approach for enabling that, but it is at least something.

Changing options requires that I open the developer console and set a magic item on our storage item with the JSON representation of those items I want to override.

`localStorage.setItem('graphiql:editorOptions', JSON.stringify({ autoCloseBrackets: false }))`

It might be nice if there was a cleaner way to expose this, though we'd still likely want to persist these override options via storage. I'm happy to explore suggestions for other approaches.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -103,7 +103,8 @@ export class GraphiQL extends React.Component {
     variables: PropTypes.string,
     onEditQuery: PropTypes.func,
     onEditVariables: PropTypes.func,
-    getDefaultFieldNames: PropTypes.func
+    getDefaultFieldNames: PropTypes.func,
+    editorOptions: PropTypes.object
   }
 
   /**
@@ -322,6 +323,7 @@ export class GraphiQL extends React.Component {
                 value={this.state.query}
                 onEdit={this._onEditQuery}
                 onHintInformationRender={this._onHintInformationRender}
+                editorOptions={this._getEditorOptions()}
               />
               <div className="variable-editor" style={variableStyle}>
                 <div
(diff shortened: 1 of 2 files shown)


src/components/GraphiQL.js
```diff
@@ -322,6 +323,7 @@ export class GraphiQL extends React.Component {
                 value={this.state.query}
                 onEdit={this._onEditQuery}
                 onHintInformationRender={this._onHintInformationRender}
+                editorOptions={this._getEditorOptions()}
```
leebyron: This should be `this.props.editorOptions` based on the added prop type, correct?

leebyron:
This seems like a good idea, but I don't think you need local storage to accomplish this. Instead you just need to pass through the prop you defined on `<GraphiQL />`.

Then you could write `<GraphiQL editorOptions={{ autoCloseBrackets: false }} />` in your installation of GraphiQL to accomplish this.

Bjvanminnen:
I was originally taking that approach - hence the editorOptions prop on GraphiQL's propTypes that I neglected to removed.

However, it seems unfortunate to have to create my own GraphiQL component to be able to have different editor options. In my particular case, I'm using GraphiQL via express-graphql, so I would then need to find a way to plumb the change through there as well.

It seems more appealing for a user to have some way to customize the editor themselves. Maybe the better approach would be to expose some UI for customizing editor preferences rather than using localStorage as this mechanism.

asiandrummer:
I'd prefer having `editorOptions` managed outside of GraphiQL scope, probably before the installation of GraphiQL component, since that local storage entry may cause an overwrite to `graphiql:editorOptions` entry if another GraphiQL instance with different editor options is installed (another browser tab maybe). I think it'd be a lot simpler to manage the custom options if a prop is passed down to GraphiQL for every installation - GraphiQL would simply pass the props down to QueryEditor component.

Addition editor preference UI sounds better, but I'm still not sure which options we'd provide to be customizable, which is why I'm not 100% comfortable with that approach. Also saving the options would cause a similar problem as described above.

For your use case, we could make changes to `express-graphql` to include editorOptions with an option to render GraphiQL. How does this sound?

leebyron:
Closing this PR since it's aging and we identified two alternative options available to potentially pursue:

For those hosting GraphiQL, it may be useful to allow for more customization via provided props.

For those using GraphiQL, perhaps there is some UI which allows for this kind of customization.

PR #95, opened 2016-02-18
Warn when trying to build before having run `npm install`

steveluscher:
This should prevent struggles like this one: http://stackoverflow.com/questions/35096635/error-in-building-graphiql/35470669#35470669

### DIFF
--- a/resources/build.sh
+++ b/resources/build.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+if [ ! -d "node_modules/.bin" ]; then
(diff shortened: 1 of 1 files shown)


steveluscher:
tl;dr: if you don't `npm install` before running `npm run build`, then the build will use any old version of Babel that it can find (say… version 6, installed globally).

asiandrummer:
This makes sense. I'll look into upgrading babel to the latest, but nonetheless this seems like a good warning. Thanks!

PR #98, opened 2016-03-01
Case insensitive search + allow regexp in query

nodkz:
Step 1:
Take out search method to particular method `SearchDoc._isMatch` with old code.

Step 2:
Change `indexOf` to case insensetive `regexp`. Do not escape user pattern for regexp, for allowing to use regexp in search query string.

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -221,6 +221,14 @@ class SearchDoc extends React.Component {
       this.props.searchValue !== nextProps.searchValue;
   }
 
+  _isMatch(sourceText, searchValue) {
+    try {
+      return sourceText.search(new RegExp(searchValue, 'i')) !== -1;
+    } catch (e) {
+      return sourceText.toLowerCase().indexOf(searchValue.toLowerCase()) !== -1;
+    }
+  }
+
   render() {
     const searchValue = this.props.searchValue;
     const schema = this.props.schema;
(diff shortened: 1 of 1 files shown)


src/components/DocExplorer.js
```diff
@@ -221,6 +221,10 @@ class SearchDoc extends React.Component {
       this.props.searchValue !== nextProps.searchValue;
   }
 
+  _isMatch(sourceText, searchValue) {
+    return sourceText.search(new RegExp(searchValue, 'i')) !== -1;
```
asiandrummer: This will show an uncaught error in console without valid search value I think. Not sure what the best way to handle it would be, but let's at least catch this `invalid regular expression` error.

leebyron:
Nice! Would you mind squashing these into a single commit, then I'm happy to merge this.

leebyron:
@asiandrummer has a good point. I think just passing the raw string to a Regexp is probably too fragile, and I don't think we actually want people searching via Regexp right now, in part to maintain our options of adding more "search-engine-like" methods in the future.

I think it would probably be better to come up with another method for performing the case insensitive match - either by the naive methods of calling `toLowerCase()` on both sides before the match, or by escaping the `searchValue` before passing to RegExp.

nodkz:
It was a bit difficult to up working test environment. So sorry for delay, I didn't have enough time yesterday. 

Fix @asiandrummer notice with `try/catch`. If regexp has error, then switched to (more expensive) lowercase search.

Also leave regexp search to be able to use anchors of start/end. Eg. `at$` for searching `createdAt`, `updatedAt` etc. This is main purpose of using regexp. But I think there will be crazy peoples, which will use it with brackets, metacharacters and qantifiers.

@leebyron commits are squashed and ready for merging.

leebyron:
Great thank you!

Issue #81, opened 2016-01-23
Execute named operation under a cursor


OlegIlyenko:
When I'm working with GraphiQL, more often than not, I have several queries in the editor pane. A typical use-case is to have a mutation and query around in order to make changes and see their effects on different queries.

At the moment I need to comment out all of the queries except one I would like to execute. It can become tiresome after some time. That's because I would like to suggest an alternative approach: an ability to execute query under the cursor. As far as i saw, this is pretty common feature available in similar applications. For instance, I use this feature a lot in elasticsearch [sense](https://chrome.google.com/webstore/detail/sense-beta/lhjgkmllcaadmopgmanpapmpjgmfcfig?hl=en) (which is now part of marvel plugin) and find it pretty useful.

So the idea would be to either have a separate context sensitive shortcut/button or make existing shortcut and "run" button context sensitive. If an editor cursor is positioned somewhere inside of some named query, then in addition to a query, GraphiQL will also send the operation name to a server.

leebyron:
This is a great suggestion.

KyleAMathews:
I've used Sense a lot as well and agree it works very well and would fit
easily into GraphiQL.

asiandrummer:
I love this idea! I'm leaning towards making existing shortcut and "run" button context sensitive - GraphQL operations could be huge and it might be difficult to position/find the separate `run` button, but maybe there is a way to mitigate that.

We could also do the similar background-color change for each operation context to clarify UI a bit more.

opatut:
To reduce the magic, the UI could have one execute ("play" icon) button next to each query/mutation definition (e.g. in the gutter to the left), instead of the global button. 

The current query/mutation would be indicated by highlighting one of the buttons, the Shortcut would "press" that highlighted button.

Issue #40, opened 2015-10-06
A way to pass the operationName


arunoda:
I think there is no way to pass the operationName along with request to the server. Is there any planned feature?

leebyron:
This is currently missing but planned, yes. I'm open to suggestions on how you might like to see this built.

arunoda:
I'd like to see a input box where I can enter this.
We could place this just below the query variables pane.

trbngr:
Just ran into this myself. Would love to see it.

PR #106, opened 2016-03-10
Enable multiple queries

leebyron:
This enables GraphiQL to support multiple queries in the editor pane at the same time. When multiple queries exist, the run button becomes a drop-down menu showing the names of all queries available to run - forcing a selection. Cmd+Enter still runs a query, but it runs whatever the last query was unless the cursor is within the bounds of a query - then it runs that query.

This implements #81 and #40.


opatut:
Good job :+1: Thanks.

OlegIlyenko:
Great, thenks a lot! :+1:

PR #117, opened 2016-03-24
Improve performance in Search in Documentation Explorer

asiandrummer:
- Debounce user actions typing in search box
- Limit the search results to 200 (including types and fields)
screen shot 2016-03-23 at 5 29 05 pm

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -19,6 +19,7 @@ import {
   GraphQLNonNull
 } from 'graphql';
 
+import debounce from '../utility/debounce';
 
 /**
  * DocExplorer
@@ -95,11 +96,7 @@ export class DocExplorer extends React.Component {
     } else if (schema) {
       title = 'Documentation Explorer';
       content =
-        <SchemaDoc
-          schema={schema}
-          onClickType={this.handleClickTypeOrField}
-          onSearch={this.handleSearch}
-        />;
+        <SchemaDoc schema={schema} onClickType={this.handleClickTypeOrField} />;
     }
 
     let prevName;
@@ -140,7 +137,6 @@ export class DocExplorer extends React.Component {
           <SearchBox
             isShown={shouldSearchBoxAppear}
             onSearch={this.handleSearch}
-            searchValue={navItem ? navItem.searchValue : ''}
           />
           {this.props.schema ? content : spinnerDiv}
         </div>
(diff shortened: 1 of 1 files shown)


src/components/DocExplorer.js
```diff
@@ -181,25 +177,33 @@ export class DocExplorer extends React.Component {
...
     onSearch: PropTypes.func,
   }
 
-  shouldComponentUpdate(nextProps) {
+  constructor(props) {
+    super(props);
+
+    this.state = { value: null };
+
+    this._debouncedOnSearch = debounce(300, () => {
```
leebyron: 300 may be a little slow, have you tested some other values to balance this? My intuition was that it would be closer to 100

src/components/DocExplorer.js
```diff
@@ -246,7 +255,12 @@ class SearchDoc extends React.Component {
     const matchedTypes = [];
     const matchedFields = [];
 
-    Object.keys(typeMap).forEach(typeName => {
+    const typeNames = Object.keys(typeMap);
+    for (const typeName of typeNames) {
+      if (matchedTypes.length + matchedFields.length > 200) {
```
leebyron: >=?

leebyron:
Mostly good. My only major feedback is that the notice that the search was limited is way to noisy. One option is to completely remove it, that could be fine. If you want to keep it, then it should go at the end of the list, not at the top, since you only need to know that the results were truncated if you've scrolled all the way through them and not found what you were after. 

Also, 200 seems like a huge amount of results. What would be the impact of limiting to something way less? 20 or 30? Maybe 100 at the most? When you're searching, you're typically looking for one thing and using the search box in an interactive way to continue to narrow down options until you find what you're after, so I don't think renderings hundreds of matches is worth the rendering performance cost.

asiandrummer:
Thanks for the feedback. I've removed the notice and changed the limit to 100 results.
I tested a bunch of values and thought ~400 brings out the best user experience for me - but I have a slightly slower typing speed than others, which is why I chose 300.

src/components/DocExplorer.js
```diff
@@ -246,7 +255,12 @@ class SearchDoc extends React.Component {
     const matchedTypes = [];
     const matchedFields = [];
 
-    Object.keys(typeMap).forEach(typeName => {
+    const typeNames = Object.keys(typeMap);
+    for (const typeName of typeNames) {
+      if (matchedTypes.length + matchedFields.length === 100) {
```
leebyron: I think you want >=. It's possible that in one iteration you add both types and fields so that one iteration is 99 or less and the next is 101 or more

leebyron:
This looks great. I'll let you merge when you're ready!

PR #125, opened 2016-04-21
Update query facts with a next available query

asiandrummer:
During receiving new `props`, update query facts with a next available query string. This will update `operationName` appropriately.


leebyron:
#goodcatch

PR #137, opened 2016-05-17
readme-fix

mkuts12:
fix to the readme for people that don't have babel installed globally.
Personally had this problem and npm installing everything fixed it


asiandrummer:
Nice! Thanks for the fix ;)

PR #140, opened 2016-06-01
Add button in doc explorer to print the full graphql schema to the co…

Globegitter:
…nsole.

We have had the case sometimes now where it was useful for us to print the whole graphql schema. We could add a script or some build rule in our system, but I thought this is a nice additional feature for graphiql, so here we go:

screenshot from 2016-06-01 15 21 46

Apart from the fact that the button now says 'Print full schema to console'.

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -16,7 +16,8 @@ import {
   GraphQLUnionType,
   GraphQLEnumType,
   GraphQLList,
-  GraphQLNonNull
+  GraphQLNonNull,
+  printSchema,
 } from 'graphql';
 
 import debounce from '../utility/debounce';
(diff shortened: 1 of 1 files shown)


asiandrummer:
This will freeze the browser window if your schema gets large enough. Also, I feel like the server could serve as an endpoint to print out a raw schema (something like `/graphiql/schema`?), instead of `console.log`ing the raw schema.

Globegitter:
That is a fair point @asiandrummer. How about a button that downloads a `.graphql` file? I do know that the server could do that, but then it is two separate endpoints that we have to implement in our apps/make sure they have some security around it/only exist in dev and staging environment. This would keep it neat and simple.

asiandrummer:
@Globegitter I could see that being useful if you needed a physical file. What do you think of using [an introspection query](https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js)? This is how we grab the schema from the fetcher, and perhaps for your application you could create a custom button to print to console/download the `.graphql` file?

As far as implementing this globally to `GraphiQL`, I'm having mixed feelings. Do you think this feature is globally useful for `GraphiQL` to have? What is/would be the use case(s) for the feature?

leebyron:
I agree with @asiandrummer that having a server serve a schema file rather than a local button is more useful. If that's something that should then be available be a button, the proper place for that is probably the button toolbar and not the documentation explorer. You can configure that for any installation of GraphiQL with the `<GraphiQL.Toolbar>` and `<GraphiQL.ToolbarButton>` components.

PR #144, opened 2016-06-09
Fix onToggleDocs behavior.

skevy:
Right now, if the onToggleDocs prop is not provided, GraphiQL errors. Let's check for the presence of that function before trying to call it.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -684,7 +684,9 @@ export class GraphiQL extends React.Component {
   }
 
   handleToggleDocs = () => {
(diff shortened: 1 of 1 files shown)


skevy:
@asiandrummer I don't usually ping on PR's but I thought I'd ping on this one simply because docs are fully broken in the latest release.

maxcodes:
any updates on this?

schickling:
Thanks @skevy. I totally missed that one...

asiandrummer:
I'm sorry to be late in responding guys - I'll be sure to be faster in responses in the future. This lgtm; thanks for the fix @skevy!!

PR #143, opened 2016-06-09
Default state for search box should not be null

dahjelle:
I'm getting the following warning when loading GraphiQL.

> warning.js:44 Warning: `value` prop on `input` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.

I believe the attached PR fixes the issue.

PR #145, opened 2016-06-09
Fix "value cannot be null" warning.

skevy:
Resolves the fact that the `SearchBox` component causes a warning to thrown because the value passed to the `input` is null, rather than an empty string or undefined.

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -195,7 +195,7 @@ class SearchBox extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = { value: null };
+    this.state = { value: '' };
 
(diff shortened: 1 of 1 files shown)


asiandrummer:
@skevy - #145 fixes this, and I was just super late to check these out. Thanks for the contribution!

gjtorikian:
@asiandrummer this PR _is_ #145; is there another PR you had in mind?

asiandrummer:
Oops, sorry, I meant #143.

gjtorikian:
Sounds like this ought to be closed, then.

asiandrummer:
Closing this one in favor of #143.

PR #146, opened 2016-06-15
Don't assume server is mounted at `/graphql`

joshprice:
The GraphQL server happens to be mounted at `graphql` in this case (https://github.com/graphql/graphiql/blob/master/example/server.js#L28), but we shouldn't assume that.

This fix should allow the GraphQL server to be mounted anywhere.

The reason this is important is that we use this template in [GraphQL Elixir](https://github.com/graphql-elixir/graphql) (other implementations would likely do something similar). Our template can be found here https://github.com/graphql-elixir/plug_graphql/blob/master/templates/graphiql.eex

We found some issues when we took this template verbatim before line 111.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -90,7 +90,7 @@
 
       // Defines a GraphQL fetcher using the fetch API.
       function graphQLFetcher(graphQLParams) {
-        return fetch(window.location.origin + '/graphql', {
+        return fetch(window.location.origin + window.location.pathname, {
           method: 'post',
           headers: {
             'Accept': 'application/json',


facebook-github-bot-4:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

joshprice:
CLA is signed, also I have investigated those test failures and could not figure out why the linter is broken. This is also happening on master locally for me so suspect this might be a dependency change causing this.

asiandrummer:
@joshprice - Sorry for keeping you hanging! I'm taking a look at the linter issue, and although I have no idea why this is happening yet, I'll see if I can unblock you from this as soon as possible.

joshprice:
Rebased and pushed to retrigger the build, looks like it was transient. I did see an explanation somewhere but can't seem to track it down right now.

asiandrummer:
Me neither, but props for solving the issue! Thanks for the contributions :D
I was caught up solving this linter issue, but what do you think about optionally parameterizing the custom endpoint pathname? Something like:
```
function graphQLFetcher(graphQLParams, endpointURL = '/graphql') {
  return fetch(window.location.origin + endpointURL, {});
}
```
This way we can customize the URL as we want, but it'll default to `/graphql` when not supplied. Aside form the customization, I was also thinking this change may break other applications if it relies on `/graphql` URL to use fetcher somehow.

joshprice:
As mentioned in the issue description we have the situation where we have many GraphQL endpoints mounted at various locations and your code would involve us configuring GraphiQL differently for each endpoint whereas using the current path makes it work transparently.

asiandrummer:
Sure - again, thanks for the contribution!

PR #147, opened 2016-06-15
Update React to v15 in example template

joshprice:
Looks like this might have been missed in d5f027ae851d357c6d83e6b908baba888b5b7282


joshprice:
Puzzled by the build error, it happened in #146 as well. Can't see what changes I made could have affected that.

asiandrummer:
👍  awesomesauce

PR #151, opened 2016-07-21
Fixes #150.

### LINKED ISSUE #150: Choosing an operation name from the drop-down immediately after a page load does nothing, but works afterwards
If I have two queries in the editor, and try to run them via the drop-down on the execute button, **nothing happens** the first time I try after a page load, but it works fine afterwards.

The appears to only be an issue with multiple queries and using the execute button.

I attached some logging to the `onMouseUp` handler on the `li`, and it does **not** trigger. On the other hand, the `_onMouseUp` handler that `_onOptionsOpen` attaches **does** trigger.

It looks like the issue is that the `_onMouseUp` handler in `_onOptionsOpen` gets triggered first, runs `setState`, and hides the menu *before* the `onMouseUp` event handler is called.

dahjelle:
Since the `<li>`'s `onMouseUp` handler will set `optionsOpen` to false, we don't need to do that here if we click on the `<li>`. Fixes #150.

### DIFF
--- a/src/components/ExecuteButton.js
+++ b/src/components/ExecuteButton.js
@@ -111,7 +111,13 @@ export class ExecuteButton extends React.Component {
       } else {
         document.removeEventListener('mouseup', onMouseUp);
         onMouseUp = null;
-        this.setState({ optionsOpen: false });
+        const isOptionsMenuClicked = (
+          downTarget.parentNode.compareDocumentPosition(upEvent.target) &
+          Node.DOCUMENT_POSITION_CONTAINED_BY
+        );
+        if (!isOptionsMenuClicked) { // menu calls setState if it was clicked
+          this.setState({ optionsOpen: false });
+        }
       }
     };
 


dahjelle:
And…I haven't the slightest idea why those tests are failing. They fail locally (forgot to look earlier…oops!), but they are also failing on `master`. Perhaps a dependency upgraded?

asiandrummer:
@dahjelle - Thanks for fixing this! I'll take a look at the master failure in a bit.

src/components/ExecuteButton.js
```diff
@@ -111,7 +111,10 @@ export class ExecuteButton extends React.Component {
       } else {
         document.removeEventListener('mouseup', onMouseUp);
         onMouseUp = null;
-        this.setState({ optionsOpen: false });
+        // don't run setState if we click on an <li> menu item
+        if (upEvent.target.parentNode.parentNode !== downTarget.parentNode) {
```
asiandrummer: A small nit, but `upEvent.target.parentNode.parentNode` could be named into something more readable

dahjelle:
I tried a different tactic using `compareDocumentPosition`, but I'm not necessarily convinced it is more readable. It is probably less fragile. Thoughts?

src/components/ExecuteButton.js
```diff
@@ -111,7 +111,13 @@ export class ExecuteButton extends React.Component {
       } else {
         document.removeEventListener('mouseup', onMouseUp);
         onMouseUp = null;
-        this.setState({ optionsOpen: false });
+        // don't run setState if we click on the menu
+        if (!(
```
asiandrummer: How about:
```
const isOptionsMenuClicked = !(
  downTarget.parentNode.compareDocumentPosition(upEvent.target) &
  Node.DOCUMENT_POSITION_CONTAINED_BY
);
if (isOptionsMenuClicked) {
```

asiandrummer:
So the build fails on `react-test-utils@15.2.1` because the DOM provided to `CodeMirror` doesn't have `getClientRects` function while creating a `Range` object. I'm trying to see if this is a known issue - nonetheless, I'll get the tests working soon. Thanks for your patience!

dahjelle:
>  Thanks for your patience!

Thank *you* for the hand-holding! Glad to help. :-D

> How about… <snip>

Yes, I like it. Thanks! I moved the `!` to make the variable name make more sense. It still isn't *strictly* accurate—it's something more like `isOptionsMenuClickedAndNotAlreadyClosed`—but I think it's close enough to read the code and get an idea of what is going on.

> So the build fails on react-test-utils@15.2.1 because the DOM provided to CodeMirror doesn't have getClientRects function while creating a Range object. I'm trying to see if this is a known issue…

It looks like [`js-dom`'s implementation of `Range` is a work-in-progress](https://github.com/tmpvar/jsdom/issues/317).

For whatever it's worth, I found [this page](https://discuss.codemirror.net/t/working-in-jsdom-or-node-js-natively/138) where they had a suggestion to add a mocked `document.createRange` to the tests. If I add 

```
document.createRange = function () {
  return {
    setEnd() { },
    setStart() { },
    getBoundingClientRect() {
      return { right: 0 };
    },
    getClientRects() {
      return { right: 0 };
    }
  };
};
```

to `GraphiQL-test.js`, the tests pass.

asiandrummer:
omg that's exactly what we needed - 010d47 takes care of this!

dahjelle:
Cool! I rebased this PR on master. :-D

Issue #152, opened 2016-07-26
Improvements for a better embedding several instances of GraphiQL on the same page 


OlegIlyenko:
Recently I implemented a small app that builds on top of GraphiQL and provides additional functionality like tabs, headers, etc. http://toolbox.sangria-graphql.org/graphiql 

Along the way I faced several issues when I tried to include several instances of GraphiQL on the same page. I also tried to use single instance and control it's state via props based on the selected tab, but it did not really worked very well at all. I found "fragile workarounds" for all of the issues, but I would like to list some of the issues with possible solutions in order to provide a better alternatives in future.
- [x] [top-level container uses an `id` of the element](https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L289) which is [not appropriate](http://programmers.stackexchange.com/a/127180) if one wants to include several instances on the same page. This also makes styling a bit harder. I think `class` would be a better choice here.
- [x] GraphiQL [always includes an event listener](https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L199) for a key press which also sub-optimal. It would be nice to make it optional and configurable via props. I [implemented](https://github.com/OlegIlyenko/graphql-toolbox/blob/master/app/javascripts/GraphiQLTool.jsx#L31-L39) a very fragile workaround to disable it and registered a single listener for a whole page. This also means that I had to use [`_runQueryAtCursor`](https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L586) which I'm not suppose to use from the outside (I assume that underscore == private). It would be nice to "officially" expose this method as well.
- [ ] CodeMirror inside of the GraphiQL has issues when rendered in a hidden (`display: none;`) div. The gutter size gets messed up a bit. I had to use a trick where I render an empty div on a tab change (first time only) and then, on a next "frame", i render the actual thing. I'm not 100% sure what the actual problem is :)

**A minor thing to consider**

GraphiQL uses `box-sizing: content-box;`. As far as i saw, many people and css frameworks, like bootstrap and foundation, start to adopt `box-sizing: border-box;`. it's easy to fix with 

``` css
.graphiql-container * {
  box-sizing: content-box;
}
```

I'm not very familiar with this field, so i'm not sure what would be a better choice, but still maybe it's a point worth considering

asiandrummer:
[...]

PR #154, opened 2016-07-28
Added prop `isKeyHandlerDisabled` to disable key press event listeners and exposed `runQueryAtCursor`

OlegIlyenko:
As discussed in #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,10 @@ and children.
   Accepts a GraphQLType instance and returns an array of field names.
   If not provided, a default behavior will be used.
 
+- `isKeyHandlerDisabled`: an optional boolean that disables key press event
+  listeners.
+ 
+
 **Children:**
 
 * `<GraphiQL.Logo>`: Replace the GraphiQL logo with your own.
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -89,6 +89,9 @@ import {
  *     Accepts a GraphQLType instance and returns an array of field names.
  *     If not provided, a default behavior will be used.
  *
+ *   - isKeyHandlerDisabled: an optional boolean that disables key press event
+ *     listeners.
+ *
  * Children:
  *
  *   - <GraphiQL.Logo> Replace the GraphiQL logo with your own.
@@ -120,7 +123,8 @@ export class GraphiQL extends React.Component {
     onEditVariables: PropTypes.func,
     onEditOperationName: PropTypes.func,
     onToggleDocs: PropTypes.func,
-    getDefaultFieldNames: PropTypes.func
+    getDefaultFieldNames: PropTypes.func,
+    isKeyHandlerDisabled: PropTypes.bool
   }
 
   constructor(props) {
@@ -195,8 +199,9 @@ export class GraphiQL extends React.Component {
     // Utility for keeping CodeMirror correctly sized.
     this.codeMirrorSizer = new CodeMirrorSizer();
 
-    // Add shortcut for running a query.
-    document.addEventListener('keydown', this._keyHandler, true);
+    if (!this.props.isKeyHandlerDisabled) {
+      this._addKeyPressEventListeners();
+    }
   }
 
   componentWillReceiveProps(nextProps) {
@@ -225,6 +230,16 @@ export class GraphiQL extends React.Component {
         nextQuery !== this.state.query ||
         nextOperationName !== this.state.operationName) {
       this._updateQueryFacts(nextQuery);
+    }
+
+    if (
+      !this.props.isKeyHandlerDisabled && nextProps.isKeyHandlerDisabled
+    ) {
+      this._removeKeyPressEventListeners();
+    } else if (
+      this.props.isKeyHandlerDisabled && !nextProps.isKeyHandlerDisabled
+    ) {
+      this._addKeyPressEventListeners();
     }
 
     this.setState({
@@ -256,7 +271,9 @@ export class GraphiQL extends React.Component {
     this._storageSet('variableEditorHeight', this.state.variableEditorHeight);
     this._storageSet('docExplorerWidth', this.state.docExplorerWidth);
 
-    document.removeEventListener('keydown', this._keyHandler, true);
+    if (!this.props.isKeyHandlerDisabled) {
+      this._removeKeyPressEventListeners();
+    }
   }
 
   render() {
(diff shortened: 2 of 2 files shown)


OlegIlyenko:
I tried to stick to a naming scheme established in the project, but still `isKeyHandlerDisabled` sounds like something that can be improved :) I would appreciate any suggestions for a better name.

src/components/GraphiQL.js
```diff
@@ -575,15 +590,24 @@ export class GraphiQL extends React.Component {
     return;
   }
 
+  _addKeyPressEventListeners = () => {
+    // Add shortcut for running a query.
+    document.addEventListener('keydown', this._keyHandler, true);
+  }
+
+  _removeKeyPressEventListeners = () => {
+      document.removeEventListener('keydown', this._keyHandler, true);
```
asiandrummer: nit: indenting 2 spaces instead of 4 please

src/components/GraphiQL.js
```diff
@@ -575,15 +590,24 @@ export class GraphiQL extends React.Component {
...
     // "Run Query" event.
     if ((event.metaKey || event.ctrlKey) && event.keyCode === 13) {
       event.preventDefault();
-      this._runQueryAtCursor();
+      this.runQueryAtCursor();
     }
   }
 
-  _runQueryAtCursor() {
+  runQueryAtCursor() {
```
asiandrummer: I think this function could use comments describing the usage as well as `@public` - look at [1] for the example

src/components/GraphiQL.js
```diff
@@ -227,6 +232,14 @@ export class GraphiQL extends React.Component {
       this._updateQueryFacts(nextQuery);
     }
 
+    if (!this.props.isKeyHandlerDisabled &&
```
asiandrummer: nit: I'd prefer this styling if you don't mind please :D
```js
if (
  !this.props.isKeyHandlerDisabled &&
  nextProps.isKeyHandlerDisabled
) {
```

OlegIlyenko: sure, still leaning the conventions :) I also added missing semicolons (not sure how significant they are, but as far as i saw, the rest of the code-base uses them)

src/components/GraphiQL.js
```diff
@@ -607,6 +609,12 @@ export class GraphiQL extends React.Component {
     }
   }
 
+  /**
+   * Runs query at a cursor and adds the `operationName`
+   * to a request if the operation defines one.
+   *
+   * @public
+   */
```
OlegIlyenko: @asiandrummer do you think this documentation is sufficient and good enough? Let me know and I can improve it

OlegIlyenko:
Rebased. I also squished 2 commits along the way.

asiandrummer:
I came around this and took a closer look just now - sorry for the delay.

I spent some more time thinking about this, and was curious what the use case for disabling the `keydown` event optionally. I might be missing something, but your implementation for including a `keydown` event looked similar that it was being included at (`componentDidMount`)[https://github.com/OlegIlyenko/graphql-toolbox/blob/master/app/javascripts/GraphiQLTool.jsx#L43]. Was the suggested approach to be able to define shortcuts and key events via props?

OlegIlyenko:
In my use-case I have several tabs and every tab contains an instance of GraphiQL component. Ideally i would like enable/disable the shortcuts of every individual GraphiQL based on the active tab. In the current codebase, they all will register a global listeners which makes them all react on <kbd>cmd</kbd>+<kbd>Enter</kbd>. In my case though, only one of them should react on the shortcuts and all other should be passive about it.

Proposed implementation fits nicely in this workflow, because it also contains this logic in `componentWillReceiveProps`: https://github.com/graphql/graphiql/pull/154/files#diff-2c51ca6a844e000cbc40288fbe06bd28R235

This means that i can parametrize each GraphiQL component on the page like this:

```html
<GraphiQL ... isKeyHandlerDisabled={tab.id != activeTabId} />
```

and active GraphiQL will register/unregister an event listeners based on the active tab. 

> Was the suggested approach to be able to define shortcuts and key events via props?

In my use-case I would prefer to avoid it. Ideally I would like to preserve the original look & feel of GraphiQL as much as possible, which also includes default shortcuts. I just need to disable/enable these global shortcuts based on the context (active tab in my case).

asiandrummer:
I understand. The reason why I asked about defining shortcuts and key events is because the same case applies not only for the run query shortcut (`keydown` event you're having a problem with) but also other feature shortcuts such as typeahead (`Ctrl+Space`) as well as other mouse events used for resizing the container/browser - and in addition to having an uncontrolled set of events per each instance, creating a multiple instances of GraphiQL with events attached at the document level can definitely introduce a perf regression, albeit a small one. It feels like we would have to toggle all those event listeners with your suggested flag in the props.

I'm fine merging this PR as it is since I have a feeling that your use case would be somewhat rare - the above was mainly to ask you for your opinion about the matter, as well as a brain dump of mine :D

OlegIlyenko:
I see your point. To be honest, I haven't considered other shortcuts. I guess there were not as noticeable as <kbd>cmd</kbd>+<kbd>Enter</kbd>. 

Just double-checked. Looks like CodeMirror [limits the scope of the event listeners to an input](https://github.com/codemirror/CodeMirror/blob/6d1379e29ad5a7bcef3ecfec7c6cd739a53567d9/lib/codemirror.js#L3524-L3529) where GraphiQL defines a global listener on a `document`. 

Just an idea for an alternative solution: maybe we can register a keydown listener on a CodeMirror itself? I think Codemirror provides this functionality via [Key Maps](https://codemirror.net/doc/manual.html#keymaps). I think this would help for my use-case as well.

README.md
```diff
@@ -123,6 +123,10 @@ and children.
   Accepts a GraphQLType instance and returns an array of field names.
   If not provided, a default behavior will be used.
 
+- `isKeyHandlerDisabled`: an optional boolean that disables key press event
```
OlegIlyenko: Definitely, I'll add this documentation. 

> How about disableRunQueryShortcut or disableExecuteQueryShortcut?

yeah, this sounds much batter. I think gave it more generic name just in case new shortcuts would be added in future. In this case, according my use-case, it does not really matter what shortcuts are disabled, as long as it covers all global shortcuts (added to a `document`). 

Just got one more idea: what do you think about `disableGlobalShortcuts`? I feel that it describes the use-case a bit better.

asiandrummer: Although technically not there yet, `disableGlobalShortcuts` would also work :D

asiandrummer:
> To be honest, I haven't considered other shortcuts. I guess there were not as noticeable as cmd+Enter.

Sorry for the confusion - most of them aren't really "shortcuts", they're just global events like `Cmd + Enter`.
> Just an idea for an alternative solution: maybe we can register a keydown listener on a CodeMirror itself?

I think that's a much better approach. We don't have to tackle it in this PR though (different scope-ish).

With `disableGlobalShortcuts` change, I'm ready to merge this PR!

OlegIlyenko:
I prepared another pull request: #158. I also verified and it indeed works for my use-case as well. If we will decide for solution presented in #158, then we can just close this one since it would not be relevant anymore.

asiandrummer:
#158 was merged in - closing this one as a response.

PR #158, opened 2016-08-03
Use CoreMirror's key maps instead of global event handler to run query at a cursor

OlegIlyenko:
An alternative solution for a issue discussed in #152. The original solution can be found in #154.


asiandrummer:
Hey, this looks so much better! Thanks for working with me on this - I really appreciate it.
Merging this and closing #154 in process.

OlegIlyenko:
Great! Thank you too, it was a nice brainstorming :+1:

PR #165, opened 2016-09-15
resolving broken deprecation message in docs

brandonblack:
Relates to:
https://github.com/graphql/graphiql/pull/157#issuecomment-245028499

Correcting usages of `field.isDeprecated` in docs to look at `field.deprecationReason` instead restores this functionality.

screen shot 2016-09-15 at 2 09 26 pm

--
**CC:** @davidcelis @globegitter

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -496,7 +496,7 @@ class TypeDoc extends React.Component {
                 {': '}
                 <TypeLink type={field.type} onClick={onClickType} />
                 {
-                  field.deprecationReason &&
+                  (field.isDeprecated || field.deprecationReason) &&
                     <span className="doc-alert-text">{' (DEPRECATED)'}</span>
                 }
               </div>
@@ -518,7 +518,7 @@ class TypeDoc extends React.Component {
               <div className="enum-value">
                 {value.name}
                 {
-                  value.deprecationReason &&
+                  (value.isDeprecated || value.deprecationReason) &&
                     <span className="doc-alert-text">{' (DEPRECATED)'}</span>
                 }
               </div>
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -496,7 +496,7 @@ class TypeDoc extends React.Component {
                 {': '}
                 <TypeLink type={field.type} onClick={onClickType} />
                 {
-                  field.isDeprecated &&
+                  field.deprecationReason &&
                     <span className="doc-alert-text">{' (DEPRECATED)'}</span>
                 }
               </div>
@@ -518,7 +518,7 @@ class TypeDoc extends React.Component {
               <div className="enum-value">
                 {value.name}
                 {
-                  value.isDeprecated &&
+                  value.deprecationReason &&
                     <span className="doc-alert-text">{' (DEPRECATED)'}</span>
                 }
               </div>


facebook-github-bot-7:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

brandonblack:
I've signed the CLA previously, but let me know if you need anything else here.

gjtorikian:
This doesn't seem like it'll show the deprecation reason though. Without getting too UI :neckbeard:y, is it possible to show that inline somehow?

src/components/DocExplorer.js
```diff
@@ -496,7 +496,7 @@ class TypeDoc extends React.Component {
                 {': '}
                 <TypeLink type={field.type} onClick={onClickType} />
                 {
-                  field.isDeprecated &&
+                  field.deprecationReason &&
```
asiandrummer: >  is it possible to show that inline somehow?

It'll probably be something like:
```js
{
  field.deprecationReason && [
    <span className="doc-alert-text">{' (DEPRECATED)'}</span>,
    <MarkdownContent
      className="doc-alert-text"
      markdown={field.deprecationReason}
    />
  ]
}
```
Instead of reusing `doc-alert-text`, though, it'd be nice to get a grey-ish text for the deprecation reason.

asiandrummer:
Thanks for the contribution @brandonblack! I'd love to get the deprecation reason as well if that's not too much troubles for you.

src/components/DocExplorer.js
leebyron: Hmm, what's happening here?

It's entirely possible that a field is deprecated but also fails to provide a reason.

The previous code looks more correct, so we may need to look elsewhere for the source of the issue.

brandonblack: Looks like we're no longer passing in `isDeprecated` to the DocExplorer. This was working, but broke in the last release. Nothing else references `isDeprecated` so it wasn't noticed. Everywhere else we're simply interacting with the `deprecationReason` only.

This patch sorts the issue out, but as you noted is probably just a symptom of the actual  issue (eg. we're not passing in _both_ of these field values).

How about I modify this to be `field.isDeprecated || field.deprecationReason` since that's really what we want anyway?

leebyron: I like that solution

leebyron: Did it get removed from the introspection query?

leebyron:
Also, this doesn't have to be done in this PR, but I think we should actually filter out fields into two groups, such that all the deprecated fields appear in their own section after the supported fields. We could have this section collapsed by default if we wish.

src/components/DocExplorer.js
brandonblack: @leebyron Hmm. Yeah, I really don't know -- definitely not removed. 

Digging through commits here and in `graphql-js` I don't see anything that would have caused this to change recently. I may be that this never worked as advertised.

In any case, I think checking the presence of either of these fields is what we want since `deprecationReason` is an optional field.

I've resolved this in https://github.com/graphql/graphiql/pull/165/commits/0aa91637d26bb98a217eed17cb6f115f28175239

leebyron:
Yeah this is ideal. Hyo and I think we've found the reason behind the confusion and breakage

PR #167, opened 2016-09-19
Handle null/undefined variables

OlegIlyenko:
Pretty often I end up in this kind of situation:

graphiql-null-vars

Steps to reproduce it:

1. When creating a GraphiQL component, provide a prop: `variables: null`
2. Refresh a page couple of times
3. Remove `variables: null`
4. Refresh the page (the result is shown on the screenshot)

This PR tries to address this issue by preventing null/undefined variables from being stored.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -247,7 +247,7 @@ export class GraphiQL extends React.Component {
   // that when the component is remounted, it will use the last used values.
   componentWillUnmount() {
     this._storageSet('query', this.state.query);
-    this._storageSet('variables', this.state.variables);
+    this._storageSet('variables', this.state.variables || '');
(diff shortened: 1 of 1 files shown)


src/components/GraphiQL.js
```diff
@@ -247,7 +247,8 @@ export class GraphiQL extends React.Component {
   // that when the component is remounted, it will use the last used values.
   componentWillUnmount() {
     this._storageSet('query', this.state.query);
-    this._storageSet('variables', this.state.variables);
+    this._storageSet('variables',
+      this.state.variables ? this.state.variables : '');
```
asiandrummer:
```js
this.state.variables || ''
```
?

OlegIlyenko: Done :)

PR #172, opened 2016-09-30
single interface for fetch

DxCx:
since the fetcher shouldn't know if it is expected to return promise or observable,
it added a small logic to convert observable into promises when promise is a must.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -897,21 +897,19 @@ function isPromise(value) {
   return typeof value === 'object' && typeof value.then === 'function';
 }
 
+// Duck-type Observable.take(1).toPromise()
 function observableToPromise(observable) {
-  if ( isPromise(observable) ) {
-    return observable;
-  }
   if ( !isObservable(observable) ) {
     return observable;
   }
   return new Promise((resolve, reject) => {
-    let lastValue;
-    observable.subscribe(v => {
-      lastValue = v;
+    const subscription = observable.subscribe(v => {
+      resolve(v);
+      subscription.unsubscribe();
     }, e => {
       reject(e);
     }, () => {
-      resolve(lastValue);
+      reject(new Error('no value resolved'));
     });
   });
 }
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -420,7 +420,7 @@ export class GraphiQL extends React.Component {
 
     const fetcher = this.props.fetcher;
 
-    const fetch = fetcher({ query: introspectionQuery });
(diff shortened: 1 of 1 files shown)


leebyron:
I'm concerned that this will break functionality for subscriptions which use Observables to provide multiple payloads for a single request over time.

Could you clarify what the problem/bug is that you're trying to solve?

src/components/GraphiQL.js
```diff
@@ -895,6 +897,25 @@ function isPromise(value) {
   return typeof value === 'object' && typeof value.then === 'function';
 }
 
+function observableToPromise(observable) {
+  if ( isPromise(observable) ) {
+    return observable;
+  }
+  if ( !isObservable(observable) ) {
```
leebyron: why both conditions above? Are both needed or is only one sufficient?

DxCx:
Actually ive written a fetch function which returns observable (and hence, tested it with it) and the issue is that the schema cannot be observable only promise, so, no, its not breaking its just means that the schema returned can be observable which emits one value.

As you know from the other thread im creating a full flow example of subscriptions using observables and this a small issue ive found on the way. (Fetch function expects to return both promise and observables on different places)

src/components/GraphiQL.js
DxCx: Only the second, the first is more for readability.
If you want ill remove it..

leebyron: Yeah - only the second is necessary.

leebyron:
Ah thanks for the extra info. I see you're only wrapping the schema fetching so this looks good

src/components/GraphiQL.js
```diff
@@ -895,6 +897,25 @@ function isPromise(value) {
...
+    return observable;
+  }
+  return new Promise((resolve, reject) => {
```
leebyron: I think you actually want the first value, a "takeOne" kind of logic, rather than waiting for the last. If fetcher returned some sort of unclosing socket, this could never complete.

Perhaps:

```js
if (isObservable(promiseOrObservable)) {
  return new Promise((resolve, reject) => promiseOrObservable.subscribe(resolve, reject, reject));
}
return promiseOrObservable;
```

leebyron: Yeah I think we want a sort of inlined .take(1).toPromise()

DxCx: Ok, Gotcha, so in that case unsubscribe should be called as well.
also, reject won't give any helpful information.
I've added a commit with both CR notes. :)

asiandrummer:
This lgtm as well - thanks for the contribution!

PR #178, opened 2016-10-13
fix custom graphiql example

cmatheson:
Hey, this is kind of a trivial pull, but the CustomGraphiQL example fails to compile because of the error:

    Syntax error: 'this' is not allowed before super()
    ...
    constructor(props) {
      this.state = {
      ^
    ....

Thanks!


asiandrummer:
@cmatheson oopsies, good find - thanks!

PR #182, opened 2016-10-31
move setState before subscribing to observable in case that observer …

aihornmac:
…emits data in the same event loop

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -554,6 +554,13 @@ export class GraphiQL extends React.Component {
     }
 
     try {
+      this.setState({
+        isWaitingForResponse: true,
+        response: null,
+        subscription,
+        operationName,
+      });
+
       // _fetchQuery may return a subscription.
       const subscription = this._fetchQuery(
         editedQuery,
@@ -568,13 +575,6 @@ export class GraphiQL extends React.Component {
           }
         }
       );
-
-      this.setState({
-        isWaitingForResponse: true,
-        response: null,
-        subscription,
-        operationName,
-      });
     } catch (error) {
       this.setState({
         isWaitingForResponse: false,
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -557,7 +557,6 @@ export class GraphiQL extends React.Component {
       this.setState({
         isWaitingForResponse: true,
         response: null,
-        subscription,
         operationName,
       });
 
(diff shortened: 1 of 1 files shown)


asiandrummer:
Thanks for the contribution @aihornmac! I'm not sure if I understand the fix though - I think you could achieve the same effect by calling `.next()` on your observable stream, right?

aihornmac:
@asiandrummer This issue happens when I am mocking data in sync way:
```js
return Observable.create(observer => { // callback triggered right after creating
  observer.next({
    data: {
      user: {
        id: 'id'
      }
    }
  });
});
```
Related code when it runs query:
```js
// _fetchQuery may return a subscription.
const subscription = this._fetchQuery(
  editedQuery,
  variables,
  operationName,
  result => { // triggered right after
    if (queryID === this._editorQueryID) {
      this.setState({
        isWaitingForResponse: false,
        response: JSON.stringify(result, null, 2),
      });
    }
  }
);

this.setState({
  isWaitingForResponse: true,
  response: null,
  subscription,
  operationName,
});
```
So in this event loop, `setState` actually is executed in this order:
```js
this.setState({
  isWaitingForResponse: false,
  response: JSON.stringify(result, null, 2),
});

this.setState({
  isWaitingForResponse: true,
  response: null,
  subscription,
  operationName,
});
```
In result, the result viewer keeps waiting for result (spinning) while there is actually result.

asiandrummer:
I see. I agree that this won't cause (or I can't think of any at the moment) the side effects. Thanks for the contribution!

asiandrummer:
I understand - unfortunately this will break since `subscription` is declared after setting the state. It also looks like the conflict comes from resetting `response` to null only. I suggest we do:
```js
this.setState({
  response: null,
  isWaitingForResponse: true,
  operationName,
});

// _fetchQuery may return a subscription.
const subscription = this._fetchQuery(
  editedQuery,
  variables,
  operationName,
  result => { // triggered right after calling next in the same event loop
    if (queryID === this._editorQueryID) {
      this.setState({
        isWaitingForResponse: false,
        response: JSON.stringify(result, null, 2),
      });
    }
  }
);

this.setState({ subscription });
```
Also could you make sure this gets tested (or let me know that you already did)? React batches the state changes, and although I'm not so sure how it resolves the conflicts between two states with different attributes, I'm thinking it looks at the order of the state changes called, so this should be okay.

aihornmac:
@asiandrummer You are right, I was missing the subscription in state, will check on this.

asiandrummer:
@aihornmac cool! And after that we can merge this :D

aihornmac:
@asiandrummer committed with update

asiandrummer:
Did a simple Observable testing with this change applied and saw no issues. Thanks @aihornmac!

Issue #160, opened 2016-08-07
Allow searching of the full response


ruleb:
When using GraphiQL as a development companion, it fails when searching through the graphql response, when the response is large.

Codemirror (which is used to render the data response) only adds a slice of the response to the HTML dom for performance reasons, causing the browser-native search to fail and not find the text you are looking for in the response.

It would be great to intercept the find event and override using a custom search box to walk through the data response.

It's a trivial optimization, but will save a lot of headache when using it as a development companion. I for one had the feeling of going crazy that sometimes data was missing and sometimes it was not, when falling back to curl, it showed that the response always contained the data I was searching for.

Reproduce:
- Query a large GQL response
- Use CMD+F to find some string at the end of your response, browser will not find it
- Use select+all (which codemirror intercepts?) and paste in another window textbox, use cmd+f again and the data is there.

joelgriffith:
Added in https://github.com/graphql/graphiql/pull/188, uses the CodeMirror search dialog that they expose via an addon (also some other lint fixes in the example code)

PR #188, opened 2016-11-04
Adds in search functionality to result-viewer, bugfixes for example

joelgriffith:
Fixes https://github.com/graphql/graphiql/issues/160

Also fixes a few issues found when using the `example` directory for testing. I could also write some scripts for watching fs changes to tighten the feedback loop when developing (currently you have to build and restart the example server).

### DIFF
--- a/css/codemirror.css
+++ b/css/codemirror.css
@@ -333,3 +333,36 @@ div.CodeMirror-dragcursors {
 
 /* Help users use markselection to safely style text background */
 span.CodeMirror-selectedtext { background: none; }
+
+.CodeMirror-dialog {
+  position: absolute;
+  left: 0; right: 0;
+  background: inherit;
+  z-index: 15;
+  padding: .1em .8em;
+  overflow: hidden;
+  color: inherit;
+}
+
+.CodeMirror-dialog-top {
+  border-bottom: 1px solid #eee;
+  top: 0;
+}
+
+.CodeMirror-dialog-bottom {
+  border-top: 1px solid #eee;
+  bottom: 0;
+}
+
+.CodeMirror-dialog input {
+  border: 1px solid #d3d6db;
+  outline: none;
+  background: transparent;
+  width: 20em;
+  color: inherit;
+  font-family: monospace;
+}
+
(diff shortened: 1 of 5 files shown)


css/codemirror.css
```diff
@@ -333,3 +333,36 @@ div.CodeMirror-dragcursors {
 
 /* Help users use markselection to safely style text background */
 span.CodeMirror-selectedtext { background: none; }
```
joelgriffith: Couldn't find a better spot to add these additions, which come straight from CodeMirror

example/build.sh
```diff
@@ -2,7 +2,7 @@
 
 rm -rf dist/ && mkdir -p dist/ &&
 babel server.js -o dist/server.js &&
```
joelgriffith: According to README here, these _shoudln't_ pull from `node_modules`...

example/server.js
```diff
@@ -23,20 +23,19 @@ import {
   GraphQLList,
 } from 'graphql';
 
-var app = express();
+const app = express();
 app.use(express.static(__dirname));
```
joelgriffith: This isn't the right path...

css/codemirror.css
asiandrummer: This is probably the best place to put these, so good choice.

example/build.sh
asiandrummer: If we make this change, we can probably remove the installation of `GraphiQL` from example/package.json

joelgriffith: Hrm, I see that `graphiql` is referenced via `../graphiql`. I'm not sure why this wasn't working entirely locally... but I feel it's more obvious this way.

asiandrummer:
@joelgriffith let me fix the lint errors really quick - you can rebase after that

asiandrummer:
@joelgriffith I tried to attend to this lint issue, but it seems like on master there are none? Maybe we're running different lint configs?

asiandrummer:
Never mind - I didn't install the correct version of eslint. Master should be clean now.

joelgriffith:
Pulled in master, build passing 🎉

PR #189, opened 2016-11-04
Fixes IE10/11 by polyfilling the Promise global and vendor prefixing

joelgriffith:
Polyfill `Promise` for IE11 and other, less-fortunate browsers


joelgriffith:
Just pushed up another delta that gets this rendered properly in IE10, and fixes numerous issues in that browser as well. It still could use a _tad_ more polish, but it works

asiandrummer:
@joelgriffith awesome - thanks for this!

PR #197, opened 2016-11-07
Feature/example watch

joelgriffith:
I'll annotate the review as much as possible, but the gist of it is:

- Clean out as many packages in `examples` as possible.
- Watch the `src/*.js` files for changes, output as fast as possible.
- Read css individually for better debugging.
- Load _both_ the `example` and `root` folders for easier file fetching (no need to `cp`).
- Limit packages needed for all of this.


example/index.html
```diff
@@ -20,15 +20,25 @@
         height: 100vh;
       }
     </style>
```
joelgriffith: Load css files individually for better debugging.

example/package.json
```diff
@@ -1,29 +1,8 @@
 {
```
joelgriffith: Most, if not all, of these deps are handled in the root of the project (or loaded externally), so need for them here.

example/server.js
```diff
@@ -21,10 +22,12 @@ import {
   GraphQLString,
   GraphQLID,
   GraphQLList,
-} from 'graphql';
+} = require('graphql');
 
 const app = express();
+
 app.use(express.static(__dirname));
```
joelgriffith: Adding the root of the project as another static directory for css and other compiled assets (graphiql.js)

package.json
```diff
@@ -32,13 +32,14 @@
     "mocha": "--full-trace --require resources/mocha-bootload src/**/*-test.js"
   },
   "scripts": {
-    "test": "npm run lint && npm run check && npm run testonly",
-    "testonly": "babel-node ./node_modules/.bin/_mocha $npm_package_options_mocha",
-    "lint": "eslint src",
-    "check": "flow check",
     "build": ". ./resources/build.sh",
```
joelgriffith: I alphabetized this list, and only added `dev` script

resources/dev.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+# Clean up dist in example
+rm -rf example/dist && mkdir -p example/dist
+
+# Boot the server
+node example/server.js &
+
+# Watch JS source files and recompile
```
joelgriffith: Main changes are here. This will watch the `src` file, output changes, babelify it, shim, export `GraphiQL`, and output to the standard `graphiql.js` file.

joelgriffith:
@asiandrummer here's the build PR, comb through carefully :)

example/build.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/sh
```
asiandrummer: Isn't this still needed to build/publish?

asiandrummer:
cc @leebyron

example/build.sh
joelgriffith: I don't see anything in the root `package.json` that references this. Is this example file deployed on `npm`?

joelgriffith:
Looking through the issues on this project, I'm starting to think it _would_ be nice to have a css preprocessor due to things like theming, vendor-specific bugs, and just other nice-to-haves for development.

It would also be nice to take advantage of things like hot-reload, sourcemaps, tree-shaking, and more; which sort of lends itself to using `webpack` as a bundler instead of `browserify` (full disclosure: I prefer `browserify` over `webpack`, but seems like the ecosystem generally prefers `webpack`).

Obviously this now touches the build pipeline and other mission-critical tech. I think it's worth the small amount of risk as long as project maintainers are ok with it? It'll allow me to more quickly turn-around issues and features coming in

joelgriffith:
I should not that build changes went from nearly a minute to milliseconds when actively developing.

leebyron:
This is awesome @joelgriffith - thanks for doing all this!

I think swapping out browserify and adding a css preprocessor are probably overkill though to your points. If there's substantial benefit, then we should certainly consider it, but otherwise no need.

I'll take another review pass through, though I know @asiandrummer already covered a lot :)

example/README.md
```diff
@@ -2,12 +2,15 @@ Example GraphiQL Install
...
 This example uses the version of GraphiQL found in the parent directory, rather
-than depending on npm, so that it is easier to test new changes. In order to use
-the compiled version of GraphiQL, first run build in the parent directory before
-installing and starting the example.
+than depending on npm, so that it is easier to test new changes. In order to use this
+example as a basis for development, you must have at least Node 6 installed.
 
-1. Run `npm install && npm run build` in the parent directory
-2. Navigate to this directory (example) in Terminal
+Instructions:
```
leebyron: This is still quite a few instructions - maybe we should just merge these packages together?

leebyron:
As a result - the steps required for starting the test environment should just be (in the root directory) `npm install && npm run dev`

joelgriffith:
Closing in favor of https://github.com/graphql/graphiql/pulls

PR #200, opened 2016-11-08
Feature/dark theme

joelgriffith:
Super hot dark theme, super slick css, all in one nice PR:

screen shot 2016-11-08 at 1 52 11 pm

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -89,6 +89,8 @@ import {
  *     Accepts a GraphQLType instance and returns an array of field names.
  *     If not provided, a default behavior will be used.
  *
+ *   - theme: A theme to use for overall UI, currently only dark is available.
+ *
  * Children:
  *
  *   - <GraphiQL.Logo> Replace the GraphiQL logo with your own.
@@ -120,7 +122,9 @@ export class GraphiQL extends React.Component {
     onEditVariables: PropTypes.func,
     onEditOperationName: PropTypes.func,
     onToggleDocs: PropTypes.func,
-    getDefaultFieldNames: PropTypes.func
+    getDefaultFieldNames: PropTypes.func,
+    theme: PropTypes.oneOf([ 'dark' ]),
+    syntaxTheme: PropTypes.string
   }
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


src/components/GraphiQL.js
```diff
@@ -120,7 +122,9 @@ export class GraphiQL extends React.Component {
     onEditVariables: PropTypes.func,
     onEditOperationName: PropTypes.func,
     onToggleDocs: PropTypes.func,
-    getDefaultFieldNames: PropTypes.func
+    getDefaultFieldNames: PropTypes.func,
+    theme: PropTypes.oneOf([ 'dark' ]),
```
joelgriffith: Figured we could keep this open to future theme(s)

leebyron:
Could this be done in a simpler way? For example, if there are more themes in the future - this sets a precedent that they can only be added by PR to this repo. Ideally anyone can write and share their own themes and host them in whatever repo they like.

What do we need to do for GraphiQL to enable that? Are the CSS class names descriptive enough that the only necessary thing is to just add another CSS file to the page? Do we need the ability to set an additional classname on the GraphiQL root node?

joelgriffith:
I think it's easy enough to add your own body/html class and overwrite the current theme via css selector specificity. I've already done this myself in a few implementations. We could also expose an optional `className` prop to make this more explicit to users. Mostly this just needs to be documented and, potentially, we have a pre-done `dark` theme for users.

My final thought is that colors and other backgrounds are defined in numerous places, so theming can be kind of painful since you have to overwrite numerous selectors. This begs for either refactoring the css to make theming easier, or implement some preprocessor so consumers can overwrite variables and generate the css themselves.

joelgriffith:
I'm going to close this PR and work on another branch that more descriptively shows how to theme, and offers a `dark` theme as an example

joelgriffith:
Hrm, doing some digging on color definitions:

`ack '\#.*;' css | grep -o '\#.*;' | sort | uniq | tee >(wc -l)` => results in around `72` unique color definitions. 

Dropping the `uniq` in my command shows that _each_ color is defined around ~2 times: `139`. I'm not sure what the threshold is in moving to a css-preprocessor, but this amount of colors and definitions makes it hard(er) for consumers to apply custom themes as there's a _ton_ of rules to rewrite.

sw-yx:
i would be interested in this

rjrobinson:
And another year later

prakashn27:
And two more years later...

acao:
@joelgriffith did you ever re-open this PR? this accidentally was cross referenced when merging other repos

PR #201, opened 2016-11-10
Update TypeDoc when schema changes

xuorig:
@asiandrummer this is something I found while trying to dynamically set schemas in a custom graphiql.

Changing the schema does not currently trigger an update on TypeDoc components. I feel like it should ? Right now the Doc pane shows the old fields even if the schema has been updated.

Let me know what you think!

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -434,7 +434,10 @@ class TypeDoc extends React.Component {
   }
 
   shouldComponentUpdate(nextProps) {
-    return this.props.type !== nextProps.type;
+    return (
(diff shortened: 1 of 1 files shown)


leebyron:
Nice! Yeah - I think we had assumed that if the schema changes, that the type would also change, though I'm fine with this for completeness.

xuorig:
🎉

any plans to cut a release soon 🐱  ?

asiandrummer:
@xuorig thanks :D Let me tidy things up and release soon. I think we have a couple of breaking changes - this release will be a minor one.

PR #203, opened 2016-11-10
Test directory w watchify and more for actively developing

joelgriffith:
Following feedback from https://github.com/graphql/graphiql/pull/197, this adds a `test` directory and a small server to load up a graphiql interface to actively test changes. Currently proxies graphql requests over to swapi-graphql (however the introspection query fails...).


package.json
```diff
@@ -32,13 +32,14 @@
     "mocha": "--full-trace --require resources/mocha-bootload src/**/*-test.js"
   },
   "scripts": {
```
joelgriffith: Alphabetizing

test/server.js
```diff
@@ -0,0 +1,45 @@
+/**
```
joelgriffith: Bundling happens here to help consolidate files

test/index.html
```diff
@@ -0,0 +1,151 @@
...
+          window.location.origin =
+            window.location.protocol + "//" +
+            window.location.hostname + (
+              window.location.port ?
+              ':' + window.location.port:
+              ''
+            );
+        }
+
+        return fetch(origin + window.location.pathname + 'swapi', {
```
asiandrummer: Noticed we're trying to use `swapi` schema and proxy the swapi parse app we're using as an example - is this intentional? Don't we want to use the mocked server schema, how we had it in the `./examples` directory?

Speaking of that, maybe a right path is to define a test schema used in `GraphiQL` testing and use here, and do the proxying you seem to be doing in the `./test/server.js` in examples. What do you think?

joelgriffith: Thanks for the comments! We could probably just export the schema in the `example` directory and I can just use it here. Probably the most flexible solution I can think of.

PR #209, opened 2016-11-14
.on('update') fires with a collection of identifiers, not functions

joelgriffith:
https://github.com/substack/watchify#bonupdate-function-ids-

This addition causes bugs when saving more than once, fixed by checking typeof `then`.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -48,7 +48,7 @@ function makeBundle(then) {
       return;
     }
     bundleBuffer = buffer;
-    then && then();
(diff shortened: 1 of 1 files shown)


joelgriffith:
This also isn't a Promise returning... so maybe `done` or something else more callback style? :bike-shed:

leebyron:
Whoops! Thanks for the fix, I manually merged and also added the renaming of then to "callback" to not create promise confusion

PR #231, opened 2016-12-05
History

tessalt:


tessalt:
this was made by mistake I thought I was on my own fork :(

asiandrummer:
@tessalt took a quick look and this seems to be an awesome contribution if it were to be one ;) are you planning to submit a real PR for this?

tessalt:
@asiandrummer I was thinking about it but I've got some real UX issues to think through... basically i have no idea where to put the panel.

asiandrummer:
@tessalt some idea I can throw:
* Expandable left-side window just like the doc explorer
* A toolbar button that opens a dialog that contains the query history

tessalt:
@asiandrummer re-opened over here with some cleaned-up code https://github.com/graphql/graphiql/pull/232 (had a different history from this branch because of long story two computers)

PR #236, opened 2016-12-08
Fix lint errror in ExecuteButton.js

wincent:
Travis is currently complaining about this ([example build](https://travis-ci.org/graphql/graphiql/builds/181706048)).


wincent:
With that lint error fixed, it exposes [another error further on down the build](https://travis-ci.org/graphql/graphiql/builds/182412683#L774), which I'll look into now.

PR #238, opened 2016-12-09
Upgrade dependencies to fix build breakages

wincent:
We've had some [lint-related build failures](https://travis-ci.org/graphql/graphiql/builds/181453346) lately. Unfortunately, these masked test failures introduced by 35d8d384f. When we fixed the lint (1eeb36cba8), the test failures [showed up](https://travis-ci.org/graphql/graphiql/builds/182431531).

This commit fixes that by upgrading our deps:

- Replace "react-addons-test-utils", which access a path that doesn't  exist in current "react-dom", with "react-test-renderer".

This required some API updates in the tests, but also some changes to use ref-based DOM node access rather than `ReactDOM.findDOMNode`, which doesn't work in "react-test-renderer".

See:

- https://github.com/facebook/react/issues/7371
- https://github.com/facebook/react/issues/8324

> Yea sorry, refs can work but `findDOMNode()` can't (we tried).


wincent:
Test plan: updated a small app I have to use this build (with `yarn link graphiql`), hit http://localhost:3000/graphql and clicked around, ran some queries.

src/utility/CodeMirrorSizer.js
```diff
@@ -20,7 +17,7 @@ export default class CodeMirrorSizer {
 
   updateSizes(components) {
     components.forEach((component, i) => {
-      const size = ReactDOM.findDOMNode(component).clientHeight;
+      const size = component.getClientHeight();
```
asiandrummer: Could you check if `DocExplorer` also needs to expose this method?

PR #243, opened 2016-12-09
Split DocExplorer into smaller files

wincent:
Just to make things easier to find: rather than thinking "where does the search box live?" I can just type "search" into my editor's fuzzy finder and go straight to the component.


asiandrummer:
:D Thanks @wincent! I'm taking a closer look now but I'm pretty sure this is fine as it is.

Issue #139, opened 2016-06-01
Docs can go back to root


crapthings:

asiandrummer:
Hello @crapthings - do you mean docs cannot go back to root? I'm not sure I understand your issue. Would you mind being more specific please?

crapthings:
i think we should add a convenient method to go back to root node

asiandrummer:
@crapthings I see. I'm assuming you're coming across the use case having to navigate into multiple documentation entries, and clicking back button many times instead of one is a bit weird. I'm thinking maybe `Documentation Explorer` title could be linked to the root node of the doc explorer.

crapthings:
@asiandrummer yeah, it can be very handy!

joelgriffith:
Can we close this? seems like the related PR is done

wincent:
I think so. https://github.com/graphql/graphiql/pull/244 ended up making a smaller change than proposed here, but I don't think the original proposal is going to make it, so we may as well close this. Thanks to everybody who participated!

PR #244, opened 2016-12-09
Clicking on documentation title takes you back to the root

wincent:
As suggested on #139 by @asiandrummer.

Closes: https://github.com/graphql/graphiql/issues/139

### DIFF
--- a/css/doc-explorer.css
+++ b/css/doc-explorer.css
@@ -27,6 +27,10 @@
           flex: 1;
 }
 
+.graphiql-container .doc-explorer-root-link {
+  cursor: pointer;
+}
+
 .graphiql-container .doc-explorer-back {
   color: #3B5998;
   cursor: pointer;
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -133,7 +133,11 @@ export class DocExplorer extends React.Component {
             </div>
           }
           <div className="doc-explorer-title">
-            {title}
+            <span
+              className="doc-explorer-root-link"
+              onClick={this.handleTitleClick}>
+              {title}
+            </span>
           </div>
           <div className="doc-explorer-rhs">
             {this.props.children}
@@ -188,6 +192,10 @@ export class DocExplorer extends React.Component {
       name: 'Search Results',
       searchValue: value
     });
+  }
+
+  handleTitleClick = () => {
+    this.setState({ navStack: [] });
   }
 }
 


wincent:
Note that this solution isn't the most discoverable thing in the world, but its presence isn't actively harmful either. We might still want to add something visible (like a home icon, or breadcrumbs) in a separate PR.

leebyron:
Dunno about this, it's pretty unintuitive to click the name of the type you're currently looking at documentation for and clear your nav stack. If someone does this unintentionally they'd probably be upset about it.

Let's come up with a nicer design for this functionality

leebyron:
Also - I'd push against the need for this in the first place - the way you end up with a deep nav stack is by navigating deeply in the first place, so heading back is more important.

What's the drive to get back to "home", is it to get back to the search bar? Maybe we should just make the search bar more omni-present

asiandrummer:
I agree with more omni-present search bar, but I do see some value with going back to the doc's main page, maybe with a button. It's an easy way to look at the root types/fields if you aren't necessarily sure which types are the root types.

leebyron:
I'm not saying there's no value in it, but we need to weigh that against the pixel real-estate a button would consume or the interaction complexity cost of sticking it somewhere undiscoverable or surprising.

You can get back to "home" now by repeatedly clicking the back arrow, which sure is more clicks than one if you nav'd deeply, but I'm not sure that getting back to learn if you named your query root type "Query" or "QueryRoot" is worth reducing a few clicks into one click and a persistent button

wincent:
> I'm not saying there's no value in it, but we need to weigh that against the pixel real-estate a button would consume

Given media queries, I don't think we need to worry too much about that, although I agree it's not *super*-valuable to have this (I was just picking off some issues to become familiar with the codebase).

Here's what I would probably want in a design at a few different breakpoints:




Wide screen:

```
--\----------\----------------------------------------------------------------
   \          \
/\  \   Type   \                     Child Type                            X
||  /          /
   /          /
--/----------/----------------------------------------------------------------
```


Medium screen:

```
-------\------------------------------------------------
        \
  Type   \              Child Type                   X
         /
        /
-------/------------------------------------------------
```

Tiny screen:

```
--------------------------------------------

  /          Child Type                   X
  \

--------------------------------------------
```

Hashtag: #NotADesigner

leebyron:
That's an interesting idea, but maybe less helpful if it's only available when you expand the explorer extra wide - I think something like that would be really useful for a full-screen doc explorer as something that's persistently available regardless of where you are in the docs that's like a link back to index.

For the smaller screen space of the in-tool doc explorer I'd be worried about [Fitt's Law](https://www.interaction-design.org/literature/article/fitts-s-law-the-importance-of-size-and-distance-in-ui-design) with the back button and home button necessarily close to each other.

I like the idea of dropping the previous type's name from the back button when the explorer is small though! Right now it truncates it with an ellipsis, but that gets pretty messy looking.

leebyron:
Also @wincent that's some pro ASCII art

wincent:
> I like the idea of dropping the previous type's name from the back button when the explorer is small though! Right now it truncates it with an ellipsis, but that gets pretty messy looking.

Ok, I'll upload something that does that:

graphiql

graphiql_and_yarn___glh-mbp___masochist___graphiql__tmux_

Issue #250, opened 2016-12-12
Use flexbox standard syntax


guzgarcia:
It seems you guys have forgotten to place `display: flex` in the css classes.

This is the look under Firefox 45.x with default styles:

![](https://i.imgur.com/6DclS6T.png)

After add `display: flex`:

![](https://i.imgur.com/tDsMLBn.png)

wincent:
Thanks for the report! I couldn't repro this on Firefox 49 or 50 (didn't have access to 45), so I downloaded v45 and I can repro.

I suspect most people allow the auto-updater to keep them on a more recent version, but v45 is still pretty recent (April 2016, I think), and the cost of fixing the issue seems low, so we may as well do it.

PR #252, opened 2016-12-13
Fix display issues on older Firefox

wincent:
Regression introduced in f9de00729998225a.

Tested on Firefox 45 (old, was broken), 49 (previous, still works), 50 (latest, still works) and Chrome 54 (current, still works).

Closes: https://github.com/graphql/graphiql/issues/250


asiandrummer:
👍  thanks for the fix!

Issue #133, opened 2016-04-29
Docs should display defaultValue of parameters


chibimagic:

murielsilveira:
I'll do it, PR soon :)

PR #272, opened 2017-01-13
Display defaultValue of parameters on the Docs

murielsilveira:
As requested on issue #133, this PR displays the default value of a parameter on the Docs sidebar.

About the tests, I was thinking about adding `enzyme` to test the components, or we should keep the `react-test-renderer` only?

That's how it looks, let me know what do you think or if I should do something different:

arguments
search
query

### DIFF
--- a/src/components/DocExplorer/Argument.js
+++ b/src/components/DocExplorer/Argument.js
@@ -15,11 +15,6 @@ export default class Argument extends React.Component {
     onClickType: PropTypes.func.isRequired,
   }
 
-  shouldComponentUpdate(nextProps) {
-    return this.props.arg !== nextProps.arg ||
-      this.props.onClickType !== nextProps.onClickType;
-  }
-
   render() {
     const arg = this.props.arg;
 
--- a/src/components/DocExplorer/Argument.js
+++ b/src/components/DocExplorer/Argument.js
@@ -16,8 +16,7 @@ export default class Argument extends React.Component {
   }
 
   shouldComponentUpdate(nextProps) {
-    return this.props.arg !== nextProps.arg ||
-      this.props.arg.name !== nextProps.arg.name ||
+    return this.props.arg.name !== nextProps.arg.name ||
       this.props.arg.type !== nextProps.arg.type ||
       this.props.arg.defaultValue !== nextProps.arg.defaultValue ||
       this.props.onClickType !== nextProps.onClickType;
--- a/src/components/DocExplorer/Argument.js
+++ b/src/components/DocExplorer/Argument.js
@@ -1,3 +1,11 @@
+/**
+ *  Copyright (c) Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE file in the root directory of this source tree.
+ */
+
 import React, { PropTypes } from 'react';
 import TypeLink from './TypeLink';
 
--- a/src/components/DocExplorer/Argument.js
+++ b/src/components/DocExplorer/Argument.js
@@ -9,35 +9,38 @@
 import React, { PropTypes } from 'react';
 import TypeLink from './TypeLink';
 
-const Argument = ({ arg, onClickType }) => {
-  return (
-    <span className="arg">
-      <span className="arg-name">{arg.name}</span>
-      {': '}
-      <TypeLink type={arg.type} onClick={onClickType} />
-      { renderDefaultValue(arg) }
-    </span>
-  );
-};
-
-function renderDefaultValue(arg) {
-  if (arg.defaultValue === undefined) {
-    return null;
+export default class Argument extends React.Component {
+  static propTypes = {
+    arg: PropTypes.object.isRequired,
+    onClickType: PropTypes.func.isRequired,
   }
 
-  return (
-    <span>
-      {' = '}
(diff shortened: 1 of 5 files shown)


src/components/DocExplorer/Argument.js
```diff
@@ -0,0 +1,35 @@
...
+    <span className="arg">
+      <span className="arg-name">{arg.name}</span>
+      {': '}
+      <TypeLink type={arg.type} onClick={onClickType} />
+      { renderDefaultValue(arg) }
+    </span>
+  );
+};
+
+function renderDefaultValue(arg) {
```
asiandrummer: Can't this just be:
```js
  {
    arg.defaultValue !== undefined &&
    <span>
      {' = '}
      <span className="arg-default-value">
        { arg.defaultValue === '' ? '""' : arg.defaultValue }
      </span>
    </span>
  }
```
within `render()` in `Argument` component?

src/components/DocExplorer/FieldDoc.js
```diff
@@ -10,6 +10,7 @@ import React, { PropTypes } from 'react';
 
 import MarkdownContent from './MarkdownContent';
 import TypeLink from './TypeLink';
+import Argument from './Argument';
```
asiandrummer: I'd appreciate if we could alphabetize by imported module names

murielsilveira: sure, done

src/components/DocExplorer/Argument.js
```diff
@@ -0,0 +1,35 @@
+import React, { PropTypes } from 'react';
+import TypeLink from './TypeLink';
+
+const Argument = ({ arg, onClickType }) => {
```
murielsilveira: I think that it could be done in any way, but it's important to maintain consistency through the project.

When I refactored the component to this state:

```js
export default class Argument extends React.Component {
  static propTypes = {
    arg: PropTypes.object.isRequired,
    onClickType: PropTypes.func.isRequired,
  }

  render() {
    const arg = this.props.arg;

    return (
      <span className="arg">
        <span className="arg-name">{arg.name}</span>
        {': '}
        <TypeLink type={arg.type} onClick={this.props.onClickType} />
        {
          arg.defaultValue !== undefined &&
          <span>
            {' = '}
            <span className="arg-default-value">
              { arg.defaultValue === '' ? '""' : arg.defaultValue }
            </span>
          </span>
        }
      </span>
    );
  }
}
```

the linter was showing that it should be done in a pure function.

screen shot 2017-01-17 at 14 30 43

But then I realize that I was probably missing the `shouldComponentUpdate` method. After that it was green.

src/components/DocExplorer/Argument.js
murielsilveira: yes, done :)

src/components/DocExplorer/Argument.js
```diff
@@ -0,0 +1,35 @@
+import React, { PropTypes } from 'react';
```
murielsilveira: done

murielsilveira:
I did all your suggestions, take a look :)

When the line gets too long it wraps to the next line:

screen shot 2017-01-17 at 14 38 38

And about the tests, I'll make another PR adding enzyme with some examples, than we can talk there about it :)

asiandrummer:
@murielsilveira hmm, I think it's good enough for the first pass, but I'm not sure I like how it looks when it gets wrapped to the next line... @leebyron would you have a better suggestion how to render this case?

The code lgtm - thanks for the contribution

src/components/DocExplorer/Argument.js
```diff
@@ -0,0 +1,46 @@
...
+import TypeLink from './TypeLink';
+
+export default class Argument extends React.Component {
+  static propTypes = {
+    arg: PropTypes.object.isRequired,
+    onClickType: PropTypes.func.isRequired,
+  }
+
+  shouldComponentUpdate(nextProps) {
+    return this.props.arg !== nextProps.arg ||
```
asiandrummer: Wait, wouldn't `this.props.arg !== nextProps.arg` check all properties? Aren't the below checks redundant?

leebyron:
I think we need a separate change to the docs renderer which renders arguments on separate lines when they don't fit well on one line. It's an existing issue that this change just accentuates. There are a number of ways we could approach that, and I'm not sure which may be best. We could attempt to be dynamic based on the width available or could wrap when there are more than one argument, or perhaps some other metric

src/components/DocExplorer/Argument.js
murielsilveira: It doesn't check all the properties. I used this clause to treat only the `undefined` to `some_object` case, but taking a closer look it's wrong anyway.

I dig a bit more, and I think we don't even need the `shouldComponentUpdate` method defined in this case, I was trying to simulate it and it's never called.

What do you think?

murielsilveira: I fixed the properties validation. I think that it's fine now. :)

leebyron:
I added a few updates to fix some minor issues, but this is great thanks for your hard work!

One thought for a follow up would be proper styling of the default value itself such that numbers, enums, strings, and complex values get different and more appropriate styles. However that's a worthy follow up.

PR #276, opened 2017-01-18
Add Jump-to-docs and Info tooltips

### LINKED ISSUE #97: Automatically navigate to current type in DocExplorer
I'd love to enable automatic documentation, so when I navigate within the query code, i.e. move the cursor in various places, the corresponding type is openend in DocExplorer. 

Could be enabled/disabled by some toggle/checkbox at the bottom of the DocExplorer panel...

leebyron:
This leverages https://github.com/graphql/codemirror-graphql/pull/96 to provide cmd+click to jump to the docs for tokens in the query and mouse-over tool tops to provide in context documentation as well.

Closes #97

### DIFF
--- a/css/info.css
+++ b/css/info.css
@@ -1,0 +1,111 @@
+.CodeMirror-info {
+  background: white;
+  border-radius: 2px;
+  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.45);
+  box-sizing: border-box;
+  color: #555;
+  font-family:
+    system,
+    -apple-system,
+    'San Francisco',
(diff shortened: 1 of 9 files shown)


css/info.css
```diff
@@ -0,0 +1,69 @@
+.CodeMirror-info {
```
asiandrummer: Just wondering if we should have CSS rules specific to `codemirror-graphql` in that repo, wdyt?

leebyron: I think packaging the css files across repos would get challenging

Issue #273, opened 2017-01-16
Display error or warning on deprecated fields inside query editor


stubailo:
Thankfully now we display information about deprecated fields in the docs sidebar, but most people just use autocompletion to write queries. It would be great if there were a way to see inline in the query when you are using a deprecated field so that you don't write new queries that use them.

asiandrummer:
Thanks for the suggestion @stubailo - @1aurabrown also pointed out that typeahead should also contain deprecation information. Deprecated GraphQL syntax is a special one since using it won't block the query execution - maybe a yellow squiggly line would suggest a warning for deprecation. For typeahead entries, showing `(DEPRECATED)` in the description box would be sufficient.

1aurabrown:
@asiandrummer thanks. the change to the ui requires a change to codemirror-graphql in order to pass those fields along with the hints in the typeahead.

1aurabrown:
@asiandrummer like a yellow squiggly underline? i'll try to do that, styling can easily be updated later.

PR #282, opened 2017-01-19
Show deprecation warnings in hint typeahead

leebyron:
This adds additional rendering in the hint typehead to indicate when a field hinted is deprecated. The styling matches that used in the info hover tooltips.

As mentioned in #273

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -417,7 +417,7 @@
 }
 
 .CodeMirror-hint-information .infoType {
-  color: #30a;
+  color: #CA9800;
   cursor: pointer;
   display: inline;
   margin-right: 0.5em;
--- a/src/utility/onHasCompletion.js
+++ b/src/utility/onHasCompletion.js
@@ -19,6 +19,7 @@ export default function onHasCompletion(cm, data, onHintInformationRender) {
 
   let wrapper;
   let information;
+  let deprecation;
 
   // When a hint result is selected, we touch the UI.
   CodeMirror.on(data, 'select', (ctx, el) => {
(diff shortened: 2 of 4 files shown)


leebyron:
screenshot 2017-01-19 13 21 21

css/app.css
```diff
@@ -417,10 +417,10 @@
 }
 
 .CodeMirror-hint-information .infoType {
-  color: #30a;
+  color: #CA9800;
   cursor: pointer;
   display: inline;
-  margin-right: 0.5em;
+  margin-right: 1ch;
```
wincent: One thing to consider: `ch` is not super well supported in IE (http://caniuse.com/#search=ch).

css/app.css
```diff
@@ -417,10 +417,10 @@
 }
 
 .CodeMirror-hint-information .infoType {
-  color: #30a;
+  color: #CA9800;
```
wincent: Could we use lowercase `a` for consistency with elsewhere? (I know we have a few mixed examples throughout; would be could to standardize, and add a lint to enforce.)

leebyron: Good idea to lint it - it's an easy one to get wrong

src/utility/onHasCompletion.js
```diff
@@ -82,7 +90,7 @@ export default function onHasCompletion(cm, data, onHintInformationRender) {
 
     // Now that the UI has been set up, add info to information.
     const description = ctx.description ?
-      marked(ctx.description, { smartypants: true }) :
+      marked(ctx.description, { sanitize: true }) :
```
wincent: No more "smart" quotes? I guess the curmudgeon in me is quite pleased by that.

wincent:
lg2m

src/utility/onHasCompletion.js
leebyron: Just making this consistent with everywhere else we use marked, in my experience smartypants has too many false-positive conversions which can look worse, and especially isn't great for things other than long-form content. sanitize is important to add though :)

css/app.css
leebyron: TIL. That's sad that even IE11 has this flaw

css/app.css
leebyron: Most of the rest of the file is uppercase hex colors, so I'll leave this for now - but we should definitely lint and unify around some standard

PR #285, opened 2017-01-20
Run introspection again if fetcher changes.

### LINKED ISSUE #223: [Feature] Switch to different schema/endpoint on the fly
I have an input field for a custom graphQL host/url endpoint and I would like the docs to update accordingly. I tried setting the schema passed as a prop to undefined when the host changes to trigger the fetcher to do the introspection request but nothing happens.

Could you advise me what is the expected behaviour and how could I achieve what I need ? (docs updating on host/url change)

Thanks

leebyron:
Closes #223

Right now, if a new fetching function is provided, GraphiQL doesn't refetch the schema with it. That can result in different schema running queries and used within the tool for insights.

This change causes the new fetcher to be called with introspection to get a new schema.

If GraphiQL is rerendered by the fetcher function is the same, then nothing will happen.

BREAKING: Note that if you provide a fetching function that is generated on every render, such as a non-autobound function in a parent React component, this will cause it to re-fetch the schema every time it's rerendered. To avoid this, it's important to ensure the fetching function changes only when intentionally using new behavior (such as a new endpoint to fetch from).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,6 +6,7 @@
   "plugins": [
     "syntax-async-functions",
     "transform-class-properties",
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "transform-regenerator"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "babel-plugin-syntax-async-functions": "6.13.0",
     "babel-plugin-transform-class-properties": "6.19.0",
     "babel-plugin-transform-object-rest-spread": "^6.20.2",
+    "babel-plugin-transform-regenerator": "^6.22.0",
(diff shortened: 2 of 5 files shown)


asiandrummer:
lgtm - one use case I was thinking is when `null` is passed in as a schema initially, but GraphiQL then receives a new fetcher function to fetch the schema. In this case the new props should remove the schema prop or set it `undefined`, and that might seem a little confusing if it were to be a truly common use case, which I'm not sure of. What do you think?

Other than that I love this!

leebyron:
If `null` is passed initially, then this bit of code won't run. We'll treat it as intentional lack of a schema, probably because you're loading it yourself.

Issue #302, opened 2017-01-26
GraphiQL Component Usage Example is incorrect


joelgriffith:
```          // Some other possible toolbar items
          <GraphiQL.Toolbar.Menu title="File">
            <GraphiQL.Toolbar.MenuItem title="Save" onClick={...}>
          <GraphiQL.Toolbar.Menu>
```
should actually be:

```
          // Some other possible toolbar items
          <GraphiQL.Menu title="File">
            <GraphiQL.MenuItem title="Save" onClick={...}>
          <GraphiQL.Menu>
```

wincent:
I think you're right. Thanks for pointing it out.

PR #303, opened 2017-01-26
Fix Component names in usage example

wincent:
Closes: https://github.com/graphql/graphiql/issues/302


wincent:
@joelgriffith, this fixes the issue you pointed out in #302. Additionally also looks like we had a closing tag that was malformed (`<GraphiQL...>` rather than `</GraphiQL...>`) so I fixed that too.

joelgriffith:
Thanks @wincent

PR #313, opened 2017-02-04
Reset docs to root when fetcher is updated. Closes #312

### LINKED ISSUE #312: Fetcher change doesn't rerender docs
After this PR: https://github.com/graphql/graphiql/pull/285, you can now change the `fetcher` and the introspection query refreshes correctly.

The issue is if you navigate into the docs at all (e.g. into Query or Mutation), and then update the fetcher, the introspection query runs but then the docs rerender the old schema and page that you were on.  If you navigate back in the docs then click into something again, it shows the correct schema.

I believe to fix this, the docs should be returned to the root page when the introspection query runs.

gdub22:
Adds a private `reset` method to `DocExplorer` which resets its `navStack`, returning it to the root doc page.

When `fetcher` is changed and introspection query re-runs, the docs need to be reset (#312)

### DIFF
--- a/src/components/DocExplorer.js
+++ b/src/components/DocExplorer.js
@@ -18,6 +18,11 @@ import SearchBox from './DocExplorer/SearchBox';
 import SearchResults from './DocExplorer/SearchResults';
 import TypeDoc from './DocExplorer/TypeDoc';
 
+const initialNav = {
+  name: 'Schema',
+  title: 'Documentation Explorer',
+};
+
 /**
  * DocExplorer
  *
@@ -41,11 +46,6 @@ export class DocExplorer extends React.Component {
 
   constructor() {
     super();
-
-    const initialNav = {
-      name: 'Schema',
-      title: 'Documentation Explorer',
-    };
 
     this.state = { navStack: [ initialNav ] };
   }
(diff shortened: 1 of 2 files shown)


wincent:
Thanks @gdub22!

I think this is a good start, although I am not sure it will actually address the issue, because the call to `_fetchSchema` returns immediately after enqueuing some async work, so your `setState` will reset the navigation before the introspection has actually run.

I'd suggest changing `_fetchSchema` to return a promise, so that you can do something like:

```
this._fetchSchema()
  .then(() => this.docExplorerComponent.reset());
```

instead of:

```
this._fetchSchema();
this.docExplorerComponent.reset();
```

I haven't looked into the question of whether merely resetting the nav stack is enough or whether some additional internal state would need to be blown away. Could you confirm that this actually fixes the underlying issue?

gdub22:
Thanks for having a look @wincent.

Whenever the introspection query is going to be re-fetched, the doc explorer nav stack should be reset right away.  I'd say it even makes more sense to move the `reset` call before the `fetchSchema` call. `fetchSchema` kicks off a loading state in the doc explorer UI anyway, so whether that succeeds or fails, it should always be back at the `initialNav` state.   I can change it your suggestion if that is what is desired, however.

This does fix the underlying issue, as I currently patch this in userland code by digging down to the `docExplorerComponent` and manually reset that private state when my fetcher is changed.

wincent:
> I can change it to your suggestion if that is what is desired, however.

Nah, I don't think that is necessary, but I do like your idea of moving the `reset` call up front. (If you do that, can you verify that it still fixes the issue?)

gdub22:
I swapped the execution order as suggested.  Confirm it still resolves the issue.

feb-14-2017 22-21-37

wincent:
Sweet. Thanks!

PR #315, opened 2017-02-06
Fix bug where the dropdowns are only scrollable inside the toolbar

AGS-:
githubgraphiqlscrollbug

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -61,7 +61,7 @@
 }
 
 .graphiql-container .toolbar {
-  overflow-x: auto;
+  overflow-x: visible;
 }
 
 .graphiql-container .docExplorerShow {


css/app.css
```diff
@@ -61,7 +61,7 @@
 }
 
 .graphiql-container .toolbar {
-  overflow-x: auto;
+  overflow-x: visible;
```
wincent: Seems reasonable, but isn't `overflow: visible` the default? (Not sure if it's cascading down here though, haven't checked).

AGS-: It's defaulting to scroll for some reason, so I think hard coding it would be the best option.

wincent: Sounds fair.

AGS-:
Thanks @wincent !

Issue #318, opened 2017-02-08
npm run build succeeds on Mac, fails on Windows (Windows_NT 6.3.9600)


mattferrin:
```
C:\Dev\graphiql>npm install && npm run build

> graphiql@0.9.2 prepublish C:\Dev\graphiql
> . ./resources/prepublish.sh

'.' is not recognized as an internal or external command,
operable program or batch file.
```

```
npm ERR! node v6.9.5
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! graphiql@0.9.2 prepublish: `. ./resources/prepublish.sh`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the graphiql@0.9.2 prepublish script '. ./resources/prepublish.sh'.
```

mattferrin:
I've changed the `.sh` files to Node `.js` files, but get the following error:

```
Uncaught ReferenceError: GraphiQL is not defined
    at (index):136
(anonymous) @ (index):136
```

```js
      ReactDOM.render(
        React.createElement(GraphiQL, {
          fetcher: graphQLFetcher,
          query: parameters.query,
          variables: parameters.variables,
          operationName: parameters.operationName,
          onEditQuery: onEditQuery,
          onEditVariables: onEditVariables,
          onEditOperationName: onEditOperationName
        }),
        document.getElementById('graphiql')
      );
```

wincent:
Thanks for bringing this up @mattferrin. We'd love to get the build working on Windows machines, but I personally don't have a machine to do any testing on. If you can identify the changes necessary to get it working there, we can fix this together.

> I've changed the `.sh` files to Node `.js` files, but get the following error:

Can you share the changes you made, and also what you did to get the error?

mattferrin:
The root issue is really that running any `.sh` file from the command prompt fails. runing Node `.js` scripts seems the right solution.

package.json
```json
  "scripts": {
    "build": "node ./resources/build.js",
    "check": "flow check",
    "dev": "babel-node test/server.js",
    "lint": "eslint src",
    "prepublish": "node ./resources/prepublish.js",
    "preversion": ". ./resources/checkgit.sh && npm test",
    "test": "npm run lint && npm run check && npm run testonly",
    "testonly": "babel-node ./node_modules/.bin/_mocha $npm_package_options_mocha"
  },
```

./resources/build.sh
```js
var exec = require('child_process').exec;

var stdout = function (command) {
[...]

PR #320, opened 2017-02-09
Change non-git shell script files to node files

mattferrin:
I think this is a good step in the right direction. I tested it on my Windows an Mac machines.

The most controversial change, I switched the example to run on port 8081 so that it is different from the port used in the main non-example.


example/server.js
```diff
@@ -15,4 +15,4 @@ const app = express();
 app.use(express.static(__dirname));
 app.use('/graphql', graphqlHTTP(() => ({ schema })));
 
-app.listen(8080, () => console.log('Started on http://localhost:8080/'));
+app.listen(8081, () => console.log('Started on http://localhost:8081/'));
```
wincent: As this is unrelated, should probably be pulled out into a separate commit.

In any case, I actually think we should use a port number of 0 here, which will cause the OS to assign a random port number. I've just sent a PR for that: https://github.com/graphql/graphiql/pull/321

wincent:
Thanks for taking the time to get things working on Windows, @mattferrin! I might add a comment or two in line, but here are some high level remarks to start with:

This is a good start. I think, however, that we should address my suggestions from https://github.com/graphql/graphiql/issues/318 before merging this. It seems quite wrong, for example, to be shelling out to the `echo` executable from node when we could instead just print to standard out with `process.stdout.write`. Likewise, there are "native" Node equivalents for much of the shell stuff we are doing in here (the `fs.*` methods I suggested in the other issue) and existing, mature abstractions to replace forking out to platform-specific commands  (for example, we could use `rimraf` instead of `rm -rf`/`rd`).

This isn't so much a "port to node" so much as a "writing bash in node", which is pretty painful to write, and even read (due to all the escaping). I'm happy to help out with doing an a node-centric rewrite if you're not up for it.

Thanks once again for your time on this!

resources/prepublish.js
```diff
@@ -0,0 +1,38 @@
...
+if (process.env.npm_config_argv) {
+  let npmConfigArgv = JSON.parse(process.env.npm_config_argv);
+  // # Because of a long-running npm issue (https://github.com/npm/npm/issues/3059)
+  // # prepublish runs after `npm install` and `npm pack`.
+  // # In order to only run prepublish before `npm publish`, we have to check argv.
+  if (npmConfigArgv.original[0] === 'publish') {
+    // # Publishing to NPM is currently supported by Travis CI, which ensures that all
+    // # tests pass first and the deployed module contains the correct file structure.
+    // # In order to prevent inadvertently circumventing this, we ensure that a CI
+    // # environment exists before continuing.
```
wincent: Can you remove this Bash comment syntax from inside your JavaScript comments?

example/server.js
mattferrin: Undone.

mattferrin:
I've removed some bash style comments because I could swear I was asked to, though I can no longer seem to find that request to comment that it's done. 

As for the port to Node file system functions, I do not have the time for it myself. My goal is to keep my code as close as possible to the existing and working solution and not break things.

wincent:
> I've removed some bash style comments because I could swear I was asked to, though I can no longer seem to find that request to comment that it's done.

[Hidden away here](https://github.com/graphql/graphiql/pull/320#discussion_r100392463).

Thanks for the updates!

resources/prepublish.js
mattferrin: Gotcha. Done :)

mattferrin:
FYI, I'm doing a little more work on this even though I said I didn't have time, and I don't! :)

mattferrin:
@wincent Your suggestion 

> Well, it is literally just concatenating the files, so could get the listing with fs.readdirSync, read the files with fs.readFileSync, concat in JS, then fs.writeFileSync the result.

wasn't implemented because I don't know how to do it while file globbing.

mattferrin:
@wincent Apologies. I'm blind. You explicitly state I can use `readdirSync`. I'll do that. Thanks.

mattferrin:
I think I've completed all the suggestions. If you would like I can soft reset all my commits and make it one commit, after we're done collaborating on this of course.

wincent:
> FYI, I'm doing a little more work on this even though I said I didn't have time, and I don't! :)

I know the feeling well.

> I think I've completed all the suggestions. If you would like I can soft reset all my commits and make it one commit, after we're done collaborating on this of course.

That won't be necessary. GitHub has a shiny "Squash and merge" button. I have reviewed the new changes yet, but will do so shortly. Thanks for taking this extra stab at it!

resources/build.js
```diff
@@ -0,0 +1,45 @@
...
+var isWindows = /^win/.test(process.platform)
+
+var out = function (command) {
+  execSync(command, {stdio:'inherit'});
+};
+
+console.log('"Running node ./resources/build.js"');
+
+if (!isWindows) {
+  out('set -e'); // exit on console error
```
wincent: I don't think this will do anything, will it?

`out` will fork a child process that lives for the duration of the `set` command, then exit, so the attempt to change the setting will have no subsequent effect.

I think we can delete this.

resources/build.js
```diff
@@ -0,0 +1,45 @@
...
+if (!fs.existsSync('node_modules/.bin')) {
+  console.log('"Be sure to run \`npm install\` before building GraphiQL."');
+}
+
+var distDir = './dist';
+rimraf.sync(distDir);
+fs.mkdirSync(distDir);
+
+out('babel src --ignore __tests__ --out-dir dist/');
+console.log('"Bundling graphiql.js..."');
```
wincent: Why the double quotes here?

resources/build.js
```diff
@@ -0,0 +1,45 @@
...
+out('browserify -g browserify-shim -s GraphiQL dist/index.js > graphiql.js');
+console.log('"Bundling graphiql.min.js..."');
```
wincent: And here (etc; I won't bother commenting on the other places).

resources/build.js
```diff
@@ -0,0 +1,45 @@
...
+if (isWindows) {
+  out('browserify -g browserify-shim -g uglifyify -s GraphiQL dist/index.js | uglifyjs -c --screw-ie8 > graphiql.min.js');
+} else {
+  out('browserify -g browserify-shim -g uglifyify -s GraphiQL dist/index.js 2> /dev/null | uglifyjs -c --screw-ie8 > graphiql.min.js 2> /dev/null');
```
wincent: I think we can do this without piping to the shell but using real pipes in a platform-agnostic way, so we won't need the `isWindows` check here.

resources/build.js
```diff
@@ -0,0 +1,45 @@
+var execSync = require('child_process').execSync;
```
wincent: Just a thought: you could use ES6 features in this file if you executed it with `babel-node` instead of `node`.

mattferrin: It's a good idea. I'll definitely do that on personal projects moving forward, but am unsure it's best here.

resources/build.js
mattferrin: They are accidental remnants from past echo commands. I can remove them, but noticed they seemed to help contrast between other console writes and made these easier to read even if the quotes lacked meaning. (There are JS filenames logged to console during the build without quotes.)

resources/build.js
mattferrin: Confirm you'd like me to remove them and the double quotes will be gone :)

resources/build.js
mattferrin: You're right. It might take a while for me to get around to doing this.

resources/build.js
mattferrin: Say to do it and I'm game though.

resources/build.js
mattferrin: I remember it halting program execution upon an error, but I need to find time to test and prove my memory is either correct or incorrect.

wincent: Sounds good. I would be extremely surprised if it has any effect.

resources/build.js
wincent: Yeah, I think you should.

resources/build.js
wincent: We have Babel in the project precisely so that we can use ES6 features, so doing so would be consistent with the other usage in the repo.

asiandrummer: Let's do `babel-node` for consistency, as @wincent suggested!

asiandrummer:
I don't feel qualified enough to review this as I have no prior experience working with Windows env - I'll trust @wincent and @mattferrin's judgements on merging this one.

mattferrin:
@asiandrummer It looks like I have mostly minor changes to make, but am unsure how to tackle the platform agnostic piping thing in Node off hand. 

So basically, I just haven't prioritized my time to include finishing this as of yet.

holm0563:
Not sure about mac but i can verify that this is the only way i can build on a windows machine.

AGS-:
Hey @mattferrin, sorry we let this PR go a little stale. Would you be willing to rebase your changes onto master? I do have both a Windows and OSX machine I can test the PR on. Let me know if you're still willing to work on this. Thanks for contributing!

mattferrin:
Hi @AGS-, my Windows machine was a prior company's computer. While it seems strange to actually say it, I don't have access to one. I know my code was functional, just slightly hacky. Being from a busy small team, didn't make sense to convert it all to pure Node. Sorry, but without a Windows machine and little incentive, I think I'm a dead end on this thread 🙂 Thanks for the renewed interest though.

AGS-:
Oh, that's unfortunate. Closing for now and hoping someone picks it up later on. Thanks for having taken the time to contribute!

Issue #169, opened 2016-09-19
Show inline field description in documentation explorer


OlegIlyenko:
At the moment fields look like this:

![graphiql-fields](https://cloud.githubusercontent.com/assets/156569/18650838/d000ecb4-7ec7-11e6-95cf-e32957d8f4cd.png)

In contrast, enum values, for instance, which have an inline documentation for all of the values:

![graphiql-enum-description](https://cloud.githubusercontent.com/assets/156569/18650855/ee726fa6-7ec7-11e6-93f8-60a31976c3c1.png)

I find it very helpful to see the documentation for all of the fields or enum values in one view, especially when I'm trying to learn new API.

So my suggestion is display descriptions inline for fields as well:

![graphiql-fields-description](https://cloud.githubusercontent.com/assets/156569/18650918/46c4ca28-7ec8-11e6-85bd-fb1f9be7ce34.png)

This raises a concern: I guess field descriptions tend to have much more information in them than an enum values. I can also imagine putting an example code snippet in the description. Maybe in this case we can truncate the description and show only a first paragraph or something along these lines?

I would be really interested to know your opinion on this :)

asiandrummer:
@OlegIlyenko - I think what you have in mind is great. We can also do an expand/collapse style of inlined descriptions, but showing the first paragraph (or a sentence or two really) if it gets too long sounds like a great start.

Something to keep in mind is that `DocExplorer` doesn't limit the number of fields to be shown in the `div`, so if a type defines some large number of fields then the browser might suffer. Obviously this issue exists even without the description texts, and we should probably limit the number of field definitions to be shown in the type documentation to a reasonable number, but this sounds like a whole another issue to track.

@leebyron and I were talking about something similar to this (in context of `deprecationReason`) - cc-ing since I'd love to get Lee's input as well ;)

hepin1989:
Could we have an button aside the field name,and follow with the first line of the description，after we click the `+` button,it show all the description,and display a `-`？

joelgriffith:
I have an MVP at this, simple expands a section when you hover over it (it's not very slick). Otherwise we have to keep track of expansion state which gets a little noisy:

![doc-expansion](https://cloud.githubusercontent.com/assets/1499985/20186505/2a754c6e-a725-11e6-97e7-5991dd00de36.gif)

joelgriffith:
Here's another pass, this one I prefer more over the last
[...]

PR #322, opened 2017-02-09
First-pass at inline-field descriptions in doc-explorer

joelgriffith:
Wubalubadubdub! First pass at the old inline-field descriptions. There is an edge-case though, since the doc-explorer has a `min-width` of 300px. So when the explorer is <300px, the ellipsis gets lost. I'd hate to alter that behavior as it can introduce regressions elsewhere:

inline-field-desc

Addresses: #169

### DIFF
--- a/css/doc-explorer.css
+++ b/css/doc-explorer.css
@@ -121,6 +121,14 @@
   color: #1F61A0;
 }
 
+.graphiql-container .field-short-description {
+  margin-left: 5px;
+  color: #999;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
+
 .graphiql-container .enum-value {
   color: #0B7FC7;
 }
--- a/css/doc-explorer.css
+++ b/css/doc-explorer.css
@@ -122,11 +122,11 @@
 }
 
 .graphiql-container .field-short-description {
-  margin-left: 5px;
   color: #999;
-  white-space: nowrap;
-  text-overflow: ellipsis;
+  margin-left: 5px;
   overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
 .graphiql-container .enum-value {
--- a/src/components/DocExplorer/TypeDoc.js
+++ b/src/components/DocExplorer/TypeDoc.js
@@ -197,6 +197,10 @@ function Field({ type, field, onClickType, onClickField }) {
       {': '}
       <TypeLink type={field.type} onClick={onClickType} />
       {
+        field.description &&
+        <p className="field-short-description">{field.description}</p>
+      }
+      {
         field.deprecationReason &&
         <MarkdownContent
           className="doc-deprecation"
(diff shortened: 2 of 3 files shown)


css/doc-explorer.css
```diff
@@ -121,6 +121,14 @@
   color: #1F61A0;
 }
 
+.graphiql-container .field-short-description {
+  margin-left: 5px;
+  color: #999;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
```
wincent: nit: would alphabetize this.

wincent:
Thanks!

This looks reasonable to me. I think that losing the ellipsis sub-300px is probably fine, as the UI already starts to truncate there intentionally anyway.

As this is a visual change, CC'ing @asiandrummer and @leebyron to take a look too. Code looks fine though.

asiandrummer:
@joelgriffith sorry we've been stale with this one! Reviewing now.

package.json
```diff
@@ -71,9 +71,9 @@
     "eslint-plugin-react": "6.9.0",
     "express": "4.14.1",
     "express-graphql": "0.6.3",
-    "flow-bin": "0.38.0",
+    "flow-bin": "0.39.0",
```
asiandrummer: nit: We could probably deal with this in a separate PR for clarity

src/components/DocExplorer/TypeDoc.js
```diff
@@ -197,6 +197,10 @@ function Field({ type, field, onClickType, onClickField }) {
       {': '}
       <TypeLink type={field.type} onClick={onClickType} />
       {
+        field.description &&
+        <p className="field-short-description">{field.description}</p>
```
asiandrummer: Should it be markdown-compatible?

test/README.md
```diff
@@ -6,4 +6,4 @@ and offers watching/compiling of the `src` folder. To utilize this, simply:
 
 1. Run `npm install`
 2. Run `npm run dev`
-3. Open your browser to [http://localhost:8080/]()
+3. Open your browser to the address listed in your console. e.g. `Started on http://localhost:49811/`
```
asiandrummer: nit: same as above

package.json
joelgriffith: Yeah, not sure how this leaked through, I'll remove it.

joelgriffith:
@asiandrummer I just got done playing with this feature and using Markdown content... and supporting markdown for this feature makes it rather difficult as markdown supports things like ul's, ol's, and blockquotes, which hose the formatting and nature of this feature (just a short snippet of the description).

I'd almost prefer doing something like `deburr` to remove the special markdown characters and show a small preview of the description, and let the actual description page render the markdown since it's difficult to condense inline.

Thoughts?

asiandrummer:
@joelgriffith that sounds like a good UX decision to me. We can move on with a small preview for now and have the markdown support if we decide later it's a good addition.

joelgriffith:
Sounds good, then I think this is ready for prime-time!

Issue #129, opened 2016-04-26
Provide a `dark theme` for GraphiQL


asiandrummer:
Probably a low-priority item (wishlist even), but a dark UI theme has been requested - I'm documenting as an issue here.

leebyron:
If someone wants to tackle this, it should be as simple as writing a new css file to include - but this isn't something we're going to prioritize on our team.

brene:
@asiandrummer Are you interested in doing this? If so, I will help you out ;-)

joelgriffith:
I've written a custom theme for my own implementation elsewhere. It's pretty easy to override the current theme via stronger css selectors. Would it be better to expose an interface via `props` on the root component?

I think documenting _how_ to override the current theme might suffice in place of widening an API surface area.

asiandrummer:
@brene - This continues to be a low priority item on our list of feature requests, so I'm not sure if our team could tackle this in a near future. I'd love to review and help merging this if you'd like to give a hand 😄 

@joelgriffith - Documentation for how to customize would be appreciated, but since I think some users of GraphiQL aren't as familiar with css rules, they would also like to have a stock-version of dark theme if it could be as easy as doing `<GraphiQL theme="dark" />` or something similar to this.

joelgriffith:
I can likely implement something like this then (and add docs on how to implement your own theme).

joelgriffith:
https://github.com/graphql/graphiql/pull/199 exposes a `theme` prop where consumers can use any of the themes found in CodeMirror (https://codemirror.net/demo/theme.html).

joelgriffith:
Spilling over from my initial PR for this: there's _so_ many color definitions and they're redefined nearly 2x throughout the css. This makes it hard for folks to bolt on their themes (especially when we don't expose setting the CodeMirror theme underneath).

I feel that if we want to accomplish this for future cases, we need to:

- Consolidate color definitions so things _aren't_ redefined all over the place, forcing consumers to overwrite numerous definitions.
- Potentially merge similar colors. There's quite a few similar variants.
- Expose 2 top-level props: `classNames` and `syntaxTheme`. One for folks to use to override css via specificity, the other to apply other CodeMirror themes.
- BONUS: source all color definitions in a css-preprocessor. Consumers can then use _those_ definitions and do all the fancy-pants `lighten` `mix` or what-have-you's to get the theme their ❤️ so desires.

Thoughts? @leebyron @asiandrummer

[...]

PR #323, opened 2017-02-09
Feature/code mirror theme exposed

joelgriffith:
First pass on exposing a `codeMirrorTheme` prop, plumbing it through, and docs on how to load/use css for the theme.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -9,7 +9,6 @@
  /* eslint-disable no-console */
 import express from 'express';
 import path from 'path';
-import fs from 'fs';
 import browserify from 'browserify';
 import browserifyShim from 'browserify-shim';
 import watchify from 'watchify';
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ GraphiQL supports customization in UI and behavior by accepting React props and 
 
 - `getDefaultFieldNames`: an optional function used to provide default fields to non-leaf fields which invalidly lack a selection set. Accepts a GraphQLType instance and returns an array of field names. If not provided, a default behavior will be used.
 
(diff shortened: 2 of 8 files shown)


joelgriffith:
See discussion in https://github.com/graphql/graphiql/issues/129

test/server.js
```diff
@@ -9,7 +9,6 @@
  /* eslint-disable no-console */
 import express from 'express';
 import path from 'path';
-import fs from 'fs';
```
wincent: Naughty! Sneaking in an unrelated change.

wincent:
Oh, forgot to mention: I applied some small fixes directly on the PR branch rather than incurring another round trip. If you do more work on this, be sure to pull those down.

joelgriffith:
> The fact that the editor views are powered by CodeMirror is really just an implementation detail, and using codeMirror in the prop name leaks that out.

This is a great point, and something that I thought about in some capacity. Here's my thoughts:

- `theme` is a bit broad in my opinion, as new adopters will likely assume that this theme would apply to the entirety of GraphiQL (which isn't the case).
- Minor, but I think that at this point GraphiQL is pretty tied to the CodeMirror implementation (there's even sub-packages that were written to marry GraphiQL to CodeMirror). Not saying that won't change, but would be non-trivial work.

I think maybe `syntaxTheme` would suffice since it succinctly addresses the fact that this only applies to code portions of the UI, and is not a global UI-theme.

Would that be ok @wincent?

wincent:
Or maybe `editorTheme`?

Agreed that for all practical purposes CodeMirror is likely to remain the implementation, but that doesn't mean we shouldn't try to practice good hygiene and abstract over it.

joelgriffith:
:boom: `editorTheme` is great. I'll update my PR tomorrow.

wincent:
Looks good. Thanks for your work on this!

joelgriffith:
Thanks for catching that JSX error....again!

PR #326, opened 2017-02-12
Update address step in test/README.md

AGS-:
Kind of a trivial PR but I was working on something else and noticed that it now randomizes the port number.


wincent:
Thanks for noticing. I randomized the ports but didn't remember to check the README.

PR #330, opened 2017-02-15
Menu doesnt stay open on click

AGS-:
Made select and open have more typical behavior.

### DIFF
--- a/src/components/ToolbarMenu.js
+++ b/src/components/ToolbarMenu.js
@@ -22,13 +22,11 @@ export class ToolbarMenu extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { visibility: 'hidden' };
+    this.state = { visible: false };
   }
 
   render() {
-    const ulStyle = {
-      visibility: this.state.visibility
-    };
+    const visible = this.state.visible;
     return (
       <a
         className="toolbar-menu toolbar-button"
@@ -40,7 +38,7 @@ export class ToolbarMenu extends React.Component {
         <svg width="14" height="8">
           <path fill="#666" d="M 5 1.5 L 14 1.5 L 9.5 7 z" />
         </svg>
-        <ul className="toolbar-menu-items" style={ulStyle}>
+        <ul className={'toolbar-menu-items' + (visible ? ' open' : '')}>
           {this.props.children}
         </ul>
       </a>
@@ -50,13 +48,14 @@ export class ToolbarMenu extends React.Component {
   handleClick(e) {
     if (this._node !== e.target) {
       e.preventDefault();
-      this.setState({ visibility: 'hidden' });
+      this.setState({ visible: false });
+      document.removeEventListener('click', this.handleClick.bind(this));
     }
   }
 
   handleOpen = e => {
     e.preventDefault();
-    this.setState({ visibility: 'visible' });
+    this.setState({ visible: true });
     document.addEventListener('click', this.handleClick.bind(this));
   };
 }
--- a/src/components/ToolbarMenu.js
+++ b/src/components/ToolbarMenu.js
@@ -47,14 +47,14 @@ export class ToolbarMenu extends React.Component {
 
   handleClick(e) {
     if (this._node !== e.target) {
-      e.preventDefault();
+      preventDefault(e);
       this.setState({ visible: false });
       document.removeEventListener('click', this.handleClick.bind(this));
     }
   }
 
   handleOpen = e => {
-    e.preventDefault();
+    preventDefault(e);
     this.setState({ visible: true });
     document.addEventListener('click', this.handleClick.bind(this));
   };
--- a/src/components/ToolbarMenu.js
+++ b/src/components/ToolbarMenu.js
@@ -14,54 +14,57 @@ import React, { PropTypes } from 'react';
  *
  * A menu style button to use within the Toolbar.
  */
- export class ToolbarMenu extends React.Component {
-   static propTypes = {
-     title: PropTypes.string,
-     label: PropTypes.string,
-   }
+export class ToolbarMenu extends React.Component {
+  static propTypes = {
+    title: PropTypes.string,
+    label: PropTypes.string,
+  }
 
-   constructor(props) {
-     super(props);
-     this.state = { visibility: 'hidden' };
-   }
+  constructor(props) {
+    super(props);
+    this.state = { visibility: 'hidden' };
+  }
 
-   render() {
-     const ulStyle = {
-       visibility: this.state.visibility
-     }
-     return (
-       <a
-         className="toolbar-menu toolbar-button"
-         onClick={this.handleOpen.bind(this)}
-         onMouseDown={preventDefault}
-         ref={() => {
-           document.addEventListener('click', this.handleClick.bind(this), false);
-         }}
-         title={this.props.title}>
-         {this.props.label}
-         <svg width="14" height="8">
-           <path fill="#666" d="M 5 1.5 L 14 1.5 L 9.5 7 z" />
-         </svg>
-         <ul className="toolbar-menu-items" style={ulStyle}>
-           {this.props.children}
-         </ul>
-       </a>
-     );
-   }
+  render() {
+    const ulStyle = {
+      visibility: this.state.visibility
+    };
+    return (
+      <a
+        className="toolbar-menu toolbar-button"
+        onClick={this.handleOpen.bind(this)}
+        onMouseDown={preventDefault}
+        ref={() => {
+          document.addEventListener('click',
+            this.handleClick.bind(this),
+            false
+          );
+        }}
+        title={this.props.title}>
+        {this.props.label}
+        <svg width="14" height="8">
+          <path fill="#666" d="M 5 1.5 L 14 1.5 L 9.5 7 z" />
+        </svg>
+        <ul className="toolbar-menu-items" style={ulStyle}>
+          {this.props.children}
+        </ul>
+      </a>
+    );
+  }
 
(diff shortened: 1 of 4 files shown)


AGS-:
Fixing errors brought up by travis

AGS-:
@wincent I was trying to pass in the DOM element by using ref however I couldn't figure out a good way to do it since I also need to pass in the event via `this.handleClick.bind(this)` to do the comparison. Any way I could pass both?

src/components/ToolbarMenu.js
```diff
@@ -14,54 +14,57 @@ import React, { PropTypes } from 'react';
...
+      <a
+        className="toolbar-menu toolbar-button"
+        onClick={this.handleOpen.bind(this)}
+        onMouseDown={preventDefault}
+        ref={() => {
+          document.addEventListener('click',
+            this.handleClick.bind(this),
+            false
+          );
+        }}
```
wincent: This isn't exactly how `ref` props are supposed to be used. The pattern is:

```
// In `render()`:
ref={node => this._node = node}
```

And then elsewhere (like in `componentDidMount()` or an event handler), access `this._node` instead of using `ReactDOM.findDOMNode`.

So your `document.addEventListener` should go in `handleOpen`, and your `handleClick` should be able to do an `if (this._node !== e.target)` or similar.

src/components/ToolbarMenu.js
```diff
@@ -14,54 +14,57 @@ import React, { PropTypes } from 'react';
...
+  }
 
-   handleClick(e) {
-     if (ReactDOM.findDOMNode(this) !== e.target) {
-       e.preventDefault();
-       this.setState({ visibility: 'hidden' });
-     }
-   }
+  handleClick(e) {
+    // eslint-disable-next-line no-undef
```
wincent: You won't need this suppression if you don't use `ReactDOM` (and if you did have to use it, I think the right way to deal with the lint would be to `import` it).

src/components/ToolbarSelect.js
```diff
@@ -29,25 +30,31 @@ export class ToolbarSelect extends React.Component {
...
     return (
       <a
         className="toolbar-select toolbar-button"
         onClick={this.handleOpen.bind(this)}
         onMouseDown={preventDefault}
         ref={() => {
-          document.addEventListener('click', this.handleClick.bind(this), false);
+          document.addEventListener('click',
+          this.handleClick.bind(this),
+          false);
```
wincent: Similar issue here; I don't think this is the pattern we want to follow.

src/components/ToolbarSelect.js
```diff
@@ -63,6 +70,7 @@ export class ToolbarSelect extends React.Component {
   }
 
   handleClick(e) {
+    // eslint-disable-next-line no-undef
```
wincent: As above.

wincent:
Awesome. Thanks for fixing this. I left a comment or four inline but I think this is looking close. Thanks!

src/components/ToolbarMenu.js
AGS-: I agree, that's a better pattern.

src/components/ToolbarMenu.js
AGS-: Haha yeah I felt bad about throwing the suppression in there

AGS-:
@wincent Fixed!

src/components/ToolbarMenu.js
```diff
@@ -14,27 +14,52 @@ import React, { PropTypes } from 'react';
...
+
+  render() {
+    const ulStyle = {
+      visibility: this.state.visibility
+    };
+    return (
+      <a
+        className="toolbar-menu toolbar-button"
+        onClick={this.handleOpen.bind(this)}
+        onMouseDown={preventDefault}
```
wincent: (Non-blocking feedback.)

I think we should either just kill the `preventDefault` function, or use it everywhere. It's a bit of a smell to have two ways of preventing a default.

src/components/ToolbarMenu.js
```diff
@@ -14,27 +14,52 @@ import React, { PropTypes } from 'react';
...
+    if (this._node !== e.target) {
+      e.preventDefault();
+      this.setState({ visibility: 'hidden' });
+    }
+  }
+
+  handleOpen = e => {
+    e.preventDefault();
+    this.setState({ visibility: 'visible' });
+    document.addEventListener('click', this.handleClick.bind(this));
```
wincent: For hygiene, we should make sure we remove this event listener on unmount or close. The currently implementation will re-bind repeatedly.

src/components/ToolbarSelect.js
```diff
@@ -14,42 +14,71 @@ import React, { PropTypes } from 'react';
...
+    (child, i) => {
+      if (!selectedChild || child.props.selected) {
+        selectedChild = child;
+      }
+      const onChildSelect =
       child.props.onSelect ||
-      onSelect && onSelect.bind(null, child.props.value, i);
-    return <ToolbarSelectOption {...child.props} onSelect={onChildSelect} />;
-  });
+      this.props.onSelect && this.props.onSelect.bind(null,
```
wincent: Indentation here looks funky; line continuations should be indented 2 spaces.

src/components/ToolbarSelect.js
```diff
@@ -14,42 +14,71 @@ import React, { PropTypes } from 'react';
...
+  }
 
-ToolbarSelect.propTypes = {
-  onSelect: PropTypes.func,
-  title: PropTypes.string,
-};
+  handleOpen = e => {
+    e.preventDefault();
+    this.setState({ visibility: 'visible' });
+    document.addEventListener('click', this.handleClick.bind(this));
```
wincent: Same issue with unmount here. Should always have a `removeEventListener` to balance every `addEventListener` call.

src/components/ToolbarMenu.js
```diff
@@ -14,27 +14,52 @@ import React, { PropTypes } from 'react';
...
-  label: PropTypes.string,
-};
+  constructor(props) {
+    super(props);
+    this.state = { visibility: 'hidden' };
+  }
+
+  render() {
+    const ulStyle = {
+      visibility: this.state.visibility
```
wincent: (Non-blocking feedback.)

I probably would have gone with a class here. It would result in a smaller change to the CSS file, and would mean we'd be more consistently using one method (predominantly class selectors) of applying presentation throughout the app.

wincent:
Left a few small comments but this is very close now. Would be good to fix those, then get this in; we can always iterate more on it later if we need.

src/components/ToolbarMenu.js
AGS-: Alright, yeah I'll change this to use a css class instead.

src/components/ToolbarSelect.js
AGS-: Whoops!

AGS-:
Alright, I'll update this PR in a bit.

AGS-:
Alright, I made it use classname instead of style like we do in ToolbarButton. I also have it remove eventListeners now.

AGS-:
graphiqlselectmenu

src/components/ToolbarMenu.js
AGS-: Ah forgot to put this in, let me do another commit really quick

src/components/ToolbarMenu.js
```diff
@@ -40,7 +38,7 @@ export class ToolbarMenu extends React.Component {
         <svg width="14" height="8">
           <path fill="#666" d="M 5 1.5 L 14 1.5 L 9.5 7 z" />
         </svg>
-        <ul className="toolbar-menu-items" style={ulStyle}>
+        <ul className={'toolbar-menu-items' + (visible ? ' open' : '')}>
```
wincent: Could use the `classnames` package for this, but I think it's good to keep the dependency footprint small, so this is good.

src/components/ToolbarMenu.js
```diff
@@ -50,13 +48,14 @@ export class ToolbarMenu extends React.Component {
   handleClick(e) {
     if (this._node !== e.target) {
       e.preventDefault();
-      this.setState({ visibility: 'hidden' });
+      this.setState({ visible: false });
+      document.removeEventListener('click', this.handleClick.bind(this));
```
wincent: Don't forget to also remove it, if present, in `componentWillUnmount`.

wincent:
I think we're good to land this after the `componentWillUnmount` change that I commented on inline.

src/components/ToolbarMenu.js
```diff
@@ -25,6 +25,10 @@ export class ToolbarMenu extends React.Component {
     this.state = { visible: false };
   }
 
+  componentWillUnmount() {
+    document.removeEventListener('click', this.handleClick.bind(this));
```
wincent: I am not sure that any of these `removeEventListener` calls will do what you want, because when you call `bind` you get a new function that is not `===`, which means that you won't remove anything here. (I could be wrong about that, but that's what I think will happen.)

I think the way to do this is as follows:

```
_subscribe() {
  if (!this._listener) {
    this._listener = this.handleClick.bind(this);
    document.addEventListener('click', this._listener);
  }
}

_release() {
  if (this._listener) {
    document.removeEventListener('click', this._listener);
    this._listener = null;
  }
}
```

And then call those methods.

AGS-: Ok, makes sense. Yeah I was afraid of that but I wasn't really sure. Implemented _subscribe and _release and tested that everything works.

wincent:
Sweet!

PR #336, opened 2017-02-18
Make Toolbar Tabbable

AGS-:
This is still far from done but I wanted to do a quick sanity check to make sure the approach I'm taking is what we want. So far it's possible to tab and open items in the toolbar however it's still not possible to tab through their children.

### DIFF
--- a/src/components/ExecuteButton.js
+++ b/src/components/ExecuteButton.js
@@ -74,12 +74,15 @@ export class ExecuteButton extends React.Component {
       <path d="M 11 9 L 24 16 L 11 23 z" />;
 
     return (
-      <div className="execute-button-wrap">
+      <div
+        className="execute-button-wrap"
+        tabIndex="-1">
         <button
           type="button"
           className="execute-button"
           onMouseDown={onMouseDown}
           onClick={onClick}
+          tabIndex="0"
           title="Execute Query (Ctrl-Enter)">
           <svg width="34" height="34">{pathJSX}</svg>
         </button>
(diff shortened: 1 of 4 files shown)


src/components/ExecuteButton.js
```diff
@@ -74,12 +74,15 @@ export class ExecuteButton extends React.Component {
       <path d="M 11 9 L 24 16 L 11 23 z" />;
 
     return (
-      <div className="execute-button-wrap">
+      <div
+        className="execute-button-wrap"
+        tabIndex="-1">
```
asiandrummer: What's this for?

AGS-: I put that there because since it's a surrounding div I think it's supposed to be marked as non tabbable with a -1? I may be totally wrong though.

src/components/ToolbarMenu.js
```diff
@@ -71,11 +79,17 @@ export class ToolbarMenu extends React.Component {
...
+    preventDefault(e);
+    this.setState({ visible: true });
+  }
+
   handleOpen = e => {
     preventDefault(e);
     this.setState({ visible: true });
     this._subscribe();
   };
+
```
asiandrummer: nit: lets get rid of this

AGS-: Yeah for sure, I'll include that in the next commit I make sometime in the next few days when I figure out how to have menus not close when I try to go through their items heh.

AGS-:
Still going to work on this soon™, just haven't had time.

PR #340, opened 2017-02-21
feat(subscriptions): added graphql-subscriptions support

Urigo:
Adding subscriptions support with the [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) library.

Replaced dummy subscription with timeSubscription and publish it every second.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
   "dependencies": {
     "codemirror": "5.23.0",
     "codemirror-graphql": "^0.6.3",
-    "marked": "0.3.6"
+    "marked": "0.3.6",
+    "subscriptions-transport-ws": "^0.5.2"
   },
   "peerDependencies": {
     "graphql": "^0.6.0 || ^0.7.0 || ^0.8.0-b || ^0.9.0",
--- a/test/index.html
+++ b/test/index.html
@@ -25,6 +25,7 @@
     <script src="./vendor/fetch.min.js"></script>
     <script src="./vendor/react-15.4.2.js"></script>
     <script src="./vendor/react-dom-15.4.2.js"></script>
+    <script src="./vendor/subscriptions-transport-ws-0.5.2.js"></script>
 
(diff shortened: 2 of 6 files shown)


asiandrummer:
Hi @Urigo - I'm not sure if `graphiql` repo is the right place for this. What is the intention of this addition?

leebyron:
I agree with @asiandrummer - If `subscriptions-transport-ws` wishes to include test coverage that it works with GraphiQL, then you should add this sort of test to that repo, but I don't think it's appropriate for GraphiQL's test cases to be subscription-specific when the large majority of usage is not subscriptions and subscriptions are still in the early phase of RFC where changes are possible. 

I'd prefer that we don't introduce volatility into GraphiQL's test harness until subscriptions are supported in a more first-class way, and that when we do so we don't upset the existing test coverage.

Urigo:
@asiandrummer @leebyron sorry got the late response.
I totally agree that while we are in RFC process this should not be merged.
Also, I agree about making the transport not subscriptions specific and we are working on that (updates soon).

Maybe we could leave this PR open as kind of RFC like thing for GraphiQL and subscriptions?
While working on that a lot of questions came up, also in simple terms of the right UI for subscriptions (When new results come - should they override the previous ones or append? , when we fire new subscription request, should we close the previous?, UI indicator for open connection? etc...)

Anyway for people who want to use GraphiQL with subscriptions today, I wrote a blog post about it - https://dev-blog.apollodata.com/how-to-use-subscriptions-in-graphiql-1d6ab8dbd74b
and I will keep improving and iterating on that.

asiandrummer:
cc @robzhu for context

AGS-:
Hey @Urigo, this seems to have been sitting here for a while and since we haven't really been able to come to a consensus on this, and I'm not sure what direction we're taking with this, I'm going to close this for now. Thanks for the contribution.

PR #343, opened 2017-02-23
fix to properly check the index

asiandrummer:
`prevIndex` should be compared to `-1`, not to the falseyness.


leebyron:
goodcatch

PR #344, opened 2017-02-23
handle the race condition in updating operationName

asiandrummer:
If a new query string is passed in (and `componentWillReceiveProps` is called) without manually editing query (e.g. programmatically modifying the content within the editor without key events), it's possible to have both `_updateQueryFacts` and `componentWillReceiveProps` trying to update `operationName` by calling the `setState` method, most often times the one in `componentWillReceiveProps` winning incorrectly.

The fix is to treat the outcome from `_updateQueryFacts` as truth and properly updating what to update during the component lifecycle.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -167,7 +167,18 @@ export class GraphiQL extends React.Component {
     if (nextSchema !== this.state.schema ||
         nextQuery !== this.state.query ||
         nextOperationName !== this.state.operationName) {
-      this._updateQueryFacts(nextQuery);
+      const updatedQueryAttributes = this._updateQueryFacts(
+        nextQuery,
+        nextOperationName,
+        this.state.operations,
+        nextSchema,
+      );
+
+      if (updatedQueryAttributes !== undefined) {
+        nextOperationName = updatedQueryAttributes.operationName;
+
+        this.setState(updatedQueryAttributes);
+      }
     }
 
     // If schema is not supplied via props and the fetcher changed, then
@@ -650,7 +661,14 @@ export class GraphiQL extends React.Component {
 
   handleEditQuery = debounce(100, value => {
     if (this.state.schema) {
-      this._updateQueryFacts(value);
+      this.setState(
+        this._updateQueryFacts(
+          value,
+          this.state.operationName,
+          this.state.operations,
+          this.state.schema,
+        ),
+      );
     }
     this.setState({ query: value });
     if (this.props.onEditQuery) {
@@ -658,26 +676,26 @@ export class GraphiQL extends React.Component {
     }
   })
 
-  _updateQueryFacts = query => {
-    const queryFacts = getQueryFacts(this.state.schema, query);
+  _updateQueryFacts = (query, operationName, operations, schema) => {
+    const queryFacts = getQueryFacts(schema, query);
     if (queryFacts) {
       // Update operation name should any query names change.
-      const operationName = getSelectedOperationName(
-        this.state.operations,
-        this.state.operationName,
+      const updatedOperationName = getSelectedOperationName(
+        operations,
+        operationName,
         queryFacts.operations
       );
 
       // Report changing of operationName if it changed.
       const onEditOperationName = this.props.onEditOperationName;
-      if (onEditOperationName && operationName !== this.state.operationName) {
-        onEditOperationName(operationName);
+      if (onEditOperationName && operationName !== updatedOperationName) {
+        onEditOperationName(updatedOperationName);
       }
 
-      this.setState({
-        operationName,
+      return {
+        operationName: updatedOperationName,
         ...queryFacts
-      });
+      };
     }
   }
 


src/components/GraphiQL.js
```diff
@@ -678,7 +689,11 @@ export class GraphiQL extends React.Component {
         operationName,
         ...queryFacts
       });
+      
+      return operationName;
     }
+    
+    return null;
```
leebyron: This line isn't necessary, by default returning undefined will behave as expected

src/components/GraphiQL.js
```diff
@@ -649,10 +654,16 @@ export class GraphiQL extends React.Component {
   }
 
   handleEditQuery = debounce(100, value => {
+    let operationName = this.state.operationName;
     if (this.state.schema) {
-      this._updateQueryFacts(value);
+      const newOperationName = this._updateQueryFacts(value);
+      // If `operationName` is updated as a result from updating query facts,
+      // treat it as a source of truth and update here accordingly.
+      if (newOperationName && newOperationName !== operationName) {
```
leebyron: This is repeating the logic on lines 684 below - that seems like this could introduce *more* possibilities for race conditions or split logic

src/components/GraphiQL.js
```diff
@@ -678,7 +689,11 @@ export class GraphiQL extends React.Component {
         operationName,
         ...queryFacts
       });
+      
+      return operationName;
```
leebyron: What if instead of calling `setState` directly, this function *returned* the object. That way you could replace calls to `this._updateQueryFacts(query)` with `this.setState(this._updateQueryFacts(query))`, however if you expect that multiple setState calls will occur in a row, then you can merge them before calling setState.

Another root issue seems to be that this function uses `this.state` as the source of truth, however clearly there are cases where `operationName` should not be read from `this.state` and instead should be passed into the function. I would suggest accepting a 2nd argument for operationName and using that instead of `this.state. operationName` within this function.

src/components/GraphiQL.js
```diff
@@ -167,7 +167,12 @@ export class GraphiQL extends React.Component {
     if (nextSchema !== this.state.schema ||
         nextQuery !== this.state.query ||
         nextOperationName !== this.state.operationName) {
-      this._updateQueryFacts(nextQuery);
+      const newOperationName = this._updateQueryFacts(nextQuery);
```
leebyron: Notice that schema can also differ from what is in `this.state` which will also cause a race condition. `schema` should also be provided as an argument to `_updateQueryFacts` to avoid this.

asiandrummer:
@leebyron done!

src/components/GraphiQL.js
```diff
@@ -167,7 +167,17 @@ export class GraphiQL extends React.Component {
...
+      const updatedQueryAttributes = this._updateQueryFacts(
+        nextQuery,
+        nextOperationName,
+        nextSchema,
+      );
+
+      if (updatedQueryAttributes !== undefined) {
+        nextOperationName = updatedQueryAttributes.operationName;
+
+        this.setState({ ...updatedQueryAttributes.queryFacts });
```
leebyron: shouldn't this be `this.setState(updatedQueryAttributes)`? `.queryFacts` isn't a property set anywhere, the variable by that name is spread into the return value

src/components/GraphiQL.js
```diff
@@ -650,34 +660,37 @@ export class GraphiQL extends React.Component {
...
 
-  _updateQueryFacts = query => {
-    const queryFacts = getQueryFacts(this.state.schema, query);
+  _updateQueryFacts = (query, operationName, schema) => {
+    const queryFacts = getQueryFacts(schema, query);
     if (queryFacts) {
       // Update operation name should any query names change.
-      const operationName = getSelectedOperationName(
+      const updatedOperationName = getSelectedOperationName(
         this.state.operations,
```
leebyron: This line should be `queryFacts.operations` shouldn't it... Relying on `this.state.operations` assumes the facts from the previous query.

src/components/GraphiQL.js
```diff
@@ -650,34 +660,37 @@ export class GraphiQL extends React.Component {
...
       const onEditOperationName = this.props.onEditOperationName;
-      if (onEditOperationName && operationName !== this.state.operationName) {
-        onEditOperationName(operationName);
+      if (onEditOperationName && operationName !== updatedOperationName) {
+        onEditOperationName(updatedOperationName);
       }
 
-      this.setState({
-        operationName,
-        ...queryFacts
```
asiandrummer: > shouldn't this be this.setState(updatedQueryAttributes)?

We were destructuring that object in here - we could probably be more specific with what we're setting the state with.

leebyron: Yes exactly. Since it's spread in here, `updatedQueryAttributes.queryFacts` above won't exist, `updatedQueryAttributes` will contain the data spread in from the `queryFacts` var here

asiandrummer: oi I see what you mean - sorry! Changed.

src/components/GraphiQL.js
```diff
@@ -650,34 +661,41 @@ export class GraphiQL extends React.Component {
...
-    const queryFacts = getQueryFacts(this.state.schema, query);
+  _updateQueryFacts = (query, operationName, operations, schema) => {
+    const queryFacts = getQueryFacts(schema, query);
     if (queryFacts) {
       // Update operation name should any query names change.
-      const operationName = getSelectedOperationName(
-        this.state.operations,
-        this.state.operationName,
+      const updatedOperationName = getSelectedOperationName(
+        operations,
```
leebyron: this should be `queryFacts.operations` - the operations passed in from the outside (via `this.state.operations`) will be derived from old information. `getQueryFacts` is responsible for determining that value from the `query`

src/components/GraphiQL.js
```diff
@@ -650,34 +661,41 @@ export class GraphiQL extends React.Component {
...
+      this.setState(
+        this._updateQueryFacts(
+          value,
+          this.state.operationName,
+          this.state.operations,
+          this.state.schema,
+        ),
+      );
     }
     this.setState({ query: value });
```
leebyron: Combine this with the above to ensure one call to setState?

```js
const queryFacts = this.updateQueryFacts(...);
this.setState({
  query: value,
  ...queryFacts
})
```

leebyron:
I'll send a follow-up PR

src/components/GraphiQL.js
leebyron: Ah I'm wrong on this one. This is the previous operations, just was confused by the names

PR #351, opened 2017-02-28
Update GraphiQL storage prop type

### LINKED ISSUE #350: Update GraphiQL storage prop type
The current prop type for storage is:
```
storage: PropTypes.shape({
      getItem: PropTypes.func,
      setItem: PropTypes.func
    }),
```

but now GraphiQL also leverage the `localStorage.removeItem` method as seen in:
> https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L485

So the prop type for storage should read:
```
storage: PropTypes.shape({
      getItem: PropTypes.func,
      setItem: PropTypes.func,
      removeItem: PropTypes.func
    }),
```

tlvenn:
Fixes #350

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -53,7 +53,8 @@ export class GraphiQL extends React.Component {
     response: PropTypes.string,
     storage: PropTypes.shape({
       getItem: PropTypes.func,
-      setItem: PropTypes.func
+      setItem: PropTypes.func,
+      removeItem: PropTypes.func
     }),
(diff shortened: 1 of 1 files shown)


tlvenn:
I believe Travis CI has failed because of S3 being down at the time of commit. Is there any way to restart the build ?

AGS-:
Hi @tlvenn! I restarted the build!

asiandrummer:
Thanks for the fix @tlvenn!

PR #358, opened 2017-03-03
Remove Autocomplete on onKeyUp event from Editor

### LINKED ISSUE #353: Problems with autocomplete and german keyboard layouts
`Alt + Any Keypress` invokes the autocomplete in GraphiQL.

On a german keyboard, the curly brace `{` is typed by `Alt + 8`. 

That means that every time you type a curly brace in GraphiQL, the autocomplete will open. 

The most annoying case is when autocomplete is invoked when typing something like `query test {`, after which you would usually want a line break. When pressing enter, would select something from the autocomplete list. 

I would suggest invoking autocomplete only on a more narrow combination, like `Alt + Space`.

ejoebstl:
Having the Autocomplete at each onKeyUp event that does not map to certain letters and braces on an english keyboard essentially shows autocomplete very frequently and inappropriately on keyboards with other languages. Fixes #353.

### DIFF
--- a/src/components/QueryEditor.js
+++ b/src/components/QueryEditor.js
@@ -120,7 +120,6 @@ export class QueryEditor extends React.Component {
     });
 
     this.editor.on('change', this._onEdit);
-    this.editor.on('keyup', this._onKeyUp);
     this.editor.on('hasCompletion', this._onHasCompletion);
   }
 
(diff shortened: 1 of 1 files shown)


ejoebstl:
*bump* 

Please either follow up on the discussion or consider merging this PR.

Issue #361, opened 2017-03-05
Support editor options overrides


BrianGenisio:
Graphiql currently supports a theme editor override with `editorTheme` but I'm hoping to have finer control over the CodeMirror options.  Ideally, a `editorOptions` property would allow any property to be overridden so that as a user, I can control things like the tab spacing:

```
editorOptions: {
    tabSize: 4,
    indentUnit: 4
}
```

PR #362, opened 2017-03-05
Support passing editorOptions override to the CodeMirror component

BrianGenisio:
Resolves issue: https://github.com/graphql/graphiql/issues/361

This change allows the user to override ANY CodeMirror setting, not just the `editorTheme`.  Specifically, I wanted to support 4-space tabs but figured it was arbitrary to add another specific override.  Instead, this PR adds an optional `editorOptions` object which overrides any of the settings in the editor.  

This allows me to set the tab rule to 4 spaces:
```
editorOptions: {
    tabSize: 4,
    indentUnit: 4
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,9 @@ GraphiQL supports customization in UI and behavior by accepting React props and 
 
 - `getDefaultFieldNames`: an optional function used to provide default fields to non-leaf fields which invalidly lack a selection set. Accepts a GraphQLType instance and returns an array of field names. If not provided, a default behavior will be used.
 
-- `editorTheme`: an optional string naming a CodeMirror theme to be applied to the `QueryEditor`, `ResultViewer`, and `Variables` panes. Defaults to the `graphiql` theme. See below for full usage.
+- `editorTheme`: an optional string naming a CodeMirror theme to be applied to the `QueryEditor`, `ResultViewer`, and `VariableEditor` panes. Defaults to the `graphiql` theme. See below for full usage.
+
+- `editorOptions`: an optional object for overriding CodeMirror options to be qpplied to the `QueryEditor`, `ResultViewer`, and `VariableEditor` panes.
 
 **Children:**
 
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -63,6 +63,7 @@ export class GraphiQL extends React.Component {
     onToggleDocs: PropTypes.func,
     getDefaultFieldNames: PropTypes.func,
     editorTheme: PropTypes.string,
+    editorOptions: PropTypes.object,
   }
 
   constructor(props) {
@@ -298,6 +299,7 @@ export class GraphiQL extends React.Component {
                 onClickReference={this.handleClickReference}
                 onRunQuery={this.handleEditorRunQuery}
                 editorTheme={this.props.editorTheme}
+                editorOptions={this.props.editorOptions}
               />
               <div className="variable-editor" style={variableStyle}>
                 <div
@@ -314,6 +316,7 @@ export class GraphiQL extends React.Component {
                   onHintInformationRender={this.handleHintInformationRender}
                   onRunQuery={this.handleEditorRunQuery}
                   editorTheme={this.props.editorTheme}
+                  editorOptions={this.props.editorOptions}
                 />
               </div>
             </div>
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -96,7 +96,9 @@ GraphiQL supports customization in UI and behavior by accepting React props and
 
 - `getDefaultFieldNames`: an optional function used to provide default fields to non-leaf fields which invalidly lack a selection set. Accepts a GraphQLType instance and returns an array of field names. If not provided, a default behavior will be used.
 
-- `editorTheme`: an optional string naming a CodeMirror theme to be applied to the `QueryEditor`, `ResultViewer`, and `Variables` panes. Defaults to the `graphiql` theme. See below for full usage.
+- `editorTheme`: an optional string naming a CodeMirror theme to be applied to the `QueryEditor`, `ResultViewer`, and `VariableEditor` panes. Defaults to the `graphiql` theme. See below for full usage.
+
+- `editorOptions`: an optional object for overriding CodeMirror options to be qpplied to the `QueryEditor`, `ResultViewer`, and `VariableEditor` panes.
```
asiandrummer: nit: `applied` ;)

asiandrummer:
@BrianGenisio - thanks for the suggestion. It's an interesting idea; however, this change will force all 3 editors to share one editor config - is this what we want? Do we want to separate the editor configurations per each editor (e.g. `queryEditorConfig`, `variableEditorConfig`, and `resultViewerConfig`), or would that be too noisy? Curious to know what you think.

BrianGenisio:
@asiandrummer Good question.  I had considered it, to be honest. I think it makes sense and is the most flexible option.  The only reason I didn't do it that way was because I was following the precedent set by `editorTheme`.  I could very easily see a use case where it would be useful to control the three views separately.

In my use case, I write my queries/variables and copy them into my project, but we use 4-space tabs, so I wanted to make my life easier.  In that case, I'd apply that style to the QueryEditor and VariableEditor, but not the ResultViewer.

This object also eclipses `editorTheme` (if the property is set) so it would allow someone to colorize the windows separately if they'd like.

I'm more than happy to make this change if this is what you prefer :)

asiandrummer:
@BrianGenisio For one example, `ResultViewer` and `QueryEditor/VariableEditor` have different formats/purpose already; therefore I think having three separate editor config overrides is more desirable. You make a good point about eclipsing `editorTheme` though - should they be merged together, and to change the colors we pass an `editorConfig` object instead of `editorTheme`?

Another thought I had was that maybe we should limit the configuration we can change to minumum - we can start from your use case:
```
editorConfig: { tabSize: Number };
```
And then whenever we see another use case, we add to this list of configuration options we can modify.

joelgriffith:
In my mind, if we open up options for editors it might be better to either include `editorTheme` in the props-object for `editorConfig`, or we just flatten `editorConfig` into the root-level props (this would avoid a breaking change).

Sounds like right now just exposing a tab-style and tab-size are most important... so maybe just have them be top-level props?

BrianGenisio:
To be clear, when I say "This object also eclipses `editorTheme`", it only does so if you define it both ways:

```
editorTheme={theme1}
editorOptions={{theme: theme2}}
```

The implementation just does an object extension `{yourProps, ...editorOptions}`, so it doesn't really break anything if you support it generically.  It is up to the caller to override something if they want.

IMO, in a perfect world, `editorTheme` would not exist and it would just be set with `editorOptions.theme`... but it doesn't seem worth it to break existing users for this. You woudnl't have any breaking changes UNLESS a user defined `editorTheme` AND `editorOptions.theme`.

I believe this is a more extensible option, but if you are wary of this change, I'd be happy to re-write it as a specific `tabSize` and `indentUnit` override.

Neitsch:
@BrianGenisio - sorry that this PR has been sitting dormant for so long. Are you still interested in merging it?

BrianGenisio:
TBH, I don't even remember doing this.  Its funny to go back and read my own things that I have no recollection of.  Internally, it looks like we used my fork for a while but have reverted to the public version for quite some time now... and nobody is complaining about the two-space default... I didn't even notice it myself.

So, I guess, if this still merges cleanly and you want to merge it, then I'll probably update our usage of it and move back to 4 spaces... but I don't have the motivation or desire to spend much time on it.  So if you want to just close this, it is fine with me as well. :)

PR #406, opened 2017-04-02
Prettier

AGS-:
Added prettier to the repo and ran everything through it. I currently set some of the eslint rules to be warnings since prettier doesn't have a lot of configuration options yet. This shouldn't really matter since we're going to have consistent formatting anyway, however I would like to add these back in as soon as we can.


resources/pretty.js
```diff
@@ -0,0 +1,58 @@
...
+];
+const glob = '{packages/*/{example,resources,src},example,resources,src}/**/*.js';
+const root = join(__dirname, '..');
+const executable = join(root, 'node_modules', '.bin', 'prettier');
+
+const check = process.argv.indexOf('--check') !== -1;
+const mode = check ? '--list-different' : '--write';
+process.chdir(root);
+
+// eslint-disable-next-line
```
wincent: What's the lint that's firing here? Would be nice to mention the specific rule.

resources/pretty.js
```diff
@@ -0,0 +1,58 @@
...
+  if (check) {
+    print(`\n${YELLOW}The files listed above are not correctly formatted.`);
+    print(`Try: ${INVERSE} npm run pretty ${RESET}`);
+  }
+}
+if (error) {
+  print('error', error);
+}
+// eslint-disable-next-line
+process.exit(status != null ? status : 1);
```
wincent: Same here, what's the lint? I wonder what is different about our `.eslintrc` files between repos that I didn't have any lints in this file when I originally wrote it [here](https://github.com/graphql/graphql-language-service/blob/master/resources/pretty.js).

wincent:
Cool! One thing to bear in mind about landing this: it will require pretty much every pending PR to be rebased/redone, so we should have a look through them and see if there's anything we should/can get in before this one.

Why exclude the test files? If we're going to remove code-formatting as something-we-have-to-think-about, it would be nice to do that everywhere.

resources/pretty.js
AGS-: This one fires 'no-unused-vars' because of the stderr

resources/pretty.js
AGS-: This one fires eqeqeq

AGS-:
Yeah, makes sense. I'll go through the PR's and wait a bit before landing this. I excluded the test files because it was unminifying the vendor tests, but I could just have it exclude ./test/vendor.

AGS-:
I'm abandoning this because it's way too aggressive, I have a much better way to go about this that I'll set up in another PR

PR #430, opened 2017-04-12
Replace React.PropTypes with prop-types package

Hypnosphi:


asiandrummer:
@Hypnosphi seems like we have a conflict :)

Hypnosphi:
> Seems like https://facebook.github.io/react/docs/typechecking-with-proptypes.html encourages this - thanks for the contribution!

As well as React's warnings do since 15.5 =)

Hypnosphi:
I've removed `yarn.lock` from commit's tree to prevent further conflicts

Hypnosphi:
One more release maybe?

AGS-:
@Hypnosphi I may cut another release today

Issue #439, opened 2017-04-18
v0.10.0 doesn't compatible with create-react-app


cpunion:
I created a repo at https://github.com/cpunion/reproduce-import-graphiql-css-error to reproduce this issue. It created by `create-react-app`, and `create-react-app` uses `css-loader` and `postcss-loader` to process .css files. So maybe it doesn't compatible with `css-loader` or `postcss-loader`.

Steps:
1. Checkout the 2nd commit https://github.com/cpunion/reproduce-import-graphiql-css-error/commit/9e1681c55da453633edc25efaa917468b384f38e, it imports css of graphiql@0.9.3 by `import 'graphiql/graphiql.css'`, and it works fine.

2. Checkout the last commit https://github.com/cpunion/reproduce-import-graphiql-css-error/commit/3ac8408011baf3fa34a6e95f1913eabe9dcb52a0, it uses graphiql@0.10.0 and doesn't work.

Errors:
```
Failed to compile.

Error in ./~/graphiql/graphiql.css
Syntax error: Unexpected token  in JSON at position 17977
    at JSON.parse (<anonymous>)
 @ ./~/graphiql/graphiql.css 4:14-107 13:2-17:4 14:20-113
```

cpunion:
I found a problem in `node_modules/graphiql/graphiql.css` by binary search (Cut a half every time :-) )，I found there is a line at `graphiql.css:699` breaks *loaders*.

Please see https://gist.github.com/cpunion/f60fcca11569ce7a6a958a8a4407e20f#file-graphiql-css-L699

gdub22:
Also seeing 0.10.0 fail with postcss-loader when doing `@import 'graphiql/graphiql.css';`

AGS-:
Let me take a look at this, and if necessary I'll cut another release today.

wmertens:
I think the easiest way forward is to remove the source map lines from the built graphiql.css. I don't think they are very useful anyway…

AGS-:
Fixed in v0.10.1 via #440. Thanks!

PR #440, opened 2017-04-18
Disable source map

cpunion:
Disable source map to remove long line. Fix https://github.com/graphql/graphiql/issues/439.


AGS-:
Thanks!

PR #461, opened 2017-05-08
Remove option screw-ie8 from build.sh uglify command

mjvanleeuwen:
`npm run build` currently fails due to an uglify option that is no longer supported.

The option `screw-ie8` is no longer supported with 3.0.x version of uglify. It has been superseded by the option `ie8` which has the opposite effect so does not need to be specified.

### DIFF
--- a/resources/build.sh
+++ b/resources/build.sh
@@ -12,7 +12,7 @@ babel src --ignore __tests__ --out-dir dist/
 echo "Bundling graphiql.js..."
 browserify -g browserify-shim -s GraphiQL dist/index.js > graphiql.js
 echo "Bundling graphiql.min.js..."
-browserify -g browserify-shim -g uglifyify -s GraphiQL dist/index.js 2> /dev/null | uglifyjs -c --screw-ie8 > graphiql.min.js 2> /dev/null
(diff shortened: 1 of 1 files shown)


wincent:
Thanks @mjvanleeuwen. If it's failing, is it exiting with a non-zero exit code? I would have expected `set -e` to make this fail loudly in that case, even without pipefail set.

mjvanleeuwen:
@wincent Yes it was failing with an exit code of 1. Here is an extract from the console output:
```
src/utility/onHasCompletion.js -> dist/utility/onHasCompletion.js
Bundling graphiql.js...
Bundling graphiql.min.js...
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! graphiql@0.10.2 build: `. ./resources/build.sh`
npm ERR! Exit status 1
```
It was failing on this line
`browserify -g browserify-shim -g uglifyify -s GraphiQL dist/index.js 2> /dev/null | uglifyjs -c --screw-ie8 > graphiql.min.js 2> /dev/null`
but because the error out is piped to null, the error reported by uglify was lost.

When I removed the `2> /dev/null` pipes, it printed this to the console:
```
Bundling graphiql.min.js...

  error: unknown option `--screw-ie8'

npm ERR! code ELIFECYCLE
```

wincent:
Would be nice to catch this in Travis CI rather than on running `build.sh` manually.

PR #465, opened 2017-05-09
Run build in CI as part of test run to confirm it works

wincent:
Would have caught:

https://github.com/graphql/graphiql/pull/461


wincent:
Run with a deliberate error introduced [failed as expected](https://travis-ci.org/graphql/graphiql/builds/230404361).

PR #486, opened 2017-05-24
replace unicode whitespace with single space. fixes #399

### LINKED ISSUE #399: Input fails on special characters
### Description
Input fails and Graphiql becomes completely (almost permanently?) unusable when certain whitespace characters are input

### Expected Behaviour
The reverse of the description

### Context
When certain whitespace characters are input (I have tested with unicode characters u2000 to u200a, u2028, u202f, could be more) errors are logged to the console (see attached photo), and typing is no longer visible to the user. As it saves the state to localStorage on `componentWillUnmount`, this makes it almost unusable even after clearing `localStorage.clear()`.  
[The spec](https://facebook.github.io/graphql/#sec-White-Space) says only `tab` and `space` white spaces are accepted, so a better job should be done to handle unaccepted characters.

### Possible solution
Perhaps better error handling? Or warning that an invalid character was inserted.

screen shot 2017-03-29 at 16 03 35

joprice:


joprice:
There may still be some edge case with this, but these changes allow me to copy paste arbitrary invalid characters, and undo the insertion without issue. The range of characters may not be complete, however.

wincent:
Awesome. This looks like a good start. We can adjust it in the future as more edge cases are discovered (and maybe at that point `normalizeWhitespace` won't be the right name for this, but for now this is good). Thanks!

joprice:
Sounds good! Not married to the name or the method of filtering the values. The 'beforeChange' part is the only essential piece. Just trying to help a coworker who's inadvertently pasting unicode from hipchat.

PR #502, opened 2017-06-09
Prettier pre commit hook

AGS-:


AGS-:
Please squash

package.json
```diff
@@ -74,9 +81,12 @@
     "express-graphql": "0.6.6",
     "flow-bin": "0.47.0",
     "graphql": "^0.10.1",
+    "husky": "^0.13.4",
```
wincent: #lazyweb: How does husky actually work? Is it literally installing a precommit hook in the local repo?

AGS-: Yeah, pretty much.

asiandrummer: Feel like this should be a `devDependencies` entry

package.json
```diff
@@ -36,11 +36,18 @@
...
+    "precommit": "lint-staged",
     "prepublish": ". ./resources/prepublish.sh",
     "preversion": ". ./resources/checkgit.sh && npm test",
     "test": "npm run lint && npm run check && npm run testonly && npm run build",
     "testonly": "babel-node ./node_modules/.bin/_mocha $npm_package_options_mocha"
   },
+  "lint-staged": {
+    "*.js": [
+      "prettier --jsx-bracket-same-line --single-quote --trailing-comma=all --write",
+      "git add"
```
asiandrummer: let's also run `npm test` to make sure we didn't break anything, although `prettier` shouldn't have in the first place.

asiandrummer:
I'm approving this to make merging this PR easier

package.json
AGS-: oh yeah, I think it would be pretty good to run `test` just in case

package.json
AGS-: It is, it just got folded right in between dependencies and dev dependencies in the diff view 😛

astorije:
Hi there! Should this actually update the commit, or warn the committer? I just met an issue when opening #568 that I understand may have to do with this, [explained here](https://github.com/graphql/graphiql/pull/568#issuecomment-321988358).
Commenting here to keep track of the related issue. Thanks for your help and work :)

wincent:
@astorije: I'll reply on #568.

PR #512, opened 2017-06-16
chore(package): update react-dom to version 15.6.1

AGS-:


sdemjanenko:
This broke the peer dependencies again?  Can we please be more careful about leaving the peerDependencies more open-ended?

wincent:
Yes, we should be more careful. The trouble is this kind of mistake is easy to make with Greenkeeper flooding multiple repos with PRs: the ones which affect peerDependencies don't always stand out (although in this case at least the diff view did show the "peerDependencies" header). Thanks for pointing this out @sdemjanenko. I'll revert this one.

sdemjanenko:
It would be rad if Greenkeeper could test out the new versions of peerDependencies and just add it to an approved list.

PR #515, opened 2017-06-18
catch the error fired when the storage quota exceeds

asiandrummer:
There are two possible scenarios where the storage quota may exceed:
- Obviously if we end up saving too many/large things
- When the mobile safari gets into the private browsing mode (the quota is set to 0)

GraphiQL started noticing this issue when there were too large of a query saved in the query history (`graphiql:queries`). Guard against this case by cleaning up the `HistoryStore` entries one by one until the saving is safe.

### DIFF
--- a/src/components/QueryHistory.js
+++ b/src/components/QueryHistory.js
@@ -29,7 +29,7 @@ const shouldSaveQuery = (nextProps, current, lastQuerySaved) => {
   return true;
 };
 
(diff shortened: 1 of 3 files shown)


asiandrummer:
cc @AGS- @wincent

AGS-:
@asiandrummer This is awesome! Let's cut a release tomorrow, I want to try to get something in today.

PR #516, opened 2017-06-19
Quick fix because graphiql was freezing

AGS-:


src/utility/HistoryStore.js
```diff
@@ -33,10 +33,10 @@ export default class HistoryStore {
 
   save() {
     const items = this.items;
-    let value = JSON.stringify({[this.key]: items});
-    while (items.length >= 0 && !this.storage.set(this.key, value)) {
+    let value = JSON.stringify({ [this.key]: items });
+    while (items.length > 10 && !this.storage.set(this.key, value)) {
```
wincent: This seems like a mysterious and magical constant. Why not just `> 0`?

AGS-: I set it to 10 so it would match the max history length constant. If it's just set to 0 it just won't add anything to the history

Issue #528, opened 2017-06-27
[0.11] Empty graphiql.min.js


eugene1g:
Hi,

Looks like the most recent build is missing `graphiql.min.js` file in the distribution. That is https://unpkg.com/graphiql@0.11.0/graphiql.min.js is empty but https://unpkg.com/graphiql@0.10.2/graphiql.min.js is fine. We reference it from the CDN as `<script src="//unpkg.com/graphiql/graphiql.min.js"></script>` which now redirect to 0.11, and thus breaks some of the pages.

Is it possible to still ship the .min file?

wincent:
Thanks for reporting! We should be able to cut a corrected release for this shortly.

wincent:
I've fixed the build, so will cut a new release presently.

wincent:
[v0.11.1](https://github.com/graphql/graphiql/releases/tag/v0.11.1) is now out on NPM and you can see the min build at https://unpkg.com/graphiql@0.11.1/graphiql.min.js

srtucker22:
Seems to not have a min file in unpkq in [v0.11.2](https://unpkg.com/graphiql@0.11.2/graphiql.min.js)

PR #529, opened 2017-06-27
Fix min build

wincent:
Closes: https://github.com/graphql/graphiql/issues/528


wincent:
One thing to note about switching from `-g` (undocumented, as far as I can tell, "global" transform) to `-t` is that the built size may be different. I'm seeing a 30KB difference. Given the kinds of devices/networks where GraphiQL is used, this is an acceptable regression (and I am not even sure how much of it is due to the switch), so let's move forward with this in order to correct the breakage. Can loop back later if we think it's worthwhile.

Issue #500, opened 2017-06-09
Add "headers" pane, similar to the graphql variables pane


ryancole:
I see several past issues asking for some way to modify headers, via graphiql. I really think it'd be great if you could add some custom headers. It's kind of a speed bump when using express-graphql, with graphiql, and you need to add a custom header to proceed testing / writing your graphql schema.

Would it be acceptable to add a pane, similar to the graphql variables pane, that just lets you provide some custom headers, in JSON-esque format? No auto complete needed. Just an expandable pane that takes in a JSON payload and interpolates the values into the headers being sent along with the request.

Would this be a pretty minimal addition?

I see in past issues, most responses are to use additional third party tools, or write your own graphiql plugins, etc. This is an issue that comes up, for me, every time I use graphiql in a new project. Wouldn't this just make sense as a pane at the bottom, below the variables pane?

asiandrummer:
@ryancole first of all thanks for the suggestion. I'll try to elaborate why I still think GraphiQL shouldn't be associated with concepts other than what it is now.

> Just an expandable pane that takes in a JSON payload and interpolates the values into the headers being sent along with the request.

How I'd like to think of what GraphiQL is in a minimal sense is that it's just a React component that needs to be rendered in some other DOM to render a web IDE for query/variables. In my opinion GraphiQL should not deal with any other things than rendering itself - it's intentionally meant to be lean and simple for the purpose of easy extensibility/usability.

One of the features of this IDE is an ability to execute a query and gets back the result, and although this may suggest that GraphiQL does operations over the wire, it actually doesn't know/care about the network at all. All GraphiQL requires is a `fetcher` function to be passed in at the time of the component mount, and this `fetcher` function could work completely local if users want it to be. For this reason, a "custom headers" pane does not constitute as an absolutely necessary component to render/use GraphiQL in my honest opinion, hence we decided not to include it. We think any network-related configuration/operation should be dealt with before/in the fetcher level.

Having said all this, I do see how including the ability to add custom headers in `express-graphql` would be useful, and believe that `express-graphql` is the correct place to add such a behavior as a sort of an "add-on". Would that help you to simplify your workflow?

ryancole:
@asiandrummer Sounds reasonable to me. Thanks for the explanation.

Durisvk:
[...]

PR #540, opened 2017-07-13
Custom Headers Feature

Durisvk:
I really like GraphQL and I can't tell how much I love GraphiQL. But there is a major constraint in the current version. Let me explain...

I have a `login` query and it returns a `jwt` token as string, after that in every request I check if that token is present. I can't do it right now with `graphiql` in the current version. So I decided to make it as new feature. Hopefully you'll like it.

Let me give you the photo of how the final feature looks like:
http://imgur.com/a/3VQDT

Thank you in advance for feedback...


Durisvk:
Btw this failing checks are your mistakes, I fix it, then I do commit and it the issues comes back

wincent:
Thanks for this @Durisvk. Please see #500 for prior discussion on why we probably don't want to add a custom headers UI in GraphiQL (and if you search you will find some other duplicate issues as well).

Durisvk:
Sorry I was angry because your system (to be concrete the `prettier` command in `lint-staged`) made my eslint always fail after `eslint`.. in the future you could take a look at this and improve it so that you won't make angry somebody else :)

Issue #545, opened 2017-07-24
Cursor color style overrides one from CodeMirror


Alexsey:
Use [css version 0.10.2](http://cdn.jsdelivr.net/graphiql/0.10.2/graphiql.css). This rules
```
739 /* CURSOR */
740
741 .CodeMirror div.CodeMirror-cursor {
742     border-left: 1px solid black;
743 }
```
in line 742 makes cursor black whatever CodeMirror theme I load later

<img width="147" alt="screen shot 2017-07-24 at 17 07 42" src="https://user-images.githubusercontent.com/6392013/28527399-0a634376-7093-11e7-92ff-6a901807e393.png">

For example themes like [icecoder](https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.27.4/theme/icecoder.css) I make screenshot from

wincent:
Thanks for the report @Alexsey. We'd be open to a PR to fix this, in the event that one of us doesn't get to this before you do.

Alexsey:
https://github.com/graphql/graphiql/pull/547

PR #547, opened 2017-07-25
fix too strong css for cursor color for CodeMirror so it could be overriden

Alexsey:
https://github.com/graphql/graphiql/issues/545


wincent:
Thanks for this @Alexsey.

PR #570, opened 2017-08-14
Allow to specify result tooltip as prop (was: Implement hover trace display)

### LINKED ISSUE #569: Tracing hover information
I am working on integrating [apollo tracing](https://github.com/apollographql/apollo-tracing) information into results display of graphiql.

image

currently it is working only partially and is not a clear solution. But once I have it working where would be apropriate place to do merge request against? This repo? What would be the requirements to get this kind of feature merged? Alternative would be to provide enough extension points to allow to implement this behaviour externally without forking this repo.

jskorepa:
Fixes #569 

If you specify environment variable `TRACING=true yarn dev` it adds tracing information to dev server. ResultViewer component removes tracing info from response and shows it as tooltip when you hover part of response.

For this to work patch to codemirror-graphql is needed, I will submit PR there in just a moment.

Screenshot:
image

### DIFF
--- a/src/components/ResultViewer.js
+++ b/src/components/ResultViewer.js
@@ -28,7 +28,6 @@ export class ResultViewer extends React.Component {
   };
   constructor() {
     super();
-    this._tooltip = document.createElement('div');
   }
 
   componentDidMount() {
@@ -44,13 +43,14 @@ export class ResultViewer extends React.Component {
 
     if (this.props.ResultsTooltip) {
       require('codemirror-graphql/utils/info-addon');
+      const tooltipDiv = document.createElement('div');
       CodeMirror.registerHelper(
         'info',
         'graphql-results',
         (token, options, cm, pos) => {
           const Tooltip = this.props.ResultsTooltip;
-          ReactDOM.render(<Tooltip pos={pos} />, this._tooltip);
-          return this._tooltip;
+          ReactDOM.render(<Tooltip pos={pos} />, tooltipDiv);
+          return tooltipDiv;
         },
       );
     }
--- a/src/components/ResultViewer.js
+++ b/src/components/ResultViewer.js
@@ -7,6 +7,7 @@
  */
 
 import React from 'react';
+import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 
 /**
@@ -23,7 +24,12 @@ export class ResultViewer extends React.Component {
   static propTypes = {
     value: PropTypes.string,
     editorTheme: PropTypes.string,
+    ResultsTooltip: PropTypes.any,
   };
+  constructor() {
+    super();
+    this._tooltip = document.createElement('div');
+  }
 
   componentDidMount() {
     // Lazily require to ensure requiring GraphiQL outside of a Browser context
@@ -36,6 +42,19 @@ export class ResultViewer extends React.Component {
     require('codemirror/keymap/sublime');
     require('codemirror-graphql/results/mode');
 
+    if (this.props.ResultsTooltip) {
+      require('codemirror-graphql/utils/info-addon');
+      CodeMirror.registerHelper(
+        'info',
+        'graphql-results',
+        (token, options, cm, pos) => {
+          const Tooltip = this.props.ResultsTooltip;
+          ReactDOM.render(<Tooltip pos={pos} />, this._tooltip);
+          return this._tooltip;
+        },
+      );
+    }
+
     this.viewer = CodeMirror(this._node, {
       lineWrapping: true,
       value: this.props.value || '',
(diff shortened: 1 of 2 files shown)


wincent:
Thanks @jskorepa. I really love this feature, and I think it's a great example of the kind of powerful things that can be built on top of GraphiQL.

However, I am pretty reluctant to increase the dependency footprint of GraphiQL and I'd love to figure out a way to make this functionality possible as an addition *outside* of GraphQL. This would likely making some changes to GraphiQL to make this kind of extension possible, but not actually baking the extension into GraphiQL itself. I haven't thought deeply about what that might look like, but I'd be interested to hear any ideas that you might have for how it could be done.

cc: @asiandrummer, who I know has a strong interest in keeping GraphiQL small.

jskorepa:
@wincent I see two options for enabling this functionality without including it in GrapiQL itself.

1. Allow to override ResultViewer with custom one.
2. Add prop renderReultsTooltip which accepts function to render tooltip. This would allow extensions to add different types of tooltips.

Another thing this PR does is that it removes tracing data from view but this can be implemented inside alternative fetcher function.

I am open to implementing any of above.

jskorepa:
I implemented option 2 (and moved original code [here](https://github.com/jskorepa/graphiql/tree/old))

This will allow to implement hover trace tooltip (or any tooltip in results view).
See [example branch](https://github.com/jskorepa/graphiql/blob/example/test/index.html#L138) for example usage.

src/components/ResultViewer.js
```diff
@@ -23,7 +24,12 @@ export class ResultViewer extends React.Component {
   static propTypes = {
     value: PropTypes.string,
     editorTheme: PropTypes.string,
+    ResultsTooltip: PropTypes.any,
   };
+  constructor() {
+    super();
+    this._tooltip = document.createElement('div');
```
asiandrummer: We shouldn't need the class variable, right? We could just render `<div>` using ReactDOM.render?

src/components/ResultViewer.js
```diff
@@ -36,6 +42,19 @@ export class ResultViewer extends React.Component {
     require('codemirror/keymap/sublime');
     require('codemirror-graphql/results/mode');
 
+    if (this.props.ResultsTooltip) {
+      require('codemirror-graphql/utils/info-addon');
+      CodeMirror.registerHelper(
+        'info',
+        'graphql-results',
```
asiandrummer: nit: `graphql-results-tooltip` may be a better name I think

jskorepa: Second parametr to registerHelper is a bit confusing, but what it does is specify in which mode that helper should be used. So anything other than `graphql-results` would not work.

src/components/ResultViewer.js
jskorepa: `codemirror-graphql` expects to recieve actual DOM element (not virtual dom) and I did not want to create new one every time tooltip is rendered (which happens every time mouse moves from character to different one). This has benefit that if tooltip remains unchanged when it moves it should not touch DOM (apart from moving it). We would loose that if we created new div every time.

asiandrummer: The DOM creation could happen when `componentDidMount` is called. I have a separate (but similar with yours in nature) concern with creating a class variable that would just act as a placeholder.

asiandrummer: Also, if `codemirror-graphql` expects an actual DOM element, why can't we just create an actual DOM element when passing in props? From your previous code that's what you'd do to insert tracing information, right?

jskorepa: `componentDidMount` is indeed  better place to create the tooltip element. The reason it accepts function returning component not function returning actual DOM element is because I think that it looks cleaner from API consumer perspective - graphiql currently shield user from working with dom it does under the hood. With this API graphiql could potentially even change to something other that codemirror without breaking its API. But in the end I am indifferent to what the final solution is.

asiandrummer: For this case, since you'd just be passing a reference of the created DOM element, I think it'll actually make the API/implementation cleaner because then you wouldn't have to "render" the DOM in here (the implementation). But yeah, it also depends on the use cases I guess

asiandrummer:
Sorry for the hassle @jskorepa, but I just remembered that [a similar feature exists in `codemirror-graphql`](https://github.com/graphql/codemirror-graphql/blob/master/src/info.js). I'll accept & merge this PR here, but we should move this to `codemirror-graphql` repo to consolidate the codemirror helpers in one place.

I'm planning to do some cleanup work/release on codemirror-graphql soon, so maybe I can move it out. Let me know if you get to it earlier than I do please. Thanks for your contribution!

PR #568, opened 2017-08-12
Reset initial size of main and doc panels when double-clicking their resize area

astorije:
It is fairly common to see tools with this capability (e.g. left/right panels in Atom), so I mechanically tried to double-click the resize area to reset the main section to 50%/50% before realizing I was not on a desktop tool :)

After adding this to the main section, I figured docs section could benefit from it as well. Because of how toggling variables section works (single click), it is not possible to add this to the variables section at the moment. I did not have enough time to investigate more as this would require a bigger UI change, which I would not do without starting a discussion anyway, so it is better left to a future work. This already adds value as it is, I think.

graphiql_reset_resize

(The circle around the mouse on click is a thing from the screencast tool, not my PR obviously 😅)

PR #577, opened 2017-08-21
Follow-ups to doc explorer resize PR

wincent:
Follow-ups to PR #568.


astorije:
Argh, sorry, I should have done something like that!

wincent:
No problem @astorije. Figured I would just fix the nits rather than round-trip through another feedback/revision cycle. Thanks for your PR!

Issue #555, opened 2017-08-03
Prettify remove comments


slorber:
Hi,

I think it could be valuable if the prettify function would not remove comments. Currently all lines starting by # are removed.

I actually am working on migrating queries (because for an MVP we are migrating from GraphCool backend to custom GraphQL server) and being able to annotate my queries and prettify them without loosing my comments would be very helpful. Currently I have to do all the formatting manually otherwise all comments are gone.

astorije:
I would love this as well! Project owners, if you agree with this, do you know if it would be fairly easy to achieve? I'd love to help if so.

> (By the way, providing UX to comment user selection/multiple lines at once would also be very handy)

I don't know if it depends on platform and keyboard layout, but on macOS, Chrome, with a `French - PC` layout, I can do this with `<kbd>cmd</kbd>`+`<kbd>!</kbd>`.
I understand my layout is very weird though (AZERTY keyboard on a Mac, yet with PC settings because why simple when it can be complicated? 😅) so you might want to try a few combinations.

slorber:
I have a french mac with azerty keyboard but i'm not sure how to comment multiple lines at once. 

As far as I know if graphiql doesn't setup custom comment shortcuts, it's unlikely to work no?

AGS-:
@slorber If you're trying to batch prettify multiple queries have you considered using [prettier](https://github.com/prettier/prettier) instead of doing each one in GraphiQL?

slorber:
Hi,

When I had this issue, I was only running in Graphiql, not any real frontend project repository. I had to write down graphql queries done against GraphCool, transpose them to a new Ruby backend, and annotate the parts that were not working as expected during the migration from GraphCool to custom backend so that backend developer can uncomment the wrong parts and ensure the new backend supports the same query abilities as what we were using from GraphCool.

I'm not sure how to plug Prettier into such a manual process, and I'm certainly not using GraphiQL to format real world queries of my app (that are split into many static fragments anyway). I would just have appreciated the ability to both comment and prettify queries in Graphiql but it's certainly the most common usecase.

astorije:
@AGS-, I understand and agree with your suggestion that prettier is a great tool for this, but having  GraphiQL not remove comments when clicking on <kbd>Prettify</kbd> is still desirable, I think. What do you think?

AGS-:
@slorber ah ok I see, I misunderstood what you were trying to do.

[...]

Issue #379, opened 2017-03-15
Shortcut to prettify the query


astorije:
After typing a query and before executing it by hitting <kbd>⌘</kbd><kbd>↩</kbd>, it is often useful to prettify the query, at least to make sure it's correctly formatted before sending it to the server.

At the moment, one has to click on the button but it would feel very natural to use a shortcut for that.
Not sure what shortcut specifically though, but if it could use the same modifier than running the query, that would make for an even nicer experience.

Thoughts?

asiandrummer:
@astorije yeah, I can see how a shortcut would be helpful! Not sure if we want to start right away with prettifying every query runs - I'd like to think of the query prettifying as code linting, which serves a good enough purpose for optional suggestions, and is not invasive. Is there a common IDE shortcut for automatically reformatting code?

AGS-:
@asiandrummer, basically all IDEs use different hotkeys for autoformat. ⇧⌘F (VS family), ⌥⇧⌘L for Intellij, ⌥⌘L for Nuclide (I think).

asiandrummer:
`Shift-Ctrl-L` and `Shift-Cmd-L` sounds legit to me then. We can start with that for now.

I'd love for a contribution here if you're down for that @astorije ;)

astorije:
@asiandrummer, I was debating between completely focusing on preparing my talk happening in less than a week, and give this a try considering I could really make use of it during my talk. Also, I pretty much know nothing about React.
So in the end, I still gave it a shot in #385 😅 

Hope you'll like it!

astorije:
@asiandrummer, pretty ping? Did you get a chance to look at my PR? :)
Thanks!

PR #578, opened 2017-08-22
Use prettier to prettify GraphQL queries from the editor

astorije:
Resolves https://github.com/graphql/graphiql/issues/555 (keeping comments intact when prettifying) as well as other goodies introduced by actually prettifying the input instead of re-printing the parsed query.

It turns out, it worked really well during my tests, but...

## Problems

@AGS-, or any other project member, would you be able to help me with those?

### Built files are too big

Products of `npm run build` before this PR:

```sh
-rw-r--r--    1 jeremie  staff   1.4M Aug 21 23:09 graphiql.js
-rw-r--r--    1 jeremie  staff   824K Aug 21 23:09 graphiql.min.js
```

Products of `npm run build` after this PR:

```sh
-rw-r--r--    1 jeremie  staff   5.4M Aug 21 23:11 graphiql.js
-rw-r--r--    1 jeremie  staff     0B Aug 21 23:11 graphiql.min.js
```

Unminified version is almost 4x bigger, urgh!

However, the minified version is veeery lightweight, would load really fast on mobile! :trollface: 
But...

### `npm run build` fails

When running `npm run build`, it fails when building the minified version. When removing the erroring to `/dev/null`, I get this, which I'm not able to explain:

```js
> bash ./resources/build.sh

src/components/DocExplorer/Argument.js -> dist/components/DocExplorer/Argument.js
[...]
src/utility/onHasCompletion.js -> dist/utility/onHasCompletion.js
Bundling graphiql.js...
Bundling graphiql.min.js...
Parse error at 0:51170,33
  return { type: "concat", parts };
                                 ^
ERROR: Unexpected token: punc (})
    at JS_Parse_Error.get (eval at <anonymous> ([...]/graphiql/node_modules/uglify-js/tools/node.js:21:1), <anonymous>:79:23)
    at fatal ([...]/graphiql/node_modules/uglify-js/bin/uglifyjs:268:52)
    at run ([...]/graphiql/node_modules/uglify-js/bin/uglifyjs:225:9)
    at Socket.<anonymous> ([...]/graphiql/node_modules/uglify-js/bin/uglifyjs:163:9)
    at emitNone (events.js:91:20)
    at Socket.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

---

Prettifying with `prettier` has a lot of potential IMO. For example, I wanted to play with `prettier.formatWithCursor` so that once #379 makes it in, the cursor could be keep at the best possible location.
But I'd rather have input from maintainers first, and help if possible, before proceeding further.

❤️

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -9,7 +9,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
-import { buildClientSchema, GraphQLSchema, parse, print } from 'graphql';
+import { buildClientSchema, GraphQLSchema } from 'graphql';
+import prettier from 'prettier';
 
 import { ExecuteButton } from './ExecuteButton';
 import { ToolbarButton } from './ToolbarButton';
@@ -688,7 +689,8 @@ export class GraphiQL extends React.Component {
 
   handlePrettifyQuery = () => {
     const editor = this.getQueryEditor();
-    editor.setValue(print(parse(editor.getValue())));
+    const pretty = prettier.format(editor.getValue(), { parser: 'graphql' });
+    editor.setValue(pretty);
   };
 
   handleEditQuery = debounce(100, value => {


AGS-:
Thanks for the contribution, this seems like quite a bit of overhead to add to GraphiQL, let me take a look at it throughout the week to see if we can figure out a more lightweight way to do it.

astorije:
@AGS-, agreed, that's pretty heavy. #sad
Is there a way to do some tree shaking or whatever to pick only what we need from `prettier`? Surely we don't need the full library, we should be able to drop a good chunk of it (but still, that would be a heavy addition even if we managed to trim 80% off of it!).

@wincent, I think that's a bit convoluted :/ I like to think of GraphQL as a "works 👌 out-of-the-box" kind of tool, and, while I totally understand what you mean and why, it seems a bit overkill for this very feature. I think prettifying should work as you expect it to as a user without having to configure it.

I'll try to think of alternatives as well, but it's kind of a bummer we can't use a perfectly good `prettier` for that 😅.

astorije:
Actually, that extra 4MB checks out when I look at the file sizes in https://unpkg.com/prettier@1.5.3/ !

And when you look at it, parsers irrelevant to us take most of the cake! If there is a way to include only `index.js` (406.84 kB, unminified) and `parser-graphql.js` (28.35 kB, 57x smaller than the biggest parser lol), then we'd end up with "only" 435 kB of overhead (10x less than original attempt), and around half that if we minify `index.js`.

Would that order of magnitude be acceptable to you? Yes, it's still overhead compared to right now, but I think it would be beneficial to have a 💯 prettifier with regards to what GraphiQL solves/offers :)

Also, I'm still pretty lame/new at this project/React, forgive me if how I see dependency management is nonsense here!

astorije:
Also, if that's acceptable to you, or if you want me to give it a shot before a definitive opinion... do you know how I can achieve this? 😅

slorber:
Shouldn't this feature be build directly in graphql lang parser? cc @vjeux @jnwng

https://github.com/graphql/graphql-js/tree/master/src/language

It is used under the hood by prettier  [here ](https://github.com/prettier/prettier/blob/master/src/parser-graphql.js#L37)

see pr https://github.com/prettier/prettier/pull/2187

AGS-:
@slorber maybe

@astorije I think ultimately the best way to go about this would be to make it pluggable. The way to do it imo would be to have it as an optional prop and have it use that if it's available and if not have it default to the graphql-js one. I don't think it would be too complicated since you just have to be able to change the editor value on click, only this time it would be done externally from GraphiQL vs your PR.

wincent:
> @wincent, I think that's a bit convoluted :/ I like to think of GraphQL as a "works 👌 out-of-the-box" kind of tool, and, while I totally understand what you mean and why, it seems a bit overkill for this very feature. I think prettifying should work as you expect it to as a user without having to configure it.

@astorije, "convoluted" might be a bit of an exaggeration. We're talking about how to avoid taking on a potentially massive non-optional dependency here. GraphiQL definitely *is* a "works out-of-the-box" kind of tool, but that doesn't mean it should be all things to all people. It is React component, and the idea is that you should be able to compose/configure/combine it with other things if the basic functionality that it provides doesn't meet all of your needs. Prettifying already *does* work in a reasonable way; what we're talking about here is making it incrementally better. If we can figure out how to do that for "free", we absolutely should; but if we can't do it without bringing the cost down in a significant way, then making it configurable is a good trade-off.

In any case, let's continue to explore just how small this could be crushed down if we were able to pull in just the specific parts of prettier that mattered. If that solution can be made practical, great. If not, I think configuration is probably the right way for now.

How small is small enough? I don't actually know. The current minified bundled is about 820K, I think. Ideally we'd stay under 1M. It's a shame that it's as big as it is, TBH, but we are able to abrogate some of our responsibility for keeping things small here due to the fact that we are a developer-facing tool, less likely to be used on mobile devices and shitty networks. But still, it's desirable to keep things lean if we can.

vjeux:
I agree that the current situation is not great regarding integration of prettier in a web tool like GraphiQL. If you are interested in doing the work inside of prettier codebase, I think that you should be able to generate a `prettier-graphql` module that:
- Uses graphql as a peer dependency, so you don't pay the cost twice.
- Only includes the files from prettier that are needed. GraphQL is really simple compared to the rest of prettier so that shouldn't be a lot of code.

asiandrummer:
Hi all, sorry I'm a little late to the party. I agree with @slorber that `graphql-js`'s parser and printer should support this, and the new implementation would be something similar to what @vjeux has in Prettier. I believe we have a `CommentNode` already in `graphql-js` that we can use to express comments, but the language package doesn't incorporate it. If we can collaborate to have `graphql-js` support comments, then we'd be able to solve this problem in a correct way that benefits everyone, even `prettier` repo.

slorber:
That makes sense to me to wait for `graphql-js` to support comments (if possible) instead of embedding Prettier. 

@wincent @leebyron do you think it's feasible to support comments parsing/printing in `graphql-js` directly?

Btw it would perhaps be possible to lazy load Prettier if no other choice?

wincent:
> do you think it's feasible to support comments parsing/printing in graphql-js directly?

Technically it is feasible, but at the same time we probably don't want to race to reinvent a wheel here. Prettification is in prettier's "wheel house", but not really in the core of graphql-js's set of responsibilities.

graphql-js is intended to be a minimal, complete reference implementation of the GraphQL Spec, which is itself basically silent on the question of exactly how documents should be formatted (other than by implicitly modeling a "conventional" style in the code examples). The fact that you can do a crude form of pretty-printing using graphql-js is not so much fruit of having pretty-printing ability as a design goal, but a simple consequence of being able to round trip text → AST → text by composing functions (and yes, it is a lossy "round trip", because the conversion to AST *is* lossy; it's called an *abstract* syntax tree for a reason...)

Having said all that, GraphQL syntax is much less complicated than JS, and building in the ability to properly preserve and print comments is likely not too large a task. I wouldn't be averse to including a clean, minimal implementation which would bloat the graphql-js package. Given that the package is low-level and intended to be a building block suitable for use in a wide range of contexts, its important that we keep an eye on performance, size, modularity etc.

vjeux:
Fwiw, GraphQL was dead simple to print and doesn't use any of prettier exotic features. Improving the existing GraphQL printer to be 80-columns aware and handle comments should be easy if you are interested.

leebyron:
> @wincent @leebyron do you think it's feasible to support comments parsing/printing in graphql-js directly?

Absolutely. In fact, I think that's probably the right solution for this issue.

I love prettier, but it's not really intended to be integrated as a web tool and it both contains far more than GraphQL related code and if we ever expand syntax for GraphQL in the future we may find ourselves in a tricky dependency situation for maintaining the prettify feature. 

As for parsing/printing GraphQL with comments, the parsing is already supported. GraphQL ASTs made with graphql-js include reference to the comments in the token stream and could be accessed for printing. Printing with comments doesn't exist yet, but would be a happy addition

astorije:
Hi all, sorry for the silence!

> @astorije, "convoluted" might be a bit of an exaggeration. We're talking about how to avoid taking on a potentially massive non-optional dependency here.

@wincent, I'm truly sorry if I offended you in any way. It wasn't my intent at all!
My understanding is that you were saying the current prettifier is good enough to not need changing, and any further improvement would require consumers to BYOPrettifier, but clearly that wasn't the case, so that's my fault.
Even if it was, I am involved in enough OSS projects to know that, as an external contributor, I am at best providing an uneducated suggestion and not trying to convince or push you of any sort. I apologize if my phrasing made seem any different.

> If we can figure out how to do that for "free", we absolutely should; but if we can't do it without bringing the cost down in a significant way, then making it configurable is a good trade-off.

I totally agree with this statement and your previous ones.

> Improving the existing GraphQL printer to be 80-columns aware

This was actually a nice side-effect of bringing prettier here along with keeping the comments: so far, clicking in "Prettify" also flattens argument lists. I wasn't aware we could maintain this nice-to-have without the whole prettier deal :)
(The other "free" side-effect of prettier is `formatWithCursor` but surely another refinement that can be achieved without it anyway.)

> if we ever expand syntax for GraphQL in the future we may find ourselves in a tricky dependency situation for maintaining the prettify feature

This is another very good point I overlooked.

---

I am very glad I opened this PR early with 6 lines of change rather than going down the rabbit hole with the few changes I was planning to make, then realize it was the wrong way to go.

Everyone here seems to be open to further contribution for making this better, and I would be more than happy to do so. I'll close this PR for archive purpose (changes in closed PRs are history) but will reference this PR in any further work, whether it lives in `graphql-js`, this repo, or another.

So, just to summarize correct me if I'm wrong: the correct fix for this would be to enhance [`graphql-js`'s printer](https://github.com/graphql/graphql-js/blob/master/src/language/printer.js) so it doesn't get rid of comments, right? Is this URL a good place for me to start? Are there any other requirements to this change you'd like to express so I don't open another PR that misses the point? 😅
I'll also look into the 80-char limit ability if that's desirable. @vjeux, re: [this comment](https://github.com/graphql/graphiql/pull/578#issuecomment-325009968), it seems you have been there recently with prettier, is there anything in particular or non-obvious I should know?

vjeux:
I would highly recommend stealing the intermediate representation that prettier uses, it makes it almost trivial to have a 80-column aware printer. The files starting with doc- in the prettier codebase and they have no dependencies: https://github.com/prettier/prettier/tree/master/src

astorije:
I will look into that and ask questions if I'm stuck, thanks @vjeux!

asiandrummer:
@astorije great resolution, thanks for your contribution/interest! I look forward to seeing your PR.

astorije:
Alright, so I started to look into this (keeping comments in this thread until I get to actually open a PR in `graphql-js`), and though I am slowly getting acclimated to the parsing code, this is still all very confusing for now :)

So at the moment, I understand that (forgive me if what follows make no sense!) the lexer understands comments (i.e. has [a token kind for it](https://github.com/graphql/graphql-js/blob/b84b69fd723e703f089f6524db9851341aa98d50/src/language/lexer.js#L104)) but [removes them](https://github.com/graphql/graphql-js/blob/b84b69fd723e703f089f6524db9851341aa98d50/src/language/lexer.js#L190) from the tokens it sends to the parser. In fact, I [could not find a `Kind` for comments](https://github.com/graphql/graphql-js/blob/b84b69fd723e703f089f6524db9851341aa98d50/src/language/kinds.js) nor any reference to "comment" in `src/language/parser.js`.

Then I played with the tests a bit, trying to get comments to appear anywhere:

```js
const queryAstShorthanded = parse(`query {
  # Test
  id }`
);
console.log(JSON.stringify(queryAstShorthanded, null, 2));
```

This produced the following:

<details>
<summary>Result <em>(click to expand/collapse)</em></summary>

```json
{
  "kind": "Document",
  "definitions": [
    {
      "kind": "OperationDefinition",
      "operation": "query",
      "variableDefinitions": [],
      "directives": [],
      "selectionSet": {
        "kind": "SelectionSet",
        "selections": [
          {
            "kind": "Field",
            "alias": null,
            "name": {
              "kind": "Name",
              "value": "id",
              "loc": {
                "start": 19,
                "end": 21
              }
            },
            "arguments": [],
            "directives": [],
            "selectionSet": null,
            "loc": {
              "start": 19,
              "end": 21
            }
          }
        ],
        "loc": {
          "start": 6,
          "end": 23
        }
      },
      "loc": {
        "start": 0,
        "end": 23
      }
    }
  ],
  "loc": {
    "start": 0,
    "end": 23
  }
}
```
</details>

I'm not seeing any reference to the comment here.

Seeking advice before I dig deeper: should the lexer not skip comments? I'm guessing any such change to the lexer would break the 💩 out of backwards compatibility, no?
And assuming lexer doesn't skip comments anymore, is it reasonable to add a `COMMENT` kind for the parser?
I am very much interested in getting to the bottom of this, but any help will be greatly appreciated 😅

Also, I did look a bit at the prettier code, but for now I'm having a hard time making the connection between both codebases, it just adds to the confusion so far... 😕 

Thanks!

wincent:
@astorije: [this test](https://github.com/graphql/graphql-js/blob/713e0e8dc5f4f4782578312b0184ae2f29738b8a/src/language/__tests__/lexer-test.js#L665-L700) makes it clear what is going on. Comment tokens are skipped in the token stream, but you can still traverse to them using `prev`/`next` properties on them.

astorije:
Hey @wincent! Thanks for the link, and yeah I did take a look at this before and this does show me that the lexer keeps everything including comments, but the parser skips them. Since the printer does everything based on the parsed AST, as far as I understand, it does not have access to comment tokens anymore.
Of course, correct me if I'm wrong :)

I opened an early PR at https://github.com/graphql/graphql-js/pull/1029 (potentially all wrong, but I'm trying things until something works). Do you have any suggestions?

PR #580, opened 2017-08-24
export core components of GraphiQL

asiandrummer:


src/index.js
```diff
@@ -6,5 +6,10 @@
...
 
-// The primary React component to use.
-module.exports = require('./components/GraphiQL').GraphiQL;
+import { GraphiQL } from './components/GraphiQL';
+import { QueryEditor } from './components/QueryEditor';
+import { VariableEditor } from './components/VariableEditor';
+import { ResultViewer } from './components/ResultViewer';
+
+export default GraphiQL;
+export { QueryEditor, VariableEditor, ResultViewer };
```
wincent: Why re-export all of this stuff which is already on the default `GraphiQL` export anyway? We haven't done that with the other component properties, so this is inconsistent.

asiandrummer:
@wincent I've reverted the inconsistency you mentioned

PR #581, opened 2017-08-24
switch to markdown-it

caub:
markdown-it is more maintained and reliable


asiandrummer:
@caub thanks for this - have we looked at the differences in bundle sizes?

caub:
@asiandrummer yes it's quite bigger 1.6M / 930KB (min) with markdown-it, 1.4M 828KB (min) with marked

I caught some bugs with marked, but if it's not critical, this lib is ok

asiandrummer:
@caub I'm okay with that I think. @wincent would like to hear your thoughts please

wincent:
> markdown-it is more maintained and reliable

@caub: Being more maintained is definitely a desirable property, although in the case of something like `marked` which is already very mature and widely used (3m downloads in the last month, compared with 1m for `markdown-it`), one can make a reasonable "if it ain't broke don't fix it" kind of argument against switching, especially in light of the increased bundle size.

In order to actually justify switching, I'd like to hear more detail about the "more reliable" part. You mentioned some bugs, @caub: can you share some concrete examples? Those might be enough to warrant making the switch. In the absence of them, I am a bit reluctant to increase the minified bundle size by 12.6% (would be interesting to see post-gzip numbers, though).

caub:
Well, sorry, I think we could freeze this PR until someone experience bugs. I didn't have any with graphiql, it's used only for descriptions

I had in the past, but I forgot why I didn't post an issue or commented an existing one, there are a few of issues: https://github.com/chjj/marked/issues and lack of maintenance, but it still does the job for simple things

wincent:
> Well, sorry, I think we could freeze this PR until someone experience bugs.

Sounds like a good plan. I'll close this for now. If anything crops up, let's re-open and re-visit this.

wincent:
See discussion on #628; maybe it is time to revive this.

asiandrummer:
From #628 there's an expected slowdown from a bug in `marked` lib - I looked at `markdown-it` briefly and it seemed like a reasonable substitute. If anyone wants to tackle this we'd welcome the contribution!

caub:
yes, I'll update it

PR #593, opened 2017-09-09
Default values for input object types are now displayed in Doc Explorer

excitement-engineer:
GraphiQL can now show default values for input object types. Currently, GraphiQL only shows default value for args that are scalar values, not input objects. This PR aims to solve this problem.

I think this solution is especially useful for a relay-compliant schemas where mutations are performed using a single `input`object argument.

Let me know what you think!

Here is a screenshot of what it looks like with the new addition:

screen shot 2017-09-09 at 10 19 48

### DIFF
--- a/src/components/DocExplorer/DefaultValue.js
+++ b/src/components/DocExplorer/DefaultValue.js
@@ -1,0 +1,31 @@
+/**
+ *  Copyright (c) Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { astFromValue, print } from 'graphql';
+
+export default function DefaultValue({ field }) {
+  const { type, defaultValue } = field;
+  if (defaultValue !== undefined) {
+    return (
+      <span>
+        {' = '}
+        <span className="arg-default-value">
+          {print(astFromValue(defaultValue, type))}
+        </span>
+      </span>
+    );
+  }
+
+  return null;
+}
+
+DefaultValue.propTypes = {
+  field: PropTypes.object.isRequired,
+};
--- a/test/schema.js
+++ b/test/schema.js
@@ -28,12 +28,12 @@ const TestEnum = new GraphQLEnumType({
   values: {
     RED: { description: 'A rosy color' },
     GREEN: { description: 'The color of martians and slime' },
-    BLUE: { description: 'A feeling you might have if you can\'t use GraphQL' },
+    BLUE: { description: "A feeling you might have if you can't use GraphQL" },
     GRAY: {
       description: 'A really dull color',
-      deprecationReason: 'Colors are available now.'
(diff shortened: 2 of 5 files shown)


test/schema.js
```diff
@@ -28,12 +28,12 @@ const TestEnum = new GraphQLEnumType({
   values: {
     RED: { description: 'A rosy color' },
     GREEN: { description: 'The color of martians and slime' },
-    BLUE: { description: 'A feeling you might have if you can\'t use GraphQL' },
+    BLUE: { description: "A feeling you might have if you can't use GraphQL" },
```
asiandrummer: nit: `'` instead

src/components/DocExplorer/DefaultValue.js
```diff
@@ -0,0 +1,31 @@
...
+import { astFromValue, print } from 'graphql';
+
+export default function DefaultValue({ field }) {
+  const { type, defaultValue } = field;
+  if (defaultValue !== undefined) {
+    return (
+      <span>
+        {' = '}
+        <span className="arg-default-value">
+          {print(astFromValue(defaultValue, type))}
```
asiandrummer: Does `print` function always print with quotes around the default value? Could we check this with some more test cases, maybe for scalar types at least? (e.g. boolean, int, etc)

Also, the color difference slightly bothers me :D I'm even for using another color hmm... is `arg-default-value` an existing classname?

asiandrummer:
@excitement-engineer Sorry we've been stalled giving you reviews! Lets work together on this next day or so to get this in! I'm super excited about it.

src/components/DocExplorer/DefaultValue.js
excitement-engineer: I will add some test cases! This code is similar to the code uthat is currently used to display default values for scalar arguments. The classname that I ised was the same as before. I will take a look to see if there is a more suited color for the default value

excitement-engineer:
Thanks for taking a look! I will update the pr as soon as I find some tome!

test/schema.js
excitement-engineer: The project has lint-staged enabled, this is changing the single quote to a double quote each time I commit. Given all the changes in the diff seems that prettier has not been run on all the files yet. I think that it's better to open another PR where prettier is run on the entire project so that new PRs will have clean diffs.

excitement-engineer:
Hey @asiandrummer I updated the PR. I added some test cases for boolean and integers and I can confirm that no quotes are printed for the default values. 

I also changed the color of the default value to green to increase the contrast between the field and the default value. See the screenshots below. 

Let me know what you think about the changes!

screen shot 2017-09-30 at 15 04 21
screen shot 2017-09-30 at 15 04 39

src/components/DocExplorer/DefaultValue.js
excitement-engineer: I ran prettier over the files in the test folders using PR #608.

asiandrummer:
cc @wincent @leebyron for visibility. This PR looks good to me, thanks again for your contribution @excitement-engineer!

Issue #565, opened 2017-08-09
feature request


Noor0:
wouldn't it be great to be able to

- add request headers.
- view response headers.
- create and set environment variables.

wincent:
Thanks for your feature request!

Please see #500 for most recent discussion around setting request headers. For visualizing response headers, are the Chrome dev tools sufficient for your needs? For "environment variables", I am not even sure what that means in the context of GraphiQL (it's just a React component), but the answer is likely the same: you can configure it with an arbitrary `fetcher` function, but GraphiQL itself has no business knowing or caring how you transport data to and from the schema; if you care about that you should be able to get pretty far by composing GraphiQL within your React UI, and doing whatever you want/need in your custom `fetcher`.

Closing this for now, but feel free to comment again if you have further questions/thoughts etc.

AlexanderChen1989:
@wincent We just want a easy way to test api!

asiandrummer:
@AlexanderChen1989 I'm sorry if our responses felt like that, but I just want to make sure we're ignoring your requests. In #500, I listed a couple of options that you can look into if you really wanted setting custom headers:
- Examples from individual contributors in [this issue](https://github.com/graphql/graphiql/issues/59)
- http://toolbox.sangria-graphql.org/ (powered by Sangria)

Hopefully those options fulfill your needs. @wincent has again made a great explanation as to why we wouldn't want the custom header feature in GraphiQL, so I'll avoid being repetitive here.

About environmental variables, in which cases would you need to use env vars? What would you like to configure/customize, and how would the env variables help? If you're using our example code to fire up a GraphiQL instance to test your queries, then you should be able to alter the server code however you need to solve your use case. If you're embedding GraphiQL as a standalone React component, then there should be a far more freedom for you to do whatever/however you wanna do. Please let us know if we're misunderstanding something.

imolorhe:
@AlexanderChen1989 You can also checkout https://github.com/imolorhe/altair

Issue #148, opened 2016-06-20
Feature request: Add ability to specify custom headers


eugenehp:
It would be great to have functionality to add custom headers for authorisation purposes within GraphQL.

eugenehp:
I'm working on the PR at the moment.

![screen shot 2016-06-22 at 20 03 38](https://cloud.githubusercontent.com/assets/1857263/16290708/8010ec58-38b4-11e6-8a49-675f95614218.png)

asiandrummer:
Hi @eugenehp - thanks for the suggestion! We have discussed a custom header feature request in the past: #114 for one of the examples. Would it be more plausible for your use case to define a custom `graphQLFetcher` that deals with custom headers?

eugenehp:
Hi @asiandrummer yes, absolutely. I'll take a look into it. I already played with the example directory and got it working. What would be the best way to define custom `graphQLFetcher` to work with the HTTP headers?

![screen shot 2016-06-22 at 20 31 25](https://cloud.githubusercontent.com/assets/1857263/16291787/80aebb62-38bf-11e6-8908-f04b9c3f6a23.png)
<img width="1550" alt="screen shot 2016-06-22 at 20 31 27" src="https://cloud.githubusercontent.com/assets/1857263/16291788/80d5ca54-38bf-11e6-9e04-94f0b9b02e35.png">

asiandrummer:
I think you may have already looked at it, but refer to [this example](https://github.com/graphql/graphiql/blob/master/example/index.html) from the example code we have. It defines a custom `graphQLFetcher` with a set of custom headers.

The main reason why we'd prefer this is because we want `GraphiQL` to be independent of the network configurations. See #96 and #59 for more conversations about this please.

eugenehp:
@asiandrummer Thanks Hyo. I'm moving this [here](https://github.com/graphql/express-graphql/issues/101) so we can have a way to place GraphiQL and custom versions of it as middleware.

webmobiles:
how i can use the middleware to input the headers ? I've tried plugins for chrome and app for mac, and none of them is working ? it's a pity no input headers in the original graphiql, looks dogmatic decision

asiandrummer:
@webmobiles it's difficult to answer your question without knowing which middleware you're talking about and the plugins you've tried/without mentioning them. I heard that GraphiQL for Mac already supports a custom header, but have not seen it/used it myself. 

[...]

PR #602, opened 2017-09-23
Add the headers to the interface

antoinealej:
having the possibility to change the headers in the interface could be great.

I've added it, please let me know if I went in the right direction.

Thanks.


andrewtpoe:
I am very interested in this feature, and hope to see it merged soon. Trying to build an api that requires auth headers with a tool that doesn't allow editing them has been challenging at best.

wincent:
See previous incarnations of this proposal:

- PR #540
- Issue #500 
- Issue #565 
- Issue #148

leebyron:
Thank you for your work on this, however unfortunately this is not something that belongs in the core GraphiQL project, but absolutely is something that should be supported as an add-on. I recommend you create this as a separate project and share it via npm.

GraphiQL is network agnostic, where the network implementation is provided as part of installation to best fit the use case. Not all networks even support headers, and even many of those that do would not want their users changing the header configuration.

GraphiQL supports a [customizable UI](https://github.com/graphql/graphiql#usage-examples) in the header and footer of the tool, and even includes a UI toolkit to help you build headers that look cohesive. Any installation of GraphiQL on an API which requires or desires header customization or other auth customization should rely on these customization features to provide them.

yangli1990:
@leebyron thanks for the detailed explanation

> Not all networks even support headers, and even many of those that do would not want their users changing the header configuration

Out of curiosity which networks does GraphiQL support that doesnt support headers. Also, we can make the header section configurable? But honestly, not having a headers section doesn't stop a user from making a call with headers - it just makes it harder for all the people that need headers.

leebyron:
> Out of curiosity which networks does GraphiQL support that doesnt support headers. 

At Facebook we've used MQTT, WebSockets, and others. Especially as subscriptions becomes more widely used, non-HTTP transports are becoming common.

Also, in my experience the most common desire to alter headers is for providing authentication tokens. That's great, but it probably deserves a UI designed for changing that authentication token rather than a generic header editor - something that would be different for every user of GraphiQL depending on how their particular auth works, which is why it's not built-in.

Instead, we sought to make it easy for people to add their own by making the header and footer of GraphiQL configurable and providing a UI toolkit you can use for easily making new elements like an auth-token addition that look cohesive.

We're not trying to intentionally make things difficult for people, instead we're trying to keep GraphiQL a focused and high quality tool for as many use cases as possible, while making it straight-forward to adjust for particular use cases. I still think that's the right direction for the project.

jimscarver:
There seems to be enough interest in this to maintain a branch. Use of Graphql is growing fast and everyone needs graphql accessing their protected information resources. Is someone working on this? Who can help?

asiandrummer:
@jimscarver I forget the links, but there are multiple branches/implementations that wrapped the custom header support around GraphiQL. I can only remember @OlegIlyenko's Sangria app having this capability.

OlegIlyenko:
Indeed, meanwhile there are a lot of implementations that offer headers and many other features on top of the base GraphiQL functionality. just a few examples:

https://github.com/OlegIlyenko/graphiql-workspace
https://github.com/prismagraphql/graphql-playground
https://insomnia.rest/

PR #608, opened 2017-09-30
Run prettier over test/ folder

excitement-engineer:
I ran prettier over the test folder because prettier added unnecessary formatting changes to the diff in PR #593. Hopefully this will mitigate this issue going forward.


asiandrummer:
Awesome, thanks for doing this @excitement-engineer! We should just include this in the lint/test runs though...

PR #609, opened 2017-09-30
Initial POC for alphabetic sorting of the info exposed in the docs.

excitement-engineer:
In a large schema it can sometimes be challenging to find a specific field/type.

I created this PR to simplify the process of finding info in the Doc Explorer by introducing alphabetical sorting for the info. 

Note, this is just a POC that only enables sorting for fields for now. Before continuing I am curious to hear what everyone thinks of this idea.

BEFORE SORTING: 

screen shot 2017-09-30 at 15 50 33

AFTER SORTING:

screen shot 2017-09-30 at 15 51 55

### DIFF
--- a/src/components/DocExplorer/TypeDoc.js
+++ b/src/components/DocExplorer/TypeDoc.js
@@ -19,6 +19,7 @@ import {
 import Argument from './Argument';
 import MarkdownContent from './MarkdownContent';
 import TypeLink from './TypeLink';
+import sortAlphabetically from '../../utility/sortAlphabetically';
 
 export default class TypeDoc extends React.Component {
   static propTypes = {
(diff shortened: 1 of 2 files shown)


asiandrummer:
@excitement-engineer Thanks for the PR first of all! The documentation explorer honors the ordering of the entities that the schema has provided, just in case that there's a specific meaning for it. I think my initial reaction would be the schema should sort the fields before being introspected by the client.

Also, would searching for the specific field resolve this?

excitement-engineer:
Hey @asiandrummer! I completely agree with you, the right place to organize the fields is in the actual graphql schema itself! Thanks for your feedback:)

Searching is a good way to finding things but I find that sorting the API in alphabetical ordering generally makes it easier to navigate around the documentation.

Thanks for you help!

PR #616, opened 2017-10-12
Fix CustomGraphiQL example

dlredden:
The `...this.state` needed curly braces around it


wincent:
Thanks for this @dlredden. I actually merged #618, which is identical this one, because I saw it first. Sorry about that. Going to close this one as it is no longer needed.

PR #623, opened 2017-10-25
Prevent toolbar menu drop-downs from being clipped

wincent:
Fixes regression introduced in fc4f1aae.


wincent:
CSS is treacherous. More context: https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue

PR #626, opened 2017-10-27
 update readme to match button

adampaulberg:
I'm not sure it was ever just button, but it changed from PrettifyButton to ToolbarButton in this commit. https://github.com/graphql/graphiql/commit/4dfa12b677f55fb7931712cc08ac19c7f88c282b#diff-a18c87224a76fc29336b97c2bf38aaaa


wincent:
Thanks for this, but `ToolbarButton` is exported as `GraphiQL.Button`:

https://github.com/graphql/graphiql/blob/e078850bfa4ec8131703cde533fb640cfe583d4a/src/components/GraphiQL.js#L976

PR #629, opened 2017-11-05
Sort fields and enum values by name in Doc Explorer

dominicbarnes:
I noticed that the current doc explorer displays fields and enum values in no particular order, since the underlying value is an object which has no guaranteed enumeration order for keys. I generally get the behavior of following the order defined in my source file, but that's just a coincidence.

This PR sorts fields and enum values by their name in alphabetical order, using [String.localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). My primary goal is that this large schemas would be easier to scan.

~~The diff is a lot bigger than I expected, which came from `prettier` being run automatically during `precommit`. (I can point out the relevant parts of my PR if desired)~~

### DIFF
--- a/src/components/DocExplorer/TypeDoc.js
+++ b/src/components/DocExplorer/TypeDoc.js
@@ -65,14 +65,12 @@ export default class TypeDoc extends React.Component {
     if (types && types.length > 0) {
       typesDef = (
         <div className="doc-category">
-          <div className="doc-category-title">
-            {typesTitle}
-          </div>
-          {types.map(subtype =>
+          <div className="doc-category-title">{typesTitle}</div>
+          {types.map(subtype => (
             <div key={subtype.name} className="doc-category-item">
               <TypeLink type={subtype} onClick={onClickType} />
-            </div>,
-          )}
+            </div>
+          ))}
         </div>
       );
     }
@@ -82,23 +80,23 @@ export default class TypeDoc extends React.Component {
     let deprecatedFieldsDef;
     if (type.getFields) {
       const fieldMap = type.getFields();
-      const fields = Object.keys(fieldMap).map(name => fieldMap[name]);
+      const fields = Object.keys(fieldMap)
+        .sort((a, b) => a.localeCompare(b))
+        .map(name => fieldMap[name]);
       fieldsDef = (
         <div className="doc-category">
-          <div className="doc-category-title">
-            {'fields'}
-          </div>
+          <div className="doc-category-title">{'fields'}</div>
           {fields
             .filter(field => !field.isDeprecated)
-            .map(field =>
+            .map(field => (
               <Field
                 key={field.name}
                 type={type}
                 field={field}
                 onClickType={onClickType}
                 onClickField={onClickField}
-              />,
-            )}
+              />
+            ))}
         </div>
       );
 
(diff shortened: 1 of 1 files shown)


dominicbarnes:
I'm pretty new to GraphQL in general, and this project in particular, so I couldn't think of other places to apply sorting off the top of my head. I'm happy to apply this change more consistently if there are other places that would benefit.

asiandrummer:
@dominicbarnes thanks for the contribution! However, the order of fields/values is from the schema and we'd like to think that the provided order is intentional by the creator of the schema - I do think this is an useful "filter" option once we come up with one.

dominicbarnes:
@asiandrummer In JSON, object are by definition **unsorted**, it's in the [spec](http://www.json.org/):

> An object is an unordered set of name/value pairs.

With that said, you cannot assume the iterated order of object keys is the same as the developer intended. My goal with this change is to make the order deterministic, since there are no guarantees otherwise.

It looks like my branch has fallen out of date, I'll rebase it so it is merge-able again.

asiandrummer:
@dominicbarnes So since I always thought that object traversal order was deterministic for some reason, I did some digging around this topic. I found the following:
- JSON objects are **unordered** set of name/value pairs by definition
- When JavaScript creates objects (like from [JSON.parse](https://tc39.github.io/ecma262/2017/#sec-json.parse)) and calls the functions reading object properties/keys, it does have a certain traversal order (I found a helpful documentation part from [ECMAScript latest draft](https://tc39.github.io/ecma262/2017/#sec-object.keys))
  - pretty interesting sentence to decide the ordering in https://tc39.github.io/ecma262/2017/#sec-enumerableownproperties

I asked some of my peers around and also found out that the browsers implemented the standard behavior first and then the ECMAScript spec later adopted this order as a "standard". There are a couple of edge cases in this ordering (e.g. integer keys always come before string/symbol keys) but I'd say this ordering is deterministic.

The presented ordering from the schema is usually meaningful, and if a particular ordering is necessary, we encourage the server to do some sort of sorting on the schema before passing it down to the client. With this said, because of the above "edge cases", I do see how the ordering from the server may not be preserved in some cases.

Lastly, if for some reason manipulating the schema isn't possible (like when you're using a third-party schema from an URL or something), I can see how sorting would be beneficial for the doc explorer, maybe as one of the filtering/organizing options should we have something like a "search options" in the doc explorer. The said feature doesn't exist yet, but it's one of the things I'd be excited to add in the current GraphiQL if anyone's interested.

I'd love to hear what you think about this - I actually wasn't even aware of the fact that traversal ordering wasn't in the spec until fairly recently, so there's a possibility that I might be missing something. I did a quick experiment just now using the example schema and etc to see if what I said above about the ordering is true, and so far I haven't seen anything that goes against. But if you have some examples I'd love to see it - let's discuss a bit more if that's okay with you.

kevinsimper:
I understand that the JSON.parse should preserve the order but from a user perspective I am not sure that I would understand that I had to order from the server-side. Also the consumer of graphiql can often be a different developer/user than the person who exposes the graphql endpoint.

I think it should preserve the order on first load and have a button called `sort` that will sort it up and down.

dominicbarnes:
JavaScript will probably iterate in the order defined in the JSON body, but as mentioned before the JSON spec does not guarantee that, in fact it specifically says the opposite. So, it's odd to me that the GraphQL spec goes out of it's way to encourage implementations to do this, but that's a separate issue and beyond the scope of a UI change in this particular GraphQL client.

In Golang, serializing a JSON object from a map will result in the keys being sorted. (not in natural alphabetical order, but deterministically nonetheless) In that case, no matter what "order" you define your schema in, the default JSON marshaller will probably change that order, unless you implement some custom marshaller. This is 1 case where the original order may not be preserved, making GraphQL's assumption here shaky in my opinion.

Just because JS happens to follow this convention _now_, there is no guarantee that it always will and that should probably be regarded as a coincidence that could change at any time.

It sounds like the only way we would add sorting here is with a more complete UI that also handles filtering and perhaps a bit more. I don't have the time to implement something more sophisticated, so I will simply close this PR if that is the case.

AGS-:
Hey, I'm trying to sort through the open PR and since it doesn't seem like we could come to a good consensus on this I'm going to close it for now. I think this would be a nice feature to have if we ever got to the point that we have filtering and sorting options, but for now I think the current behavior is fine. Thanks for taking the time to contribute to GraphiQL!

PR #635, opened 2017-11-17
fix MenuItem example in README

mvayngrib:
- onSelect instead of onClick
- add labels, otherwise menus just appear as triangles


asiandrummer:
Thanks @mvayngrib!

Issue #88, opened 2016-02-04
GraphiQL expects 200 status for errors


sporto:
In order to render an error on the right panel, GraphiQL expects the response to have a 200 status. In case of a bad query,  the server needs to respond with 200 in order the see the error in the UI.

We had a case where bad queries were send to the server e.g. asking for something that doesn't exists. Looking at the server logs wouldn't show the issue as everything was 200. So we changed our graphql server to return 422 on bad queries.

But now GraphiQL doesn't show anything on the right, as it expects 200. 

So I am not sure about this, is this a graphql spec issue or should GraphiQL handle some errors and try to render the response?

asiandrummer:
Hi @sporto - sorry it took a while to get to this. I'm not sure if I understand the problem; when you say "bad queries", do you mean a invalid query with a correct syntax (e.g. a field that isn't defined is queried), or a query that returns `null` because the result was empty?

GraphiQL does not know about the network status/http status code/etc - instead it handles errors if the defined fetcher rejects the query and/or an error is thrown during the execution ([defined here] (https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L376)). So it is possible to render the error logs without a specific status code - [express-graphql implementation] (https://github.com/graphql/express-graphql/blob/master/src/index.js) is an example of this.

sporto:
hi @asiandrummer 	using graphiql 0.5 this is the behaviour I see.

Consider a query where one attribute doesn't exist e.g. `unknown` in the screenshot below

![qraphiql-200](https://cloud.githubusercontent.com/assets/1005498/13557616/70b72eea-e448-11e5-9ac5-11000e2f0381.png)

In this case I have setup my graphql server to return a 200 with the errors. GraphiQL shows the error on the right panel.

But returned 200 when an error happens is not what we think should happen. 

![qraphiql-422](https://cloud.githubusercontent.com/assets/1005498/13557655/1f4d1b22-e449-11e5-930f-bb93ed57fc2e.png)


So in the screenshot  above I have set up the server to return 422 with the error. GraphiQL just shows `[object Object]` on the right.

The expected behaviour would be to have GraphiQL show the error on the right if there is an error on the response body, even if the status is 422.

asiandrummer:
[...]

Issue #17, opened 2015-08-27
Expose errors in building schema


leebyron:
As #16 encountered, if an introspection result causes issues with building a client schema today we just silently continue without a schema - instead we should present some error so the developer can quickly find and fix the issue.

asiandrummer:
I'll get either a simple dialog or info button on the top bar to expose errors to begin the effort

skevy:
@asiandrummer @leebyron what about displaying a redbox error (like on React Native):

![Image of redbox](https://s3.amazonaws.com/f.cl.ly/items/0o232A1s2I421n191F1u/Screen%20Shot%202015-09-05%20at%2011.33.45%20AM.png)

skevy:
With this method, I feel like you could bubble up fatal errors from the lower components and display this as well.

kroepke:
Even logging something to the console can be helpful, in my case it was an incompatible field definition between an interface and an implementation type.

KyleAMathews:
:+1: this would be great. @skevy has a working version of this for the electron wrapped version of GraphiQL.

leebyron:
It's not quite true that without a schema nothing works. You can at least still use the editor with syntax highlighting and you can still send queries (we just won't validate them first). For that reason I'd rather not red-box the whole screen, but certainly outputting a clear error in the right pane is a good idea. 

Our current behavior of printing a JS error with stack trace is clearly lacking though.

From first-principles, GraphiQL's role is to be a valuable tool for exploring and testing queries against a GraphQL server. I believe that should include diagnosing a server which has a broken server. So GraphiQL should do it's best to report with clarity why a server's introspection result is considered broken (relevant code here: https://github.com/graphql/graphql-js/blob/master/src/utilities/buildClientSchema.js)

This has obviously not been a priority for Facebook because our server is never broken like this, but it's understandable that someone could encounter a broken server and having clearer messaging about how to fix it is high value.

joelgriffith:
Having an error in the right pane I think is a good idea initially, but if a user runs a query that pane will show new data (Unable to fetch or whatever).

I've been thinking it might be nice to signal a potential problem by decorating the "play"/run-query button, similar to how the prettify button hanges hue when there's syntax issues. Once the user hovers over execute query button, we can show a message underneath by using the tooltip. Obviously if a query returns results/200 we can remove any decoration.

I can try and tackle this over the next few days if this sounds like a decent approach @leebyron

vjpr:
[...]

PR #636, opened 2017-11-28
[WIP] Fix subscriptions errors output

lebedev:
Hi!

I've been using GraphiQL with `subscriptions-transport-ws` package and discovered, that in case of subscriptions if there're some issues on server side, GraphiQL shows `[object Object]` in output window, despite a message sent over WebSocket having correct stringified JSON.

This happens because in case of an error (of errors), the error (or `error.stack`) just gets String casted, and if there's no `.stack`, an error is effectively swallowed by showing `[object Object]` instead. And the only way to debug this kind of server issues is to look at WebSocket messages directly, which isn't very convenient.

There're 3 places in code where errors get String casted. Despite me actually tracing this `[object Object]` issue to only one place, I think that all of them could lead to such inconvenience.

I propose that something should be done with this kind of situations (and with errors just being String casted). My PR isn't merge-ready, it's a matter of discussion.

So, what do you think?

This issue is indirectly mentioned in #17 and #88, and mentioned in apollographql/subscriptions-transport-ws#236 and apollographql/subscriptions-transport-ws#305.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -530,7 +530,9 @@ export class GraphiQL extends React.Component {
       .catch(error => {
         this.setState({
           schema: null,
-          response: error && String(error.stack || error),
+          response: error && error.stack
+            ? String(error.stack)
+            : JSON.stringify(error, null, 2),
         });
       });
   }
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -563,7 +563,9 @@ export class GraphiQL extends React.Component {
       fetch.then(cb).catch(error => {
         this.setState({
           isWaitingForResponse: false,
-          response: error && String(error.stack || error),
+          response: error && error.stack
+            ? String(error.stack)
+            : JSON.stringify(error, null, 2),
         });
       });
     } else if (isObservable(fetch)) {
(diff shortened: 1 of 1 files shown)


angly-cat:
I wasn't sure about the format of errors, but I saw what Github's GraphQL endpoint returns, and judging by that, it returns an array of objects (stringified error messages from the server), even with a single error.

So I think that the corresponding code either receives an instance of `Error`, if an error happened on client, which should have `.stack` and can be `String` casted, or an array of objects came from server, which should be `JSON.stringify`ed before showing.

asiandrummer:
Hi @angly-cat - thanks for firing up the discussion. Your observation looks valid except for the case where the server is already serving up a JSON-stringified error message, which GraphiQL then would show a doubly-stringified error message.

I agree that we're probably operating on a non-standard feature such as `error.stack` (probably Mozilla only - haven't actually checked), which is why we're relying on the server side to provide either a `error.stack` property or an error message that can be `String` casted. Perhaps `subscriptions-transport-ws` needs to be updated with regards to how it serves up errors?

Neitsch:
@lebedev - I'd be fine with just JSON.stringify ing the whole error. Would you mind rebasing and doing so? :)

PR #639, opened 2017-12-11
Add react@^15.5.0 and react-dom@^15.5.0 as peer dependencies

keyworks:
There doesn't seem to be any breaking changes from 15.5.0 to 15.6.0 so thought I could add this in.


keyworks:
Actually. I don't need this anymore.

PR #650, opened 2018-01-15
Add request options

MateuszLeo:
Sometimes you need to choose request options, headers, token stuff like that, here is a feature
which makes it possible
image

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -7,137 +7,141 @@
 -->
 <!DOCTYPE html>
 <html>
-  <head>
-    <style>
-      body {
-        height: 100%;
-        margin: 0;
-        width: 100%;
-        overflow: hidden;
+
+<head>
+  <style>
+    body {
+      height: 100%;
+      margin: 0;
+      width: 100%;
+      overflow: hidden;
+    }
+
+    #graphiql {
+      height: 100vh;
+    }
+  </style>
+
+  <!-- Included here to allow for testing offline -->
+  <script src="./vendor/es6-promise.auto.js"></script>
+  <script src="./vendor/fetch.min.js"></script>
+  <script src="./vendor/react-15.4.2.js"></script>
+  <script src="./vendor/react-dom-15.4.2.js"></script>
+
+  <!-- Load each file individually for better debugging  -->
+  <!-- Note that in practice, a compiled graphiql.css is included -->
+  <link rel="stylesheet" href="css/app.css" />
+  <link rel="stylesheet" href="css/codemirror.css" />
+  <link rel="stylesheet" href="css/doc-explorer.css" />
+  <link rel="stylesheet" href="css/foldgutter.css" />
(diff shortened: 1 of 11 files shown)


test/index.html
```diff
@@ -7,137 +7,141 @@
...
-      if (parameters.variables) {
-        try {
-          parameters.variables =
-            JSON.stringify(JSON.parse(parameters.variables), null, 2);
-        } catch (e) {
-          // Do nothing, we want to display the invalid JSON as a string, rather
-          // than present an error.
-        }
-      }
+<head>
```
kevinsimper: This should be indented like before 👍

kevinsimper:
It is cool with the idea to add some options but I am not sure it fits into this repository as GraphQL does not have an opinion about authentication and authorization and with Graphiql you can easily provide your own fetcher to match your authentication scheme.

MateuszLeo:
Yes but we should be at least able choose methods and content type.

kevinsimper:
@MateuszLeo Yes, but isn't it possible for you to configure here?    https://github.com/graphql/graphiql/blame/master/README.md#L36-L37

gardner:
It is possible to configure this in code and to implement a UI for this. It is understandable this project wants to leave that aspect up to implementors. The reason this keeps coming up in issues and now a PR is that GraphiQL is shipped as part of other applications. One example is in postgraphile: https://github.com/graphile/postgraphile/issues/83

This means that some new GraphiQL users may not necessarily know what GraphiQL is or that it is even a separate project. We could leave the authentication up to the projects that ship GraphiQL or we could add basic UI that allows developers to hack in what they need. The swiss-army-knife nature of GraphiQL makes me think that the latter approach would be more fitting.

While I have your attention: Thanks for making GraphiQL. We use it every day.

acao:
@gardner features like this are why we have decided to go with a plugin system for GraphiQL 1.x. This feature in particular came up in the Working Group as a candidate for a plugin rather than a core feature. I don't think we will get this adopted in 0.x, unfortunately, but its not unlikely we will use some of this code when making a plugin for this in 1.x!

@benjie is a maintainer of GraphiQL and Postgraphile, hopefully the thread over there gave you some workable options.

Neitsch:
@MateuszLeo - would you still be interested in merging this PR? We will likely have to restructure it, since we are trying to move to a plugin driven architecture. A PR like this will help us figure out how to properly structure it 👍

acao:
@Neitsch we may end up with somewhat of a rewrite for 1.0.0-alpha. So this PR would be a good candidate for graphiql 0.x's LTS as we begin innnovating on the 1.0.0-alpha with the plugin system. This is probably one of the single most requested features

MateuszLeo:
@Neitsch yeah it would be great to merge it. I will go through this pr and resolve conflicts. Thanks for letting me know.

MateuszLeo:
@Neitsch @acao I've created new PR #923 for this change.

PR #660, opened 2018-02-12
Remove font-family for hover error message

binaryseed:
Hi, I came across this issue where an overridden style prevents the GraphQL error message from being displayed in `monospace`. This prevents the alignment of source code columns from working properly..

For example:

screen_shot_2018-02-12_at_3_26_54_pm

The illegal char is the silly "smart" quote, but the `^` points to the `m`

It works just fine if the font override is removed....

screen shot 2018-02-12 at 3 21 25 pm


acao:
A good version 0 bugfix. @orta  do you approve of the minor design change here?

orta:
Yes!

PR #662, opened 2018-02-20
[Proof of concept] Expose plugins prop

ismail-syed:
This is a proof of concept PR to expose an optional `plugins` prop for the GraphiQL component. With the use of the `plugins` prop, users of the GraphiQL component can create, share, and use `plugins` that can leverage the data returned in the `ResultsViewer`. 

A great use case of this is shown in this talk [here](https://www.youtube.com/watch?v=ykp6Za9rM58#t=27m13s). Using the [apollo-tracing](https://github.com/apollographql/apollo-tracing) GraphQL extension and the `plugins` prop in this PR, we can open doors to create plugins components that can provide teams/developer more insight into the data returned from the graphql apis. 

Here is an simple example of the `plugins` prop being used: 

screen shot 2018-02-19 at 7 41 32 pm
 
Let us know what you think!


src/components/GraphiQL.js
```diff
@@ -64,6 +65,12 @@ export class GraphiQL extends React.Component {
     editorTheme: PropTypes.string,
     onToggleHistory: PropTypes.func,
     ResultsTooltip: PropTypes.any,
+    plugins: PropTypes.arrayOf(
+      PropTypes.shape({
+        title: PropTypes.string.isRequired,
```
ismail-syed: * `title` is what will be displayed as the tab title in the `PluginsPane`. 
* `plugin` is a component with the render prop that passes in a function that takes the result string as an argument and returns whatever jsx it wants to be rendered.

test/index.html
```diff
@@ -26,6 +26,8 @@
     <script src="./vendor/react-15.4.2.js"></script>
     <script src="./vendor/react-dom-15.4.2.js"></script>
 
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-core/5.8.23/browser.js"></script>
```
ismail-syed: This is just for the context of the example.

test/index.html
```diff
@@ -125,6 +128,18 @@
         });
       }
 
+      const Plugin = <div />;
+      const viewer = function(results) {
```
ismail-syed: This is the function that takes in the results as a string. Just like the `value` in `ResultsViewer` [here](https://github.com/graphql/graphiql/blob/master/src/components/ResultViewer.js#L25)

test/index.html
```diff
@@ -134,7 +149,13 @@
           operationName: parameters.operationName,
           onEditQuery: onEditQuery,
           onEditVariables: onEditVariables,
-          onEditOperationName: onEditOperationName
+          onEditOperationName: onEditOperationName,
+          plugins: [
+            {
+              title: 'Example',
+              plugin: <Plugin render={viewer} />,
```
ismail-syed: This is just a dummy `Plugin` component, but in theory one could create multiple different types of Plugin components for different types. This could even be it's own library of components. 
eg. one with CodeMirror contained in the component

kevinsimper:
Hi @ismail-syed 

Just an outsider contributer, really cool idea with the idea of allowing plugins. I think there is a benefit to keep the core of GraphiQL small and as I see a wrapper could be made around `<GraphiQL/>` so that plugin support can be built as a standalone component. Then the GraphiQL should provide a callback that allows to get the response from the https request.

What do you think about this approach?

Also if there is a big plugin community is could be reevaluated again.

ismail-syed:
Thanks for the feedback @kevinsimper. Yes, in theory this can also be solved as a standalone component that wraps the GraphiQL and provides 3rd party plugin support via a 3rd party GraphiQL wrapper. However, we were interested to see if the community sees this as a valuable core feature. With first class plugin support, the community can be encouraged to build a suite a valuable plugins.  
Interested to hear what others think as well.

kevinsimper:
@ismail-syed Have you some examples of plugins that you have made? :)

PR #664, opened 2018-03-26
Add operation name to introspection query

jbblanchet:
Very simply PR to add the operation name to the initial IntrospectionQuery.

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -483,7 +483,12 @@ export class GraphiQL extends React.Component {
   _fetchSchema() {
     const fetcher = this.props.fetcher;
 
-    const fetch = observableToPromise(fetcher({ query: introspectionQuery }));
+    const fetch = observableToPromise(
+      fetcher({
+        query: introspectionQuery,
+        operationName: 'IntrospectionQuery',
+      }),
+    );
     if (!isPromise(fetch)) {
       this.setState({
         response: 'Fetcher did not return a Promise for introspection.',
@@ -502,6 +507,7 @@ export class GraphiQL extends React.Component {
         const fetch2 = observableToPromise(
           fetcher({
             query: introspectionQuerySansSubscriptions,
+            operationName: 'IntrospectionQuery',
           }),
         );
         if (!isPromise(fetch)) {
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -31,6 +31,7 @@ import { getLeft, getTop } from '../utility/elementPosition';
 import { mergeAst } from '../utility/mergeAst';
 import {
   introspectionQuery,
+  introspectionQueryName,
   introspectionQuerySansSubscriptions,
 } from '../utility/introspectionQueries';
 
@@ -486,7 +487,7 @@ export class GraphiQL extends React.Component {
     const fetch = observableToPromise(
       fetcher({
         query: introspectionQuery,
-        operationName: 'IntrospectionQuery',
+        operationName: introspectionQueryName,
       }),
     );
     if (!isPromise(fetch)) {
@@ -507,7 +508,7 @@ export class GraphiQL extends React.Component {
         const fetch2 = observableToPromise(
           fetcher({
             query: introspectionQuerySansSubscriptions,
-            operationName: 'IntrospectionQuery',
+            operationName: introspectionQueryName,
           }),
         );
         if (!isPromise(fetch)) {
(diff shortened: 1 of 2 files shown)


IvanGoncharov:
@jbblanchet Thanks for PR 👍 
Can you please describe your use case for this change?
Do you need it for logging, whitelisting or something else?

jbblanchet:
The exact use case for us was filtering out this query from server side logging and performance stats, but I can imagine other uses. And since the operationName is set correctly for every other query with one sent by GraphiQL, I assume this was simply an oversight.

IvanGoncharov:
> And since the operationName is set correctly for every other query with one sent by GraphiQL, I assume this was simply an oversight.

Makes a lot of sence 👍

src/components/GraphiQL.js
```diff
@@ -476,7 +476,12 @@ export class GraphiQL extends React.Component {
   _fetchSchema() {
     const fetcher = this.props.fetcher;
 
-    const fetch = observableToPromise(fetcher({ query: introspectionQuery }));
+    const fetch = observableToPromise(
+      fetcher({
+        query: introspectionQuery,
+        operationName: 'IntrospectionQuery',
```
IvanGoncharov: I think it's dangerous to hard-code the name of introspection query here.
Especially since `introspectionQuery` is imported from `graphql-js` which is one of `peerDependencies`.

It's make sense to add something like this into [introspectionQueries.js](https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js).
```js
export const introspectionQueryName = getOperationAST(introspectionQuery).name.value;
```

src/components/GraphiQL.js
```diff
@@ -495,6 +500,7 @@ export class GraphiQL extends React.Component {
         const fetch2 = observableToPromise(
           fetcher({
             query: introspectionQuerySansSubscriptions,
+            operationName: 'IntrospectionQuery',
```
IvanGoncharov: Same here but in this case, it can be hardcoded as `introspectionQuerySansSubscriptionsName` in `introspectionQueries.js`.

src/components/GraphiQL.js
jbblanchet: Ok, I'll update the PR.

jbblanchet:
@IvanGoncharov I've reused the operation name in the SansSubscription query itself, let me know if that's acceptable.

I strongly dislike the import and export of `introspectionQuery` in separate statements, but it seemed the cleanest way to do it.

acao:
I think this makes sense for better logging and visibility, and thus a good candidate to round out version 0

src/utility/introspectionQueries.js
```diff
@@ -4,14 +4,17 @@
  *  This source code is licensed under the MIT license found in the
  *  LICENSE file in the root directory of this source tree.
  */
+import { getOperationAST, parse, introspectionQuery } from 'graphql';
 
 export { introspectionQuery } from 'graphql';
+export const introspectionQueryName = getOperationAST(parse(introspectionQuery))
+  .name.value;
```
benjie: Can we guarantee that the `introspectionQuery` from `graphql` will always be named, @IvanGoncharov?

If not, we should add a fallback here:

```suggestion
  .name.value || "IntrospectionQuery";
```

benjie:
Ugh I didn't mean request changes.

src/utility/introspectionQueries.js
benjie: Can we guarantee that the `introspectionQuery` from `graphql` will always be named, @IvanGoncharov?

If not, we should add a fallback here:

```suggestion
  .name.value || "IntrospectionQuery";
```

IvanGoncharov: > Can we guarantee that the introspectionQuery from graphql will always be named?

Yes, I think we need consistent naming.
I will add `graphql-js` that will check this.

> If not, we should add a fallback here:

You can't do that. If the query doesn't have a name you can't just send `operationName` it will result in an error.

BTW. It's already ~2 years since subscription was added so it would be safe to drop `introspectionQuerySansSubscriptions` in the next major release.

benjie: > If the query doesn't have a name you can't just send operationName it will result in an error.

Oh yeah, I was thinking of the `introspectionQuerySansSubscriptions` query which we name ourselves, but you're right that we use the same variable with the normal introspection query too.

> safe to drop `introspectionQuerySansSubscriptions` in the next major release.

Noted.

Thanks Ivan!

Issue #665, opened 2018-03-27
ReferenceError: b is not defined


marcgreenstock:
Apologies if this is out of scope, but I have a bug in webpack 4 with `--mode production`.

It appears it is a result of the `optimization.minimize` config, turning minimize off fails to reproduce the issue. I couldn't identify which uglify option specifically causes it though.

```JSON
"dependencies": {
  "graphiql": "^0.11.11",
  "graphql": "^0.11.7",
  "codemirror": "^5.36.0",
  "codemirror-graphql": "^0.6.12",
  "react": "^16.2.0",
  "react-dom": "^16.2.0"
}
```

<img width="491" alt="screen shot 2018-03-27 at 14 49 06" src="https://user-images.githubusercontent.com/287739/37968311-119bf546-31ce-11e8-82c4-eee0cf4d0442.png">

jimmybrawn:
any updates on this?

marcgreenstock:
@jimmybrawn are you seeing this too?

jimmybrawn:
Yes and it drives me crazy.. Its only in production.

marcgreenstock:
@jimmybrawn I haven't resolved it, the best I could do was disabling `optimization.minimize` entirely. I've moved on from this for now. I will revisit this again once  #663 is resolved since my codebase depends on graphqljs v0.13 now.

whtsky:
running into the same problem with graphqljs 0.13 & graphiql 0.11.11

whtsky:
https://github.com/mishoo/UglifyJS2/issues/3010#issuecomment-373957512
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/265

jimmybrawn:
@whtsky so is it now fixed with uglify-es then?

whtsky:
@jimmybrawn nope. uglify-es will not be maintained so I forked uglifyjs-webpack-plugin to make it use uglify-js ( which does not support ES6 by the way): https://github.com/whtsky/uglifyjs-3-webpack-plugin

jimmybrawn:
So we've fixed it by doing the following in webpack 4:
```
 minimizer: [
      new UglifyJsPlugin({
        cache: true,
        parallel: true,
        uglifyOptions: {
          compress: false,
          ecma: 6,
          mangle: true
        }
      })
    ]
```
this seems to work on production

sekharp:
[...]

PR #669, opened 2018-04-11
fix: check `this.editor` exist before `this.editor.off` in QueryEditor

whtsky:
`this.editor` might be undefined if error occured during `componentDidMount`, like #665

### DIFF
--- a/packages/graphiql/src/components/QueryEditor.js
+++ b/packages/graphiql/src/components/QueryEditor.js
@@ -193,10 +193,12 @@ export class QueryEditor extends React.Component {
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


jimmybrawn:
Yes PLSS

whtsky:
this does not fix #665 though, just fixed another error caused by #665

jimmybrawn:
I had both error on the "off" function and the #665 which is an error caused by optimization.minimize in webpack 4 i suppose @whtsky

Neitsch:
@whtsky - sorry for not looking at this earlier. If this is still relevant, would you mind rebasing?

whtsky:
@Neitsch rebased.

PR #675, opened 2018-05-01
Fix persistent search box shortcut key to Alt+F

umlx5h:
[This pull request](https://github.com/graphql/graphiql/pull/600) implements a persistent search box and binds its shortcut to Ctrl or Cmd+F key.
But that conflicts Emacs cursor shortcut Ctrl+F on Mac. Mac users often use emacs keybindings.

CodeMirror extraKeys option seems to have no ability to distinguish between Mac and Windows.
So I changed the shortcut Ctrl or Cmd+F to Alt+F separately.
I think it seems naturally because [CodeMirror official demo](https://codemirror.net/demo/search.html) also binds its shortcut to Alt+F.

### DIFF
--- a/src/components/QueryEditor.js
+++ b/src/components/QueryEditor.js
@@ -129,8 +129,7 @@ export class QueryEditor extends React.Component {
         },
 
         // Persistent search box in Query Editor
-        'Cmd-F': 'findPersistent',
-        'Ctrl-F': 'findPersistent',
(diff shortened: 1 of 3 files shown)


Neitsch:
@umlx5h - not everybody uses emacs bindings. Let's stick with the previous default :) Feel free to reopen if you disagree

PR #689, opened 2018-06-22
fixed pipe character being removed when pasting

albertlockett2:
When some text gets pasted into the editor, the pipe character ( | ) would be removed from the text.

I think the regex that normalizes the whitespace uses the character set so it doesn't need the pipe to mean "match any character in the set"


AGS-:
Hey, thanks for the contribution. Sorry it took so long to respond!

PR #715, opened 2018-09-08
friendly hotkey for switch panels & schema search for non-english developer

timshaw9791:
1.set hotkeys for switch panels (history, docs , query variables) , it'll be friendly for our programmers.
ctrl+d for switching docs panel,
ctrl+h for switching history panel,
and i think you guess out,ctrl+q for switching query variables panel,

2.when a developer can't remember the mutation field name, he/she can search by a piece of description, may be useful for non-english developer since they write description with native language. yes. i'm chinese dev, and some of my guys want this feature definitely.
thinking further in the way,description in search result is also helpful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
   "dependencies": {
     "codemirror": "^5.26.0",
     "codemirror-graphql": "^0.6.11",
-    "markdown-it": "^8.4.0"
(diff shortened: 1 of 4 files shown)


timshaw9791:
image

PR #718, opened 2018-09-20
Add readOnly prop and pass to `QueryEditor` and `VariableEditor`

TheSharpieOne:
This change allows `readOnly` to be passed to GraphiQL (the default export) which will pass it to `QueryEditor` and `VariableEditor`. Both `QueryEditor` and `VariableEditor` already accept `readOnly`, this just exposes it.


acao:
I think this is a good version 0 candidate because it is essentially exposing functionality available in the underlying components. @benjie  @orta ?

benjie:
Definitely 👍

PR #722, opened 2018-10-03
Stringify subscription errors elegantly to avoid `[object Object]` errors

wawhal:


src/components/GraphiQL.js
```diff
@@ -574,7 +574,7 @@ export class GraphiQL extends React.Component {
         error: error => {
           this.setState({
             isWaitingForResponse: false,
-            response: error && String(error.stack || error),
+            response: this.makeSubscriptionErrorResponse(error),
```
Neitsch: Would you mind keeping the original null check for `error`?

acao:
there are several issues regarding this, would be good to get this one released very soon

benjie:
The idea of accessing `originalError` is an interesting one; but I'm going to leave it out of the default implementation in #934 - you can always override `GraphiQL.formatError`

Issue #741, opened 2018-11-08
Exposing variableEditorOpen


christieb7007:
Could we expose the variableEditorOpen option as a prop so that we can decide whether or not to show the variable editor by default?
It looks like it currently opens up if there are variables passed in, but would not open by itself otherwise
https://github.com/graphql/graphiql/blob/master/src/components/GraphiQL.js#L111

christieb7007:
I have committed the change to https://github.com/christieb7007/graphiql/commit/b86f3844a54db5a2a55a0f4fd125af01e086f592 but I cannot run the tests or start the app up locally (see https://github.com/graphql/graphiql/issues/318)

christieb7007:
successfully ran and passed tests

PR #744, opened 2018-11-14
expose variableEditorOpen to set a default value

christieb7007:
https://github.com/graphql/graphiql/issues/741


christieb7007:
can this be merged?

acao:
@christieb7007 looks good to me, other than the unneeded diff to `package.json`. can you remove that change from the PR?

christieb7007:
I've updated the PR so that the `variableEditorOpen` prop is evaluated in the `render()` method as well as update tests/README/example

christieb7007:
Apologies for the delay.
The functionality I was thinking of was actually just the `defaultVariableEditorOpen` - maybe the controlled component part can be in a separate PR.

packages/graphiql-examples/cdn/index.html
```diff
@@ -143,6 +143,7 @@
           operationName: parameters.operationName,
           onEditQuery: onEditQuery,
           onEditVariables: onEditVariables,
+          variableEditorOpen: true,
```
benjie:
```suggestion
          defaultVariableEditorOpen: true,
```

packages/graphiql/src/components/GraphiQL.js
```diff
@@ -120,6 +121,9 @@ export class GraphiQL extends React.Component {
       docExplorerOpen = this._storage.get('docExplorerOpen') === 'true';
     }
 
+    // initial variable editor pane open
+    const variableEditorOpen = props.defaultVariableEditorOpen !== undefined ? props.defaultVariableEditorOpen : Boolean(variables);
```
benjie: (Note to self: comparing against `undefined` should be safe in modern browsers; I think this is okay.)

packages/graphiql-examples/cdn/index.html
christieb7007: fixed

PR #759, opened 2018-12-30
preserve `ctrl-f` for macOS for moving forward

pd4d10:
1. Extract common keys to utility
2. Preserve `ctrl-f` for macOS for moving forward (Emacs style key binding)

### DIFF
--- a/packages/graphiql/src/components/QueryEditor.js
+++ b/packages/graphiql/src/components/QueryEditor.js
@@ -11,6 +11,7 @@ import { GraphQLSchema } from 'graphql';
 import MD from 'markdown-it';
 import { normalizeWhitespace } from '../utility/normalizeWhitespace';
 import onHasCompletion from '../utility/onHasCompletion';
+import commonKeys from '../utility/commonKeys';
 
 const md = new MD();
 const AUTO_COMPLETE_AFTER_KEY = /^[a-zA-Z0-9_@(]$/;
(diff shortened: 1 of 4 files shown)


src/utility/commonKeys.js
```diff
@@ -0,0 +1,16 @@
+// keep consistent with codemirror:
+// https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2/src/util/browser.js#L17
+const isMacOs = /Mac OS X 1\d\D([8-9]|\d\d)\D/.test(navigator.userAgent);
+
+const commonKeys = {
+  // Persistent search box in Query Editor
+  [isMacOs ? "Cmd-F" : "Ctrl-F"]: "findPersistent",
```
Neitsch: The previous behavior is definitely clowny. However, I don't quite trust the codemirror regex, since it only returns true for mountain lion and higher. Let's keep the former way for now.

src/utility/commonKeys.js
```diff
@@ -0,0 +1,16 @@
+// keep consistent with codemirror:
+// https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2/src/util/browser.js#L17
+const isMacOs = /Mac OS X 1\d\D([8-9]|\d\d)\D/.test(navigator.userAgent);
```
benjie: Rather than sniffing the user agent, let's just use [the `platform` API](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform):

```suggestion
const isMacOs = window.navigator.platform === 'MacIntel';
```

I don't think we need to worry about 'Mac68K' and 'MacPPC' which are other valid options for Mac platforms.

src/utility/commonKeys.js
benjie: Let's use `navigator.platform` - much less clowny :)

acao: this UA detection seems needless here. Just make two entries, one for Cmd-F and one for Ctrl-F, with the same value. Thats the usual codemirror keymap MO

acao:
ohhh i see. so if its mac, keep the default ctrl-f behavior?

benjie:
@acao Exactly; many Mac users (including, until recently, me) expect the standard Emacs-style bindings to work. It's safe to bind `Cmd-F` everywhere, but on Mac we shouldn't bind `Ctrl-F`.

src/utility/commonKeys.js
pd4d10: Updated

benjie:
@acao Would you like to dismiss your review and then merge? (@Neitsch, do you have any further feedback?)

PR #762, opened 2019-01-24
Add merge feature

RamonSaboya:
Create a button on the toolbar to defragment a query.

All fragments found on the editor pane will be stripped and their contents will be inserted as InlineFragment in all of their usage on queries and mutations.

This feature will improve readability and debugging, since it's hard to understand how strongly-fragmented queries relate to each other.

Before:
```
query TestQuery {
  ...Fragment
}

fragment Fragment on Test {
  id
}
```

After:
```
query TestQuery {
  ... on Test {
    id
  }
}
```

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -28,6 +28,7 @@ import debounce from '../utility/debounce';
 import find from '../utility/find';
 import { fillLeafs } from '../utility/fillLeafs';
 import { getLeft, getTop } from '../utility/elementPosition';
+import { mergeAst } from '../utility/mergeAst';
 import {
   introspectionQuery,
   introspectionQuerySansSubscriptions,
@@ -251,6 +252,11 @@ export class GraphiQL extends React.Component {
           label="Prettify"
         />
         <ToolbarButton
+          onClick={this.handleMergeQuery}
+          title="Merge Query (Shift-Ctrl-M)"
+          label="Merge"
+        />
+        <ToolbarButton
           onClick={this.handleToggleHistory}
           title="Show History"
           label="History"
@@ -336,6 +342,7 @@ export class GraphiQL extends React.Component {
                 onHintInformationRender={this.handleHintInformationRender}
                 onClickReference={this.handleClickReference}
                 onPrettifyQuery={this.handlePrettifyQuery}
+                onMergeQuery={this.handleMergeQuery}
                 onRunQuery={this.handleEditorRunQuery}
                 editorTheme={this.props.editorTheme}
               />
@@ -355,6 +362,7 @@ export class GraphiQL extends React.Component {
                   onEdit={this.handleEditVariables}
                   onHintInformationRender={this.handleHintInformationRender}
                   onPrettifyQuery={this.handlePrettifyQuery}
+                  onMergeQuery={this.handleMergeQuery}
                   onRunQuery={this.handleEditorRunQuery}
                   editorTheme={this.props.editorTheme}
                 />
@@ -694,6 +702,18 @@ export class GraphiQL extends React.Component {
     editor.setValue(print(parse(editor.getValue())));
   };
 
+  handleMergeQuery = () => {
+    const editor = this.getQueryEditor();
+    const query = editor.getValue();
+
+    if (!query) {
+      return;
+    }
+
+    const ast = parse(query);
+    editor.setValue(print(mergeAst(ast)));
+  };
+
   handleEditQuery = debounce(100, value => {
     const queryFacts = this._updateQueryFacts(
       value,
@@ -1018,6 +1038,8 @@ const defaultQuery = `# Welcome to GraphiQL
 # Keyboard shortcuts:
 #
 #  Prettify Query:  Shift-Ctrl-P (or press the prettify button above)
+#
+#     Merge Query:  Shift-Ctrl-M (or press the merge button above)
 #
 #       Run Query:  Ctrl-Enter (or press the play button above)
 #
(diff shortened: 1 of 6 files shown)


src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
     editor.setValue(print(parse(editor.getValue())));
   };
 
+  handleDefragQuery = () => {
```
asiandrummer: Feel like this logic could be under another file

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+    const editor = this.getQueryEditor();
+    const ast = parse(editor.getValue());
```
asiandrummer: What if `editor.getValue()` returns empty/null? We need to guard against that

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+
+    let operations = ast.definitions.filter(elem => {
+      return elem.kind !== 'FragmentDefinition';
```
asiandrummer: nit:
```
const operations = ast.definitions.filter(elem => elem.kind !== 'FragmentDefinition');
```

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+    });
+
+    const fragments = {};
```
asiandrummer:
```
const fragments = ast.definitions.filter(...).map(...);
```

src/components/GraphiQL.js
asiandrummer: Actually, we don't need to compute operations here at all - we could just filter at line 730 where we do actually need the operations. This is unnecessary

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+    ast.definitions
+      .filter(elem => {
+        return elem.kind === 'FragmentDefinition';
+      })
+      .forEach(frag => {
+        frag.kind = 'InlineFragment';
+        fragments[frag.name.value] = frag;
+      });
+    operations = operations.map(op => this.resolveDefinition(fragments, op));
```
asiandrummer:
```
const operations = ast.definitions.filter(...).map(op => this.resolveDefinition(fragments, op));
```

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+
+    ast.definitions = operations;
```
asiandrummer: I'd be more comfortable if we don't mutate `ast` - let's just create a new GraphQL ast tree from operations and print the value.

src/components/GraphiQL.js
```diff
@@ -694,6 +710,42 @@ export class GraphiQL extends React.Component {
...
+    editor.setValue(print(ast));
+  };
+
+  resolveDefinition = (fragments, obj) => {
+    let definition = obj;
+    if (definition.kind === 'FragmentSpread') {
```
asiandrummer: I think we have an enum value for `FragmentSpread` and keywords like these above. Let's use that instead

src/components/GraphiQL.js
RamonSaboya: Agreed

src/components/GraphiQL.js
RamonSaboya: Yes, you're right. And I just realized that we have a bug with this on the prettify feature

src/components/GraphiQL.js
```diff
@@ -250,6 +251,11 @@ export class GraphiQL extends React.Component {
           title="Prettify Query (Shift-Ctrl-P)"
           label="Prettify"
         />
+        <ToolbarButton
```
asiandrummer: Could we get a screenshot for this?

src/components/GraphiQL.js
```diff
@@ -250,6 +251,11 @@ export class GraphiQL extends React.Component {
...
+          onClick={this.handleMergeQuery}
+          title="Merge Query (Shift-Ctrl-M)"
+          label="Merge"
```
asiandrummer: Eh, I'm leaning more towards not having this as a button... I understand a button would help to show this feature exists, but:
- We need to consider the space at the top bar; it's not too bad right now, but I'd like us to think really carefully about which features to put in the top bar - it's gotta be a feature that sees the major usage overall.
- I don't know what to call this feature :D I definitely don't think "Merge" is a good title for this feature/button since it's "Merge fragments into queries", which is then too long. 

Could we keep this as a shortcut-only feature for now?

src/components/GraphiQL.js
```diff
@@ -694,6 +702,18 @@ export class GraphiQL extends React.Component {
     editor.setValue(print(parse(editor.getValue())));
   };
 
+  handleMergeQuery = () => {
+    const editor = this.getQueryEditor();
+    const query = editor.getValue();
+
+    if (!query) {
+      return;
```
asiandrummer: This is too forgiving - if merging fragments into a query isn't possible, we need to either show that it's not possible and/or disable this feature. A couple of scenarios where this feature isn't meaningful imo:
- No query string, like this code
- A query string without fragments - no fragments to merge then
- An invalid query string - this will break the query editor and ultimately GraphiQL as a whole

I think we **at least** need to handle the last "invalid query string" scenario. Could we do:
- catch the invalid query string and either just return (if we're getting rid of the button) and/or
- change the UI (I think "prettify" feature does something to make the button red)?

src/components/GraphiQL.js
```diff
@@ -694,6 +702,18 @@ export class GraphiQL extends React.Component {
...
+    }
+
+    const ast = parse(query);
```
asiandrummer: This will throw if the query is invalid - please handle this case by `try/catch`ing and consider implementing the suggestion above :)

AGS-:
Merging!

PR #764, opened 2019-02-04
fix(graphiql): better quota management

jbblanchet:
While I understand the intent of the original code, the problem with "acknowledg[ing] QuotaExceededError only if there's something already stored" is that no one upstream is managing the exception, so the result right now if I fill up my quota with queries (which will happen on very large queries) is that the page crashes. This is not an ideal solution, because as a user I have no clear message that my query wasn't saved, but it's a better experience right now than a blank page with now explanation.

Ideally the code would be smart enough to remove the oldest items in the history one at a time and retry saving the queries, but the StorageAPI is not the place to put this. I can look into it if you guys are interested in this feature, but I don't want to put in the time you're not.


src/utility/StorageAPI.js
```diff
@@ -26,8 +26,10 @@ export default class StorageAPI {
     if (this.storage) {
       const key = `graphiql:${name}`;
       if (value) {
-        if (isStorageAvailable(this.storage, key, value)) {
```
Neitsch: Clearly it is bad that it silently fails and that it literally writes the item already. However, could we maybe design a friendlier message? Some ideas:
* Tell the user to clear their local storage
* Tell the user that no storage is available and ask if we can clean up some old queries in his storage.

benjie: I always assumed that the GraphiQL history was LRU; I think we should just delete old queries to make space for new ones (with the exception of starred queries, which should never be deleted). I don't think we need to ask permission. We should also probably have a byte-limit on the size of an actual item; if it's over, say, 100kB I don't think it should even attempt to be stored.

benjie:
Hey @jbblanchet; we're really sorry it took so long to review your PR ─ this repo is now under new management (namely the GraphQL Foundation) with a new team of maintainers so hopefully we'll be keeping things much more timely once we're up to date with the backlog. Are you interested in updating the PR based on the feedback above?

jbblanchet:
So the logic would be:

```
if (value.length > 50000) { // not sure if memory representation in memory is constant across browsers, so I'd go on a string length condition instead of size
  // return without attempting to save
} else {
  try {
    storage.setItem(key, value);
  } catch (e) {
    // on error, check if it's a quota thing
    if (isQuotaError(e)) {
      try {
        removeOldestEntry()
        // try again once? Or do we want to loop on this?
        storage.setItem(key, value);
      }
      catch (e) {
        // fail silently?
      }
    }
  }
}
```

Is that the behavior you'd expect?

benjie:
Awesome; thanks for getting back to me so fast!

> string length

Yep; this is fine. I'm not sure what a reasonable number is; lets err on the side of "too big" and go for `100000` (I understand this could be more than 100kB due to unicode).

> try again once?

Let's do 5 attempts so long as we can trust the `isQuotaError(e)`. That should deal with fairly standard variance in query sizes.

> fail silently

Sounds reasonable.

---

For code quality, try and ensure there's only one line that calls `storage.setItem(...)`.

jbblanchet:
Ok, I've asked my boss for a little time to look into this, I should have something hopefully early next week.

jbblanchet:
@benjie I'm not sure there's a good way to have automated tests for this since there's no dependency injection nor browser tests. I'm tested most of it in both Chrome and Firefox, still working on breaking that quota.

The addition of favorites makes it a little awkward, because now I'm potentially removing favorites, but without a re-architecture that would let me pass messages up and down the stack to the user, I don't think there's a good way to do this (i.e. ask for confirmation and stuff). The 100000 character limit should make it easier to stay below the quota, which is I think 5MB on Chrome and 10MB on Firefox.

benjie:
@jbblanchet Starred queries should never be deleted. If there's only starred queries left we should just give up trying to save new ones.

We're looking at potentially adding Cypress tests but we've not got around to implementing them yet.

jbblanchet:
@benjie I'm adding some unit tests, but ideally I'd be able to use mocks or stubs for the storage. I can't find anything for that in the current test stack. I'm tempted to add sinon and sinon-chai, but I want to make sure it's ok first.

Also, tests are using chai and that means a lot of `not.toBe.undefined` and such, which triggers the `no-unused-expressions` eslint rule, which is defined in `/packages/graphiql/eslintrc.yml`. Normally I would use a file pattern override to disable that rule for `*.spec.js` files, but I can't find how to do that in yml config files. I've disabled the rule in my own new test files with a comment for now, if you have a better way to solve that let me know.

packages/graphiql/src/utility/QueryStore.js
```diff
@@ -64,13 +65,38 @@
...
+    const items = [...this.items, item];
+
+    if (this.maxSize && items.length > this.maxSize) {
+      items.shift();
+    }
+
+    let done = false;
+    let success = false;
+    let tries = 0;
+    while (!done) {
```
Neitsch: Can we use a for loop here to make the termination a little clearer? i.e.
```
for(let i = 0; i < 5; i++) {
  const response = this.storage.set(this.key, JSON.stringify({ [this.key]: items }));;
  if (response.error === null) {
    this.items = items;
  } else if (response.quotaError) { // Why are we nullchecking `maxSize` here? I think we can skip that check?
    items.shift();
  } else {
    return; // We don't know what happened in this case, so just bailing out
  }
}

File: packages/graphiql/src/utility/QueryStore.js [hunk 3]
```diff
@@ -64,13 +65,38 @@
...
+        done = true;
+      }
+    }
+
+    if (success) {
+      this.items = items;
+    }
   }
 
   save() {
```
Neitsch: Can we delete the `save` method now or is it still used someplace else?

packages/graphiql/src/utility/__tests__/QueryStore.spec.js
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      expect(store.items.length).toBe(20);
+      // keeps the more recent items
+      expect(store.items[0]).toBe('item80');
+      expect(store.items[19]).toBe('item99');
+    });
+
+    it('tries to remove on quota error until it succeeds', () => {
+      let shouldTrow;
```
benjie: Typo:
```suggestion
      let shouldThrow;
```

packages/graphiql/src/utility/QueryStore.js
jbblanchet: Null check is for favorites store. I'm using maxSize as a proxy for knowing if I'm in a LRU (queries) or a persistent store (favorites). I'll update the code.

packages/graphiql/src/utility/QueryStore.js
jbblanchet: `save` is still used by the `delete` and `edit` functions since we don't check for quota on these.

packages/graphiql/src/components/QueryHistory.js
```diff
@@ -11,6 +11,9 @@ import PropTypes from 'prop-types';
 import QueryStore from '../utility/QueryStore';
 import HistoryQuery from './HistoryQuery';
 
+const MAX_QUERY_SIZE = 100000;
+const MAX_HISTORY_LENGTH = 20;
```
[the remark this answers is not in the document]
jbblanchet: I agree it should be user configurable. I also think that eventually we should have user-definable error handlers so they can decide what to do on quota errors and such, or even make the entire history a separate module so users can decide to save the history and favorites to a database instead of inside the browser, so it persists across browsers or organizations. But that's out of the scope of this PR :smile:

jbblanchet:
Implemented required fixes and answered questions. Should be ready to merge at your convenience.

benjie:
Awesome; I aim to review tomorrow, but failing that it'll be first half of next week unless someone gets to it before me :+1:

packages/graphiql/src/components/QueryHistory.js
```diff
@@ -53,7 +58,7 @@ export class QueryHistory extends React.Component {
 
   constructor(props) {
     super(props);
-    this.historyStore = new QueryStore('queries', props.storage);
+    this.historyStore = new QueryStore('queries', props.storage, MAX_HISTORY_LENGTH);
     this.favoriteStore = new QueryStore('favorites', props.storage);
```
benjie: I just noticed that the favoriteStore doesn't get passed the length, so it can grow unbounded (I think)?

```suggestion
    this.favoriteStore = new QueryStore('favorites', props.storage, MAX_HISTORY_LENGTH);
```

Are there any more QueryStore's around?

jbblanchet: Favorite is a not LRU cache. From your requirements up in this thread: "Starred queries should never be deleted. If there's only starred queries left we should just give up trying to save new ones." That's my there's not a max length.

There is no other QueryStory.

jbblanchet:
@benjie Question answered and conflict resolved.

benjie:
Awesome; I plan to review this on Monday.

packages/graphiql/src/components/QueryHistory.js
benjie: Oh, of course. Let's explicitly pass a value here (null? Infinity?) and maybe add a comment to explain why?

jbblanchet: Done.

PR #774, opened 2019-03-01
Capture Cmd/Ctrl-G

captbaritone:
My usage of search in Chrome/Firefox is generally:

1. Cmd-F to open the search input
2. Type my query
3. Type Cmd-G to iterate through the search results

When I do this out of muscle memory in GraphQL, I get suck since the
`Cmd-G` gets captured by the browser and puts my focus in the browsers
search input.

With this fix, my normal flow works as expected.

Tested on Chrome/Mac


asiandrummer:
Not sure if it's feasible to have two shortcuts for one feature, but if this makes your workflow easier.... :)

PR #776, opened 2019-03-11
Capture and disable Cmd/Ctrl-S

cvqprs:
Steps to reproduce:
1. Open editor
2. Type query
3. Press Cmd/Ctrl+S

Actual result:
Lots of people (including me) have a habit to press Cmd/Ctrl+S when writing queries or editing stuff. 
In graphiql query editor, after every Cmd/Ctrl+S browser Save as prompt will be open.

Expected result:
This simple PR is going to capture and disable Cmd/Ctrl+S to prevent Save as prompt. I find it super annoying when I have Save as prompt open multiple times during my work. I know that we can avoid this simply pressing execute button, but sometimes it is hard to get rid of this kind of habits.


acao:
I think the only thing to remove are the `if (this.props.onRunQuery)` conditions, since we are trapping these key combinations always with non-behavior, they aren't really serving a purpose.  Otherwise this makes sense, I would actually consider it a bugfix because I've had the exact same issue you've mentioned with having too many save dialogs

benjie:
I was thinking about this (because I have the exact same issue) and was wondering if maybe Ctrl-s should trigger the star action in history? That’s like a save. I worry that history could get flooded if we do this though, which is especially problematic if people don’t even use the history button.

acao:
sorry had to push to master to merge this the right way, the conflict was very simple to resolve, but it looks like we had approved this so yeah!

Issue #778, opened 2019-03-14
Regression: explorer title is not selectable anymore


kayhadrin:
I had made the explorer title in the "Docs" section selectable back in 2017 (see #373) but it seems that some new changes have reverted it. :-/

acao:
thanks @kayhadrin for your PR! it's been merged and released!

PR #232, opened 2016-12-06
History pane

tessalt:
graphql queries can be verbose, I find I have a scattered collection of queries saved in notes.app when I'm debugging an API 🙃. So I added a "history" panel like Postman has so I can easily re-run queries. 

I'm not a designer and I'm not sure this is the best UX, but I've tried to make it as unintrusive as possible. I've had to re-arrange the toolbar a bit to fit it in logically. 

image

image

Also wrapped the localStorage access in a class so it can be shared between user settings and storing queries. 

@asiandrummer

PR #779, opened 2019-03-14
Fix: explorer title is not selectable anymore #778

kayhadrin:
Fix issue that was introduced in 0e85c75 (#232)

### DIFF
--- a/css/doc-explorer.css
+++ b/css/doc-explorer.css
@@ -21,7 +21,7 @@
   padding: 10px 0 10px 10px;
   text-align: center;
   text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


kayhadrin:
Ping ping ping! :-)

kayhadrin:
@captbaritone EZ fix :-)

daern91:
Thanks a lot, @kayhadrin! We are also facing this problem and can confirm that this PR solves it.

ping @AGS- and @asiandrummer

orta:
Alright, let's get this in - feels like a safe PR. :+1:

PR #789, opened 2019-04-18
[rfc] Add image preview functionality

dunnbobcat:
This adds a tooltip-like component for previewing images in GraphQL
responses. The `ImagePreview` component tries to detect if a string is an
image URL and, if so, renders the image plus some metadata (dimensions
and mime type).

I found that if you specify a `ResultsTooltip` on the `ResultViewer`, it
unconditionally renders no matter what. So I added a hacky way to
prevent that from happening (the static `shouldRender` function). I don't
really like this, but I don't see an obvious workaround without
introducing a new codemirror addon.

To test this, I added an image asset (the GraphQL logo in SVG) and hard-
coded it in the schema as `test.image`. Then I fired up GraphiQL and 
moused over the image URL in the results panel.


example/schema.js
```diff
@@ -26,16 +26,16 @@ const TestEnum = new GraphQLEnumType({
   values: {
     RED: { description: 'A rosy color' },
     GREEN: { description: 'The color of martians and slime' },
-    BLUE: { description: 'A feeling you might have if you can\'t use GraphQL' },
-  }
+    BLUE: { description: "A feeling you might have if you can't use GraphQL" },
```
dunnbobcat: Not sure why all these formatting changes got pulled in…

PR #791, opened 2019-04-30
hint DOM nodes liveoutside the codemirror DOM tree.

codeStryke:
The problem I'm having is that the `hint` DOM elements get created as children of `<body>`. If we pass in the container, they are created as children of the container which is more appropriate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     ]
   },
   "dependencies": {
-    "codemirror": "^5.26.0",
+    "codemirror": "^5.47.0",
     "codemirror-graphql": "^0.8.3",
     "markdown-it": "^8.4.0"
   },
--- a/src/components/QueryEditor.js
+++ b/src/components/QueryEditor.js
@@ -106,11 +106,11 @@ export class QueryEditor extends React.Component {
       },
       gutters: ['CodeMirror-linenumbers', 'CodeMirror-foldgutter'],
       extraKeys: {
-        'Cmd-Space': () => this.editor.showHint({ completeSingle: true }),
-        'Ctrl-Space': () => this.editor.showHint({ completeSingle: true }),
-        'Alt-Space': () => this.editor.showHint({ completeSingle: true }),
-        'Shift-Space': () => this.editor.showHint({ completeSingle: true }),
-        'Shift-Alt-Space': () => this.editor.showHint({ completeSingle: true }),
+        'Cmd-Space': () => this.editor.showHint({ completeSingle: true, container: this._node }),
+        'Ctrl-Space': () => this.editor.showHint({ completeSingle: true, container: this._node }),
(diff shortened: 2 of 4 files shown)


codeStryke:
CC @dunnbobcat @RamonSaboya @AGS-

codeStryke:
ping @IvanGoncharov, can I get some eyes on this please 🙏

acao:
@codeStryke I'll take a look today! thanks for this.
we plan on releasing soon

acao:
@codeStryke off the cuff, is the `codemirror` upgrade a requirement? we will need a lockfile change, or this will fail the `yarn check --integrity` we added to  CI yesterday

package.json
```diff
@@ -51,7 +51,7 @@
     ]
   },
   "dependencies": {
-    "codemirror": "^5.26.0",
+    "codemirror": "^5.46.0",
```
acao: is the `codemirror` upgrade a requirement? we will need a lockfile change, or this will fail the `yarn check --integrity` we added to  CI yesterday

codeStryke:
@acao thanks for looking into it. I did some more testing locally and we need to hold off until https://github.com/codemirror/CodeMirror/pull/5857 is released. It will probably be `5.47.0`. I can ping again when it is ready. Sorry for the trouble

acao:
@codeStryke no bother! This is important as it potentially impacts many use cases in the wild. I'll gladly review again once it lands.

package.json
codeStryke: @acao I've updated the lock file

codeStryke:
@acao this is ready for review again. I've updated the dependency on CodeMirror to the latest version `5.47.0`.

acao:
thanks! @benjie I wonder if this would be a breaking change for anyone downstream?

acao:
@codeStryke can you also rebase with the latest upstream master? it looks like your lockfile is out of date

benjie:
@acao I can’t think of a non-contribed situation where this would be a breaking change; I’m all for it.

codeStryke:
@acao I've rebased on latest master. The latest commit on master for me locally (25006623dc56c78259cd1ab1401eea6b96343b42) matches what I see on github. Could the problem be with my yarn version? I am on 1.10.1

acao:
@codeStryke my bad, yeah its fine, all those new lines in the lockfile are downstream dependencies of `codemirror`, so they should be expected. gradually adding the integrity hashes still it looks like

acao:
thanks @codeStryke! I'll be sure to reference this PR once we get a release out on npm

codeStryke:
@acao thanks!!

codeStryke:
@acao any idea when the next release will be?

benjie:
The transfer of the ability to publish modules to the GraphQL foundation is still underway, but the wheels are in motion. There is no ETA.

acao:
This was realeased on June 21st fyi

PR #810, opened 2019-05-09
upgrade husky

acao:
- upgrade `husky` to 2.2
- switch to `husky.hooks` and raw githooks as per husky@0.14+


acao:
as per discussion with @IvanGoncharov, another possibility would be to remove `husky` entirely

benjie:
+1 for removing husky; it's great for company-wide tooling, but I think it may be a little unexpected tweaking things on install for arbitrary contributors.

acao:
@benjie closed for #817

Issue #743, opened 2018-11-13
valid license declaration within package.json


pli2014:
Please use "license": "MIT" description instead of "license": "SEE LICENSE IN LICENSE" within package.json. following standard declaration.

imolorhe:
@acao Closed!

acao:
@imolorhe thank you!

PR #814, opened 2019-05-10
Sets MIT in the package.json now that GraphiQL is MIT

orta:
Now that the license is MIT, then I think this should be fine to move back to MIT so tooling can pick it up.


acao:
solves https://github.com/graphql/graphiql/issues/743

acao:
@orta you can do the honors!

orta:
<3

PR #819, opened 2019-05-13
Upgrade to react 16 in tests

acao:


benjie:
I’m concerned this risks us accidentally relying on React 16 features, and thus breaking backwards compatibility without realising. I think we should update to 16 in a new major and switch to using hooks, but for now I’m not sure of the impetus for this change?

acao:
Good point! We currently support peerDependencies of 15 and 16, but you're right, we should keep the testing react version at 15. Using hooks would mean dropping support for peers below 16, which we should definately wait on. 

I was doing this as the first step towards switching to jest/enzyme (I have a working local branch) but we should definately keep the test version of react at 15 til we decide to make a major breaking version! Good point! What would we do about peerDependencies in that case?

benjie:
👍 I think the peerDependencies can remain unchanged.

PR #823, opened 2019-05-13
remove browserify-shim

acao:
- codemirror is commonjs now.
- manually tested with minified version.
- annoying for development with linked packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,10 +24,6 @@
     "README.md",
     "LICENSE"
   ],
-  "browserify-shim": {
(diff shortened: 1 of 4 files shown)


orta:
The Jest PR made this need a rebase I'm afraid

acao:
@orta good, it should! any time we merge something with a lockfile change this should happen. we should actually put this on hold, as I'm not 100% certain this won't be a breaking change. I'm looking towards `react-apollo`'s rollup config as our potential endgame here.

PR #824, opened 2019-05-14
switch out mocha for jest

acao:
Why jest and enzyme? Because it the test harness/renderer pair that react developers expect nowadays

- upgrade test version of react to 15.6.2
- replace mocha bootload with enzyme adapter config
- port test pattern directly, for simplicity
- configure eslint for jest
- add test specific babel config, because ie 9/ios 9 browser targets break babel-jest
- replace mocha config with jest config
- move mock document.createRange to enzyme config


orta:
Yep. I think there's no reason to delay on this one 👍

PR #831, opened 2019-05-17
Add unit test coverage for DocExplorer components

acao:
Seemed like a good place to start
- create ExampleSchema for tests
- full coverage for `TypeLink`, `FieldDoc`
- extensive coverage for `TypeDoc`
- starting point for `DocExplorer`

### DIFF
--- a/src/components/DocExplorer/__tests__/TypeDoc-test.js
+++ b/src/components/DocExplorer/__tests__/TypeDoc-test.js
@@ -1,0 +1,127 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { expect } from 'chai';
+
+import { GraphQLString } from 'graphql';
+
+import TypeDoc from '../TypeDoc';
+
+import {
+  ExampleSchema,
+  ExampleQuery,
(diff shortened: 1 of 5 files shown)


src/components/DocExplorer/__tests__/TypeDoc-test.js
```diff
@@ -0,0 +1,72 @@
...
+import {
+  ExampleSchema,
+  ExampleQuery,
+  ExampleUnion,
+  ExampleEnum
+} from '../../__tests__/ExampleSchema'
+
+describe('TypeDoc', () => {
+  it('renders a top-level query object type', () => {
+    const onClickType = jest.fn();
```
benjie: Would be nice to have an assertion on this.

acao: i added that because its a required prop, that would essentially be searching `TypeLink`, but why not, sure!

acao: i do need an assertion for `onClickField` thoough

PR #832, opened 2019-05-19
Add a toolbar button to copy the existing query to clipboard

jaebradley:
### Summary

This PR adds a `Copy` button to the default toolbar.

image

Often times a query or mutation executed in GraphiQL will be copied to another destination. 

By adding a `Copy` button (as well as a keyboard shortcut), this allows a user a quick way of copying the specified query / mutation (in all fairness, `Command + A, Command + C` is not exactly a particularly time-intensive set of operations).

### Discussion

I can see a number of reasons why a `Copy` button in and of itself might be ambiguous (is this copying the query? the response?). That being said, I still think that there is _some_ utility in adding this button.

I can also see an argument to label the button `Copy Query`, but I was also trying to stay consistent with the `Prettify` and `Merge` buttons.

I also made the decision to copy a formatted query to clipboard - I thought it made the most sense from the end user's perspective.

Finally, to support copying to clipboard, I added [the `copy-to-clipboard` `npm` package](https://www.npmjs.com/package/copy-to-clipboard). It seems to be widely used, have good browser support, and a [reasonable bundle footprint](https://bundlephobia.com/result?p=copy-to-clipboard@3.2.0).

### TODO

- [x] Complete the CLA

### DIFF
--- a/src/components/GraphiQL.js
+++ b/src/components/GraphiQL.js
@@ -9,6 +9,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
 import { buildClientSchema, GraphQLSchema, parse, print } from 'graphql';
+import copyToClipboard from 'copy-to-clipboard';
 
 import { ExecuteButton } from './ExecuteButton';
 import { ImagePreview } from './ImagePreview';
@@ -57,6 +58,7 @@ export class GraphiQL extends React.Component {
       removeItem: PropTypes.func,
     }),
     defaultQuery: PropTypes.string,
+    onCopyQuery: PropTypes.func,
     onEditQuery: PropTypes.func,
     onEditVariables: PropTypes.func,
     onEditOperationName: PropTypes.func,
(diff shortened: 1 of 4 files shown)


src/components/GraphiQL.js
```diff
@@ -741,6 +749,23 @@ export class GraphiQL extends React.Component {
...
 
+  handleCopyQuery = () => {
+    const editor = this.getQueryEditor();
+    const query = editor.getValue();
+
+    if (!query) {
+      return;
+    }
+
+    const formattedQuery = print(parse(query));
```
benjie: Since the user can click prettify and then copy easily, and since prettify currently drops comments, I suggest you just copy the query without formatting.

acao:
@benjie  i think this is another great example of something that makes sense as a plugin. worth considering that this could be provided via an option that's disabled by default? either way, very well done! i think this is a great case for wrapping our heads around how we would be able to provide this via a plugin in the new 1.x version.

personally, I'm not averse to adding simple little (especially optional) features like this or auto-prettify to the 0.x version, and then providing them via plugins in the 1.x version

benjie:
This could be accomplished by passing the copy button externally using the existing component interface; but I think it is worth the increased bundle-size to add it to core. Definitely a plugin in v1 though 👍

jaebradley:
@acao and @benjie thanks for the reviews! Should I implement making this an optional feature?

benjie:
I don't think that's necessary; people can already opt out by overriding buttons. @acao ?

acao:
@benjie  @jaebradley sorry for the delay guys, I concur with @benjie ! I reviewed the addition of `copy-to-clipboard` and it looks fine by me.

Issue #835, opened 2019-06-01
🚀 IDE Monorepo or bust! 🚀


acao:
So, as per discussions amongst the GraphiQL Maintainers Team, we are about to move towards a monorepo pattern, whereupon:

- `graphiql`
- `graphql-language-service` - a monorepo with:
  - `graphql-language-service-parser`
  - `graphql-language-service-server`
  - `graphql-language-service-interface`
  - `graphql-language-service-types`
  - `graphql-language-service-utils`
- `codemirror-graphql`

Will all be combined into this repo, as a lerna + yarn workspaces monorepo, living side by side.

We will migrate all git and github history from the two projects to this repository, and `graphiql` will be the reference implementation for the official GraphQL Foundation-owned IDE ecosystem.

Package names will be retained, however we are likely going to bump the next major versions of all of them to `14.0`, to match `graphql-js`.

This is all precedent to the forthcoming `graphiql` 1.0 effort (maybe graphiql 14.0 ?), though discussions and deliberation will continue concurrently to this effort.

Most of this was discussed at the last GraphiQL WG meeting, however if anyone has concerns, we can deliberate on it again.

We will provide instructions on how to port your PRs from other repositories to this one.

Here is a preview, currently with just `graphiql` and `codemirror-graphql`:
https://github.com/acao/graphiql/tree/ide-monorepo

PR #839, opened 2019-06-06
GraphQL IDE monorepo

acao:
This combines `codemirror-graphql` and `graphql-language-service` with this repo as per the plan laid out in https://github.com/graphql/graphiql/issues/835

All the original refs should be there, including for files who were moved and renamed and in some cases reformatted to make this possible.

Viewing the actual file tree is much easier to look at:
https://github.com/acao/graphiql/tree/ide-monorepo

*NOTE*: As @benjie stated in the WG, we will be open for one week of feedback on the monorepo migration starting today.  If we don't hear anything, we will get ready to go live with this beginning Thursday, June 13th.


orta:
Looking good 👍

acao:
thanks @orta !

packages/graphiql/example/index.html
```diff
@@ -107,9 +107,12 @@
       // use fetch, and could instead implement graphQLFetcher however you like,
       // as long as it returns a Promise or Observable.
       function graphQLFetcher(graphQLParams) {
-        // This example expects a GraphQL server at the path /graphql.
+        // When working locally, the example expects a GraphQL server at the path /graphql.
+        // In a PR preview, it connects to the Star Wars API externally.
         // Change this to point wherever you host your GraphQL server.
-        return fetch('/graphql', {
+        const isDev = window.location.hostname === 'localhost'
```
benjie: I normally alias my `localhost:1234` addresses to e.g. `graphiql.meh` using [mehserve](https://github.com/benjie/mehserve) (and I don't think I'm the only person who does similar, there was a lot of anger when `.dev` was registered by Google!). Can we use an allow-list approach rather than a block-list approach? Is there a regexp we can use against the domain that'll recognise netlify?

```suggestion
        const isDev = !window.location.hostname.endsWith('.netlify.com')
```

acao: agreed. as in my PR review as well, we need to extract this example and/or somehow leverage swapi-graphql.

acao: this hostname may not end with `.netlify.com` - for example, we might have `graphql.org/swapi` be a consumable GraphQL Service soon

benjie: `/(^|\.)netlify\.com$|(^|\.)graphql\.org$/`? We can always update it later, I assume?

Issue #834, opened 2019-05-30
Live demo link is broken in README


RussianCow:
[The link](https://graphql.org/swapi-graphql) gives a Page Not Found error:

![image](https://user-images.githubusercontent.com/1589155/58652327-25289f00-82c8-11e9-9caf-ece177579fb5.png)

acao:
Oh dear, that's a new one. Possibly part of the facebook -> graphql foundation transition. I'll ask around.

acao:
Can you report that here @RussianCow ? https://github.com/graphql/swapi-graphql
Might have to do with updated deps

PR #845, opened 2019-06-08
Fix swapi demo link in README

corysimmons:
https://graphql.github.io/swapi-graphql/ works. Old one doesn't. It's taken more than 5 seconds to fix. Use this one instead, even if it's just temporary until you fix the old one.

Broken for at least 9 days: https://github.com/graphql/graphiql/issues/834

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 GraphiQL
 ========
 
-*/ˈɡrafək(ə)l/* A graphical interactive in-browser GraphQL IDE. [Try the live demo](http://graphql.org/swapi-graphql).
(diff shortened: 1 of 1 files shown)


acao:
thanks @corysimmons , this might be a good temporary fix.

The official issue we are tracking right now is here:
https://github.com/graphql/swapi-graphql/issues/143

I got the github.io link working myself, and we are currently trying to track down the related DNS issues, and see which organization owns what. Very slow going and the active maintainers have limited access. Just a hiccup in the transition to GraphQL Foundation ownership[

acao:
I think I can fix this on graphql.org for now as well

PR #860, opened 2019-06-19
improved root readme with headers and badges

acao:
also, GLS doesnt have its own build status anymore

pretty, huh? https://github.com/graphql/graphiql/blob/59c1393c999bf6aa28e413a3c648718a3f5a315a/README.md


orta:
Yep, this looks good to me - let's do it!

PR #871, opened 2019-06-20
move GLS tests to jest, add codecov

acao:
- [x] - jest for GLS repositories
- [x] - introduce codecov
- [x] - per subpackage tests/test coverage


packages/codemirror-graphql/src/variables/__tests__/hint-test.js
```diff
@@ -32,10 +29,13 @@ function getHintSuggestions(query, variables, cursor) {
   const editor = createEditorWithHint(query);
   return new Promise(resolve => {
     const graphqlVariablesHint = CodeMirror.hint['graphql-variables'];
+
     CodeMirror.hint['graphql-variables'] = (cm, options) => {
+      console.log('\n\n\n\n\n\nHINT HIT\n\n\n\n\n');
```
Neitsch: ?

packages/graphiql/package.json
```diff
@@ -43,7 +29,7 @@
     "dev": "babel-node --root-mode upward test/server.js",
     "prepublish": ". ./resources/prepublish.sh",
     "preversion": ". ./resources/checkgit.sh && npm test",
-    "test": "jest --no-cache spec"
+    "test": "node ../../resources/runTests ENZYME=true"
```
Neitsch: It would be good if we could eventually move away from enzyme. It's not recommended for React development.

packages/graphql-language-service/README.md
```diff
@@ -3,6 +3,7 @@
 [![NPM](https://img.shields.io/npm/v/graphql-language-service.svg?style=flat-square)](https://npmjs.com/graphql-language-service)
 [![License](https://img.shields.io/npm/l/graphql-language-service.svg?style=flat-square)](LICENSE)
 
+
```
Neitsch: ?

resources/runTests.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+'use strict';
+
+const { exec } = require('./util');
+
+exec(
+  'jest',
+  '--config',
+  '../../jest.config.js',
```
Neitsch: Use a path relative to this script, i.e. `path.join(__dirname, '..', 'jest.config.js')`

acao:
we actually already have jest and mocha running side by side!

packages/graphql-language-service/README.md
acao: 😭

packages/graphiql/package.json
acao: lol, I just adopted it, but I'm all about what FB engineers have to say about testing react, haha! so the built-in https://reactjs.org/docs/test-renderer.html instead?

packages/codemirror-graphql/src/variables/__tests__/hint-test.js
acao: sorry yeah, this PR is still very much a WIP if you see in the title

acao:
as per codecov, we are in progress working with the codecov folks to be able to have labels for subpackages, so that the LSP libs and the codemirror mode don't look bad just because graphiql is lacking in coverage.

good idea to remove codecov from the scope. 

and yeah, i felt like i was so close with codemirror-graphql, but it seems that codemirror + jsdom + jest is not an easy mix, according to my research.

packages/graphiql/package.json
Neitsch: There's a nice wrapper around it with react-testing-library (https://github.com/testing-library/react-testing-library). It's fine to have enzyme for now, but it often lags behind React, because it uses unsafe/unstable internal APIs.

acao:
@ganemone see here, we will not be able to replace mocha entirely yet

acao:
I can clean this PR up tonight after work

acao:
@Neitsch should be good to go. Left code-mirror tests as is, but changed the configuration to make codemirror-graphql the only mocha harnessed subpackage. Left codecov in place for now

Neitsch:
@acao, looks like Travis is not passing. Would you mind checking on that first? :)

acao:
@Neitsch this happens every time i rebase! I think its because new tests keep getting added upstream using mocha

acao:
Screen Shot 2019-09-07 at 6 32 38 PM

interesting, seems like legit tests failures converting new tests from `to.Equal` to `toEqual` for proper deep comparison?

Neitsch:
Nice!

Issue #873, opened 2019-06-20
Empty tooltip on results


acao:
In results viewer, hovering over query results always shows an empty tooltip:
![Screen Shot 2019-06-20 at 4 29 40 PM](https://user-images.githubusercontent.com/1368727/59879848-ab8d4980-9379-11e9-82d5-b810245f3339.png)

Additionally, if you include an image in the query, then the image preview works, however it appears when you hover over _all_ fields, and continues to appear even when you remove the fields with images. I haven't tried with multiple image fields, but I'm guessing it might not work well in that context either.

acao:
this was resolved in 0.13.2

PR #874, opened 2019-06-20
Destroy tooltip when it isn't needed

acao:
Hard to demo the image bugfix without the local demo server, since it requires an image URI


acao:
Fixes #873

PR #875, opened 2019-06-21
Add regenerator runtime

acao:
It seems we might not be getting a working build without it. Acceptance tests are nice 😢


acao:
i just confirmed that this isn't required without all the package bumps, so I'm closing this as an independent PR

Issue #657, opened 2018-02-02
Docs can be hard to discover


html5cat:
Few people at the workshop today didn't know about docs after using GraphiQL for a while. Wondering if it makes sense to have the Docs open by default or until hidden once?

kevinsimper:
Yes, I have the same idea and it could maybe also take more of the screen to allow more focus.

PR #878, opened 2019-06-23
make `docExplorerOpen` a config prop, solves #657

acao:


acao:
there was a ticket with a few +1s with this feature request, but whatever you think is clever! I'm not great at gauging that yet.

PR #884, opened 2019-06-25
Add completion priority logic for tagged template completion requests

ganemone:
This has the sortText in completion items reflect the order in which they appear in the schema. This appears to add more value then alphabetical sorting, since fields are often ordered and grouped semantically.

This also ensures completion items returned from the graphql language server are sorted at the top of the completion list when working inside tagged templates.

### DIFF
--- a/packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.js
+++ b/packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.js
@@ -50,6 +50,26 @@ describe('getAutocompleteSuggestions', () => {
       });
   }
 
+  it('provides correct sortText response', () => {
+    const result = getAutocompleteSuggestions(schema, `{ h`, new Position(0, 3))
+      .map(({sortText, label, detail}) => ({sortText, label, detail}))
+    expect(result).to.deep.equal([{
+      sortText: '0hero',
+      label: 'hero',
+      detail: 'Character',
+    },
+    {
+      sortText: '1human',
+      label: 'human',
+      detail: 'Human',
+    },
+    {
+      sortText: '6__schema',
+      label: '__schema',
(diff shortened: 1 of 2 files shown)


packages/graphql-language-service-server/src/MessageProcessor.js
```diff
@@ -401,11 +401,20 @@ export class MessageProcessor {
     if (range) {
       position.line -= range.start.line;
     }
-    const result = await this._languageService.getAutocompleteSuggestions(
+    let result = await this._languageService.getAutocompleteSuggestions(
       query,
       position,
       textDocument.uri,
     );
+    if (textDocument.uri.endsWith('js')) {
```
Neitsch: What about `jsx`, `ts` or `tsx`?

packages/graphql-language-service-server/src/MessageProcessor.js
```diff
@@ -401,11 +401,20 @@ export class MessageProcessor {
...
+      // This ensures completion items are brought to the top of the list when inside a tagged template
+      result = result.map(item => {
+        return {
+          ...item,
+          sortText: '0',
```
Neitsch: Use something like `sortText: '0' + item.label` so that there is a clear tie breaker

packages/graphql-language-service-server/src/MessageProcessor.js
ganemone: Good point. I'm actually going to update this PR to handle this in a way that should be generic for all source file types (including graphql). Updates coming soon

packages/graphql-language-service-server/src/MessageProcessor.js
ganemone: Will do

ganemone:
ping @Neitsch how does this look?

PR #872, opened 2019-06-20
peer resolutions and upgrades

acao:
- don't allow < 0.12 anymore because graphql-language-service-utils depends on `ExecutableDefinitions` (see https://github.com/graphql/graphiql/commit/895e68537fd802b8b6ddf2578a1f76f85982c773)
- Fix peer resolutions,
- a clone of graphql/codemirror-graphql#259
- resolves  #863
- minor version updates to test and production dependencies
- add `regeneratorRuntime` as a polyfill for CDN bundle only
- ensure `dist` build is a more progressive build
- ~~add `common-shakify` to tree shake and reduce bundle size~~ (dynamic requires downstream in the interface caused a global eject :()

PR #885, opened 2019-06-25
Update babel configuration to target node for server packages

ganemone:
This updates the root babel configuration to target node for the server packages. This means the server packages can use native language features like async await, object spread, destructuring, etc.

This requires overriding configuration in the graphiql and codemirror-graphql packages to use the default configuration in @babel/preset-env.

### DIFF
--- a/packages/codemirror-graphql/.gitignore
+++ b/packages/codemirror-graphql/.gitignore
@@ -1,7 +1,7 @@
 node_modules/
 coverage/
 /*.js
-!.babelrc.js
+!babel.config.js
 /utils
 /variables
 /results
--- a/packages/codemirror-graphql/babel.config.js
+++ b/packages/codemirror-graphql/babel.config.js
@@ -1,0 +1,9 @@
+
+module.exports = {
+  presets: [
+    require.resolve('@babel/preset-env'),
+    require.resolve('@babel/preset-flow'),
+    require.resolve('@babel/preset-react'),
+  ],
+  plugins: [require.resolve('@babel/plugin-proposal-class-properties')],
+};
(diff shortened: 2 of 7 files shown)


ganemone:
@acao - Thoughts on having separate targets for browser vs server packages? The main benefit of doing this is that it allows native language features which should be faster and is easier to debug if you end up debugging transpiled code.

I'm also not sure if I correctly captured the packages which are used in server vs browser. This has browser config only for `graphiql` and `codemirror-graphql`

acao:
@ganemone good suggestion, however it's complicated because every package but the server is used in browsers as well. we have an upcoming PR that introduces a seperate babel config for graphiql. I will take a look at all of this later tonight

acao:
youll see that the babel config for graphiql here #872 is a bit more nuanced. we can't just go with the preset-env defaults because we also need to decide when and where to inject regeneratorRuntime, and to have different targets for the library and CDN build. we might just end up introducing Rollup to solve this problem

ganemone:
I will close this for now, since you are in the process of other build customization

acao:
@ganemone thanks for opening it though, because it raises an important point, that we should be using `node: true` for server packages.

AFAIK, only `graphql-language-service-server` is the only package in this repo that runs *only* on a server. Everything else is designed to be isomorphic, even graphiql is built to be able to be rendered server side. Javascript is funny like that huh? 

It's just hard to decide how to distribute these packages for both server and browser runtimes. The `main` and `module` entrypoint pattern falls a bit short here IMO, as that has more to do with module systems than targets.

acao:
@Neitsch re: `node: true`

Issue #868, opened 2018-01-12
type lookup in hint fails


timsuchanek:
Hi,

we're using the `codemirror-graphql` in the GraphQL Playground and experience problems, that the `type` property of a suggestion is undefined.
It happens in  this line of code:
https://github.com/graphql/codemirror-graphql/blob/master/src/hint.js#L66

The problem is, that `item.detail` can contain `!` in the case of a non-nullable type.
`item.detail` comes from here https://github.com/graphql/graphql-language-service/blob/master/packages/interface/src/getAutocompleteSuggestions.js#L97

`schema.getType()` however is just returning `getTypeMap()[name]` - as there is no type with the name `User!`, `undefined` is returned.
https://github.com/graphql/graphql-js/blob/master/src/type/schema.js#L182

Now is the question what the solution for this could be.
The quick-fix, that would work for us, is just not calling `.getType()` in `hint` anymore.
But I guess that the type lookup has been added for a reason.

@asiandrummer would be awesome to hear your take on that.

Thanks!

acao:
@timsuchanek thanks for your fix for this, I'm going to consider this resolved

acao:
@yoshiakis is this not resolved yet? i think i mistook the PR above as having solved this

yoshiakis:
@acao This issue is not solved yet. Sorry for stopping fixing this issue. Can I continue to be responsible to solve this issue? I think I can throw fixed PR in 1 or 2 weeks.

acao:
would love your help, you are right on track already, that timeline sounds perfect!

acao:
@yoshiakis this is in codemirror-graphql@0.12.0-alpha.5 which now has graphql 15 support! enjoy

yoshiakis:
@acao would you please reopen this issue? Unfortunately, this issue hasn't been fixed yet. I'm currently modifying the test code in [a branch](https://github.com/graphql/graphiql/pull/902) which needs to be merged to fix this issue.

acao:
@yoshiakis i merged as is, and I can add more tests in another PR, is that ok?

yoshiakis:
@acao yeah, it's ok. and thank you for merging! I want to do other PRs for refactoring and adding test cases and changing code of hint.js a little bit because the way how the value of documentation in completion list is determined has been slightly changed in graphql-language-servide-interface by #1541.

acao:
@yoshiakis @timsuchanek do we feel this bug is resolved now?

[...]

PR #902, opened 2019-06-28
fix: make list type and non-nullable type available

yoshiakis:
fixes of #868 .


acao:
thank you so much! i think this might fix a few bugs that were reported for graphiql as well. will take a look tomorrow

packages/codemirror-graphql/src/hint.js
```diff
@@ -80,3 +81,17 @@ CodeMirror.registerHelper('hint', 'graphql', (editor, options) => {
 
   return results;
 });
+
+function getType(schema, name) {
```
Neitsch: Could the string `name` ever contain spaces? If so, it doesn't look like the code can handle this.

packages/codemirror-graphql/src/hint.js
```diff
@@ -80,3 +81,17 @@ CodeMirror.registerHelper('hint', 'graphql', (editor, options) => {
...
+  if (name.slice(-1) === '!') {
+    const nestedType = getType(schema, name.slice(0, -1));
+    if (isNullableType(nestedType)) return new GraphQLNonNull(nestedType);
+    return;
```
Neitsch: Why is this not `return nestedType;`? In the line above you are returning if it is nullable.

yoshiakis: It's because I wanted `undefined` to be returned if something like `foo!!` is passed into `name` parameter.

packages/codemirror-graphql/src/hint.js
yoshiakis: `name` can contain spaces, but in that case, `undefined` is returned.

yoshiakis: `name` can contains spaces. If `name` contains spaces and `schema` contains the type of the `name`, this function returns that type. If `name` contains spaces and `schema` doesn't contain the type of the `name`, this function returns `undefined`.

yoshiakis:
@Neitsch Thanks for your feedback! I agree with you so I'll try to add a new property to the output of `getAutocompleteSuggestions` that holds GraphQL specific type information.

acao:
@yoshiakis would you like someone to finish this for you or do you plan on coming back around to it? whatever you decide we are very happy with the current direction :)

yoshiakis:
@acao I'm sorry for taking time to finish this PR because I've said 2 weeks before that I would finish in 1 or 2 weeks (https://github.com/graphql/graphiql/issues/868#issuecomment-560404279). I'm in progress to change code for this, although it's not finished yet since I wasn't able to make enough time. Do you need to finish this PR soon? If not, Would you give me other 2 weeks?

acao:
@yoshiakis my bad, hard to keep track of all the wonderful initiatives around, please take your time! lmk if you have any questions. Nigel should be back by next week as well.

yoshiakis:
Now I'm thinking `graphql-language-service-interface` exports additional functions so that `codemirror-graphql` can use them. Also I've noticed I have to export flow type definition for that functions because `codemirror-graphql` is still using flow. So I'm thinking to change the code of [index.js.flow](https://github.com/graphql/graphiql/blob/graphql-language-service-interface%402.3.3/packages/graphql-language-service-interface/src/index.js.flow), but should I need to add code manually or some commands creates it automatically?

@Neitsch It seems you've created this file `index.js.flow`. I'd really appreciate it if you give me any information :)

yoshiakis:
@acao I know there is a project that converts the js files in `codemirror-graphql` to ts files. Do you think it is better to finish that project in advance instead of exporting more type definitions for flow from `graphql-language-service-interface`?

acao:
@yoshiakis if you’d like to, sure! GraphiQL will be moving to monaco soon, so we were considering skipping converting the codemirror mode, as it will probably be within 6 months or so be moved to a new contributors github org. however this will be very useful. when we move to monaco it will be typescript first. the current POC is in typescript

acao:
also yes @yoshiakis index.js.flow is maintained manually in interface for now. we havent found a tool to automate generating it from the typescript definition, do you have any reccomendations?

yoshiakis:
@acao Sure! I have other things in graphql project I want to focus on more, so I'll skip converting `codemirror-graphql` to ts for now. I'll choose to export type definitions for flow from `gql-interface` which I need for `codemirror-graphql`.

yoshiakis:
@acao Unfortunately not. I don't know any useful tools for generating index.js.flow automatically from ts definition either :(

acao:
In fact I might just add this as part of the current typescript rewrite, or as an addendum PR
https://github.com/joarwilk/flowgen. it looks pretty robust and like it will be compatible with most of what we have.

yoshiakis:
It seems like the test files within `src/__test__` directory in `codemirror-graphql` haven't been tested anymore since merging [this PR](https://github.com/graphql/graphiql/pull/871). Is it just a mistake or does it have any reasons?

acao:
@yoshiakis the tests are still run via mocha, they just dont show up in coverage yet. just re adding them in a PR thats open now.

yoshiakis:
@acao When I run a test command, some test files within `codeimirror-graphql`are read but the others aren't. I thinks it's because glob is incorrect here.
https://github.com/graphql/graphiql/blob/c735b2227c676005607b5ed75109992df53eb0d8/packages/codemirror-graphql/package.json#L37

I think the correct is `src/**/__tests__/*-test.js`.

acao:
good find @yoshiakis ! I’ll open another PR to address this

acao:
@yoshiakis fyi, a fix was merged for this test issue was merged upstream if you want to pull

yoshiakis:
Thanks a lot, @acao !

yoshiakis:
I've fixed the code, though it was not easy to read. I'm currently making the code more readable and add test for parts of the code changed. It seems that it takes a few more days to complete that.

yoshiakis:
I'm thinking adding new lines within `codemirror-graphql` which is similar to lines in `graphql-language-service-interface` while the new lines uses flow type definition instead of TypeScript. The similar lines within `graphql-language-service-interface` includes [function getSuggestionsForVariableDefinition](https://github.com/graphql/graphiql/blob/c735b2227c676005607b5ed75109992df53eb0d8/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L349) and it seems [TODO](https://github.com/graphql/graphiql/blob/c735b2227c676005607b5ed75109992df53eb0d8/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L357) inside the function is related with TypeScript. I'm not sure what the TODO message means, though I think it also relates to flow type check if you use it instead of TypeScript. @acao Could you please tell me in more detail what the TODO message means?

yoshiakis:
ping @acao .
If you aren't too busy, could you tell me what [this TODO](https://github.com/graphql/graphiql/blob/c735b2227c676005607b5ed75109992df53eb0d8/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L357) means in more detail since it may relates to this PR?

acao:
hey @yoshiakis thanks so much for your effort in this!

this todo:
https://github.com/graphql/graphiql/blob/c735b2227c676005607b5ed75109992df53eb0d8/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L358

is one of the few places where I used the`any` keyword. I couldn't find a definition for the `type` param that worked

acao:
so, programatically the code is fine, we just dont have the type inference we want here. think of this `@TODO` as similar to `$FlowFixMe`

yoshiakis:
I finished modifying the code. It changed the code not only of `codemirror-grpahql`, but also of `graphql-language-service-interface` and `graphql-language-service-type`, so I've made PRs for them, respectively (#1264 for `graphql-langauge-service-interface`, and #1265 for `graphql-language-service-type`).

I'll rebase the branch at this PR on master branch after #1264 and #1265 are merged. And then, I'll PR this branch again.

acao:
awesome work, thanks!

acao:
merging this almost a year later, haha! thanks so much for this @yoshiakis

Issue #535, opened 2017-07-08
No intellisense for __typename


kbrandwijk:
If I want to use `__typename` _on any Type_ in a query, it doesn't show up in the 'intellisense-autocomplete-popup'. This is inconsistent, because top-level introspection fields like `__type` and `__schema` do show up in the popup.
![typename](https://user-images.githubusercontent.com/852069/27988025-908af7a0-6418-11e7-985c-b488bcbdd271.gif)

wincent:
Thanks for the report, @kbrandwijk. I think this may be the result of an overly literal reading of the spec:

`__type` and `__schema` are [top-level introspection fields](http://facebook.github.io/graphql/#sec-Schema-Introspection).

`__typename` is a meta field that can be queried against [any Object, Interface or Union type](http://facebook.github.io/graphql/#sec-Type-Name-Introspection).

The `query` type, however, *is* ["an Object base type"](http://facebook.github.io/graphql/#sec-Initial-types), so does allow `__typename` (even though it is pointless to ever query for it).

kbrandwijk:
But it doesn't show up on any Type. In my screenshot, I use it on `allFilms`, not on `query` directly. Do you agree that it should show up on 'any Object, Interface or Union type'?

wincent:
Sorry I wasn't explicit. My comment basically boils down to, "Yes, it looks like it should appear in the typeahead, according to the spec, and I theorize that the reason it behaves like it currently does is that we just read things a little too narrowly." Probably. Haven't looked at the code yet.

kbrandwijk:
Thanks for clearing that up :smile:

PR #903, opened 2019-07-03
 feat(language-service-interface): add __typename field suggestion against object type

yoshiakis:
implementation of #535.


orta:
Interestingly this [didn't appear in the display preview](https://deploy-preview-903--graphiql-test.netlify.com/?query=%7B%0A%20%20allFilms%20%7B%0A%20%20%20%20__%0A%20%20%20%20planets%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D), maybe it doesn't get the latest version of the language server?

Screen Shot 2019-07-26 at 9 37 00 AM

That said, I trust the tests more 👍

benjie:
This is a great feature; but would it be better if `__typename` was at the *bottom* of the list of options. I'm concerned it could get irritating being at the top. (I'm just inferring from the tests it appears at the top?)

PR #907, opened 2019-07-08
Fix check mark not being visible

### LINKED ISSUE #906: Check mark is not visible in select lists
Check mark is not visible when an item is selected from a select list in the toolbar.

How it looks like:
Screenshot_2019-07-05 GraphiQL(1)

How it should look like:
Screenshot_2019-07-05 GraphiQL

mjaakko:
Fixes #906


orta:
👍

PR #913, opened 2019-07-09
Chore(graphiql): modify dependencies of GraphiQL example

yoshiakis:


orta:
I don't think we want this, at minimum the example should definitely be using the graphiql package from this repo - I'm not certain about the removal of react too

yoshiakis:
@orta Thanks your comment!

>I don't think we want this, at minimum the example should definitely be using the graphiql package from this repo

Even If you install graphiql into npm modules by npm or yarn, you don't use this graphiql module in the example. Instead you get graphiql.js from [here](https://github.com/graphql/graphiql/pull/872/files#diff-63e5cd2ab899816a7d5228213e307bdaR41), which is the your local working draft of graphiql so I think we can remove graphiql package from dependencies because it's not currently used at all.

>I'm not certain about the removal of react too

It's not used either. While It's used as a peerDependencies (see: https://github.com/graphql/graphiql/commit/73b133076ebc93823ab0a95765077d8f85dacad3) before because the example needed React module from node modules, the example currently gets react from CDN.

acao:
thanks for this! we have an open PR where we make similar modifications, and also move it into its own subpackage so that its dependencies are installed via yarn workspaces. I agree about removing react. I like your initial approach here however, so if you want to do the last step of:
- moving it to its own subpackage
- ensuring the netlify build still works
- updating the cdn react to the latest 15.x
- updating the docs to show the `yarn setup` step

then this PR will be super helfpul!

acao:
@yoshiakis if you don't have time this weekend, I'll create it tonight or tomorrow so I can merge #921 tomorrow, and I'll give you credit for the idea to upgrade dependencies and remove the react dependency, too!

yoshiakis:
@acao I'm sorry but I don't have time this weekend to do the last step you've mentioned. I'd appreciate it if you do the last step.

acao:
can do!

acao:
Closing this for #922

PR #917, opened 2019-07-27
Deprecate <0.12.x graphql for most packages, upgrade codemirror to interface/parser 2.x

acao:
- because utils now requires graphql-js < 0.12.x
- make sure all downstream dependents reflect that peer resolution change
- upgrade codemirror-graphql to use ^2.0.1 interface/parser

@Neitsch this plus package bumps as the icing on the cake for this round? looks like we will need a few minor/major bumps because of this, even?


acao:
hmm... might need to just close this and slice this differently yet again, haha

PR #919, opened 2019-07-27
Upgrade codemirror-graphql to gls interface/parser 2.0x

acao:
- Update tests to support `__typename` in object type hint suggestions :shrug:
- Quite a breaking change! This should bump codemirror-graphql to 0.9.0

### DIFF
--- a/packages/codemirror-graphql/package.json
+++ b/packages/codemirror-graphql/package.json
@@ -1,6 +1,6 @@
 {
   "name": "codemirror-graphql",
-  "version": "0.8.3",
+  "version": "0.9.0",
   "description": "GraphQL mode and helpers for CodeMirror.",
   "contributors": [
     "Hyohyeon Jeong <asiandrummer@fb.com>",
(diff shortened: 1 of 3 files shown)


acao:
This will probably break the netlify deployment of graphiql above (white screen), mostly just because codemirror-graphql is still at 0.8.3, graphiql still depends on that version, and will locally resolve it and try to use regenerator-runtime (which will be missing) in the babel build

acao:
fixed!

acao:
@graphql/graphiql-maintainers do you think we should upgrade graphiql to this version of codemirror-graphql so we can manually regression test this upgrade easier? it'll require figuring out the regenerator runtime issue. 

@ganemone had the idea to use `node: true` at the root in preset-env, but we will still need to add the regenerator runtime for the browsers list in graphiql (would like to be evergreen, but I promise @IvanGoncharov  we would match graphql-js's browserslist for graphiql). 

Where do we draw the line, since almost all of these libraries are browser facing? @Neitsch  see why there was so much going on in that PR? haha

IvanGoncharov:
@acao Just to clarify: I meant you shouldn't support browsers that are not supported by `graphql-js`. But it's up to you guys to decide if you want to drop some additional browsers.

acao:
@IvanGoncharov oh that's lovely news 💯 !!!

Issue #915, opened 2019-07-17
[RFC][LSP] Pluggable schema loading


Neitsch:
Hi all,

Here at FB we have some schema extensions that live in a separate file and are applied during the build. While this is not particularly relevant for GraphiQL, we really want to have this as part of the LSP. Therefore I propose the following:
* In the configuration of the LSP (`.graphqlrc`), the user can specify the following:
  * (required) a package that provides two functions: `async init()` and `async getSchema()`. The latter returns the latest GraphQLSchema. Hopefully this can work in a similar way to babel plugins or webpack loaders.
  * (optional) arguments that are provided to the `init` function, before the `getSchema` function is called for the first time.
  * (optional) a glob relative to the `.graphqlrc` that triggers the schema to be reloaded on file change.

With this, I believe we can provide a very flexible schema loading experience. The first schema loaders we implement should probably be: a file schema loader to provide the current LSP experience and an introspection query schema loader to make it work with GraphiQL.

acao:
I love the sound of all of this. Moving the introspection query schema loader down to the LSP sounds excellent, and is similar to how `vscode-json-languageservice` loads JSONSchema definitions in a pluggable fashion.

acao:
Also, I wouldn't rule out alternative schema loading patterns, even alternative protocols being more readily supported for graphiql-1.0.0-alpha. for example, using wss to locally (or over the wire?) transmit json-rpc from a language server.
https://github.com/TypeFox/monaco-languageclient/blob/master/example/src/client.ts

acao:
The interface that the typescript language service uses for adding LSP plugins is interesting, and is designed to work in browsers, vscode, monaco, etc
https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin

We could do something similar with `-interface`, or possibly a layer on top of it designed for greater parity with existing vscode-<language>-languageservice implementations

acao:
@Neitsch there are two ways we can go about this for graphiql - we can do something provisional, where GraphiQL has a prop for this and we can do that very soon, but then we would have to migrate to the new monaco implementation, ideally keeping parity from the GraphiQL side of things. the other would be of course just waiting for the new monaco implementation, which means defining the loader in the LS itself, as with monaco-vscode.

acao:
[...]

Issue #882, opened 2017-07-19
Cannot build with webpack


wmertens:
I'm trying to bundle graphiql which uses codemirror-graphql which uses graphql-language-service.

The build is failing because webpack can't parse the bundled `js.flow` files, and it is trying to parse those because of the dynamic import at https://github.com/graphql/graphql-language-service/blob/master/packages/graphql-language-service-interface/src/GraphQLLanguageService.js#L87 (probably) which could be importing anything at all.

Is it not possible to limit the require by making it something like:
```js
require(`./${somePath}.js`)
```
?

If not, is it possible to not include the flow js files in the `dist` folder?

The webpack error is:
```
Module parse failed: .../node_modules/graphql-language-service-interface/dist/getOutline.js.flow Unexpected token (11:12)
You may need an appropriate loader to handle this file type.
|  */
|
| import type {
|   Outline,
|   TextToken,
 @ ./node_modules/graphql-language-service-interface/dist ^.*$
 @ ./node_modules/graphql-language-service-interface/dist/GraphQLLanguageService.js
 @ ./node_modules/graphql-language-service-interface/dist/index.js
 @ ./node_modules/codemirror-graphql/lint.js
 @ ./node_modules/graphiql/dist/components/QueryEditor.js
 @ ./node_modules/graphiql/dist/components/GraphiQL.js
 @ ./node_modules/graphiql/dist/index.js
 @ ./lib/universal/Graphql/index.jsx
 @ ./lib/universal/Root.jsx
 @ ./lib/client/index.js
 @ multi babel-polyfill stratokit/client
```

acao:
@lostplan should we work to change this interface? this `require(`${template string}`)` implementation also breaks tree shaking tools such as common-shake, preventing a slimmer build of graphiql

JoshRobertson:
Is there no way to make this a priority? This issue has been open for 2 years. We can't use this package until this issue is resolved.

acao:
I will be providing a webpack example soon, as I haven't had an issue except with common shake. We are working on reworking it to work with tree shaking soon, might require breaking changes.

benjie:
In case it helps, here's the webpack configuration we're currently using in PostGraphile to bundle GraphiQL:

https://github.com/graphile/postgraphile/blob/7979cfb10f3246c7c198be8db2c929e82e0e4323/scripts/make-assets#L49-L124

acao:
[...]

PR #920, opened 2019-07-27
Pluggable schema loading

Neitsch:
So, to get us started, I've implemented pluggable schema loading. How this works is the following:
* We have a new extension key `schema-loader` that contains two fields `loaderPackage` and `loaderArgs`. The loader package will be dynamically loaded. Subsequently we call `init(loaderArgs)` for it to initialize. (Alternatively we could use ES6 classes)
* The dynamically loaded package exposes a function `getSchema`, that if called returns the GraphQL Schema it loaded.

A couple caveats:
* This PR per se does not add any functionality. It relies on subsequent PRs actually adding schema loaders that conform to the interface.
* I am not sure wether simply calling `require()` is a good way to do this, but at least the new test passes
* This does not add all of the functionality I suggested in the original issue (#915)
* Right now we instantiate a new schema loader every time. We should probably cache it if it becomes a bottleneck.

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLSchemaLoaderExtension.js
+++ b/packages/graphql-language-service-server/src/GraphQLSchemaLoaderExtension.js
@@ -1,0 +1,26 @@
+import {GraphQLSchema} from 'graphql';
+
+export type GraphQLSchemaLoaderPackage = {
+  init: (loaderArgs: any) => GraphQLSchemaLoader,
+}
+
+export type GraphQLSchemaLoader = {
+  getSchema: () => Promise<?GraphQLSchema>,
+}
+
+export type GraphQLSchemaConfigData = {
+  loaderPackage: string,
+  loaderArgs: any,
+}
+
+export class GraphQLSchemaLoaderExtension {
+  _raw: GraphQLSchemaConfigData;
+
+  constructor(schemaConfig: GraphQLSchemaConfigData) {
+      this.raw = schemaConfig;
+  }
+
+  getSchemaLoader(): GraphQLSchemaLoader {
+    return require(this.raw.loaderPackage).init(this.raw.loaderArgs);
+  }
+}
--- a/packages/graphql-language-service-server/src/GraphQLCache.js
+++ b/packages/graphql-language-service-server/src/GraphQLCache.js
@@ -669,7 +669,6 @@ export class GraphQLCache implements GraphQLCacheInterface {
       if (this._schemaMap.has(schemaCacheKey)) {
         return this._schemaMap.get(schemaCacheKey);
       }
-      console.log(schemaLoaderExtension);
       schema = await schemaLoaderExtension.getSchemaLoader().getSchema();
       this._schemaMap.set(schemaCacheKey, schema);
       return schema;
--- a/packages/graphql-language-service-server/src/GraphQLCache.js
+++ b/packages/graphql-language-service-server/src/GraphQLCache.js
@@ -33,6 +33,7 @@ import {
 } from 'graphql-config';
 import { getQueryAndRange } from './MessageProcessor';
 import stringToHash from './stringToHash';
+import {GraphQLSchemaLoaderExtension, GraphQLSchemaLoader} from './GraphQLSchemaLoaderExtension';
 import glob from 'glob';
 
 // Maximum files to read when processing GraphQL files.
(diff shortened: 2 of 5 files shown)


acao:
this is a great, love it! can't wait to implement this with gls-interface + monaco

only caveat of dynamic requires is that they can sometimes break build tooling. for example, common-shake will do a global exit on this, making it so that we can accomplish tree shaking, at least from what I can tell. it might introduce other issues as well, such as making things funky with webpack loaders and how they resolve paths/extensions. see this issue for details https://github.com/graphql/graphiql/issues/882

Neitsch:
Alright, alternative idea how to facilitate this:
```
global.schemaLoaders = {}
...
if(global.schemaLoaders[mySchemaLoader]) {
  // use dynamic schema loader
}
```
Then to register a schema loader the developer must import the plugin on a top level. The plugin itself then has some code like this:
```
global.schemaLoaders['cool-schema-loader'] = {
  init: () => ({ getSchema: () => null })
}
```

Alternatively, could we mark the loader with `sideEffect`? (https://webpack.js.org/guides/tree-shaking/)

Neitsch:
I guess if we think about an LSP itself it might not actually be able to require other packages that our outside of the extension. So alternatively we could just model this as a command line invocation that prints the loaded schema to stdout. Then we just need to reparse it.

Neitsch:
Digging more into that last idea, maybe a platform specific solution is necessary here: For VSCode we communicate via the extensions API (https://stackoverflow.com/questions/50058517/how-to-communicate-between-vscode-extensions). For GraphiQL we communicate via HTTP calls

acao:
@Neitsch yes 98% of the time via HTTP for graphiql, but I do want the new graphiql to be able to support folks using it in an electron/nwjs/RN?!/etc app and thus being able to use json-rpc alongside alternative, vscode-like schema loading mechanisms. Maybe thats a bit ambitious though, haha!

Either way, I agree that we need to figure out something platform specific for this. Those two patterns are great, and then in my above case, maybe we could read or stream from file or some clever cacheing mechanism. A tool like insomnia or altair, playground desktop would probably be able to leverage this as well since they are using these libraries. 

@schickling @imolorhe @gschier @martijnwalraven, thoughts on this interface?

acao:
possibly we need an interface for platform-specific extensions in general... related to the above issue I mentioned, we somewhere along the idea had the idea of making `graphql-config` optional via a plugin (that might be a bit ambitious), and here we have the extensibkle schema loader, and Gatsby wants to be able to add their common fragments behind the scenes as well (they were going to do this at the codemirror-graphql level, but I suggested taking it down to the LSP/interface level somehow).

packages/graphql-language-service-server/src/GraphQLSchemaLoaderExtension.js
```diff
@@ -0,0 +1,26 @@
...
+
+export class GraphQLSchemaLoaderExtension {
+  _raw: GraphQLSchemaConfigData;
+
+  constructor(schemaConfig: GraphQLSchemaConfigData) {
+      this.raw = schemaConfig;
+  }
+
+  getSchemaLoader(): GraphQLSchemaLoader {
+    return require(this.raw.loaderPackage).init(this.raw.loaderArgs);
```
imolorhe: Everything about this looks good to me. Love the idea as well. The only part about this that I'm not sure of is this line which you guys were already talking about. I'm not familiar with all the contexts the language server can be used in (for example a browser context might not have the `require` method), so when this is called at runtime it would need to work well for all contexts it's used in.

Having the package add itself to a global object is one way to go. Probably there are other approaches for this.

_Ps: If the language server always runs in a node context, then I would prefer the `require` to the global object approach._

imolorhe:
Another thing to consider would be if `getSchema` in the loaderPackage needs to perform asynchronous operations to fetch the schema. In that case, it would be best to have it return a promise from the get-go.

packages/graphql-language-service-server/src/GraphQLSchemaLoaderExtension.js
benjie: I think we should try and stay away from the `global` object and dynamic `require`s. All these concerns should be pushed up the stack so the calling application deals with them directly (and can then choose to do dynamic require vs not depending on what bundling they're using).

With that in mind, should `getSchemaLoader` be async?

acao: i think we are working towards allowing this to load behavior based on the platform - so in VSCode that would be via an extension, with node it should
 be an async file load not a require, and in the browser we will need to load it differently, probably dynamic import. What it needs is to dynamically decide which platform it's using and to attempt to load the schema loading plugin with the matching mechanism. @Neitsch and I were discussing this

Neitsch:
I'll close this PR for now, until I/we have a better story around plugin loading.

Neitsch:
After thinking about it a little more and based on @acao 's input on TS plugins, reopening the PR. Will address @imolorhe 's comments next :)

packages/graphql-language-service-server/src/GraphQLSchemaLoaderExtension.js
acao: we have a fix for this now! :D

acao: @imolorhe as it turns out, webpack should handle the require method just fine, we will shift towards async imports as well, as should electron/etc, as import() is supported in node 9.7+. The other thing is this is on the server side, so this code won't be loaded by webpack.

@Neitsch if you want to update this for the new file resolution utilities we merged a week or two ago (they are in gls-utils now) they should ensure that this dynamic `require()` line works across environments.

also, is it possible we can use graphql-config to load the schema? looks like it has its own capability for this already. check out the latest 3.0.0 alpha, we will be adopting this soon!

acao:
@divyenduz what do you think we should do about this? we should be leveraging graphql-config to load the schema correct?

acao:
@Neitsch i think that in order to accomplish this in the ways that you've indicated facebook and other large scale users might need, it could be that making the graphql-config library itself optional (as some have suggested), thought the basic config spec format might still be honored if that's so. one should of course be able to provide their own configuration management solution in place of graphql-config, and the interface should expect a configuration object that matches up.

PR #921, opened 2019-07-27
upgrade graphiql to codemirror-graphql 0.9.0, interface/parser 2.x

acao:
- upgrade graphiql to use codemirror-graphql 0.9.0
- because we are using parser/interface 2.x, we are using utils 2.x which deprecates support for graphql versions below 0.12.x, thus update peer resolutions to reflect this
- upgrade development graphql to latest
- add `regenerator-runtime`
- deduplicate lockfile
- extensive manual testing, looks good!

(had been broken before because it required a change to the example project which has been implemented)

(P.S. I know I'm foregoing the proper babel-runtime/etc plugin route, but requiring `regenerator-runtime` directly was much simpler, and tree shaking can take care of it in build environments that duplicate this config. we can get more clever with it at a later iteration)

### DIFF
--- a/packages/graphiql/package.json
+++ b/packages/graphiql/package.json
@@ -52,12 +52,13 @@
   },
   "dependencies": {
     "codemirror": "^5.47.0",
-    "codemirror-graphql": "^0.8.3",
+    "codemirror-graphql": "^0.9.0",
     "copy-to-clipboard": "^3.2.0",
-    "markdown-it": "^8.4.0"
+    "markdown-it": "^8.4.0",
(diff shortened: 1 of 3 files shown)


acao:
The build is succeeding because it works in yarn workspaces, but because the example package is not part of the yarn workspaces scope, it can't find it

The upstream fix for this is to move the example to its own subpackage, rather than a pseudo-sub-sub package

acao:
updated after the graphiql-example fix was merged, now everything is resolving just fine!

PR #923, opened 2019-07-28
Allow to change request options.

MateuszLeo:


acao:
hmmmm... the thing I'm worried about most with this is that so many companies have customized their fetcher/http pattern, without forking graphiql, and this will consitute a major breaking change for them. 

So, to prevent this from being a breaking change, can we make this a top level option prop? such as `<GraphiQL showRequestOption />` set to false by default? And then only add that state to each fetcher request and the introspection fetcher requests when its enabled?

benjie:
This is a really nice feature; but I don't think we want this in core because I think the majority of GraphiQL users wouldn't need/want it and it increases the surface area of the tool.

I think this will make a great plugin for the new GraphiQL version we're working on - and ensuring we have the relevant integration points for it will be a good test of the plugin/extension system.

acao:
@Peterkingalex1972 don't know who you are but its obvious we didnt want this PR merged. It doesn't even satisfy the auth pattern that most of our users have. I havent used a csrf token since my rails and django days. We wont be merging anything like this. If youd like to get involved in maintenance please join us for working groups and/or reach out on chat

PR #928, opened 2019-08-09
Invalidate schema cache on file changes to schemaPath

### LINKED ISSUE #893: Watch for schema{.json,.graphql} changes
If changes are made to the schema while the service is running, they are not taken into account since the schema gets cached on first load (`_schemaMap`).

For our use case, we emit a `schema.json` as part of our server compilation. This would help with newly added fields showing up in the autocomplete, validation, etc. without having to reload the entire window.

ganemone:
Fixes #893

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLCache.js
+++ b/packages/graphql-language-service-server/src/GraphQLCache.js
@@ -303,15 +303,19 @@ export class GraphQLCache implements GraphQLCacheInterface {
   ) => (result: Object) => {
     if (result.files && result.files.length > 0) {
       const graphQLFileMap = this._graphQLFileListCache.get(rootDir);
-      if (!graphQLFileMap) {
-        return;
-      }
       result.files.forEach(async ({ name, exists, size, mtime }) => {
+        const filePath = path.join(result.root, result.subscription, name);
+        if (projectConfig.schemaPath && filePath === projectConfig.schemaPath) {
+          this._invalidateSchemaCacheForProject(projectConfig);
+        }
+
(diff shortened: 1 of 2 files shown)


packages/graphql-language-service-server/src/GraphQLCache.js
```diff
@@ -303,15 +303,19 @@ export class GraphQLCache implements GraphQLCacheInterface {
   ) => (result: Object) => {
     if (result.files && result.files.length > 0) {
       const graphQLFileMap = this._graphQLFileListCache.get(rootDir);
-      if (!graphQLFileMap) {
-        return;
-      }
       result.files.forEach(async ({ name, exists, size, mtime }) => {
+        const filePath = path.join(result.root, result.subscription, name);
+        if (projectConfig.schemaPath && filePath === projectConfig.schemaPath) {
```
Neitsch: Is the schemaPath guaranteed to be absolute? Otherwise this condition might be false, even though it should be true.

ganemone: It gets resolved into an absolute path. My test uses a relative path in the config file to verify this works

ganemone:
> If you have time, it might also make sense to invalidate on `.graphqlrc` or `.graphqlconfig` changes. No pressure though. Requesting changes just for my one comment :)

I think that could come in a future PR, since it potentially involves more complexity.

ganemone:
@Neitsch any further comments?

Neitsch:
Sorry it took so long!!

PR #929, opened 2019-08-11
set up WG meeting agenda

acao:


acao:
@Neitsch it's been on my mind too, I'm working on a diagram that might help with architecting some possible paths

Issue #735, opened 2018-10-30
Uncaught Error: Mode graphql failed to advance stream.


darinallen:
I am trying to use graphiql with a gatsby project and I am experiencing the same issue described in #163. As the author of that issue wrote, I can't type anything into the editor. The editor space is just blank. When looking at the developer tools I see:

`Uncaught Error: Mode graphql failed to advance stream.`

I've tried clearing the local storage, and that doesn't do anything. I believe this issue originated when I accidentally hit option-space instead of ctrl-space, which I've seen reported as an issue with another plugin:

https://github.com/jimkyndemeyer/js-graphql-intellij-plugin/issues/33

darinallen:
I was able to resolve this by manually setting query=test as a URL parameter. There is likely a better solution, but at least that is available as a workaround.

edevil:
Same problem here.

SebastienTainon:
After running into this issue a few times, I've just figured out a way to reproduce it:

In the GraphQL query editor, do **Alt Gr + Space** (or Option + Space): this will trigger this issue.

I hope it will help fix it, that's pretty annoying when you do it without making attention...

andrei-anisimov:
On Mac OS this can be reproduced by pressing Shift + Option + Space

yboulkaid:
I ran into this issue on Firefox and neither the "killing the tab" nor the "localStorage.setItem" hacks worked. I finally managed to solve it by clearing localStorage, disabling javascript (through the built-in setting or by via an extension) and reloading the page.

acao:
thanks for reporting this @yboulkaid
the PR we merged for this has not been released yet, as we are still awaiting NPM credentials.
I did confirm that this does fix this bug. so, graphiql 13.1 *will* fix this, once its actually released on NPM.

In the mean time, there is a release of dist files here, if that works for anyone:
https://github.com/graphql/graphiql/releases/tag/v0.13.1

acao:
good news! 0.13.2 has been released to npm. 0.13.1 contained this fix, however we decided to just publish 0.13.2 instead. can you confirm that this new version fixes the issue?

robinmetral:
Great news! Personally I'll be waiting until [`gatsby-graphiql-explorer`](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-graphiql-explorer) updates its dependency :slightly_smiling_face:

robinmetral:
[...]

PR #932, opened 2019-08-12
Fix 'Mode graphql failed to advance stream'

benjie:
Fixes https://github.com/graphql/graphiql/issues/735

Basically the bad token wasn't being "eaten" so it literally was failing to advance the stream. This fix consumes it.

Screenshot_20190812_142416

### DIFF
--- a/packages/graphql-language-service-parser/src/onlineParser.js
+++ b/packages/graphql-language-service-parser/src/onlineParser.js
@@ -109,7 +109,12 @@ function getToken(
 
   // If there's no matching token, skip ahead.
   if (!token) {
-    stream.match(/\S+/);
+    const matchedSomething = stream.match(/\S+/);
+    if (!matchedSomething) {
+      // We need to eat at least one character, and we couldn't match any
+      // non-whitespace, so it must be exotic whitespace.
+      stream.match(/\s/);
+    }
     pushRule(SpecialParseRules, state, 'Invalid');
     return 'invalidchar';
   }


benjie:
@robinmetral can you confirm that this deploy fixes the issue for you?

https://deploy-preview-932--graphiql-test.netlify.com/?query=%7B%0A%20%20allFilms%20%7B%0A%20%20%20%20title%C2%A0%0A%20%20%7D%0A%7D

acao:
Once the user confirms this fix, we can merge this, release GLS, codemirror-graphql, and then graphiql. GraphiQL in master currently depends on unpublished dependency versions FYI!

acao:
If this is an issue with gls-parser, then we should also see this bug in insomnia and playground

Neitsch:
Is this issue specific to whitespaces or unicode characters overall? If it's the latter maybe we can  tackle unicode characters more broadly in this PR?

benjie:
Lets do that in a separate PR @Neitsch; this PR is very localised to one specific issue and it already handles `/\S+/` and `/\s/` so there's no more characters it can handle (since `/[\S\s]/` covers all possible characters).

robinmetral:
@benjie @acao sorry for my late reply, fix confirmed! :tada:

robinmetral:
@acao do you know when the next release will be? I'm looking forward to bringing this fix to Gatsby :rocket:

acao:
@robinmetral this week, ideally!

PR #934, opened 2019-08-12
Fix formatting of GraphiQL errors

benjie:
Fixes https://github.com/graphql/graphiql/pull/636  
Fixes https://github.com/graphql/graphiql/pull/722

This approach allows overriding the error handler. By default it uses `JSON.stringify` on the error; if the result is `'{}'` then it will instead create a `{message,stack}` error to stringify (since `JSON.stringify(new Error("Something"))` is `'{}'`).

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.js
+++ b/packages/graphiql/src/components/GraphiQL.js
@@ -579,7 +579,7 @@ export class GraphiQL extends React.Component {
       .catch(error => {
         this.setState({
           schema: null,
-          response: error && String(error.stack || error),
+          response: error ? GraphiQL.formatError(error) : null,
         });
       });
   }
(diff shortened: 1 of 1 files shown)


acao:
this works with subscriptions?

benjie:
It should do; we had to hack around this issue in PostGraphile:

https://github.com/graphile/postgraphile/blob/7979cfb10f3246c7c198be8db2c929e82e0e4323/postgraphiql/src/components/PostGraphiQL.js#L320-L333

I'll give it a try.

benjie:
Screenshot_20190812_163947

Yes, definitely works with subscriptions. Took me a while to get our dev server up and running with subs!

acao:
@benjie awesome, lgtm!

PR #936, opened 2019-08-12
Have arrays of errors output message if available

benjie:
Sorry @acao, turns out there's situations where arrays of errors are returned which then outputs:

```json
[
  {}
]
```

which is not super useful. This PR replaces the new formatError function with one that handles this situation better.

#### Unknown errors (e.g. when they don't return `{"errors": ...}`):

Screenshot_20190812_170029

(Note this error message comes from `subscriptions-transport-ws` rather than from us)

#### Known errors:

Screenshot_20190812_170705

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.js
+++ b/packages/graphiql/src/components/GraphiQL.js
@@ -1073,16 +1073,18 @@ GraphiQL.formatResult = function(result) {
   return JSON.stringify(result, null, 2);
 };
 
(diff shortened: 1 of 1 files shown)


acao:
hmm, yeah looks good to me! as long as we follow expectations for error formatting in the official spec. unknown errors will probably always be a thing

benjie:
Cool; ready for merge :+1:

PR #937, opened 2019-08-12
dev: add ws subscriptions support to test environment

benjie:
Only for the GraphiQL dev environment: add websocket subscriptions support powered by `subscriptions-transport-ws` (note: this only affects the `subscription` operation, other operations delegate to the previous fetcher).

### DIFF
--- a/packages/graphiql/test/index.html
+++ b/packages/graphiql/test/index.html
@@ -41,6 +41,10 @@
 
     <!-- Note: this resource is not a file, it is bundled live by the test server -->
     <script src="./graphiql.js"></script>
+
+    <!-- We use these files for testing subscriptions -->
+    <script src="//unpkg.com/subscriptions-transport-ws@0.5.4/browser/client.js"></script>
+    <script src="//unpkg.com/graphiql-subscriptions-fetcher@0.0.2/browser/client.js"></script>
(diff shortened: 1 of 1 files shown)


packages/graphiql/test/index.html
```diff
@@ -105,15 +109,16 @@
       }
 
       // Defines a GraphQL fetcher using the fetch API.
+      const graphQLEndpoint = window.location.protocol + '//' + window.location.host + '/graphql';
```
benjie: Websockets needs the full URL, so I'm being explicit. Note `host` includes the port number if non-standard.

packages/graphiql/test/index.html
```diff
@@ -105,15 +109,16 @@
...
-        return fetch('/graphql', {
+        return fetch(graphQLEndpoint, {
           method: 'post',
           headers: {
             'Accept': 'application/json',
             'Content-Type': 'application/json',
           },
           body: JSON.stringify(graphQLParams),
-          credentials: 'include',
+          credentials: 'same-origin',
```
benjie: If we need to test against a different server (e.g. one that supports subscriptions :wink:) then this is necessary to allow that. Should not change behaviour for the current situation.

PR #941, opened 2019-08-18
prepare for next round of releases

acao:
- add changelog
- remove peer deps
- minor version bumps for GLS + codemirror because of license change
- everything ready for next release but GraphiQL

PR #948, opened 2019-08-19
docs: document our release process

benjie:
From discussion with @acao we'd like a steady release process. This `RELEASE_PROCESS.md` documents how the release process works, using `lerna` for much of the automation.

Lerna automatically handles publishing packages in the right order (as can be seen with the `npx lerna changed --long --toposort` command).

Lerna will prompt for the version bump for each package:

Screenshot_20190819_122019

I've configured lerna to publish to the `@next` tag on `npm` by default; however I've not tested this.

I've made lerna forbid versioning on any branch but master.

In future we should look into using semantic versioning to automatically decide if each package should be a major/minor/patch release; but for now I think we can handle this manually.

Note: at this time 2.1.0 of various things is already out (e.g. https://unpkg.com/browse/graphql-language-service-server@2.1.0/) though master still states 2.0.1 - these seem to have been released from within this PR: https://github.com/graphql/graphiql/pull/941


lerna.json
```diff
@@ -1,10 +1,9 @@
 {
-  "lerna": "3.13.1",
```
benjie: Removed:

> `lerna`: originally used to indicate the current version of Lerna. Made obsolete and removed in v3
> -- https://github.com/lerna/lerna#legacy-fields

next2latest
```diff
@@ -0,0 +1,21 @@
...
+
+next2latest graphql-language-service-types
+next2latest graphql-language-service-parser
+next2latest graphql-language-service-utils
+next2latest graphql-language-service-interface
+next2latest graphql-language-service-server
+next2latest graphql-language-service
+next2latest codemirror-graphql
+next2latest graphiql
+next2latest graphiql-example-cdn
```
benjie: This runs in topological order. Maybe it could do with a `set -e` though?

packages/graphiql-examples/cdn/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "graphiql-example-cdn",
+  "private": true,
```
benjie: We either need to not publish this (`"private": true`), or we need to give it a version number. Is this intended to be published to NPM?

packages/graphql-language-service-server/src/__tests__/GraphQLCache-test.js
```diff
@@ -99,6 +100,7 @@ describe('GraphQLCache', () => {
       ).to.deep.equal({
         args: [],
         description: undefined,
+        isRepeatable: false,
```
benjie: These changes pulled in from https://github.com/graphql/graphiql/pull/947

packages/graphiql-examples/cdn/package.json
acao: It should be private: true

acao:
I have an issue for adding semantic release and had a PR for adding it and commitizen I was going to push after manually compiling all the release info, and `cz-conventional-changelog`

benjie:
Changed `@next` references to `@rc` to reflect the decisions in the WG.

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
+# GraphiQL Monorepo Release Process
+
+This document should guide you through doing a release of GraphiQL. Much of
+the release process is automated, so just follow this process through and
+everything should go smoothly!
+
+Note: we run all `lerna` commands with `npx lerna` to ensure that we're using
```
acao: this can also be accomplished with `yarn exec`

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
...
+1. Run `npx lerna version` and choose which version to go to for each package
+2. `git push && git push --tags` to push tags to GitHub
+3. Update changelogs _(TODO: what command pulls in the conventional changelog?)_
+
+Ref: [lerna version](https://github.com/lerna/lerna/tree/master/commands/version)
+
+## Step 2: publish to npm
+
+We always publish to the `rc` tag on npm (i.e. to install you'd do
+`npm install graphiql@rc`) so we can test releases in-situ before making
```
acao: we need to be `yarn add`ing, lockfile, yark workspaces, etc

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. `npx lerna changed --long --toposort`
+
+Ref: [lerna changed](https://github.com/lerna/lerna/tree/master/commands/changed)
+
+## Step 1: cut a version
+
+1. Run `npx lerna version` and choose which version to go to for each package
+2. `git push && git push --tags` to push tags to GitHub
+3. Update changelogs _(TODO: what command pulls in the conventional changelog?)_
```
acao: the conventional changelog is generated from the commit messages, when you run `lerna version`

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. `npx lerna publish from-package` to release unreleased packages (you'll need
+   to enter your OTP; please ensure it has the full 30 seconds left)
+2. If publish fails part way through for 2FA/OTP reasons, just run step 1 again
+
+Ref: [lerna publish](https://github.com/lerna/lerna/tree/master/commands/publish)
+
+## Step 3: test that everything went well
+
+This step left as an exercise for the reader ;)
```
Neitsch: It would be good if had a set of things to run. Maybe we can even add some webdriver tests to our test suite to automate the testing?

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
...
+
+(Pull down the latest `graphiql@rc` or related package into your
+application and test everything worked as expected.)
+
+## Step 4: promote the release to official
+
+Since we release to the `@rc` tag; when we're ready we need to promote the
+releases to `@latest`; you can do so with the handly script:
+
+1. `./rc2latest` (you will be asked for your OTP)
```
Neitsch: I'm not quite clear at which point we actually make the semver release. Is this in step 2 or 4?

RELEASE_PROCESS.md
acao: yes, we will be adding cypress soon at least!

RELEASE_PROCESS.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Step 0: see what would change
+
+1. `npx lerna changed --long --toposort`
+
+Ref: [lerna changed](https://github.com/lerna/lerna/tree/master/commands/changed)
+
+## Step 1: cut a version
+
+1. Run `npx lerna version` and choose which version to go to for each package
```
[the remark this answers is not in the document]
acao: we have some more details, such as semantic commits, etc

acao:
I'm working on a few more commits here @Neitsch to add conventional commits/changelog/release/etc

RELEASE_PROCESS.md
acao: i'm revising this

acao:
Screen Shot 2019-09-01 at 12 21 40 PM
Now _this_ is what I was looking for... :)))

acao:
Now just - commitlint or commitizen?

acao:
Here's the generated output from my PR against this PR
https://github.com/graphql/graphiql/pull/951/files#diff-cb2abf73448a6338d7f7f250233df271R9

acao:
I have it commitint-ing as an after effect of writing your commit message. It's less instrusive, but it also means that if you carefully draft a commit message and it doesn't adhere to the guidelines, that it will possibly be lost the first few times until you get used to it. so possibly a prompt is better? i dont know if i like commitlint's prompt as well as I like commitizen's. le sigh.

acao:
@Neitsch if we try this, merge a bugfix to master with `fix(graphiql): example fix message`, merge that PR and then checkout master, rebase from master and then publish that from master locally, we will see it truly in action

RELEASE_PROCESS.md
benjie: `npx` runs the local copy if there is one; it's shorter than but equivalent to `./node_modules/.bin/lerna` assuming that the modules are installed.

RELEASE_PROCESS.md
benjie: This is saying that we publish to npm under the `rc` tag, so to install it _in your own project_ you'd do `npm install graphiql@rc` (or `yarn add graphiql@rc`). You would not run this inside the GraphiQL project itself. 

This gives the ability for people to test the `@rc` release in their own projects to ensure the deploy worked correctly before promoting it to `@latest` for everyone else to use.

RELEASE_PROCESS.md
benjie: The semver release would be step 2, but it's not official until step 4 (people running `yarn add graphiql` would still get the old version until step 4, so if something goes wrong we can cut a new patch and try again).

acao:
@benjie  @Neitsch  are we good to merge this? its been hangin with no approvals

benjie:
I think the release process should be in a different document as most contributors won't need to worry about that (better to keep `CONTRIBUTING` short so it's less intimidating and people are more likely to read it).

Other than that it all looks good :+1: (though I feel like `husky` should be opt-in rather than opt-out; I'm always uncomfortable adding anything that modifies user's data to OSS projects)

acao:
good advice, we can implement those things in another branch. another way to do this without local hooks is to do the commitlint in CI

benjie:
I think we should definitely do commitlint in CI as well; preferably using something like @danger

acao:
well we have to ensure that every commit message is correct. we can use
danger too, but we need to be able to fail a PR until its git commit
messages are in line

Issue #905, opened 2019-07-04
Prettify Query shortcut (Shift-Ctrl-P) non-functional for Firefox users


justinclift:
Trying GraphiQL for the first time, the default message has a helpful line:

```
Prettify Query:  Shift-Ctrl-P (or press the prettify button above)
```

Unfortunately, at least with Firefox that won't work, as Firefox has a hard coded key binding for "Shift-Ctrl-P".  Instead it just brings up a Private Browsing window, and the key combination doesn't make it through to the GraphiQL window:

![FF1](https://user-images.githubusercontent.com/406299/60685700-e483fd00-9ee7-11e9-9117-570285b92baf.png)

Firefox doesn't natively have any ability to change key bindings either (ugh).  There is a "Saka Key" extension which can apparently allow for key binding changes, but that seems pretty non-optimal.

Is there a way to change the GraphiQL key binding instead?

benjie:
Currently this isn't easily customisable:

https://github.com/graphql/graphiql/blob/3537316fac65e9710ae00d38cae50ff6d75bd7f6/packages/graphiql/src/components/QueryEditor.js#L139-L143

Do you have a recommendation for an alternate keybinding that'd work well in firefox and relates to "prettify"?

justinclift:
Hmmm.  All of the letters in Prettify already have existing key bindings in Firefox.  It may need to be something non-obvious instead. :man_shrugging:

benjie:
What do you suggest?

justinclift:
Both <kbd>Shift</kbd>-<kbd>Ctrl</kbd>-<kbd>V</kbd>, and <kbd>Shift</kbd>-<kbd>Ctrl</kbd>-<kbd>B</kbd> seem available.  Either of those should be ok.  Probably the B one then, just coz? :smile:

benjie:
`B` for `beautify` seems like a good solution :+1:

justinclift:
Sounds good to me. :smile:

benjie:
Once https://github.com/graphql/graphiql/pull/759 is merged, adding this should be as easy as adding a line to this file:

https://github.com/graphql/graphiql/blob/aa82f547b61c8aa40aa4467613456580dd556e95/packages/graphiql/src/utility/commonKeys.js#L8

benjie:
#759 is merged

nasehim7:
@benjie Hi can we directly change the binding like from this:
'Shift-Ctrl-P': () => { 
[...]

PR #950, opened 2019-08-28
fix for #905 - Prettify query shortcut change

nasehim7:
Keyboard bind edit for Prettify shortcut. Fix for #905

### DIFF
--- a/packages/graphiql/src/components/QueryEditor.js
+++ b/packages/graphiql/src/components/QueryEditor.js
@@ -143,6 +143,14 @@ export class QueryEditor extends React.Component {
           }
         },
 
+        /* Shift-Ctrl-P is hard coded in Firefox for private browsing so adding an alternative to Pretiffy */
+
+        'Shift-Ctrl-F': () => {
+          if (this.props.onPrettifyQuery) {
+            this.props.onPrettifyQuery();
+          }
+        },
+
         'Shift-Ctrl-M': () => {
           if (this.props.onMergeQuery) {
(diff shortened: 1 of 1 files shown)


packages/graphiql/src/components/QueryEditor.js
```diff
@@ -143,6 +143,12 @@ export class QueryEditor extends React.Component {
           }
         },
 
+        'Shift-Ctrl-B': () => {
```
benjie: As per https://github.com/graphql/graphiql/issues/905#issuecomment-526861134 we're probably going to move to Monaco at some point; so if `Shift-Alt-F` works in Firefox can we use that instead please? Means less to re-learn in future :)

Also please can you add a comment above this explaining why this change was made?

```suggestion
        /* Shift-Ctrl-P is hard-coded in Firefox for private browsing mode, so add an alternative: */
        'Shift-Alt-F': () => {
```

nasehim7: @benjie Yes understandable. I will make the requested changes

acao:
If you cold revise your commit message, we are just about to begin enforcing conventional commits, makes release management easier
`fix(graphiql): prettify keybinding bug for Firefox. Fixes #905`

nasehim7:
@acao Yes Surely I will do that in a moment and update you here

nasehim7:
@acao Done the change to the commit message

acao:
Great, thank you so much!

PR #955, opened 2019-09-05
Updated broken links to packages in README files

imolorhe:
Some of the links pointing to packages were broken following the move to the monorepo.


Neitsch:
Might be resolved with #953 ?

PR #956, opened 2019-09-09
Upgrade

camray:


acao:
@camray thanks so much for this! this looks like a nice start, however we have a series of steps we want to take here, and our redesign comps put the tab sidebar on the left
can you start with a PR just upgrading react 16?

camray:
Hi @acao ! Sorry this request was made in error. I selected the wrong project as a merge target.

PR #957, opened 2019-09-14
Convert LSP packages to TS (WIP)

acao:
- [x] - `graphql-language-service-utils`
- [x] -  `graphql-language-service-types`
- [x] - `graphql-language-service-parser`
- [x] - `graphql-language-service-interface`
- [x] - shared tooling for composite build
- [x] - update linting/etc scripts for typescript, add typescript-eslint, etc


package.json
```diff
@@ -24,13 +24,14 @@
...
     "lint": "eslint packages/**/src || (printf '\\033[33mTry: \\033[7m yarn run lint -- --fix \\033[0m\\n' && exit 1)",
+    "lint:fix": "eslint 'packages/*/src/**/*.{ts,js,jsx,tsx,json}' --fix",
     "lint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
     "check": "flow check --show-all-errors",
     "prepublish": "node resources/prepublish.js",
     "pretty": "node resources/pretty.js",
     "pretty-check": "node resources/pretty.js --check",
     "version:release": "lerna version",
-    "version:prerelease": "lerna version --conventional-prerelease",
+    "version:prerelease": "lerna versiton --conventional-prerelease",
```
benjie: Typo?

packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
```diff
@@ -76,7 +75,7 @@ export class GraphQLLanguageService {
     // Perform syntax diagnostics first, as this doesn't require
     // schema/fragment definitions, even the project configuration.
     let queryHasExtensions = false;
-    const projectConfig = this._graphQLConfig.getConfigForFile(uri);
+    const projectConfig = this._graphQLConfig.getConfigForFile(uri) as GraphQLProjectConfig;
```
benjie: Prefer 
```suggestion
    const projectConfig: GraphQLProjectConfig = this._graphQLConfig.getConfigForFile(uri);
```

since the `as` syntax [looses type safety](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions). It does seem that the type is being sourced from `graphql-config` so this should be safe?

https://github.com/graphql/graphiql/blob/6878d7c9c4bf9dc586b003b8effaa8d5f892cad4/packages/graphql-language-service-types/src/index.ts#L29

packages/graphql-language-service-interface/src/__tests__/getDefinition-test.ts
```diff
@@ -43,6 +40,10 @@ describe('getDefinition', () => {
           },
         ],
       );
+<<<<<<< HEAD:packages/graphql-language-service-interface/src/__tests__/getDefinition-test.js
+=======
+
+>>>>>>> chore: Convert LSP tests to jest from mocha (#871):packages/graphql-language-service-interface/src/__tests__/getDefinition-test.ts
```
benjie: Merge conflict

packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
acao: thanks yeah, ill look into this some more. had to cast the config somehow, because graphql-config returns as Maybe<> essentially, which would give a bunch of errors where it's used. The type inference seemed to work either way but I will do this instead as a better practice

packages/graphql-language-service-interface/src/__tests__/getDefinition-test.ts
acao: thanks, theres some cleanup as i started on this before we merged the lsp jest conversion

acao:
another big commit on the way, down to about 52 compilation errors across all the packages!

packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
acao: unfortunately there isn't a way around this without breaking the refactor bubble, since the type we get back from `graphQLConfig.getConfigForFile()` will always give us `GraphQLConfig | undefined`, thus we need to check for whether it's undefined every time if we go that route.

acao:
oh so this weekend it was fully compiling, tests are mostly working too, just having an issue with the typescript and babel eslint parsers butting heads, so this won't be complete until all the tooling supports both flow and typescript! @Neitsch if you wanted to finish/polish up be my guest!

benjie:
@acao I was handling this same problem in the Graphile codebase on Monday; I've managed to make ESLint, Flow and TypeScript play fairly well together. Here's my ESLint config in case it helps:

https://github.com/graphile/graphile-engine/blob/bc047b2f86a187c6aa3621d8cd4f423f6cbaa2fc/.eslintrc.js

One thing of note is that you cannot do `extends` in the `overrides` so you have to list the rules manually.

Neitsch:
Thanks @benjie  - with that in hand it was super easy to resolve the lint errors. Let's see what travis has to say :)

benjie:
Awesome @Neitsch I'm glad it was helpful. It certainly took me a little while to get it working smoothly!

CI: looks like there's issues resolving the packages (yarn workspaces should have handled this for us?)

acao:
damn @benjie thanks for that babel eslint + ts eslint parser solution!  i spent a few hours flummoxed on that. maybe we should let [this poor soul](https://github.com/manuelbieh/experimenting-with-babel-7-typescript-and-eslint) know because they have a whole repo devoted to trying to resolve this issue.

acao:
also, thanks y'all in general for the save! please don't wait for me if you want to carry this to completion. also @rebornix on dischord from vscode might be able to give a review for this or later iterations to point out which official vscode types we might be able to leverage, reducing redunancies and making us more standard adherent i think?

acao:
also @orta what do you think? i used `type`s quite liberally instead of `interface`s mostly because i was bein kinda lazy about converting them from flow, and `types` seemed appropriate most of the time. Any chance that you might see some of them as a type smell? Or other issues in terms of official typescript best practices?

packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
```diff
@@ -0,0 +1,321 @@
+/**
```
acao: noting the path for these files wasn't rewritten, and there really weren't that many changes. not a huge deal also not ideal

packages/graphql-language-service-interface/src/__tests__/GraphQLLanguageService-test.ts
```diff
@@ -5,9 +5,9 @@
  *  This source code is licensed under the license found in the
  *  LICENSE file in the root directory of this source tree.
  *
- *  @flow
  */
 
+ import { Position } from 'graphql-language-service-types'
```
acao: format bug, ope

Neitsch:
@acao - I was able to resolve the flow type issues, but tests are failing now. Do you have any idea what could be going wrong here?

acao:
@Neitsch I’m hoping to have time to look tomorrow, looks like some typescript errors from the tests themselves? Were these working before?

Neitsch:
@acao  - it looks like these errors are stemming from `ts-jest` performing some diagnistics. I threw in ts-ignore where applicable, but some errors are just weird and I don't quite know what caused those

acao:
@Neitsch try this in the jest config:
```js
projects: ["packages/*"]
```
so that the downstream subpackages become the baseDir and their tsconfigs are picked up. Last time I had issues like this, this resolved quite a few of them. Esp in situations where the ts compile works fine, but then suddenly ts jest is throwing a bunch of typescript compilation errors. There might be a little bit more fine tuning too.

I really hope I have time to look at this and the accessibility PR tonight but it all depends on whether a soon to be ex roomate decides to get around to cleaning out their stuff tonight, unfortuantely it seems very unlikely at this point.

acao:
@benjie I can't seem to find where in chat or on github you showed the `babel-eslint` and `@typescript-eslint` parsers config working side by side in the same config??

acao:
sadly, had to do a merge commit, because there were other merge commits in the master branch upstream. If we stop doing merge commits then we can rebase much more easily!

acao:
Screen Shot 2019-10-03 at 10 24 40 AM

I'm getting these errors again after merging in the upstream? looks like we aren't linting TS anymore?

acao:
we are using eslint 6.4 now in master:
https://github.com/graphql/graphiql/blob/master/package.json#L54

in this branch we still have 5.16.0 after i merged remote in?:
https://github.com/graphql/graphiql/blob/ea7894628194b97b30207facdf9d6a241147a993/package.json#L56

something funky here

acao:
@benjie @Neitsch i have to get to work now, any chance you guys can look at this today? if not ill keep at it tomorrow. from this branch locally I did `git merge master -Xours` and i think we got some issues from that.

Neitsch:
CI is failing. Would you mind updating the yarn lockfile?

acao:
@Neitsch unfortunately its more than updating the lockfile, if you see the erroneous eslint warnings above, shows that `@typescript-eslint/parser` is not working

Issue #949, opened 2019-08-27
Proposal: improve visibility of <code> tag in doc explorer


yroseau:
When we add code with backquote in graphql documentation, it's not easy to read.

Now:
![Screenshot from 2019-08-27 15-28-09](https://user-images.githubusercontent.com/9842982/63775364-79e7b000-c8df-11e9-95c5-aeb8101581d3.png)

Proposal:
![Screenshot from 2019-08-27 15-25-01](https://user-images.githubusercontent.com/9842982/63775369-7d7b3700-c8df-11e9-8ce5-bbcce8264098.png)

```css
code, pre {
    --saf-0: rgba(var(--sk_foreground_low,29,28,29),0.13);
    font-size: 12px;
    line-height: 1.50001;
    font-variant-ligatures: none;
    white-space: pre;
    white-space: pre-wrap;
    word-wrap: break-word;
    word-break: normal;
    -webkit-tab-size: 4;
    -moz-tab-size: 4;
    tab-size: 4;
}
code {
    padding: 2px 3px 1px;
    border: 1px solid var(--saf-0);
    border-radius: 3px;
    background-color: rgba(var(--sk_foreground_min,29,28,29),.04);
    color: #e01e5a;
    background-color: white;
}
```

yroseau:
Sure, I going to make a PR this week ;)

PR #959, opened 2019-09-20
#949 Beautify code tag in doc explorer

yroseau:
Sorry for the delay :)

See https://github.com/graphql/graphiql/issues/949

### DIFF
--- a/packages/graphiql/css/doc-explorer.css
+++ b/packages/graphiql/css/doc-explorer.css
@@ -94,6 +94,33 @@
   margin-bottom: 4px;
 }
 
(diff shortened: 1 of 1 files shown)


Neitsch:
LGTM. @acao - CSS is not my strength. Any concerns here?

acao:
i was looking through the netlify example but i dont know if we have any concrete examples for this in our test schema. so ill give this a try tonight and see if i can break it, but overall it looks good!

acao:
@yroseau merged, thanks for your contribution!

PR #960, opened 2019-09-23
Add URL parameters for field and type DocExplorer links

bonniee:
Hey folks,

Thanks for the guidance in Discord. I ended up not going for a full overhaul of the DocExplorer state and instead figured out how to implement this as a fairly isolated change.

This change adds permalinks for field and type docs in the DocExplorer. In this example, they look like `http://localhost:62087/?initialTypeName=MutationType&initialFieldName=setString`.

- Add new props to `GraphiQL` and `DocExplorer`: `initialTypeName` and `initialFieldName`, which can be used to set which type/field to render when the component loads a new schema
- Add new callback prop for `GraphiQL` and `DocExplorer`: `onShowDoc`, which is called when a new doc is shown. Used to update URL parameters as the user browses the documentation.
- Update the `test/index.html` page accordingly

I've tested this locally, but let me know if there are other validation steps I should take care of. Thanks!


benjie:
> I've tested this locally, but let me know if there are other validation steps I should take care of. Thanks!

Hi @bonniee; please click through to the CI results ("Details" above) and address the linting issues :+1:

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -54,6 +57,29 @@ export class DocExplorer extends React.Component {
...
+      const typeMap = this.props.schema.getTypeMap();
+      const fullType = typeMap[this.props.initialTypeName];
+
+      if (fullType && this.props.initialFieldName) {
+        if (fullType.getFields) {
+          const fullField = fullType.getFields()[this.props.initialFieldName];
+          if (fullField) {
+            this.showDoc(fullField);
+          }
+          this.showDoc(fullType.getFields()[this.props.initialFieldName]);
```
benjie: I think you meant to delete this line

```suggestion
```

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -54,6 +57,29 @@ export class DocExplorer extends React.Component {
...
+        }
+      } else if (fullType) { // Make sure that fullType exists.
+        this.showDoc(fullType);
+      }
```
benjie: If the field doesn't exist but the type does, I think we should display the type, so something more like:

```js
      const { initialTypeName, initialFieldName } = this.props;
      const typeMap = this.props.schema.getTypeMap();
      const fullType = typeMap[initialTypeName];
      const fieldMap = fullType ? fullType.getFields() : null;
      const fullField = fieldMap && initialFieldName ? fieldMap[initialFieldName] : null;

      if (fullField) {
        this.showDoc(fullField);
      } else if (fullType) {
        this.showDoc(fullType);
      }
```

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -142,6 +168,24 @@ export class DocExplorer extends React.Component {
...
+        fieldName = '';
+      } else {
+        fieldName = typeOrField.name;
+        const navStack = this.state.navStack;
+        const lastState = navStack[navStack.length - 1];
+        typeName = lastState.name;
+      }
+
+      this.props.onShowDoc(typeName, fieldName)
+    }
```
benjie: Lets move this after the `this.setState` call, inside of the `if` statement.

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -142,6 +168,24 @@ export class DocExplorer extends React.Component {
...
   // Public API
   showDoc(typeOrField) {
+    if (!typeOrField) { return; }
+
+    if (this.props.onShowDoc) {
+      let typeName;
+      let fieldName;
+      if (isType(typeOrField)) {
+        typeName = typeOrField.name;
+        fieldName = '';
```
benjie: I think `null` is more appropriate than the empty string.

```suggestion
        fieldName = null;
```

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -142,6 +168,24 @@ export class DocExplorer extends React.Component {
...
+      } else {
+        fieldName = typeOrField.name;
+        const navStack = this.state.navStack;
+        const lastState = navStack[navStack.length - 1];
+        typeName = lastState.name;
```
benjie: I don't think there's any guarantee that the navStack will be in the right order for this; for example you could navigate to a field from hovering over the query document. Please extract the type name from `typeOrField` (if possible), or find a more reliable way of getting the type.

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -142,6 +168,24 @@ export class DocExplorer extends React.Component {
...
   showDoc(typeOrField) {
+    if (!typeOrField) { return; }
+
+    if (this.props.onShowDoc) {
+      let typeName;
+      let fieldName;
+      if (isType(typeOrField)) {
+        typeName = typeOrField.name;
+        fieldName = '';
+      } else {
```
benjie: I'd like to see an `else if (isField(typeOrField))` here, since we may add other entities such as directives to the `showDoc`.

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -142,6 +168,24 @@ export class DocExplorer extends React.Component {
...
+        fieldName = typeOrField.name;
+        const navStack = this.state.navStack;
+        const lastState = navStack[navStack.length - 1];
+        typeName = lastState.name;
+      }
+
+      this.props.onShowDoc(typeName, fieldName)
```
benjie: On second thoughts, I think `this.props.onShowDoc` should pass `typeOrField` directly to user code; it should be up to the user code to decompress it into the relevant typeName/fieldName/directiveName/etc. You can move the code above this to the example to show how to do it, perhaps.

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -54,6 +57,29 @@ export class DocExplorer extends React.Component {
     );
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.schema && 
+      (this.props.schema !== prevProps.schema) && 
+      this.props.initialTypeName) {
+      // The schema has changed, and we should render an initial type.
```
benjie: I'm not sure that we should do this on componentDidUpdate unless there was no previous schema. My reasoning is that it will clash with the desired "hot load" functionality, where the navStack is maintained over schema replacements, because it will keep pushing the type/field onto the nav stack (assuming it's not already at the top). I think maybe only do it if the nav stack is empty (one entry)?

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -54,6 +57,29 @@ export class DocExplorer extends React.Component {
     );
   }
 
+  componentDidUpdate(prevProps) {
```
benjie: React advises against the use of this method.

> Note:
> These methods are considered legacy and you should avoid them in new code:
> UNSAFE_componentWillUpdate()
> UNSAFE_componentWillReceiveProps()

packages/graphiql/src/components/DocExplorer.js
bonniee: Hmm, ok. I don't think there is a more reliable way of getting the type from `typeOrField` alone: types describe their fields, but fields contain no reference to their types.

I can update the `TypeDoc` component to include the relevant type info when calling its `onClickField` prop, if that works?

packages/graphiql/src/components/DocExplorer.js
bonniee: Hmm, is that true? It looks like componentWillUpdate is deprecated, but componentDidUpdate is not.

https://reactjs.org/docs/react-component.html#componentdidmount

packages/graphiql/src/components/DocExplorer.js
bonniee: Sure, that makes sense. I'm not sure how to include more dependencies in the example, since it just uses `<script>` tags - do you know how I could import the `graphql` package so that I can parse the `typeOrField` data?

bonniee:
Thanks for the quick review.

> I like the idea behind this PR and it definitely seems useful, but my biggest concern is over tying it to the lifecycle methods without breaking future functionality. I think it might be better to hold off and add it once we're on React v16 with our improved state management.

Just out of curiosity, around when do you think the v16 refactor would be done?
If I can rework this to be less coupled to the lifecycle methods, do you think it's something that would be worth merging before the v16 state management changes?

packages/graphiql/src/components/DocExplorer.js
benjie: Oops, I clearly looked at the wrong part of the docs somehow :flushed:

benjie:
React v16 is basically the top priority, but it needs volunteers to actually do the work so it's hard to put a timeline on it. But then at the rate @acao is getting things done, who knows? :rocket:

acao:
its more like a rewrite at this point, but it is very much in progress as RFC work!

acao:
@bonniee closing for now, but we will be sure to use this as reference, and I'll be sure to call you in when we're starting to build out these plugins! this is a fantastic idea and should be an MVP feature for any documentation, imo

PR #962, opened 2019-09-24
fix: check `window` is defined before using it

RxAssim:
Hi !

I was trying to use GraphiQL with Gatsby but I got `window is not defined` issue.

### DIFF
--- a/packages/graphiql/src/utility/StorageAPI.js
+++ b/packages/graphiql/src/utility/StorageAPI.js
@@ -7,7 +7,7 @@
 
 export default class StorageAPI {
   constructor(storage) {
-    if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


packages/graphiql/src/utility/StorageAPI.js
```diff
@@ -7,7 +7,9 @@
 
 export default class StorageAPI {
   constructor(storage) {
-    this.storage = storage || window.localStorage;
+    if (typeof window !== 'undefined') {
```
benjie: This discards the `storage` argument passed to the constructor when on the server side; please fix that:

```suggestion
    this.storage = storage || (typeof window !== 'undefined' ? window.localStorage : null);
```

RxAssim:
@benjie Yes now working. I exported the package with the fix and used it here https://www.assim.me/resume

PR #964, opened 2019-09-26
Fix yarn.lock merge conflict; prevent recurrence

benjie:
- Fixed 6 lint errors
- Use a frozen yarn lockfile on CI so that attempts to change it (e.g. when there are merge conflicts, or out of date dependencies) are caught (see [this build](https://travis-ci.org/graphql/graphiql/jobs/590013798) to see this in action)
- Fix merge conflicts in yarn.lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: node_js
-node_js: 
+node_js:
   - stable
   - lts/*
 before_install:
   - npm config set spin false --global
 install:
-  - yarn --ignore-scripts # ignore prepublish scripts
+  - yarn --frozen-lockfile --ignore-scripts # ignore prepublish scripts
 after_success:
   - codecov
 cache: yarn
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,7 +29,8 @@ module.exports = {
     es6: true,
     node: true,
     jest: true,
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -1,11 +1,11 @@
 language: node_js
-node_js: 
+node_js:
   - stable
   - lts/*
 before_install:
   - npm config set spin false --global
 install:
-  - yarn --ignore-scripts # ignore prepublish scripts
+  - yarn --frozen-lockfile --ignore-scripts # ignore prepublish scripts
```
Neitsch: I recall that `yarn` had a `--ci` flag, but I can't find docs on it any more. This example is linked as the recommended style (https://codefresh.io/docs/docs/learn-by-example/nodejs/react/). Can we maybe try setting the environment variable `CI=true`?

benjie: Doesn’t most CI set CI to true for you?

https://docs.travis-ci.com/user/environment-variables/#convenience-variables

benjie: Yarn docs:

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag

That’s the only mention of continuous integration on that page; I believe it’s the currently recommended way.

PR #965, opened 2019-09-26
chore: overhaul ESLint integration

benjie:
This PR is built on #964; so we should merge that first before reviewing this.

- Upgrades ESLint to latest version
- Replaces unmaintained eslint-plugin-dependencies with eslint-plugin-import (replacing the rules)
- Standardises on `.eslintrc.js` rather than mixing JS and YAML (see note below)
- Uses `.eslintignore` cleverly so we don't need to pass explicit paths to lint (means we can lint files outside of the `src` folder)
- Fix some lint issues that were not checked for previously

This is just the beginning; if it's better than the current status quo lets get it merged before discussing other requests. I'd really like to overhaul our ESLint/prettier integration, and standardising it is the first step.

#### Why JS and not YAML for ESLint config?

Mainly that `yaml` [gets security issues raised periodically](https://snyk.io/vuln/npm:js-yaml), so reducing our dependence on it reduces our maintenance burden; but also it's more in line with the rest of the codebase and enables us to lint the lint config with the lint config :joy:


benjie:
That’s something I plan to address in a follow up PR; I believe the reason is that GraphiQL had its own eslintrc before the monorepo merge, and merging them is non-trivial.

benjie:
I've rebased because #964 has been (squash) merged; but the rebase was clean, no changes were necessary.

acao:
would like to be sure @IvanGoncharov can review some of this for another pass, since we had worked on trying to standardize using this pattern, but yeah it was pretty messy since it involved combining eslint configs from three seperate repos!

Also, note that graphiql still has its own `eslintrc.js` after this PR

benjie:
Yep, my plan was to get them into the same language before attempting to merge them. With lockfiles like yarn.lock we don’t need to worry so much about explicitly stating the rule list as we used to, so I think moving to some standard presets (eslint:recommended, prettier, etc) would make a lot of sense. Was planning to look at this in the next couple weeks. Will tag Ivan on the next ESLint PR 👍

IvanGoncharov:
@acao @benjie Thanks for keeping me updated 👍 
I think it makes a lot of sense to keep ESLint config at the same format so we can diff them. Here how it currently looks in `graphql-js`:
https://github.com/graphql/graphql-js/blob/master/.eslintrc.yml
`.js` has a few potentials issues:
* You need to ignore it with ESLint itself but also Flow, TS, etc. since it should be native JS supported by all currently supported Node versions. 
* Also it requires a lot of unnecessary formatting:
```
  'init-declarations': 'off',
  'no-delete-var': 'error',
  'no-label-var': 'error',
```
versus YAML
```
  init-declarations: off
  no-delete-var: error
  no-label-var: error
```
* Prettier forces inconsistent formatting on keys:
```
    curly: 'error',
    'default-case': 'off',
```

IvanGoncharov:
@acao @benjie BTW. In `graphql-js` I removed `eslint-prettier` and switched to just run prettier globally on all files:
https://github.com/graphql/graphql-js/blob/21de98ea4855df8c5146580611fda8ad76f52a2f/package.json#L33-L34
So it now it formats all JSON, YAML and even markdown files.

Issue #954, opened 2019-09-05
Discussion: Editor and Documentation Accessibility


cmoody:
Howdy,

This issue is designed to open up a discussion about the accessibility in GraphiQL.  Currently at Vrbo we have a team member who is blind and since most tooling is not accessible first, it makes it difficult for them to both learn and utilize GraphQL to its fullest.

I wrote the title as Editor and Documentation because I think each could have their own challenges as we look at what is needed.

**Editor:**
I did a quick test of a few screen readers.  For the most part all of them get stuck in the editor with no way out.  This could also be due to my lack of knowledge around screen readers so warrants more investigation.

**Documentation:**
It was hard to test this too much because it tended to toss me back into the editor.  I think GraphQL lends itself nicely to make documentation a wonderful experience.  Having good descriptors on moving from parent to child I think will make navigating through a breeze.

Im a novice when it comes to accessibility but hope to help out and my team member is up for testing out any features.

acao:
thanks @cmoody for reporting! I'll be sure to check it out in greater depth this weekend.

acao:
It looks like Codemirror 6 would introduce some improvements for the editor:
https://github.com/codemirror/CodeMirror/issues/4604

However, its not certain when that will come out, and we will be switching to Monaco in a few months anyways

Neitsch:
@cmoody, accessibility is important for GraphiQL to be inclusive. I appreciate that you file this issue. Are you / your coworker aware of how to approach this problem? I recall that there are some linters to help a11y compatibility.
Anything that would help us have a more structured approach would be appreciated!

acao:
in my experience, the linters can only do so much. for example in react, we can lint the static prop names of components but not values before theyd been cast. So a linter like a11y is great to catch things statically, but then you need to do runtime checks in CI, there are tools we can use with or alongside e2e suites, and thatll help us keep track of many more issues. So I would say first step is an end to end suite, and adding the linter, and then adding a more comprehensive check against, say, our netlify@next that maybe even gates against accessibility debt introduced in a PR. I architected something like this for the new medicaid/medicare patient enrollment forms when i was consulting for CMS in baltimore. 508 had just synced up with WCAG2 at that point, so im quite familiar with using that as a standard.

acao:
Here's a really cool example, wish we had this then!
https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/blogs__a11y

[...]

PR #967, opened 2019-09-27
fix(accessibility): improve accessibility of all components

kasperisager:
This pull request brings a baseline of accessibility to the various GraphiQL components with a focus on making them usable by keyboard and/or screen readers. Some highlights worth mentioning:

- The various sections of the interface are now implemented as regions (using `<section>` elements) with proper accessible names (using `aria-label`) for identifying their purpose. This is especially important for users who rely on screen readers as it's otherwise very difficult to identify and navigate the different parts of the interface, in particular the query editor and the results viewer.

- All components are now navigable by keyboard, with especially the history pane benefitting from this. Previously, the history pane could not be navigated by keyboard at all as interactions relied solely on mouse events (`mouseenter` and `mouseleave`).

- In extension of the previous point, all interactive components are now implemented as `<button>` elements, ensuring that they correctly communicate to users of assistive technology that they can be interacted with and that standard keyboard interaction works as expected. This was especially critical for the history pane components, which were previously implemented as `<span>` elements without any indicators that users could interact with them beyond a cursor pointer.

- All interactive icons now have accessible names, ensuring that their behaviour is communicated to users of assistive technology. Previously, one would have to guess what these components did when interacted with as the only thing announced were words such as "cross" for the "close" icon or "neuter" for the "search" icon.

- The results viewer is now a live region, ensuring that users of screen readers are made aware of changes to results when queries are evaluated.

I'd love to hear from @cmoody if there are other critical issues not addressed in this pull request that could help solve #954.

### DIFF
--- a/packages/graphiql/src/components/DocExplorer/SearchBox.js
+++ b/packages/graphiql/src/components/DocExplorer/SearchBox.js
@@ -26,16 +26,21 @@ export default class SearchBox extends React.Component {
   render() {
     return (
       <label className="search-box">
+        <div className="search-box-icon" aria-hidden="true">{'\u26b2'}</div>
         <input
           value={this.state.value}
           onChange={this.handleChange}
           type="text"
           placeholder={this.props.placeholder}
+          aria-label={this.props.placeholder}
         />
         {this.state.value && (
-          <div className="search-box-clear" onClick={this.handleClear}>
+          <button
+            className="search-box-clear"
+            onClick={this.handleClear}
+            aria-label="Clear search input">
             {'\u2715'}
-          </div>
+          </button>
         )}
       </label>
     );
--- a/packages/graphiql/src/components/DocExplorer/TypeLink.js
+++ b/packages/graphiql/src/components/DocExplorer/TypeLink.js
@@ -43,7 +43,7 @@ function renderType(type, onClick) {
     );
   }
   return (
-    <a className="type-name" onClick={event => onClick(type, event)}>
+    <a className="type-name" onClick={event => {event.preventDefault(); onClick(type, event)}} href="#">
       {type.name}
     </a>
   );
--- a/packages/graphiql/css/app.css
+++ b/packages/graphiql/css/app.css
@@ -78,7 +78,6 @@
   cursor: pointer;
   font-size: 14px;
   margin: 0;
-  outline: 0;
   padding: 2px 20px 0 18px;
 }
 
@@ -151,6 +150,9 @@
   font-size: 18px;
   margin: -7px -8px -6px 0;
   padding: 18px 16px 15px 12px;
+  background: 0;
+  border: 0;
+  line-height: 14px;
 }
 
 .graphiql-container div .query-editor {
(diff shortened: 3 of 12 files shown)


acao:
@kasperisager this is a godsend, thank you!! cant wait to give this a closer look tomorrow

packages/graphiql/src/components/DocExplorer/SearchBox.js
```diff
@@ -26,16 +26,21 @@ export default class SearchBox extends React.Component {
   render() {
     return (
       <label className="search-box">
+        <div className="search-box-icon" aria-hidden="true">{'\u26b2'}</div>
         <input
           value={this.state.value}
           onChange={this.handleChange}
           type="text"
           placeholder={this.props.placeholder}
+          aria-label={this.props.placeholder}
```
Neitsch: Why use the placeholder here? Shouldn't it be `Search Input`?

kasperisager: It really shouldn't be the _placeholder_, that's for sure. Placeholders are no replacement for proper labels though they are quite often (mis)used as such, which is what I was guessing was the case here. Seeing how the placeholder was therefore really used as a label, I wanted to make sure that this label was communicated to assistive technology. 

While the `placeholder` attribute itself _should_ be used when computing the accessible name exposed to assistive technology, this functionality was only introduced recently and is still not widely supported: https://github.com/w3c/html-aam/issues/167. I therefore opted to expose the placeholder explicitly through an `aria-label`.

A better way to go about this would be to add a proper label to the search input, though this would require a slightly more involved change to the design of the interface.

acao:
@kasperisager this looks wonderful so far! I was wondering, is there a reason why `preventDefault()` is removed throughout? Is it to improve keyboard navigability?

packages/graphiql/src/components/DocExplorer/TypeLink.js
```diff
@@ -43,7 +43,7 @@ function renderType(type, onClick) {
     );
   }
   return (
-    <a className="type-name" onClick={event => onClick(type, event)}>
+    <a className="type-name" onClick={event => {event.preventDefault(); onClick(type, event)}} href="#">
```
acao: here, `event.preventDefault()` is added, whereas elsewhere it is removed, any particular reason?

kasperisager: With the exception of form submission buttons, there’s no default behavior to prevent for `<button>` elements. For `<a>` elements, their default behavior is however to navigate, which we need to prevent if we want to use it as a sort of button. The `TypeLink` is actually a case where this is somewhat appropriate as it is used for internal links within the documentation explorer.

It is also worth pointing out that `<a>` elements need an `href` attribute for them to have any behavior. This has an effect on both keyboard operability and assistive technology as an `<a>` element without an `href` attribute is neither focusable nor communicates any semantics.

benjie: Great explanation

acao:
I've evaluated the [netlify build](https://deploy-preview-967--graphiql-test.netlify.com/) extensively and everything looks great, looks like you resolved almost all of the non-codemirror accessibility issues!

kasperisager:
My pleasure! 😄

PR #971, opened 2019-10-05
chore: cleanup for publishing, move examples directory

acao:
- prepublish script assumed every directory beneath `packages/` would contain a package.json
- examples directory began with `graphiql` causing issues with tooling. plus it's about to become more generic, and include simple examples of codemirror and monaco even
- remove examples from conventional commits versioning scope
- remove codemirror prepublish for default script


acao:
Also found that `lerna publish` for whatever reason isn't working after lerna version works successfully

PR #972, opened 2019-10-05
chore: prettier reset trailingComma: all

acao:
in the eslint upgrade and typescript conversion kerfuffle we accidentally reverted to es5 commas briefly. re-adding all those es2017 compatible commas!


acao:
Screen Shot 2019-10-06 at 5 23 17 PM
this is failing because it changes code coverage :/

acao:
@Neitsch this needs rebased but that new file will already exist as of the new master
The change is to `.prettierrc`. In this PR, we change from `es5` `trailingCommas` (which is why some commas were just recently stripped, because we had enforced this rule) back to `all` `trailingCommas` (since es2017 supports trailing commas in functions).

acao:
as you can see though, adding these extra commas somehow decreases coverage significantly

acao:
@Neitsch again, everything passes but codecov's test coverage comparison :/

.eslintignore
```diff
@@ -33,5 +33,12 @@
...
 # Vendored files
-/packages/codemirror-graphql/
+**/graphiql.min.js
+**/graphiql.css
+**/graphiql.js
+
 /packages/graphiql/test/vendor
+
+# Funky codemirror-graphql build pattern
+/packages/codemirror-graphql/**
```
benjie: You have to be very careful with this pattern; if this line matches a directory then files in that directory _will not be re-included_ by the `!` rule. ESLint uses the same algorithm as gitignore; which states:

> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

To fix you have to match JUST the files explicitly:

```suggestion
/packages/codemirror-graphql/**/*.ts
/packages/codemirror-graphql/**/*.tsx
/packages/codemirror-graphql/**/*.js
/packages/codemirror-graphql/**/*.jsx
```

Maybe we should note this more strongly here: https://github.com/graphql/graphiql/blob/2da818b1f4e7adb1f1e519e314a428e846dc257e/.eslintignore#L7

.prettierignore
```diff
@@ -0,0 +1,7 @@
+**/coverage/**
+**.graphql
+.graphqlconfig
+
+# breaks prettier graphql parser
+packages/graphql-language-service-server/src/__tests__/__queries__/test.graphql
+packages/codemirror-graphql/src/__tests__/schema-kitchen-sink.graphql
```
benjie: Maybe we should ignore all the graphql files in `__tests__`; some of them might test whitespace?

resources/pretty.js
```diff
@@ -14,7 +14,7 @@ const INVERSE = '\x1b[7m';
 const RESET = '\x1b[0m';
 const YELLOW = '\x1b[33m';
 
-const options = ['--trailing-comma=es5'];
+const options = ['--config ../../.prettierrc'];
```
benjie: These options should be separate (for `spawn`), and we should pass a fixed path so that it doesn't matter where the script is called from.

```suggestion
const options = ['--config', `${__dirname}/../.prettierrc`];
```

acao: really, the script is so simple now, we could just invoke it directly

.prettierignore
acao: good point... that reminds me of at least one test "fix" i had to make that was suspicious in this PR...

.eslintignore
acao: ok, the above suggestion works except it needs to ignore everything but `src` though. the whole issue with `codemirror-graphql` is that it builds to the root for easy resolutions, and the new eslintignore file doesnt ignore the built files anymore. the subpackage `.gitignore` honors this and we need the `.eslinignore` to do the same

benjie:
Sounds like a plan. Fair warning: I am not a fan of YAML.

Neitsch:
Thanks for fixing the trailing comma stuff! Let's ship it 👍 
Then we can address @benjie improvement suggestions

acao:
@Neitsch @benjie i think we should scrap this PR personally. We already have es5 commas, so all object literals have trailing commas, etc. This just adds them to function parameters. We are gonna need to add a bunch of test coverage to the LSP parser or server to get this to pass. So it could take me a while

benjie:
I wouldn't worry about increasing code coverage when the changes are just formatting. It's okay to ignore that one status check :)

.eslintignore
benjie: Yes; that's why I left the `!.../src/**` line below :) It's only the actual `/**` line I changed to make more explicit to only match files and not folders. I believe this will achieve the outcome you desire.

benjie:
@acao  I'd like to see this PR done as two separate commits:

- Commit 1: changes the configuration of ESLint/prettier/etc
- Commit 2: autofix changes ONLY - no manual changes.

This will make it a lot easier to review.

I believe we can completely ignore codecov for this PR.

acao:
I agree on the approach, but I think we should just close it. we dont even
need this config change. we already have es5 commas

benjie:
It's certainly not a big priority. Let's close it and come back to it later.

PR #973, opened 2019-10-06
feat: resolve modules using extensions for webpack, export esm, build with webpack

acao:
This solves #882
- [x] - file util for resolving modules in a webpack friendly way where extensions are explicit (with tests)
- [x] - export esm modules from typescript projects (towards better tree shaking)
- [x] - export esm modules from babel projects (towards better tree shaking)
- [x] - make react eslint config available outside of graphiql's eslint config
- [x] - bundle graphiql using webpack, get rid of browserify
- [x] - use dynamic import
- [x] - full webpack example, showing the config workaround is no longer necessary, addressing #882
- [x] - update documentation for new build approach, cdn and webpack
- [x] - create multiple bundles for CDN export
- [x] - export dynamic imports with CDN path (unpkg)
- [x] - new `renderGraphiQL` helper function


packages/graphql-language-service-interface/src/fileUtils.ts
```diff
@@ -0,0 +1,57 @@
+function getExtension (pathParts: string[]): string | null {
+  return pathParts.pop() || null
```
benjie: Should this really mutate the argument?

```suggestion
  return pathParts.length > 0 ? pathParts[pathParts.length - 1] : null
```

packages/graphql-language-service-interface/src/fileUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function removeExtension  (filePath: string, pathParts: string[]): string {
+  // if there's a file extension
+  if (pathParts.length > 1) {
+    // pop the extension from the array
+    pathParts.pop() as string
```
benjie:
```suggestion
    pathParts.pop()
```

packages/graphql-language-service-interface/src/fileUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+  else {
+    return filePath
+  }
+}
+
+// these make webpack happy
+
+export function resolveFile (filePath: string) {
+  const pathParts = filePath.split('.')
+  const extension = getExtension(pathParts)
```
benjie: This does not seem a good way of getting the extension. For example it would mess up on the path `/home/me/my.work/foo` or `/path/to/something/../otherthing/foo` and it would be expensive on a path with lots of dots in.

packages/graphql-language-service-interface/src/fileUtils.ts
benjie: This probably shouldn't mutate the argument.

packages/examples/graphiql-webpack/src/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    <div>
+      <GraphiQL 
+        
+      />
+    </div>
+  )
+}
+
+render()
```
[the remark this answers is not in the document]
acao: oh yeah this part isn't finished, haha

packages/examples/graphiql-webpack/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+        use: ["style-loader", "css-loader"]
+      },
+      {
+        test: /\.svg$/,
+        use: [{ loader: "svg-inline-loader" }]
+      }
+    ]
+  },
+  resolve: {
+    extensions: [".js", ".json", ".ts", ".tsx", ".jsx", ".css", ".mjs"],
```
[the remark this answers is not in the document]
acao: good point. copying this over from a previous working webpack config. my goal is to trim down on the webpack config as much as possible, as this PR is meant to introduce a fix for webpack

packages/examples/graphiql-webpack/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+    new HtmlWebpackPlugin({ template: "index.html.ejs" }),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-interface[\\/]dist$/, new RegExp(`^\\./.*\\.js$`)),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-utils[\\/]dist$/, new RegExp(`^\\./.*\\.js$`)),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-parser[\\/]dist$/, new RegExp(`^\\./.*\\.js$`))
+  ],
+  devServer: {
+    hot: true,
+    // bypass simple localhost CORS restrictions by setting
+    // these to 127.0.0.1 in /etc/hosts
+    allowedHosts: ["local.graphiql.com", "graphiql.com"]
```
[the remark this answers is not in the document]
acao: that works better. it was `bob.autoparts.com` which is my usual go-to local psudeo host 😆

packages/graphiql/index.d.ts
```diff
@@ -0,0 +1,42 @@
+declare module 'graphiql' {
+  import React from 'react'
+  import { GraphQLSchema} from 'graphql'
```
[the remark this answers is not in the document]
acao: whats wrong with the formatting?

packages/graphql-language-service-interface/src/fileUtils.ts
acao: ah yes, convention over terseness here.

packages/graphql-language-service-interface/src/fileUtils.ts
acao: well `pop()` can return `null | string` so it gets mad if i dont do this

packages/graphql-language-service-interface/src/fileUtils.ts
acao: yeah good point. i was hoping just popping the last item would work, but that won't work with extensionless paths. ill write some unit tests that take this into account and rewrite accordingly. I want to use the `path` module because it has great utilities for this, but we are trying to steer GLS-interface away from being node dependent, since we need it to be able to be mounted in the browser, thus fully isomorphic in the coming months.

packages/examples/graphiql-webpack/webpack.config.js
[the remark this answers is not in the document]
benjie: Aside: `something.example.com` might be more appropriate in future:

- http://example.com/
- https://www.iana.org/domains/reserved

packages/graphiql/index.d.ts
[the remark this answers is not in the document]
benjie: A very subtle spacing error that you should not have to think about because prettier should do it for you:

```diff
-  import { GraphQLSchema} from 'graphql'
+  import { GraphQLSchema } from 'graphql'
```

packages/graphql-language-service-interface/src/fileUtils.ts
benjie: I think webpack at least has polyfills for the `path` module. Likely other bundlers do too?

packages/examples/graphiql-webpack/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+  resolve: {
+    extensions: [".js", ".json", ".ts", ".tsx", ".jsx", ".css", ".mjs"],
+    plugins: [new TsConfigPathsPlugin()]
+  },
+  plugins: [
+    new HtmlWebpackPlugin({ template: "index.html.ejs" }),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-interface[\\/]dist$/, new RegExp(`^\\./.*\\.js$`)),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-utils[\\/]dist$/, new RegExp(`^\\./.*\\.js$`)),
+    new webpack.ContextReplacementPlugin(/graphql-language-service-parser[\\/]dist$/, new RegExp(`^\\./.*\\.js$`))
```
gpoitch: Just stumbled upon this PR wondering why there were a bunch of warnings emitted with the latest version.  These lines fix it.  Curious why this is necessary and the resolve.extension order wasn't respected.

acao: @gpoitch yes this is exactly the issue we're trying to solve in this PR actually. i've introduced this working webpack config, and once i have the webpack example working all the way to netlify, i will disable these plugin invocations until it is working transparently upstream.

we believe we just have to import dynamically using `require`, or a dynamic import with an explicit extension for webpack to be happy, at least we think that might help. ideally i'd like to be doing things like this:

```js
await import(`graphiql-plugin-${plugin}.js`)
```
for now still using `require` and just iterating slowly til we get there. i guess we'll be breaking people's bundler configs no matter what

acao: https://github.com/graphql/graphiql/pull/973/files#diff-f9580b2b8f2bad829ef1defe703f8e18L162 specifically i beleive its this line, and the other `require` a few lines below it. they were introduced in the 2.x version LSP earlier this year, but we just released the LSP again after a year of no releases so yeah

acao: hey so, good news @gpoitch, i got this working locally, about to push a version with a working netlify webpack build that doesn't require the special config.

acao:
```js
const webpack = require("webpack");
const HtmlWebpackPlugin = require("html-webpack-plugin");
const path = require("path");
const isDev = process.env.NODE_ENV === "development";

module.exports = {
  entry: isDev
    ? [
        "react-hot-loader/patch", // activate HMR for React
        "webpack-dev-server/client?http://localhost:8080", // bundle the client for webpack-dev-server and connect to the provided endpoint
        "webpack/hot/only-dev-server", // bundle the client for hot reloading, only- means to only hot reload for successful updates
        "./index.jsx" // the entry point of our app
      ]
    : "./index.jsx",
  context: path.resolve(__dirname, "./src"),
  mode: "development",
  devtool: "inline-source-map",
  externals: {
    react: "React",
    "react-dom": "ReactDOM"
  },
  performance: {
    hints: false
  },
  module: {
    rules: [
      {
        test: /\.html$/,
        use: ["file?name=[name].[ext]"]
      },
      {
        type: "javascript/auto",
        test: /\.mjs$/,
        use: [],
        include: /node_modules/
      },
      {
        test: /\.(js|jsx)$/,
        use: [{ loader: "babel-loader", options: { presets: ["@babel/preset-env", "@babel/preset-react"]} }]
      },
      {
        test: /\.css$/,
        use: ["style-loader", "css-loader"]
      },
      {
        test: /\.svg$/,
        use: [{ loader: "svg-inline-loader" }]
      }
    ]
  },
  resolve: {
    extensions: [".js", ".json", ".jsx", ".css", ".mjs"],
  },
  plugins: [
    new HtmlWebpackPlugin({ template: "index.html.ejs" }),
  ],
  devServer: {
    hot: true,
    // bypass simple localhost CORS restrictions by setting
    // these to 127.0.0.1 in /etc/hosts
    allowedHosts: ["example.com"]
  },
  node: {
    fs: "empty"
  }
};

```

acao: this works as it should! :D

acao:
The one thing I'm thinking here is, why not use dynamic imports, aka `await import()`? we already have async functions where we need to invoke the require so, why not support it natively? node 8 is being retired, dynamic imports was introduced in node 10, and we can show folks how to configure their webpack/rollup/etc builds for edge/ie/etc support for it as well?

packages/examples/graphiql-webpack/webpack.config.js
acao: @gpoitch the example is ready now, seems to be working, and now we have tree shaking too!

packages/graphiql/README.md
```diff
@@ -127,44 +175,107 @@ GraphiQL supports customization in UI and behavior by accepting React props and
...
-      onEditQuery: null,
-      onEditVariables: null,
-      onEditOperationName: null,
-
-      // GraphiQL automatically fills in leaf nodes when the query
-      // does not provide them. Change this if your GraphQL Definitions
-      // should behave differently than what's defined here:
-      // (https://github.com/graphql/graphiql/blob/master/src/utility/fillLeafs.js#L75)
-      getDefaultFieldNames: null
+      fetcher: this.props.fetcher
```
benjie: Are these state fields documented elsewhere that's easy to copy/paste?

If fetcher is on props then it doesn't really need state; so this whole example doesn't need state?

packages/graphiql/README.md
```diff
@@ -223,137 +334,3 @@ In order to theme the editor portions of the interface, you can supply a `editor
...
-  }
-}
-
-fragment HumanFragment on Human {
-  name
-  homePlanet
-}
-```
-
-Read more from [GraphQL Fragment Specification](https://graphql.github.io/graphql-spec/draft/#sec-Language.Fragments).
```
benjie: I'm hesitant to remove this content.

packages/graphiql/babel.config.js
```diff
@@ -1,13 +1,36 @@
-module.exports = {
+const babelEnvConfig = {
+  // ignore browserslist ONLY when we make our main/modules babel builds
+  ignoreBrowserslistConfig: Boolean(process.env.ESM && !process.env.BUNDLE),
+  modules: process.env.ESM ? false : 'commonjs',
+  targets: process.env.ESM ? 'last 1 chrome version, last 1 firefox version': undefined,
```
benjie: Prettier should be detecting and handling this:

```suggestion
  targets: process.env.ESM ? 'last 1 chrome version, last 1 firefox version' : undefined,
```

packages/graphiql/cypress/integration/init.spec.js
```diff
@@ -39,20 +39,29 @@ describe('GraphiQL On Initialization', function() {
...
     cy.get('.execute-button').click()
+    cy.wait(3000)
     cy.window().then((w) => {
-      cy.expect(JSON.parse(w.g.resultComponent.viewer.getValue())).to.deep.equal(mockSuccess)
+      const result = w.g.resultComponent.viewer.getValue() 
+      if (result) {
+        cy.expect(JSON.parse(result)).to.deep.equal(mockSuccess)
+      }
+      else {
+        throw Error('result not received')
```
benjie: Really should construct Error rather than casting:

```suggestion
        throw new Error('result not received')
```

packages/graphiql/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+    onEditQuery?: Function,
+    onEditVariables?: Function,
+    onEditOperationName?: Function,
+    onToggleDocs?: Function,
+    getDefaultFieldNames?: Function,
+    editorTheme?: string,
+    onToggleHistory?: Function,
+    ResultsTooltip?: any,
+    readOnly?: boolean,
+    docExplorerOpen?: boolean,
```
benjie: Each of these lines should end with a semicolon rather than a comma; prettier should be doing this for you.

packages/graphiql/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  export interface GraphiQLState {
+    schema?: GraphQLSchema,
+    response?: string,
+    query?: string
+  }
+  export type RenderConfig = {
+    /**
+     * the url if no fetcher is provided
+     */
+    url: string
```
benjie: This is currently marked as required; I don't think that's intentional.

packages/graphiql/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * the headers if no fetcher is provided
+     */
+    headers: Headers
```
benjie: Ditto.

packages/graphiql/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+    containerId?: string
+    containerEl?: HTMLElement
+  } & GraphiQLProps;
+  export function renderGraphiQL(config: RenderConfig): Promise<void>
```
benjie: Is config required?

```suggestion
  export function renderGraphiQL(config?: RenderConfig): Promise<void>
```

packages/graphiql/package.json
```diff
@@ -13,21 +13,26 @@
...
     "build": "yarn run build-js && yarn run build-flow",
-    "build-js": "bash ./resources/build.sh",
+    "build-bundles": "bash ./resources/buildBundles.sh",
+    "build-js": "bash ./resources/buildBabel.sh",
     "build-flow": "node ../../resources/buildFlow.js",
     "check": "flow check",
-    "dev": "babel-node test/server.js",
-    "e2e": "yarn e2e:server && cypress run && kill -9 `cat test/pid`",
+    "dev": "PORT=8080 babel-node test/server",
+    "e2e": "kill -9 `cat test/pid` | true && yarn e2e:server && cypress run && kill -9 `cat test/pid`",
```
benjie:
```suggestion
    "e2e": "kill -9 `cat test/pid` || true && yarn e2e:server && cypress run && kill -9 `cat test/pid`",
```

packages/graphiql/package.json
```diff
@@ -13,21 +13,26 @@
   },
   "license": "MIT",
   "main": "dist/index.js",
+  "module": "esm/index.js",
+  "types": "index.d.ts",
   "files": [
     "dist",
-    "graphiql.js",
```
benjie: Are we still bundling non-minified GraphiQL or not? (We should - for better error messages at least.)

packages/graphiql/package.json
```diff
@@ -37,15 +42,12 @@
...
-    "transform": [
-      "babelify"
-    ]
-  },
   "dependencies": {
     "codemirror": "^5.47.0",
     "codemirror-graphql": "^0.11.2",
     "copy-to-clipboard": "^3.2.0",
+    "graphql": "14.5.6",
+    "isomorphic-fetch": "^2.2.1",
```
benjie: We should try to minimise dependencies; I really don't think we should add `isomorphic-fetch`.

We shouldn't pin a specific version of `graphql` as this will massively increase the likelihood of users having multiple versions of GraphQL in their `node_modules` folder. If possible, I don't think we should specify `graphql` at all (we didn't previously?). If necessary, add it as a `devDependency` instead?

packages/examples/graphiql-webpack/src/index.html.ejs
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <title>Hello React!</title>
+  <link rel="stylesheet" href="node_modules/graphiql/graphiql.css" crossorigin="anonymous" />
```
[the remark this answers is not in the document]
benjie: Why does this contain `node_modules`? How does it work?

packages/graphiql/package.json
acao: @benjie here we do load isomorphic fetch asynchronously, only in `renderGraphiQL`, so it doesn't get bundled at all with the default GraphiQL entrypoint, and as an optional chunk if they load the `render.min.js` entrypoint.

@benjie good point, i think we should make `graphql` only a peer, dev and/or optional dependency of all the modules, with ranges for peers/optional like we do already, and then bundle it with the CDN bundles.  I ended up forcing it everywhere because i found with 14.5.8 it was clashing somehow with something that required 14.5.6? need to get to the bottom of that.

packages/graphiql/package.json
acao: yep! my bad

packages/graphiql/package.json
acao: we are yeah, ill be fixing this

packages/graphiql/package.json
benjie: Why do we need `isomorphic-fetch` when we didn't in the past? Can't this be something the user adds if they need it rather than us polyfilling for them? I'm keen to avoid additional dependencies for the npm package whether or not they're bundle split/asynchronously loaded unless we absolutely need them. `node_modules` is already deep enough :wink:

acao: because of the purpose of `renderGraphiQL`.

with `graphiql.min.js` its bring your own fetch implementation for sure
but with `renderGraphiQL`, its meant to be the most minimal lines to get to a working implementation, with everything taken care of, including the DOM render and all. see the new examples in the netlify build (/render). a graphiql instance in 10 lines basically, plus that wrapper CSS

packages/examples/graphiql-webpack/src/index.html.ejs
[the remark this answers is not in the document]
acao: html-webpack-plugin! i think? or maybe this doesnt work and the `import 'graphiql/graphiql.css` below does... good catch

packages/graphql-language-service-interface/package.json
```diff
@@ -31,6 +33,7 @@
     "graphql": "^0.12.0 || ^0.13.0 || ^14.0.0"
   },
   "dependencies": {
+    "graphql": "14.5.6",
```
benjie: :warning:

packages/graphiql/src/utility/onHasCompletion.js
```diff
@@ -15,76 +15,78 @@ const md = new MD();
  * about the type and description for the selected context.
  */
 export default function onHasCompletion(cm, data, onHintInformationRender) {
-  const CodeMirror = require('codemirror');
+  import('codemirror')
+    .catch(console.err)
```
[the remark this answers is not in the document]
acao: why?

packages/graphiql/src/components/QueryEditor.js
```diff
@@ -53,28 +53,33 @@ export class QueryEditor extends React.Component {
...
+      import('codemirror/addon/search/jump-to-line'),
+      import('codemirror/addon/dialog/dialog'),
+      import('codemirror/addon/lint/lint'),
+      import('codemirror/keymap/sublime'),
+      import('codemirror-graphql/esm/hint'),
+      import('codemirror-graphql/esm/lint'),
+      import('codemirror-graphql/esm/info'),
+      import('codemirror-graphql/esm/jump'),
+      import('codemirror-graphql/esm/mode'),
+    ]);
```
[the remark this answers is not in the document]
acao: which is problematic and why?

acao:
> so you must have put a hell of a lot of time into it!

you don't even wanna know 😆 
but despite all the late nights, itll save me lots of sleep in the long run

packages/graphiql/src/utility/onHasCompletion.js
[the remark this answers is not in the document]
benjie: Because otherwise the error function isn’t bound to console, and this can cause trouble under certain circumstances (particularly when console has been replaced).

packages/graphiql/src/components/QueryEditor.js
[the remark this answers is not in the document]
benjie: Because the `componentDidUpdate` method may be called and exit before the `componentDidMount` method completes, which makes the flow of the code very unintuitive and opens us up to potential race conditions (if not now, then in the future). See #994 for a better explanation/fix.

packages/graphiql/src/utility/onHasCompletion.js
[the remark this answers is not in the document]
benjie: (Also errors are unhandled.)

acao:
Closed for #998 and it's successive cousins

PR #974, opened 2019-10-07
Fix some problems with Readme in graphiql

olexandryermilov:


acao:
oops, thanks for that

Issue #969, opened 2019-09-30
Results panel can be made inaccessible through resizing


FidelVe:
Playing around in graphiql today I hid the result frame to make some more room for the text editor window and then couldn't find a way to bring it back, this video can show what am talking about:

https://gfycat.com/playfulalarmedbird

As you can see, if you resize the result "frame" minimizing it all the way to the right, if you try to drag it back you will resize the entire browser window, there's also no button or other intuitive way of bringing the frame back, clicking on "run" should automatically bring the results frame back, or maybe adding a button to the top bar.

the only way to bring it back is by double-clicking inside the text editor frame, which is not very intuitive.

benjie:
I've renamed this issue because I don't agree with using a button to solve the issue, but I agree it's an issue.

I think the solution is to ensure that the resize bar for the result frame is always visible, perhaps by using `flex-basis` or `min-width` if appropriate, or perhaps by tweaking the JS layout logic.

I think achieving it should be _fairly_ straight forward so I've added the "good first issue" label; if someone wanted to take a look at doing this, let us know and we can give some pointers :+1:

FidelVe:
@benjie If it's no problem with you, I can do it. I haven't looked at the source code for graphiql yet, so any pointers would be appreciated.

benjie:
Of course!

The size of the editor panel is controlled by the `editorFlex` value on state:

https://github.com/graphql/graphiql/blob/2da818b1f4e7adb1f1e519e314a428e846dc257e/packages/graphiql/src/components/GraphiQL.js#L913

Which is then used as the `flex-grow` value via shorthand `flex` syntax for the CSS of the editor:

https://github.com/graphql/graphiql/blob/2da818b1f4e7adb1f1e519e314a428e846dc257e/packages/graphiql/src/components/GraphiQL.js#L307

Since this is all flexbox; I think we can fix it by making the `resultWrap` have a non-zero `flex-basis` here:

https://github.com/graphql/graphiql/blob/73a3f9016124dfae6b327ca3a0c9946eaf12638e/packages/graphiql/css/app.css#L117-L123

e.g.

![Screenshot_20191007_182806](https://user-images.githubusercontent.com/129910/66334195-73157980-e930-11e9-9813-7b2b751345ca.png)

FidelVe:
Well you basically did all the work there 😅. Either way I could make a PR, implement the changes and check if nothing else break. I'll have it done by the end of the day.

benjie:
[...]

PR #980, opened 2019-10-10
fix(css): added minimum width for result panel in GraphiQL

FidelVe:
# fix(css): Added minimum width for result panel in GraphiQl (#969), solves #969

As discussed in #969 this is a simple fix, just added a `flex-basis: 1em` value to `graphiql/packages/graphiql/css/app.css`.

Since the goal is to keep the panel from disappearing when resized all the way to the left, any value small enough to allow the panel to still be visible and "draggable" by the mouse should work.

I think is a simple solution, nothing in the UI breaks by using the value, but I'm open to listen to other ways or recommendations to implement this change.

### DIFF
--- a/packages/graphiql/css/app.css
+++ b/packages/graphiql/css/app.css
@@ -119,6 +119,7 @@
   display: flex;
   flex-direction: column;
   flex: 1;
+  flex-basis: 1em;
   position: relative;
 }
 


FidelVe:
Your welcome!, it was a pleasure for me to at least collaborate with this one line :smile:

This is a great open source project, I'll be looking out for more ways to collaborate in the future.

PR #982, opened 2019-10-18
Fix broken GraphiQL minified build

### LINKED ISSUE #976: Broken minified build for graphiql 0.16.0
Hi All,

The jsDelivr CDN distribution of GraphiQL seems broken at the moment.

I followed the official example at https://github.com/graphql/graphiql/blob/master/packages/examples/graphiql-cdn/index.html, and then changed the CDN URL from `./graphiql.js` to https://cdn.jsdelivr.net/npm/graphiql/graphiql.min.js, which results in `Minified React error #130`.

Changing it to the full version ( https://cdn.jsdelivr.net/npm/graphiql/graphiql.js) seems to fix it.

Not sure if this issue belongs here, just something to note.
Thanks!

imolorhe:
The current source used for minifying graphiql in the build script is wrong. Therefore the build is broken. After debugging for the reason while the build was failing, I figured out that in order to use the correct source for the build, we need all the packages used in the build to be transpiled to their es5 equivalents for the build process to work. Main reason being that uglify doesn't support es6 syntax and **requires** es5 to minify the file. Given that the tsconfig specified the target as `esnext`, the builds were transpiled to their es6/es7 equivalents, which is required by graphiql as a dependency, and so uglify breaks.

I'm not sure why we decided to set the tsconfig to `esnext`, but this question should be answered before merging this PR.

This closes #976

### DIFF
--- a/packages/graphiql/package.json
+++ b/packages/graphiql/package.json
@@ -35,6 +35,9 @@
       "git add"
     ]
   },
+  "browserify": {
+    "transform": [["babelify", { "presets": ["@babel/preset-env"] }]]
+  },
   "dependencies": {
     "codemirror": "^5.47.0",
     "codemirror-graphql": "^0.11.1",
--- a/packages/graphiql/package.json
+++ b/packages/graphiql/package.json
@@ -36,7 +36,7 @@
     ]
   },
   "browserify": {
-    "transform": [["babelify", { "presets": ["@babel/preset-env"] }]]
+    "transform": ["babelify"]
   },
   "dependencies": {
     "codemirror": "^5.47.0",
--- a/packages/graphiql/resources/build.sh
+++ b/packages/graphiql/resources/build.sh
@@ -13,7 +13,7 @@ babel src --ignore __tests__ --out-dir dist/
 echo "Bundling graphiql.js..."
 browserify -g browserify-shim -s GraphiQL dist/index.js > graphiql.js
 echo "Bundling graphiql.min.js..."
-browserify -g browserify-shim -t uglifyify -s GraphiQL graphiql.min.js | uglifyjs -c > graphiql.min.js
+browserify -g browserify-shim -t uglifyify -s GraphiQL graphiql.js | uglifyjs -c > graphiql.min.js
 echo "Bundling graphiql.css..."
 postcss --no-map --use autoprefixer -d dist/ css/*.css
 cat dist/*.css > graphiql.css
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "composite": false,
-    "target": "esnext",
+    "target": "es5",
     "module": "commonjs",
     "moduleResolution": "node",
     "esModuleInterop": true,
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "composite": false,
-    "target": "es5",
+    "target": "esnext",
     "module": "commonjs",
     "moduleResolution": "node",
     "esModuleInterop": true,


acao:
unfortunately we need target esnext or es2018 for node purposes. babel
should handle this in the first step of the graphiql build

imolorhe:
Babel doesn't because it only transpiles from the src to dist and didn't create the bundle. The bundle is created by browserify after babel transpiles

imolorhe:
Babelify should help in this case

imolorhe:
Adding the browserify transform config for babelify works without for bundling graphiql without changing the target from esnext. This however means that all the browserify operations (currently bundling to graphiql.js and minifying to graphiql.min.js) would apply the babelify transform as well, but this works at the moment.

imolorhe:
The reason why this is the solution I am proposing is this: https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed

Since the problematic package is gls-interface (which is in the node_modules), the browserify transform wouldn't transpile those by default.

acao:
this is why i tried for weeks to switch to rollup earlier in the year. i
think webpack is a good solution. thanks for looking into this. not sure
how this regressed, we must have broken the build scripts

imolorhe:
Yeah, not a big fan of browserify myself 😅 

If there are no other concerns, perhaps you can take a look at this PR locally to verify the implementation. For me, it seems to work fine.

packages/graphiql/package.json
```diff
@@ -35,6 +35,9 @@
       "git add"
     ]
   },
+  "browserify": {
+    "transform": [["babelify", { "presets": ["@babel/preset-env"] }]]
```
acao: we should already have a special downstream babel.config.js with our browserslist for babelify to use, beside this file
we should pass -t babelify on the CLI like we were doing before. not sure where that went, or how we went from min.js -> min.js like that. quite some regressions!

acao: oh wait, we never used babelify
https://github.com/graphql/graphiql/blob/ac274d3fb81d4e580920357f0563223c87246d23/resources/build.sh

imolorhe: Yeah the babelify package was installed but never used. I tried passing it in the CLI but it wasn't picking up the presets, which was when I saw that FAQ on browserify and applying transforms to dependencies. Following the FAQ solution (adding the config to package.json) is what worked for me. Not sure how else to do it.

acao: whatever works best, we just need it to pick up our babel.config.js for graphiql, since its also used for tests and for the development watch. That config is different from what is being passed here.

imolorhe: I agree, they should be sourced from the same file. Removed the babel presets specified in the babelify config. Babelify should use the babel.config.js for graphiql now.

acao:
thanks so much @imolorhe !! going to cut another release of graphiql here now

jbblanchet:
@acao Was this fix supposed to be part or version 0.16? Seems to still be broken for me. Only the non-minified version is working.

acao:
hmmm! well I have a PR (the huge bundle one) with passing tests that
definitely fixes that, should be ready to merge unless there are new
upstream conflicts from today.

the new netlify build in that PR directly uses the min.js, so you can
preview it for yourself. terribly sorry for the trouble, i had manually
confirmed it last release i’d thought. is this from npm?

jbblanchet:
From jsdelivr. Switching over from cdnjs since it's been dead for a month. Using `https://cdn.jsdelivr.net/npm/graphiql@0.16.0/graphiql.js` works as expected, using `https://cdn.jsdelivr.net/npm/graphiql@0.16.0/graphiql.min.js` gives an error `ReferenceError: GraphiQL is not defined`

acao:
@jbblanchet i'll see if I can't kick out a quick fix for this tonight, but otherwise the existing PR #973 will address this. terribly sorry.  the netlify preview build and e2e tests as of this new PR #973 will use the minified bundle as well. I had set them to use graphiql.js for debugging :( my bad!

jbblanchet:
@acao No worries, it happens. Thanks for taking care of it, appreciate the hard work.

PR #985, opened 2019-10-21
chore: first pass at cypress e2e tests

acao:
Some of these files were stubbed out by cypress, thinking of maybe adding some more tests as well, wanted to see how it runs in travis :D


packages/graphiql/cypress/integration/init.spec.js
```diff
@@ -0,0 +1,53 @@
...
+        "__typename": "Test"
+      }
+    }
+  }
+}
+
+
+describe('GraphiQL On Initialization', function() {
+  it('Renders without error', function() {
+    cy.visit(`/?query=${testQuery}`)
```
benjie: Every cypress test should take the form (setup) > action > assertion. [Ref](https://docs.cypress.io/guides/getting-started/writing-your-first-test.html#Write-a-real-test)

There are circumstances where this `visit` will throw errors, but I don't think it's really a sufficient assertion - we should add something like checking the query pane is visible or similar that requires that the page is of the shape we accept as opposed to an error page or a blank page.

packages/graphiql/cypress/integration/init.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+   it('Executes a query that has the expected resut', function() {
+    cy.get('.execute-button').click()
```
benjie: Using classnames is brittle; it's fine if we're guaranteeing that these classes will always be there and will always be used for this purpose, but I prefer the `data-cy` approach [detailed here](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements).

I tend to write a helper for this; in `cypress/support/commands.js` add:

```js
Cypress.Commands.add("getCy", cyName => {
  return cy.get(`[data-cy=${cyName}]`);
});
```

Then this line would become:
```suggestion
    cy.getCy('execute-button').click()
```

relating to something like `<button data-cy="execute-button">`. Note: we should commit to some kind of naming convention for these cypress-addressable names. We should also consider stripping `data-cy` from the production bundle.

packages/graphiql/cypress/integration/init.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    cy.window().then((w) => {
+      cy.expect(JSON.parse(w.g.resultComponent.viewer.getValue())).to.deep.equal(mockSuccess)
+    })
+  })
+
+  it('Toggles doc pane off', function() {
+    cy.get('.docExplorerWrap button.docExplorerHide').should('be.visible').click({ multiple: true })
```
benjie: The 'be.visible' assertion is [unnecessary](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability).

```suggestion
    cy.get('.docExplorerWrap button.docExplorerHide').click({ multiple: true })
```

What does the `multiple: true` do? I've never used it. Docs say "Serially click multiple elements" but that doesn't seem relevant here?

packages/graphiql/cypress/integration/init.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    cy.get('.doc-explorer').should('not.exist')
+  })
+
+  it('Toggles doc pane', function() {
+    cy.get('.docExplorerShow').should('be.visible').click()
```
benjie:
```suggestion
    cy.get('.docExplorerShow').click()
```

packages/graphiql/cypress/integration/init.spec.js
acao: I wanted to add to the markup in another pass. I know full well that using classes for e2e tests are problematic. My goal for this PR is to only add cypress harness, to not change the actual components/etc yet. We will do this in another pass. I think the better solution is to use ID's, which is something we should be doing already. But we will be getting rid of most of our css classes soon anyways with styled components.

acao:
@benjchristensen it does not, but the pre-push hook took care of this. There are a bunch of linting/prettier fixes that need to be added in a seperate PR it looks like.

PR #987, opened 2019-10-22
🐛 Don't use initial query on every re-render

jimmybergman:
Without this change then when using query to set the initial displayed query and then editing the query in the editor the state was reverted to the initial query on every render (since it reset to nextProps.query even if the prop was not changed).

With the change it behaves as expected (the prop is only used for the intial value and then as long as it isn't changed during the component lifetime, we'll not overwrite the state).

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.js
+++ b/packages/graphiql/src/components/GraphiQL.js
@@ -182,10 +182,10 @@ export class GraphiQL extends React.Component {
     if (nextProps.schema !== undefined) {
       nextSchema = nextProps.schema;
     }
-    if (nextProps.query !== undefined) {
+    if (nextProps.query !== undefined && this.props.query !== nextProps.query) {
       nextQuery = nextProps.query;
     }
-    if (nextProps.variables !== undefined) {
+    if (nextProps.variables !== undefined && this.props.variables !== nextProps.variables) {
       nextVariables = nextProps.variables;
     }
     if (nextProps.operationName !== undefined) {


benjie:
@jimmybergman thanks for this pull request. It needs some internal discussion between the maintainers before we can advance it; please bear with us.

---

@acao To my mind `query` and `onEditQuery` should pair to make GraphiQL a controlled component (similarly `variables` and `onEditVariables`), with `defaultQuery` allowing you to set the initial query but maintaining internal state management. This would make the component behave the same as a standard React `<input>` tag with `value`, `onChange` and `defaultValue`.

However, this has two issues:

1. this isn't what the README says: "`query`: an optional GraphQL string to use as the initial displayed query, if `undefined` is provided, the stored query or `defaultQuery` will be used."
2. if we make the component controlled due to the presence of `query` then there's no way to force GraphiQL to display particular initial text but manage its own state.

I think we've been painted into a corner by the existing code alas. We should probably address this in the next major version.

If we accept this pull request as-is then using `query` and `onEditQuery` will not make the component a fully controlled component as an `onEditQuery` that rejects the change and maintains `query` at its previous value will be ignored by this code. This could cause all sorts of confusing synchronisation issues.

I'm wondering if we should make it so the component state is treated as external only if both `query` and `onEditQuery` are provided; otherwise we treat the state as managed internally.

Maybe we need separate `query` / `defaultQuery` and `fallbackQuery` settings in the next major version?

What do you think?

jimmybergman:
@benjie Thanks for your quick response, for what it's worth I picked what I thought corresponded to the README - but agree that defaultQuery (with other semantics than the currently implemented one) is probably more reasonable naming for setting the initial query but having state internal.

acao:
oh this is great @jimmybergman thank you! @benjie all good points, and i like your suggestion of making the editor an intelligently controlled component. we do need to support both controlled and uncontrolled modalities, and what you suggest doesnt seem would be a breaking change?

benjie:
I think the cleanest solution would be to have a breaking change and have `query` make the component controlled; and raise a warning in the console or elsewhere if you provide `query` but not `onEditQuery`.

The need for `fallbackQuery` is an interesting one: it's effectively a "low priority" defaultQuery. Where `defaultQuery` would be displayed as the initial text always (assuming `query` was not provided), `fallbackQuery` would only be displayed as the initial text if both no `query` was provided _and_ there's no history or similar to restore.

We also need to decide what history restoration means for `onEditQuery` - should we call it on initial render with the initial text if `query` is undefined/null?

acao:
ah, should have brought this into 1.0.0

jimmybergman:
From my perspective I consider this done now that https://github.com/graphql/graphiql/pull/1663 is merged, IMO the breaking changes could be handled as a separate PR if needed.

PR #994, opened 2019-11-07
Refactor async lifecycle methods in #973

benjie:
The promises returned from `async componentDidMount` / `async componentWillUpdate` are ignored by React, which could lead to broken state and/or race conditions. To solve this, I've factored them out into separate functions so the flow is clearer, and stored the error state should one arise.

Note: the initial promise chain started in `componentDidMount` is stored to `this._ready`. We then add `.then` calls to this in `componentDidUpdate` which means all the `componentDidUpdate` events will play back in chronological order once the editor is loaded. This should hopefully avoid any unexpected changes (although it may induce a little stutter when the code finally loads).

### DIFF
--- a/packages/graphiql/src/components/QueryEditor.js
+++ b/packages/graphiql/src/components/QueryEditor.js
@@ -47,13 +47,78 @@ export class QueryEditor extends React.Component {
   constructor(props) {
     super();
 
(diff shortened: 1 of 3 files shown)


acao:
we can close this one for now

PR #998, opened 2019-11-15
Step 1: Webpack resolutions for #882, webpack example, faster/more stable e2e tests

acao:
A more minimal first step
- Resolves the webpack resolution issues for #882 via a file utility that resolves in a way that makes webpack happy, with explicit extensions.
- Adds webpack example as well, to prove the config is working end to end
- Fixes netlify demo (doesnt use min.js yet, fixed in the next round)
- Uses a more thin e2e server with prebuilt graphiql and schema for e2e tests
- Clean build, to prepare for build improvements in the next PR


packages/examples/graphiql-webpack/webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+        use: ['style-loader', 'css-loader'],
+      },
+      {
+        test: /\.svg$/,
+        use: [{ loader: 'svg-inline-loader' }],
+      },
+    ],
+  },
+  resolve: {
+    extensions: ['.js', '.json', '.jsx', '.css', '.mjs'],
```
benjie: OPTIONAL: Having justification for this search order in a comment would be handy.

packages/graphql-language-service-utils/package.json
```diff
@@ -21,6 +21,10 @@
   ],
   "main": "dist/index.js",
   "types": "src/index.ts",
+  "engines": {
+    "node": ">= 9.7.3"
```
benjie: This could be a breaking change? We can drop Node 8 support from 1st Jan 2020, but might be too early right now?

packages/graphql-language-service-utils/src/file.ts
```diff
@@ -0,0 +1,81 @@
...
+  switch (extension) {
+    case 'js': {
+      return require.resolve(pathWithoutExtension + '.js');
+    }
+    case 'json': {
+      return require.resolve(pathWithoutExtension + '.json');
+    }
+    default: {
+      if (fs.existsSync(filePath + `.js`)) {
+        return require.resolve(filePath + '.js');
```
benjie: I'm surprised we can bundle with these `fs.existsSync` calls?

benjie: Would it be better to use a `try` block and skip the `fs` calls?

I don't like the idea of `fs.existsSync` being called after startup either; can we memoize this function?

packages/graphql-language-service-utils/package.json
acao: It will be a breaking change. Hmm... maybe we shouldn't use dynamic imports yet, and just stick with require until then? I wanted to deprecate it now, its so close!!

packages/examples/graphiql-webpack/webpack.config.js
acao: yes, we will address this when we are bundling graphiql's min bundle with webpack in the next few rounds

packages/graphql-language-service-utils/src/file.ts
acao: good idea

acao: do you mean like this?

```ts
export function resolveFile(filePath: string) {
  const extension = getFileExtension(filePath);
  const pathWithoutExtension = getPathWithoutExtension(filePath, extension);
  switch (extension) {
    case 'js': {
      return require.resolve(pathWithoutExtension + '.js');
    }
    case 'json': {
      return require.resolve(pathWithoutExtension + '.json');
    }
    default: {
      try {
        require.resolve(filePath + '.js')
      }
      catch (err) {
        try {
          require.resolve(filePath + '.json')
        }
        catch (err) {
          if (extension) {
            throw Error(
              `cannot require.resolve() module with extension '${extension}'`,
            );
          }
        }
      }
    }
  }
}
```

packages/graphql-language-service-utils/package.json
benjie: If it's breaking anyway, might as well go whole hog?

packages/examples/graphiql-webpack/babel.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
acao: this in duplicate to be exemplary

packages/examples/graphiql-webpack/src/index.html.ejs
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
```
acao: this in quasi-duplicate to be exemplary

packages/graphql-language-service-utils/package.json
acao: I'd just thought I'd leave it to `engineStrict: false` by default, so that it will warn in the case that its a browser build with node 8 which should still be supported.

.npmignore
```diff
@@ -14,3 +14,6 @@ coverage
 resources
 src
 packages
+packages/graphiql/*.html
+cypress.json
+babel.config.js
```
benjie: Is npmignore required? We should be using the `"files"` entry in `package.json` files which should mean we can completely delete `.npmignore`.

package.json
```diff
@@ -19,28 +19,32 @@
     }
   },
   "scripts": {
-    "build": "tsc --build && lerna run build --scope codemirror-graphql --scope graphiql --scope graphql-language-service --scope graphql-language-service-server",
+    "build": "yarn run build-clean && yarn build-ts && yarn build-js",
+    "build-js": "lerna run build --scope codemirror-graphql --scope graphql-language-service-server --scope graphiql graphql-language-service",
+    "build-ts": "tsc --build",
+    "build-clean": "tsc --build --clean && rm -rf 'packages/**/src/{dist,esm,bundle}' && lerna run build-clean --parallel",
```
benjie: I'm not comfortable with `dist` / `esm` / `bundle` being inside the `src` folder long-term. We should try and refactor that in a follow-up PR.

package.json
```diff
@@ -19,28 +19,32 @@
...
     "test": "yarn run lint && yarn run check && yarn run build && yarn run testonly && yarn run e2e",
     "testonly": "jest && lerna run test --scope codemirror-graphql",
     "e2e": "cd packages/graphiql && yarn e2e",
-    "cypress:open": "cd packages/graphiql && yarn e2e:server & cypress open --project ./packages/graphiql",
+    "cypress-open": "cd packages/graphiql && yarn e2e:server & cypress open --project ./packages/graphiql",
     "t": "yarn run testonly",
-    "lint": "eslint --ext=ts,js,jsx,tsx . || (printf '\\033[33mTry: \\033[7m yarn run lint -- --fix \\033[0m\\n' && exit 1)",
-    "lint:fix": "eslint --ext=ts,js,jsx,tsx . --fix",
-    "lint:check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    "lint": "eslint --ext=ts,js,jsx,tsx . || (printf '\\033[33mTry: \\033[7m yarn format \\033[0m\\n' && exit 1)",
```
benjie: There doesn't appear to be a `yarn format` command... did you mean:
```suggestion
    "lint": "eslint --ext=ts,js,jsx,tsx . || (printf '\\033[33mTry: \\033[7m yarn lint-fix \\033[0m\\n' && exit 1)",
```

benjie: Can we use `||` and `&&` and still be Windows safe? (Genuine question, I do not know the answer.)

packages/codemirror-graphql/package.json
```diff
@@ -14,26 +14,30 @@
...
     "mocha_tdd": "--full-trace --watch --require resources/mochaBootload src/**/*/__tests__/*-test.js"
   },
   "scripts": {
     "lint": "eslint src",
     "check": "flow check",
-    "build": "babel src --root-mode upward --ignore __tests__ --out-dir .",
-    "build-js": "node ../../resources/buildJs.js",
+    "build": "yarn build-clean && yarn build-js && yarn build-flow .",
+    "build-js": "babel src --root-mode upward --ignore src/__tests__ --out-dir .",
+    "build-clean": "rimraf {mode,hint,info,jump,lint}.{js,js.flow} && rimraf esm results utils variables coverage __tests__",
```
benjie: I _think_ we should wrap this in quotes to avoid shell expansion (so rimraf does the expansion instead, in a more cross-platform way):

```suggestion
    "build-clean": "rimraf '{mode,hint,info,jump,lint}.{js,js.flow}' && rimraf esm results utils variables coverage __tests__",
```

packages/graphiql/cypress/support/commands.js
```diff
@@ -12,6 +12,33 @@
...
+  if (variables) {
+    url += `&variables=${encodeURIComponent(variables)}`;
+  }
+  return cy.visit(url);
+});
+
+Cypress.Commands.add('assertQueryResult', (op, mockSuccess) => {
+  cy.visitWithOp(op);
+  cy.clickExecuteQuery();
+  cy.wait(500);
```
benjie: If this is necessary, the test is probably flaky. We should fix this in a follow up PR; for now:

```suggestion

  // TODO: we should re-write this test so the wait isn't necessary
  cy.wait(500);

```

Though I guess it also comes up under the Cypress ESLint checks? :grin:

packages/graphiql/package.json
```diff
@@ -21,14 +21,23 @@
...
+    "build": "yarn build-clean && yarn build-js && yarn build-flow && yarn build-demo",
+    "build-clean": "rimraf 'graphiql.{css,*min.js,js}' esm dist",
     "build-js": "bash ./resources/build.sh",
     "build-flow": "node ../../resources/buildFlow.js",
+    "build-demo": "cp resources/index.html .",
     "check": "flow check",
     "dev": "babel-node test/server.js",
     "e2e": "yarn e2e:server && cypress run && kill -9 `cat test/pid`",
-    "e2e:server": "PORT=8080 babel-node test/server.js & wait-on http://localhost:8080",
+    "e2e:server": "PORT=8080 node test/e2e-server & wait-on http://localhost:8080",
```
benjie: A semicolon would make it more obvious that the `&` was intentional and not a mistyped `&&`.

```suggestion
    "e2e:server": "PORT=8080 node test/e2e-server &; wait-on http://localhost:8080",
```

I don't like that this process is never cleaned up; but I guess it's just a convenience method for us, so :man_shrugging:

packages/graphiql/cypress/support/commands.js
acao: the only way to get around this is probably to mock the schema entirely client side. @tgriesser has some utilities, as well as some patterns I've found to work as well, so it's something we will have to address more properly down the road :(

packages/graphiql/package.json
acao: I like

.npmignore
acao: ah good point, i fix this partially in the next PR, we can get rid of it totally in another PR later

packages/graphiql/package.json
acao: In step three, I switch to another suggested approach by Cypress, a node binary package called `start-server-and-test`

PR #999, opened 2019-11-17
Step 2: Publish ECMAScript modules

acao:
- publish ecmascript modules
- for typescript, add tons of extra tsconfig so we can support esm modules (is there any way to do this besides those rollup abstractions like microbundle/etc?)
- share babel config that honors ESM variable
- copy modules to codemirror-graphql in an expected relative fashion (using a new shared script for rewriting file paths and extensions)


packages/graphql-language-service-types/package.json
```diff
@@ -14,21 +14,23 @@
...
+  "module": "esm/index.js",
+  "typings": "src/index.ts",
   "scripts": {
     "build": "yarn run build-ts && yarn run build-flow",
     "build-ts": "tsc",
     "build-flow": "node ../../resources/buildFlow.js",
     "prepublish": "node ../../resources/prepublish.js"
   },
   "peerDependencies": {
-    "graphql": "^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0"
```
acao: oops, need to fix this. was going change this to ranges in another PR, definately not the move

acao: ok resolved, better peer version ranges now and all the way up the PR chain from here :D

packages/graphql-language-service-types/package.json
```diff
@@ -14,21 +14,23 @@
...
+    "dist",
+    "esm"
   ],
   "keywords": [
     "graphql"
   ],
   "main": "dist/index.js",
-  "typings": "src/index",
+  "module": "esm/index.js",
+  "typings": "src/index.ts",
```
benjie: We're not bundling `src`.

Not sure if this should be `dist` or `esm`, but you put `esm` everywhere else so I'm going to trust that.

```suggestion
  "typings": "esm/index.d.ts",
```

packages/graphql-language-service-types/package.json
```diff
@@ -14,21 +14,23 @@
...
   "scripts": {
     "build": "yarn run build-ts && yarn run build-flow",
     "build-ts": "tsc",
     "build-flow": "node ../../resources/buildFlow.js",
     "prepublish": "node ../../resources/prepublish.js"
   },
   "peerDependencies": {
-    "graphql": "^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0"
+    "graphql": ">= ^0.10.0 <= ^14.0.0"
```
benjie: Invalid semver

benjie: Also this differs from the previous definitions in that it has extra spaces. Best to try and keep the specs as similar to each other as possible so that lock files don't get too verbose.

packages/graphql-language-service-utils/package.json
```diff
@@ -33,10 +34,13 @@
     "prepublish": "node ../../resources/prepublish.js"
   },
   "peerDependencies": {
-    "graphql": "^0.12.0 || ^0.13.0 || ^14.0.0"
+    "graphql": ">=^0.12.0 <=^14.0.0"
```
benjie: Invalid semver

packages/graphql-language-service/package.json
```diff
@@ -31,7 +31,7 @@
     "prepublish": "node ../../resources/prepublish.js"
   },
   "peerDependencies": {
-    "graphql": "^0.12.0 || ^0.13.0 || ^14.0.0"
+    "graphql": ">=^0.12.0 <=^14.0.0"
```
benjie: Invalid semver

resources/renameFileExtensions.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const tempDirectory = fs.mkdtempSync(`${tmpDir}${path.sep}`);
+
+copy(src, tempDirectory, (error, files) => {
+  if (error) {
+    throw error;
+  }
+  files.forEach(file => {
+    if (file.dest) {
```
benjie: When would there not be `file.dest`? Should it throw in that case?

packages/codemirror-graphql/package.json
```diff
@@ -14,29 +14,23 @@
...
   "license": "MIT",
   "files": [
-    "dist",
-    "hint.js",
-    "info.js",
-    "jump.js",
-    "lint.js",
-    "mode.js",
+    "esm",
+    "*.js",
```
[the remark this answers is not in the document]
acao: hmmm... i think i can make this more selective.
i wanted to do something like:

`"{hint,info,jump,lint,mode}.{js,js.flow,esm.js}"`

but i couldnt get it working

this ended up being the missing piece from the last attempt, to properly publish all the esm.js files
i'll have to see if that works.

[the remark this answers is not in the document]
acao: this ended up being the secret to the tree shaking, essentially! haha

Screen Shot 2019-11-19 at 12 22 05 PM

we need to publish everything that is greyed out, essentially

packages/examples/graphiql-cdn/index.html
```diff
@@ -36,13 +36,10 @@
...
-    <script src="/graphiql.js" type="application/javascript"></script>
-    <script src="/renderExample.js" type="application/javascript"></script>
-
   </head>
   <body>
     <div id="graphiql">Loading...</div>
-    <script>
-    </script>
+      <script src="/graphiql.js" type="application/javascript"></script>
+      <script src="/renderExample.js" type="application/javascript"></script>
```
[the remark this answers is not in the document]
acao: I'm going to do a PR after this where i fix autoformatting and do a prettier/eslint fix run/etc

packages/graphiql/.babelrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./babel.config')
```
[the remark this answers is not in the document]
acao: it was there as a shim for the cypress test coverage plugin which was looking for .babelrc.js, i can remove it since i removed coverage reporting

packages/codemirror-graphql/package.json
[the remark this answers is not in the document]
benjie: Lets just list them explicitly. It's not too many files, and they're not likely to change that often. Still significantly less risky than a blocklist approach IMO.

packages/graphiql/package.json
```diff
@@ -36,7 +35,7 @@
     "build-demo": "cp resources/index.html .",
     "check": "flow check",
     "dev": "babel-node test/server.js",
-    "e2e": "yarn e2e:server && cypress run && kill -9 `cat test/pid`",
+    "e2e": "yarn e2e:server && cypress run",
```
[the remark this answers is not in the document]
acao: it was causing weird issues in CI, where the last of the two would kill the first one's tests mid-run.
i also thought it would be nice to relieve the e2e server from tracking the PID using the filesystem

packages/graphiql/.babelrc.js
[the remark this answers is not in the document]
benjie: Ah okay. A comment on weird workaround you have to do would be excellent next time; those are always entertaining to read :popcorn:

packages/graphiql/package.json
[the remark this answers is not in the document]
benjie: :man_shrugging: Whatever. Just raised it in case it was a mistake.

packages/codemirror-graphql/package.json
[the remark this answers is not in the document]
benjie: Lets just list them explicitly. It's not too many files, and they're not likely to change that often. Still significantly less risky than a blocklist approach IMO.

resources/renameFileExtensions.js
acao: good question. this script is rewritten in step 3 significantly anyways, otherwise it doesnt work on netlify. the `fs.mkTempSync()` fails miserably on netlify but works fine on travis. Should i just backport those improvements to this PR then?

packages/codemirror-graphql/package.json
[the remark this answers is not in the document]
acao: will do!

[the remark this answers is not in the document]
acao: it's probably best to be explicit in this case anyways. these are not minor details! if it's the secret sauce, this is the best way.

packages/graphiql/.babelrc.js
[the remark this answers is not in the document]
acao: hoooo boy, the haunting tales of node.js tooling experiments! that could be a whole podcast. haunted dot js 😆

.eslintrc.js
```diff
@@ -292,9 +294,19 @@ module.exports = {
...
-  plugins: ['babel', 'import', 'flowtype', 'prefer-object-spread', 'cypress'],
+  plugins: defaultPlugins,
 
   overrides: [
+    // Cypress plugin, global, etc only for cypress directory
+    // https://github.com/cypress-io/eslint-plugin-cypress
+    // cypress clashes with jest expect()
+    {
+      files: ['**/cypress/**'],
+      plugins: [...defaultPlugins, 'cypress'],
```
benjie: I think the existing plugins are still used; so the following should suffice:

```suggestion
      plugins: ['cypress'],
```


(I couldn't find this confirmed in the docs, but I just tested it on a separate project by introducing an error that needed the `graphql` plugin to work, and it still worked after this change.)

acao: that would be simpler... it makes sense, since for example presets also provide plugins that are loaded

packages/graphql-language-service-types/package.json
benjie: Yeah that sounds like a build-order error to me. We can't set `typings` to be `src/*` because it won't be in the final bundle.

resources/renameFileExtensions.js
benjie: That, or just throw an error in the `else` block here.

.eslintrc.js
benjie: With this we could go back to not using variables I think too :raised_hands:

packages/graphql-language-service-types/package.json
acao: yep, the cross-references aren't completely set up

packages/graphiql/package.json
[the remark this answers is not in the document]
acao: I'm pretty torn about it TBH. And there's probably a better solution that I'm not picking up on

packages/graphql-language-service-types/package.json
acao: now that I added the one missing downstream project reference resolution it works properly at `dist/index.d.ts` :D

.eslintrc.js
acao: yep thats what I did! it was nice to elevate the plugins list though, it's hard to see what plugins are enabled at first glance

packages/graphql-language-service-types/package.json
acao: yes, it might be helpful to use this shorter semver once we need to update it, say for 15.0.0 releases which may come soon for the alpha.1

resources/renameFileExtensions.js
acao: I think I will do both. because this PR would fail on netlify right now.

packages/codemirror-graphql/package.json
```diff
@@ -14,30 +14,36 @@
...
-    "*.flow",
+    "mode.js.flow",
     "utils",
     "variables",
-    "results",
-    "README.md",
-    "LICENSE"
+    "results"
   ],
-  "main": "dist",
```
benjie: Do we not need a `main` for this package?

packages/graphiql/graphiql.min.css
```diff
@@ -0,0 +1,10 @@
...
+.graphiql-container .history-contents{font-family:Consolas,Inconsolata,Droid Sans Mono,Monaco,monospace;margin:0;padding:0}.graphiql-container .history-contents li{align-items:center;display:flex;font-size:12px;overflow:hidden;text-overflow:ellipsis;white-space:nowrap;margin:0;padding:8px;border-bottom:1px solid #e0e0e0}.graphiql-container .history-contents li button:not(.history-label){display:none;margin-left:10px}.graphiql-container .history-contents li:focus-within button:not(.history-label),.graphiql-container .history-contents li:hover button:not(.history-label){display:inline-block}.graphiql-container .history-contents button,.graphiql-container .history-contents input{padding:0;background:0;border:0;font-size:inherit;font-family:inherit;line-height:14px;color:inherit}.graphiql-container .history-contents input{flex-grow:1}.graphiql-container .history-contents input::placeholder{color:inherit}.graphiql-container .history-contents button{cursor:pointer;text-align:left}.graphiql-container .history-contents .history-label{flex-grow:1;overflow:hidden;text-overflow:ellipsis}
```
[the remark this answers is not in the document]
benjie: Is this intended to be included in git?

packages/graphiql/package.json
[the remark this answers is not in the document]
benjie: Meh.

packages/graphql-language-service-interface/tsconfig.esm.tsbuildinfo
```diff
@@ -0,0 +1,3856 @@
+{
+  "program": {
+    "fileInfos": {
```
benjie: This shouldn't be under version control.

packages/graphql-language-service-parser/tsconfig.esm.tsbuildinfo
```diff
@@ -0,0 +1,3503 @@
+{
+  "program": {
+    "fileInfos": {
```
benjie: This shouldn't be under version control.

packages/graphql-language-service-types/tsconfig.esm.tsbuildinfo
```diff
@@ -0,0 +1,3426 @@
+{
```
benjie: This shouldn't be under version control.

packages/graphql-language-service-utils/tsconfig.esm.tsbuildinfo
```diff
@@ -0,0 +1,3554 @@
+{
+  "program": {
```
benjie: This shouldn't be under version control.

packages/codemirror-graphql/package.json
acao: actually we dont, it has no index! with this there, certain bundlers throw warnings and/or maybe I saw it break a webpack build once?

babel.config.js
```diff
@@ -1,7 +1,13 @@
-
+// for ESM don't transpile modules
```
orta: This might bite us later down the line with TypeScript?

acao: needed for tree shaking, and besides we still have a cjs build at the same path

acao:
@orta thank you! gonna resolve these conflicts from dependabot merges

PR #1001, opened 2019-11-18
Step 3: convert browserify build to webpack, fixes #976

### LINKED ISSUE #976: Broken minified build for graphiql 0.16.0

acao:
the vast amount the added lines to this PR are via the yarn.lock file, FYI! this PR is mostly a 120 line webpack config file and some tooling and configuration changes.

- [x] browserify min and graphiql bundle replaced with webpack 3, retaining window.GraphiQL interface pattern
- [x] windows compatible tooling: replace bash scripts for babel, browserify, uglify, postcss/autoprefixer, etc 'build' turned into `build` 20s and `build-bundle`. This came with replacing browserify, as we no longer needed a bash script. Also added crossenv, rimraf, etc
- [x] use new webpack mincss plugin, and generate a nanocss build at `graphiql.min.css` too! (note: temporary til styled components)
- [x] update `yarn dev` server to mount webpack dev server with test GraphQL server, instead of the exact same pattern using browserify
- [x] fix some issues with cypress + jest battling over the linting of `expect()` in cypress specs that was failing lint
- [x] HUGE reduction in minified build size. I'm not kidding.

Screen Shot 2019-11-18 at 1 36 27 PM
(parsed size, from webpack output)

Screen Shot 2019-11-18 at 2 27 58 PM

here is proof (on the far right) that the file utility from the first PR in this series is working:
Screen Shot 2019-11-18 at 6 35 47 PM


packages/graphiql/package.json
```diff
@@ -28,28 +29,28 @@
...
-    "build": "yarn build-clean && yarn build-js && yarn build-flow && yarn build-demo",
-    "build-clean": "rimraf 'graphiql.{css,*min.js,js}' esm dist",
-    "build-js": "bash ./resources/build.sh",
+    "analyze-bundle": "NODE_ENV=production CDN=1 ANALYZE=1 webpack -c resources/webpack.config.js -p",
+    "build": "yarn build-clean && yarn build-js && yarn build-flow",
+    "build-clean": "rimraf esm dist",
+    "build-js": "babel src --ignore __tests__ --out-dir dist/ && ESM=1 babel src --ignore __tests__ --out-dir esm/",
+    "build-bundles": "yarn build-bundles-clean && yarn build-bundles-dev && yarn build-bundles-min",
+    "build-bundles-dev": "NODE_ENV=development CDN=1 webpack --config resources/webpack.config.js -d",
+    "build-bundles-min": "NODE_ENV=production CDN=1 webpack --config resources/webpack.config.js -p",
```
kevinsimper: webpack --bail 

😅from past unfortunate experiences

acao: ah, thank you!

acao:
oof, in order to claim windows interoperability, we still need to add cross-env wherever environment variables are being set!

acao:
this is failing because of https://github.com/markdown-it/markdown-it/issues/616, which breaks the webpack build

packages/graphiql/test/dev-server.js
```diff
@@ -0,0 +1,60 @@
+/**
```
acao: for example, if we don't provide a new webpack dev server with the same feature parity as the browserify dev server, then this PR isn't complete

packages/graphiql/cypress/plugins/index.js
```diff
@@ -11,9 +11,9 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-module.exports = (on) => {
-  on('task', require('@cypress/code-coverage/task'));
-  on('file:preprocessor', require('@cypress/code-coverage/use-babelrc'));
+module.exports = (_on, _config) => {
+  // on('task', require('@cypress/code-coverage/task'));
+  // on('file:preprocessor', require('@cypress/code-coverage/use-babelrc'));
```
acao: the `@cypress/code-coverage` stuff was broken for us, I can move this into another PR

packages/graphiql/package.json
```diff
@@ -64,27 +67,32 @@
     "react-dom": "^15.6.0 || ^16.0.0"
   },
   "devDependencies": {
```
acao: replacing the entire bundling toolchain. thus the huge yarn lock diff

package.json
```diff
@@ -24,10 +24,10 @@
     "build-ts": "tsc --build",
     "build-clean": "tsc --build --clean && rm -rf 'packages/**/src/{dist,esm,bundle}' && lerna run build-clean --parallel",
     "build-demo": "lerna run build-demo",
-    "test": "yarn run lint && yarn run check && yarn run build && yarn run testonly && yarn run e2e",
+    "test": "yarn run lint && yarn run check && yarn run build && yarn run testonly && lerna run build-bundles && yarn run e2e",
     "testonly": "jest && lerna run test --scope codemirror-graphql",
     "e2e": "cd packages/graphiql && yarn e2e",
-    "cypress-open": "cd packages/graphiql && yarn e2e:server & cypress open --project ./packages/graphiql",
+    "cypress-open": "cd packages/graphiql && yarn cypress-open",
```
benjie: Cleaner to use the [yarn workspace](https://yarnpkg.com/lang/en/docs/cli/workspace/) command:

```suggestion
    "cypress-open": "yarn workspace graphiql cypress-open",
```

(Untested.)

packages/graphiql/cypress/support/commands.js
```diff
@@ -32,10 +32,10 @@ Cypress.Commands.add('visitWithOp', ({ query, variables }) => {
...
   cy.visitWithOp(op);
   cy.clickExecuteQuery();
-  cy.wait(500);
+  cy.wait(200);
   cy.window().then(w => {
     cy.expect(JSON.parse(w.g.resultComponent.viewer.getValue())).to.deep.equal(
       mockSuccess,
     );
-  });
+  })
```
benjie: Lint issue?

babel.config.js
```diff
@@ -1,12 +1,26 @@
...
+
+if (process.env.ESM) {
+  envConfig.modules = false
+}
+
+if (process.env.CDN) {
+  envConfig.modules = 'umd'
+  envConfig.targets = null
+  envConfig.ignoreBrowserslistConfig = false
+}
```
benjie: I'd rather this used an `if/elseif/else` chain and instead of overwriting the initial properties of `envConfig` it set variables that were then used to create the `envConfig` object afterwards, but :man_shrugging:

acao: I'm so averse to else it's problematic, haha! will take note for further iterations

package.json
acao: works a charm!

packages/codemirror-graphql/.npmignore
```diff
@@ -0,0 +1,4 @@
+babel.config.js
+.gitignore
+resources
```
caseywebdev: By not publishing `resources` here, the reference at https://github.com/graphql/graphiql/blob/98be6f5df0f2e01054df51377949d401c5e6ad01/packages/graphiql/src/components/ResultViewer.js#L45

is no longer valid. Either this line or that line should be removed.

acao: I’m confused? that import is from codemirror-graphql, which still exports that module. are you having issues? this PR was merged

acao: if what you are saying is true, then the results viewer would fail to highlight jaon here in the current build of master:

https://graphiql-test.netlify.com

caseywebdev: Sorry, I got `resources` and `results` confused so this isn't the direct issue. Regardless, there's still an issue with the latest `codemirror-graphql` package because the `results` folder, referenced in the line above, does not exist.

```sh
~ npm install codemirror-graphql@0.11.5
~ ls node_modules/codemirror-graphql/results
ls: node_modules/codemirror-graphql/results: No such file or directory
```

This broken link is causing my build to fail, so I've resorted back to `graphiql@0.17.2` and manually pinning `codemirror-graphql@0.11.4` (which does publish this folder) for now.

acao:
oh yikes! thanks for reporting this. ill have to see what happened and fix
it immediately

packages/codemirror-graphql/.npmignore
acao: @caseywebdev try now with `0.11.6`

acao: https://unpkg.com/browse/codemirror-graphql@0.11.6/ my bad! we should be doing pre-releases for all these build changes.

caseywebdev: Working again 👍🏻 Thanks for the quick turnaround!

PR #1007, opened 2019-11-23
lightouse audit workflow via gh actions

acao:
introduce a lighthouse audit workflow for each deploy preview, as per this [documentation](https://github.com/jakejarvis/lighthouse-action#pull-request-audits-with-netlify-deploy-preview)


acao:
oops, seems theres a race condition with the gh-actions to build the demo site somehow? definately something to come back to!

PR #1012, opened 2019-11-27
fix typescript typings path

chandu0101:


acao:
thanks, we have an open PR that fixes this, but we'll take it, haha!

PR #1025, opened 2019-11-28
fix: hints vertical scroll

zouxuoz:
As I understand, `graphiql` have copied CodeMirror CSS files and sorted properties by alphabet, but it causes the problem: `overflow` override `overflow-y` and disable vertical scroll for the hints.

CodeMirror source file for the hints: https://github.com/codemirror/CodeMirror/blob/master/addon/hint/show-hint.css#L21

graphiql-hints-scroll

### DIFF
--- a/packages/graphiql/css/show-hint.css
+++ b/packages/graphiql/css/show-hint.css
@@ -7,8 +7,8 @@
   margin-left: -6px;
   margin: 0;
   max-height: 14.5em;
-  overflow-y: auto;
   overflow: hidden;
+  overflow-y: auto;
   padding: 0;
   position: absolute;
   z-index: 10;


acao:
awesome, thanks!

https://deploy-preview-1025--graphiql-test.netlify.com/

going to confirm the bug is fixed in the deploy preview ^^^ here shortly

acao:
now, we don’t officially support mobiles, but im seeing this on my phone:
D4C7CD9F-698C-442A-A648-12F8E0CF1CBB
ill try with some desktop browsers later though!

zouxuoz:
@acao it's just quick fix according to latest CodeMirror styles, it shouldn't break anything :) As I know, CodeMirror doesn't provide good mobile behavior

acao:
@zouxuoz it looks great upon comparison. thanks for the fix!

Issue #909, opened 2019-07-08
Plugin: vim mode


nikitavoloboev:
It would be nice if you can run some simplified version of vim inside the editor. Specifically I wish I could do `di{` to delete inside `{}` brackets. But some other vim features would be very nice to have too.

Perhaps [neovim](https://neovim.io) could be embedded to serve this purpose.

thomasheyenbrock:
@acao what else would we need to do here in your eyes? Or can we close this issue?

dimaMachina:
In graphqil 5 with Monaco editor, you can use community plugins. Monaco Vim: https://github.com/brijeshb42/monaco-vim.

PR #1031, opened 2019-11-29
Add VIM keybindings support

semanser:
Issue: https://github.com/graphql/graphiql/issues/909

This PR brings VIM key bindings support to the Graphiql.

Features:
- [x] Vim keybindings for Query editor
- [x] Vim keybindings for Variables editor
- [x] Toggle keybinding: `Cmd-\` or `Ctrl-\`
- [x] Local storage persistance
- [x] Passing like an option to the`GraphiQL` component


packages/graphiql/src/components/QueryEditor.js
```diff
@@ -69,6 +75,7 @@ export class QueryEditor extends React.Component {
     require('codemirror/addon/dialog/dialog');
     require('codemirror/addon/lint/lint');
     require('codemirror/keymap/sublime');
+    require('codemirror/keymap/vim');
```
acao: possibly do inline requires based on the chosen option? These files can be pretty large. pretty soon we will also replace all these requires with `import().then()`

acao:
awesome, thank you! can you write a few tests for your PR so codecov is happy

packages/graphiql/src/components/QueryEditor.js
semanser: For the toggling between vim/sublime modes, we need to have these files to be preloaded.

acao:
ugh i hate to do this because you did a great job of implementing this, and i asked you to add tests.

actually, now that i think about it @semanser, this is going to have to wait for plugins in a couple months. the reason being is that we are also switching to monaco, and so we may end up defaulting to what is more of a vscode style keybinding. i already have a prototype somewhere of how we would provide plugins for vim/emacs modes (using `monaco-vim` and `monaco-emacs`) to demonstrate my proposal for the new graphiql plugin api.

for your purposes currently, is there a way you can just require `codemirror-vim` in a way that overwrites those bindings in a more minimal fashion? sounds messy but maybe it would work? just trying to keep things minimal so that we can lean on plugins for a multitude of reasons

semanser:
> ugh i hate to do this because you did a great job of implementing this, and i asked you to add tests.
That's totally fine, don't worry :)

> for your purposes currently, is there a way you can just require codemirror-vim in a way that overwrites those bindings in a more minimal fashion?

What do you mean with that? Right now, I'm using the default codemirror vim keybindings (and I guess they should work fine). I guess for now, this implementation can work pretty well, and can be replaced later with plugins.

acao:
i guess what im trying to say is that this implementation is beyond the scope of graphiql currently, before we introduce plugins. so we will introduce vim support to graphiql natively as an option _only_ via plugins, as it were now. despite how simple this PR is, it will add some size to the bundle that we are trying to reduce, so the idea is that almost all feature requests are being parlayed for plugins. this one would be an example of that.

semanser:
@acao so, I guess adding more tests has no sense, because this PR is going to be rejected?

acao:
yeah, thats what i was trying to say earlier, my bad if i was unclear. @benjie would probably say the same thing at this point. i just got so excited about vim support and how nice of a PR it was, and then i started reviewing it like we could merge it now, my mistake!

acao:
https://github.com/acao/graphiql-1-poc-monaco/blob/master/src/plugins/graphiql-plugin-mode/ModePlugin.tsx here's some half-complete psuedocode as an example, I'll be introducing some more complete proposals soon.

acao:
@benjie what do you think? maybe if we made the vim dependency an async bundle which we can do with webpack or CDN versions? that would still a context driven require `import()` or `require.ensure()`. i've implemented it before and it's unnoticable, especially since the syntax mode is loaded so early on. and besides, with the minified bundle, all the async chunks are concatenated into a single file anyways. so it would still drive up the size of the minified bundle a bit. but we just had a huge reduction so. I'm torn

acao:
[vscode vim](https://marketplace.visualstudio.com/items?itemName=vscodevim.vim): 1.3M installs
[vscode sublime text keymap](https://marketplace.visualstudio.com/items?itemName=ms-vscode.sublime-keybindings): 830K installs

i mean, they slightly apples to oranges in that the vim one isn't just a keybinding extension, but interesting to compare

benjie:
Sorry for the delay; I've been incredibly busy and am currently spending a couple days catching up with 200+ GitHub notifications!

Some evaluation criteria:

:+1: Passes this criteria

:-1: Fails this criteria

:grey_question: Yet to be evaluated

1. :grey_question:   **Bundle size**: does it add significant weight to the compressed minified bundle? If less than ~2% I'm not too worried, but above that we should think more carefully
2. :+1: **Necessity**: is it possible and desirable to _instead_ allow people to implement this in their own GraphiQL by passing in a few props? For example, I can imagine passing in custom keybindings for toggling, passing in a callback that requires the relevant package, etc. This might require fewer changes but add more power (for example other codemirror plugins could be loaded?)
3. :+1:  **Scope**: does this expand the scope of GraphiQL beyond where we want it? I think given this is planned to be supported in the new GraphiQL it's not really expanding the scope
4. :grey_question: **Timing/backwards-compatibility**: one issue of introducing this now and then later (soon) replacing it with a Monaco equivalent is that the plugins will likely work in different ways and support different parts of Vim, which may result in "X Vim action worked in version Y but doesn't work in version Z" which is something we couldn't address without working on the monaco plugin itself.

Q1 can be answered by comparing the size of the bundled result.

Q2 I'm on the fence with. I'm erring towards :+1: just adding it directly rather than adding a too-flexible props system that we're going to replace with a plugin system in the coming months.

Q3 I'm also on the fence with; since we're planning to replace codemirror it seems unwise to expand its scope currently, but also lots of people want Vim mode and we could add this easily now so people get to enjoy it months earlier

Q4 Anyone have any idea of compatibility between the monaco and codemirror vim plugins?

packages/graphiql/src/components/QueryEditor.js
acao: @semanser i think we're ready to accept this earlier than we though, but we really need to use `import()`, as we are switching all of this to use `import()` very soon. it will be pre-loaded for the minified bundle, and loaded async for more modern setups. it will work great i promise. its a great use case for contextually loading modules asynchronously, with sublime as the one thats loaded by default by the entrypoint bundle. then the VIM module makes no impact on the entrypoint bundle size, though it still impacts the min

brentluna:
Has there been any update on this. I feel so lost without vim

acao:
@brentluna this is high priority for me, so it should come early in 2.0.x as an extension hopefully. this PR will no longer apply of course because it will be using monaco. we would most likely end up using beonaco-vim](https://github.com/brijeshb42/monaco-vim).

I wish we could add it to 1.0, but if we keep doing that, the 2.0 rewrite will never be completed

acao:
I've had a change of heart regarding this PR. Mostly because we'll be adding a new feature with some complexity to 1.0.0 anyways, so we might as well add this as well.

Just note that there will be significant breaking changes in DX and keybindings between 1.0.x and 2.0.x, when we switch to `monaco-graphql`

if someone wants to get this PR updated and ready to merge, that would be grand.

i wish there was a way to get dynamic imports working for the codemirror modes. ideally, someone should only fetch the vim support bundle once they toggle the option, but if we can't have that, it's no big worry. 

bundle size is not really as much of an issue with DX I'm learning, because we aren't as concerned with first hit, and are increasingly looking to bundle assets statically for offline usage as well.

saly98:
> العدد: [# 909](https://github.com/graphql/graphiql/issues/909)
> 
> يجلب هذا العلاقات العامة دعم ارتباطات مفاتيح VIM إلى Graphiql.
> 
> سمات:
> 
> * [x]  روابط مفاتيح فيم لمحرر الاستعلام
> * [x]  رابطات المفاتيح فيم لمحرر المتغيرات
> * [x] تبديل ربط المفاتيح: `Cmd-\`أو`Ctrl-\`
> * [x]  استمرار التخزين المحلي
> * [x] تمرير مثل خيار `GraphiQL`للمكون

semanser:add-vim-keybindings

thomasheyenbrock:
Closing this in favor of #2501 which has been merged a couple of weeks ago and is released since `1.9.9`

acao:
@thomasheyenbrock we should keep this branch for reference. We only implemented 1/4 of this!

thomasheyenbrock:
@acao can you elaborate? The only thing I see here that wasn't added in your PR is a shortkey to switch keymaps and some tests.

acao:
@thomasheyenbrock I totally apologize! i thought this person's PR had more to it than that, and didn't realise that sublime's mode came with an edit mode. long term memory failed me.

a complete re-creation as you said is now possible with the new settings dialog so we can mark this as closed and not needed for reference for current efforts!

PR #1036, opened 2019-12-01
chore(ci): remove cypress coverage plugins for now

acao:
we added a plugin for codecov coverage but it wasn't quite working yet. going to make another pass at that soon. for now lets keep it simple!


acao:
ended up doing this in step 3 anyways

PR #1044, opened 2019-12-03
fix: csp headers violation @gracenoah 

acao:
graphql/codemirror-graphql#246

From the original PR

> This PR fixes a [Content Security Policy (CSP) violation](https://www.html5rocks.com/en/tutorials/security/content-security-policy/)  for those who are using CSP headers and do not have `script-src` of `unsafe-eval` enabled.

> The way the code is written now, `onHover` is `undefined` in [this call to `setTimeout`](https://github.com/graphql/codemirror-graphql/blob/master/src/utils/info-addon.js#L52). Passing `undefined` to `setTimeout` triggers a CSP violation for `'unsafe-eval'` because `setTimeout` thinks you are passing a string, which is [not allowed](https://developer.chrome.com/extensions/contentSecurityPolicy#JSEval).

> The CSP Violation: `Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: ...`

> This PR moves the usage of `onHover` to after the function is defined, so that `setTimeout` is not called with `undefined` and there are no CSP violations.

### DIFF
--- a/packages/codemirror-graphql/src/utils/info-addon.js
+++ b/packages/codemirror-graphql/src/utils/info-addon.js
@@ -50,9 +50,6 @@ function onMouseOver(cm, e) {
 
   const box = target.getBoundingClientRect();
 
-  const hoverTime = getHoverTime(cm);
-  state.hoverTimeout = setTimeout(onHover, hoverTime);
-
   const onMouseMove = function() {
     clearTimeout(state.hoverTimeout);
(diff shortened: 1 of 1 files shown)


acao:
this is long awaited and well reviewed, so i will trust it to go out after confirming the netlify build

PR #1052, opened 2019-12-03
Fix: #1045 css resolution bug, broken npm package for 0.17.1 release

### LINKED ISSUE #1045: v0.17.1 Can't resolve '../css/doc-explorer.css' in graphiql/dist/components
Topic

acao:
Fixes #1045

As noted in the bug, this breaks NPM package imports using bundlers, etc currently. Minified bundle is fine of course.

This not only fixes the issue, but also changes the CI test run to run the webpack demo to ensure that we don't have issues like this again, and that we always have a working webpack import


packages/graphiql/src/index.js
```diff
@@ -15,5 +15,8 @@ import './css/lint.css';
 import './css/loading.css';
 import './css/show-hint.css';
 
+import './css/doc-explorer.css';
+import './css/history.css';
+
```
acao: move these up to the root to avoid resolution issues

packages/examples/graphiql-webpack/webpack.config.js
```diff
@@ -8,9 +8,9 @@ module.exports = {
         'react-hot-loader/patch', // activate HMR for React
         'webpack-dev-server/client?http://localhost:8080', // bundle the client for webpack-dev-server and connect to the provided endpoint
         'webpack/hot/only-dev-server', // bundle the client for hot reloading, only- means to only hot reload for successful updates
-        './index.js', // the entry point of our app
+        './index.jsx', // the entry point of our app
```
acao: the webpack demo was broken over use of jsx/js entrypoint, oops!

package.json
```diff
@@ -24,7 +24,8 @@
     "build-ts": "tsc --build",
     "build-clean": "tsc --build --clean && rm -rf 'packages/**/src/{dist,esm,bundle}' && lerna run build-clean --parallel",
     "build-demo": "lerna run build-demo",
-    "test": "yarn run lint && yarn run check && yarn run build && yarn run testonly && lerna run build-bundles && yarn run e2e",
+    "build-bundles": "lerna run build-bundles",
+    "test": "yarn run lint && yarn run check && yarn run build && yarn run testonly && yarn build-bundles && yarn run e2e && yarn build-demo",
```
acao: add `yarn build-demo` to our CI script.

this will ensure that packages consuming graphiql as a webpack module will not run into build errors at least.

packages/graphiql/package.json
```diff
@@ -31,9 +31,10 @@
   "scripts": {
     "webpack": "webpack --config resources/webpack.config.js",
     "analyze-bundle": "cross-env NODE_ENV=production CDN=1 ANALYZE=1 yarn webpack -p",
-    "build": "yarn build-clean && yarn build-js && yarn build-flow",
+    "build": "yarn build-clean && yarn build-cjs && yarn build-esm && yarn build-flow",
     "build-clean": "rimraf esm dist",
-    "build-js": "babel src --ignore __tests__ --out-dir dist/ && cross-env ESM=1 babel src --ignore __tests__ --out-dir esm/",
+    "build-cjs": "babel src --ignore __tests__ --out-dir dist/ && mkdirp dist/css && copy 'src/css/*.css' dist/css",
```
acao: here i have to copy files to the dist/esm outputs so modules will resolve. this doesn't feel right, however it looks like it leads to a working webpack build, which is what our users want most besides CDN build

yarn.lock
```diff
@@ -6951,10 +6951,10 @@ graceful-fs@^4.1.4:
   integrity sha1-TK+tdrxi8C+gObL5Tpo906ORpyU=
 
 "graphiql@file:packages/graphiql":
-  version "0.17.0"
+  version "0.17.1"
   dependencies:
     codemirror "^5.47.0"
-    codemirror-graphql "^0.11.3"
+    codemirror-graphql "^0.11.4"
```
acao: huh, looks like we hadn't bumped our internal resolution on the last release somehow, i think this means the last release is missing a relatively obscure bugfix that doesnt impact graphiql in its suggested implementation, but may fix a bug for implementations that use that header

packages/graphiql/src/components/DocExplorer.js
```diff
@@ -5,8 +5,6 @@
  *  LICENSE file in the root directory of this source tree.
  */
 
-import '../css/doc-explorer.css';
-
```
acao: see: below

packages/graphiql/src/components/QueryHistory.js
```diff
@@ -11,8 +11,6 @@ import PropTypes from 'prop-types';
 import QueryStore from '../utility/QueryStore';
 import HistoryQuery from './HistoryQuery';
 
-import '../css/history.css';
```
acao: see: below

packages/graphiql/package.json
benjie: Can we move the css back out of the `src` folder like it was previously? This'd mean the path to it would be the same from within `src/`, `dist/` and `esm/` folders.

packages/examples/graphiql-webpack/webpack.config.js
benjie: I'm sorry; this looked like such a deliberate change I didn't vet it well enough!

packages/graphiql/package.json
acao: i wish we could but webpack doesn't make that easy now. I will look into this. either way, its a temporary solution until we replace source css files (will still have bundled graphiql.css of course, just no source .css files) so i dont want to fuss with it too much

acao:
@benjie  thank you so much for the really rapid response! pre-releases or bust

PR #1087, opened 2019-12-11
fix: prefer displayName over type equality for children overrides

### LINKED ISSUE #671: react hot loader incompatible with children customizations
This is what react-hot-loader has to say about this issue:

> We know this problem, but could not do anything to solve it. At least today.
> The only proper way to write your code is to compare the same things:
> 
> node.type === <Component />.type

Could you update GraphiQL in this way so that I can customize the toolbar, etc.?

joshmadewell:
This should fix #671.

The core issue here was that when `GraphiQL.xxx` got wrapped by some other component, `.type` would on the child would be of that wrapper causing the the type equality check to fail. 

By checking for `displayName` first, consumers can set `displayName` on their wrapped components and still be able to override graphiql components.

### DIFF
--- a/packages/graphiql/src/components/__tests__/GraphiQL.spec.js
+++ b/packages/graphiql/src/components/__tests__/GraphiQL.spec.js
@@ -143,4 +143,146 @@ describe('GraphiQL', () => {
     );
     expect(graphiQL.state().variableEditorOpen).toEqual(false);
   });
+
+  describe('children overrides', () => {
+    const MyFunctionalComponent = () => { return null; }
+    const wrap = (component) => () => <div>{component}</div>;
+
+    it('properly ignores fragments', () => {
+      const myFragment = (
+        <React.Fragment>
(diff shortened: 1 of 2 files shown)


packages/graphiql/src/components/__tests__/GraphiQL.spec.js
```diff
@@ -143,4 +143,130 @@ describe('GraphiQL', () => {
     );
     expect(graphiQL.state().variableEditorOpen).toEqual(false);
   });
+
```
acao: awesome work! 1.5%+ coverage, nice

let's add a class component test too?

joshmadewell: 👍 amended my previous commit to include a test.

acao: wow that was fast! LGTM now, thanks for shoring this up!

PR #1094, opened 2019-12-14
Fix: changed to using "echo" instead of printf in yarn tests

### LINKED ISSUE #1033: Windows: Running yarn tests reports trying to use "printf"
Running the https://github.com/graphql/graphiql/blob/master/CONTRIBUTING.md#getting-started steps on windows 10 (command line running as admin) I get an error showing at the end of the run about `printf` which doesn't exist on windows. Not sure whats throwing it as the `yarn tests` command chains a bunch of stuff together so I'll run them individual and see if I can narrow it down.

PeteDuncanson:
Fixes #1033


PeteDuncanson:
Will need testing on a Mac to make sure I've not broken that OS while getting to work on Windows.

To test just run "yarn lint" and when complete ensure no error messages shown in the last 3 lines of output

acao:
I was about to remove this in a PR for linting/formatting, but this looks fine for now!
`echo` should work fine in all OSes

acao:
ope. yeah doesn't work on osx. not sure what `echo.` command is? i'm used to `echo` in linux/osx, and i thought this was some way of passing an argument, but on osx with zsh it says: `/bin/sh: echo.: command not found`

package.json
```diff
@@ -23,7 +23,7 @@
     "e2e": "yarn workspace graphiql e2e",
     "cypress-open": "yarn workspace graphiql cypress-open",
     "t": "yarn run testonly",
-    "lint": "eslint --ext=ts,js,jsx,tsx . || (printf '\\033[33mTry: \\033[7m yarn format \\033[0m\\n' && exit 1)",
+    "lint": "eslint --ext=ts,js,jsx,tsx . || ( echo. && echo. && echo Try:  yarn format && echo. && exit 1)",
```
acao: I would suggest that we just remove the or statement here.

PeteDuncanson:
Its just a "smart" way of doing a newline in MSDos. I'll remove them and just write it out without the newlines and see if that gets through

PeteDuncanson:
Ha ha, I found that after I posted this. If you are removing it anyway I'll just leave it be. A fun afternoon messing around regardless.

PR #1095, opened 2019-12-14
Fix: added .editorconfig and .gitattributes files to control line endings across environments

### LINKED ISSUE #1032: Windows: Running tests reports errors with line endings
Following the instructions on https://github.com/graphql/graphiql/blob/master/CONTRIBUTING.md#getting-started results in line ending issue being reported when running the test (yarn test):

image

Seems by default git for windows will checkout files and convert them to CRLF line endings rather than LF as everyone else is using.

Initially I can fix this by running `npm run lint-fix` but I had to find that and google about it to know I could do it, better to not have to have Windows devs run that each time.

Seems a new rule is needed to be added to a currently non-existant file (.gitattributes) to overrule this setting when first checked out on windows machines.

https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems

Alternative is to add something to the docs.

PeteDuncanson:
Fixes #1032 

To test (one windows):

- checkout
- run "yarn"
- run "yarn lint"

Without these mods you get 100's of errors about line endings. With them though all that noise is gone.

Additionally the editorconfig file will force supporting editors to use the LF line ending so reduce any annoying errors when windows folks are developing.


.editorconfig
```diff
@@ -0,0 +1,9 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
```
acao: it appears there are still is still the no newline character in this file :/

PeteDuncanson: no newline character?

acao: oh weird, its gone now!?

acao: Screen Shot 2019-12-14 at 7 59 17 PM

when i go to the full view its there, but inline in the PR conversation tab its not 🤷‍♂

acao: @PeteDuncanson do you want to remove this character and then we can merge it? or should I re-create the PR with the fix, and give you credit?

benjie:
Thanks for this PR, @PeteDuncanson; if you can edit the two files you've added to make sure that they themselves have a terminal newline (as you've configured) then we can merge it :+1:

PeteDuncanson:
Oh the irony of having to fix new lines on a config file that is meant to fix new lines.

Thanks for the heads up and for letting me keep the PR :)

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto eol=lf
```
benjie: I must admit I don't full know what this gitattributes will do; I worry that it'll apply eol transforms against binary files or some other unexpected thing. Other projects seem to list out the whole list of file extensions manually, e.g. Drupal: https://github.com/alexkaratarakis/gitattributes/blob/master/Drupal.gitattributes - I assume there must be a reason for this?

Something like this:

```suggestion
* text=auto

*.js text eol=lf
*.jsx text eol=lf
*.ts text eol=lf
*.tsx text eol=lf
*.json text eol=lf
*.md text eol=lf
*.html text eol=lf
*.css text eol=lf
*.sh text eol=lf
*.yml text eol=lf
```

PeteDuncanson: Its all a good excuse to learn new stuff!

The drupal one is quite verbose ain't it? I read through the doc it links to (https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html) and there are some bits that jumped out which seem to valid the initial logic:

"Git recognizes files encoded in ASCII or one of its supersets (e.g. UTF-8, ISO-8859-1, …) as text files. Files encoded in certain other encodings (e.g. UTF-16) are interpreted as binary and consequently built-in Git text processing tools (e.g. git diff) as well as most Git web front ends do not visualize the contents of these files by default."

However now you've written the above might as well roll with that as its explicit. Does mean we might need to add to it in the future if new file type is added but cross that bridge when we come to it I guess.

acao:
looks good, thanks @PeteDuncanson and @benjie for the review! this looks good to go

.gitattributes
benjie: :+1: Thanks Pete

PR #1098, opened 2019-12-15
fix: componentWillReceiveProps warning messages

### LINKED ISSUE #970: Deprecation warning since React 16.9.0
Hi 😊 

Since the React 16.9.0 version, the following deprecation warning about React legacy lifecycle methods appears:

```
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: GraphiQL, QueryHistory
```

ryan-m-walker:
Fixes #970

* prefixed `componentWillReceiveProps` in `GraphiQL` with `UNSAFE_`
* removed use of `componentWillReceiveProps` in `QueryHistory` and added a class method to be called by parent component instead

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.js
+++ b/packages/graphiql/src/components/GraphiQL.js
@@ -151,6 +151,9 @@ export class GraphiQL extends React.Component {
     // Ensure only the last executed editor query is rendered.
     this._editorQueryID = 0;
 
+    // Ref to QueryHistory component
+    this._queryHistory = null;
+
     // Subscribe to the browser window closing, treating it as an unmount.
     if (typeof window === 'object') {
       window.addEventListener('beforeunload', () =>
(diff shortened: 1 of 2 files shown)


acao:
ah! so @ryan-m-walker you’ll want to run this against the local jest suite. yarn jest —watch from the root is what i prefer. looks like a schema change isn’t registering in the tests at least?

acao:
ok @ryan-m-walker lets pivot:

- deprecate react 15 peer support
- bump react in GrapiQL dev Deps
- switch enzyme adapters (soon to be replaced)
- react version detection error boundary with instructions for when someone tries to use React 15 (is this possible? at the very least a very clear console warning or something)
- and of course as originally intended, remove `componentWillReceiveProps` using React 16. if you decide to use 16 only features, make sure that you ideally use features that are still available to people still on earlier version of React 16 if possible.

acao:
@ryan-m-walker ah instead of error boundaries, lets do something like this:

```js
if (React.version.indexOf('16') !== 0) {
  console.warn("GraphiQL 0.18.0 and after is not compatible with React 15 or below.\n If you are using a CDN, follow this example:\n url to cdn example")
}
```

PR #1118, opened 2019-12-17
feat(graphql-config): add graphql config extensions

divyenduz:
This PR is a port of https://github.com/graphql/graphql-language-service/pull/240

It adds GraphQL config extension support to the language server.

This would enable the following scenarios (not-exhaustive):

1. Using environment variables (or variables in a custom format)
2. Supply customDirectives dynamically - can be tool specific like
Gatsby
3. Supply customValidationRules dynamically
4. Add endpoints, headers dynamically


packages/graphql-language-service-server/src/MessageProcessor.js
```diff
@@ -107,8 +113,15 @@ export class MessageProcessor {
       );
     }
 
-    this._graphQLCache = await getGraphQLCache(rootPath);
-    const config = getGraphQLConfig(rootPath);
+    this._graphQLCache = await getGraphQLCache(rootPath, this._extensions);
+    let config = getGraphQLConfig(rootPath);
+    if (this._extensions && this._extensions.length > 0) {
+      /* eslint-disable no-await-in-loop */
+      for (let i = 0; i < this._extensions.length; i++) {
```
acao: we have support for the latest language features now, so if you want to use for... await of here you can! I think Its node 9+.

acao:
this looks great @divyenduz, thanks!

acao:
oh dear, looks like codecov blocked it. but yeah this is right up our alley in terms of our intended route with graphql config. I’m looking forward to getting the LanguageService class and its own GraphQLCache working in the browser soon. We may end up figuring out how graphql config could be one config service, or you can bring your own, but that seems too complicated. different platforms want to have their own config formats, but maybe we can introduce something to graphql-config where they could specify different extensions and map their values to graphql-config values?

divyenduz:
> but maybe we can introduce something to graphql-config where they could specify different extensions and map their values to graphql-config values?

That is interesting indeed, are you suggesting that we move this feature from here into the GraphQL config? IIRC they had some extension work done in GraphQL Config back then, we chose to do this feature back then in LSP because we wanted dynamic config changes (my memory on this is faded, but we can surely re-think it)

packages/graphql-language-service-server/src/MessageProcessor.js
divyenduz: Bumped the loop to new syntax but kept `/* eslint-disable no-await-in-loop */` becuase lint still disables it.

Urigo:
@kamilkisiela has already introduced a start for extending GraphQL Config for different tools in the new alpha.

@divyenduz we would love to chat and see if that fits your use case or should we change anything!

packages/graphql-language-service-server/src/MessageProcessor.js
acao: @divyenduz interesting note, perhaps we should change this rule now that we have modern for...await of

acao:
@divyenduz just needs some more test coverage to get builds passing! let's get this merged before my refactor is complete?

divyenduz:
@acao When do you plan to merge your refactor? I am traveling over the weekend. I can maybe get to this on 26th? (or will see if I can find a moment from an airport or something 😅)

acao:
it’s almost finished, just two more tests, but then itll need plenty of time to review. that should be plenty of time! enjoy your holidays, thank you @divyenduz !

acao:
@divyenduz this one should be relatively easy to update? can you update docs as well?

divyenduz:
@acao Is this even valid after #1225 is merged? I need to build more context on that but it looks like they are doing away with the extensions feature, making this PR already outdated 😅

acao:
@divyenduz I think they are deprecating the endpoints extension specifically, however I think `extensions` will remain!

https://graphql-config.com/docs/introduction

acao:
@divyenduz an alpha release is now available :D

packages/graphql-language-service-server/src/startServer.ts
```diff
@@ -46,10 +47,13 @@ type Options = {
   // port for the LSP server to run on
   port?: number;
   // socket, streams, or node (ipc). if socket, port is required
-  method?: 'socket' | 'stream' | 'node';
```
divyenduz: This looks incorrect. I think this should be `'socket' | 'stream' | 'node'`

divyenduz:
@acao I wonder why CI tests that blocked me last time didn't appear this time? 🤔

packages/graphql-language-service-server/src/startServer.ts
divyenduz: Fixed via https://github.com/graphql/graphiql/pull/1118/commits/f6d8a564708bdc7682d2c76248df15cc1f5ae69f

acao:
@divyenduz that is strange, i wonder if there's an issue with github actions?

acao:
possibly because this PR pre-dates the switch to github actions?
but that wasnt an issue when we first switched?
was it a travis failure or gh actions failure before?

divyenduz:
@acao haha, weird indeed, I was talking about this comment: https://github.com/graphql/graphiql/pull/1118#issuecomment-566749106

Maybe GH actions don't run for older PRs. I can create a new PR and replace this one with that. Does that sounds good?

acao:
oh right! i have not restored the codecov behavior in github actions, my bad, forgot to do that! yikes.

divyenduz:
But still, even the tests did not run! 🤔 

I see that they are running for other PRs. I will just re-create this one to see if it runs.

divyenduz:
"Workflows in forked repositories don't run by default." via https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions

I couldn't find the docs to enable it though, looking for it. 

✅ Green in my fork: https://github.com/divyenduz/graphiql/runs/507245373?check_suite_focus=true

Haha! Let me know how to proceed, we can 

1. Merge this one and clean it up later! (by adding codecov + tests)
2. Clean it up now

I would prefer 2 for now, just to get this PR out of the door, it has been looooong living 😄

divyenduz:
Yes it works! Precioussssss! <3

packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -111,7 +117,14 @@ export class MessageProcessor {
...
 
-    this._graphQLCache = await getGraphQLCache(rootPath);
+    this._graphQLCache = await getGraphQLCache(rootPath, this._extensions);
+    let config = this._graphQLCache.getGraphQLConfig();
+    if (this._extensions && this._extensions.length > 0) {
+      /* eslint-disable no-await-in-loop */
+      for (const extension of this._extensions) {
+        config = await extension(config);
+      }
+    }
```
divyenduz: I don't think that this is required here since we do the same thing in the `getGraphQLCache` function.

divyenduz: Removed it, nice, this was a code smell with duplicate code. Moving config to cache solves this 👍

acao:
nice work! thanks @divyenduz

packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -54,8 +54,15 @@ const {
...
 export async function getGraphQLCache(
   configDir: Uri,
+  extensions?: Array<(config: GraphQLConfig) => GraphQLConfig>,
 ): Promise<GraphQLCacheInterface> {
-  const graphQLConfig = await loadConfig({ rootDir: configDir });
+  let graphQLConfig = await loadConfig({ rootDir: configDir });
+  if (extensions && extensions.length > 0) {
+    /* eslint-disable no-await-in-loop */
+    for (const extension of extensions) {
+      graphQLConfig = await extension(graphQLConfig);
```
acao: @divyenduz so this is intended to re-declare `graphQLConfig` on every invocation? I'm guessing the result of `await extension(graphQLConfig)` returns the previous config plus the extension?

acao:
maybe we can update docs, as well?

packages/graphql-language-service-server/src/GraphQLCache.ts
divyenduz: Yes, as of now, the signature of an extension is a function that takes a GraphQLConfig as input and returns a GraphQL config as its output with the modification (extension) applied.

`extensions?: Array<(config: GraphQLConfig) => GraphQLConfig>`
Which implies, `extension: (config: GraphQLConfig) => GraphQLConfig`

acao:
https://deploy-preview-1118--graphiql-test.netlify.com/lsp/modules/graphql_language_service_server.html#getgraphqlcache fyi, interesting how tsdoc handles this as `Array<function>`

divyenduz:
Added a test, figuring out where the docs live! 

Does your tsdoc comment mean that we get docs for free (after we fix the array function representation?) or do I need to do something else as well? :)

acao:
@divyenduz hood question! let’s put the docs for this in graphql-language-service for now. there will be a doc site soon.

the lsp api docs are nice but not instructive enough yet, or fully configured, and the readme’s come with

divyenduz:
@acao Added some docs to the README of `graphql-langauge-service` package. Please see if that makes sense :)

divyenduz:
@acao I did a split of CLI and Server docs as discussed. There is some duplication in the two readme files, please let me know if any further changes are needed :)

acao:
@divyenduz looking great! we are ready to merge then

PR #1128, opened 2019-12-19
Fix/enable cookies for cross-domain requests

qqvm:
This will enable user to make authenticated requests to another domain (if Cross-Origin Resource Sharing policy of this domain allows user to do so). Authentication should be done before in a separate tab.

### DIFF
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -82,6 +82,7 @@ function graphQLFetcher(graphQLParams) {
   const api = isDev ? '/graphql' : 'https://swapi.graph.cool/';
   return fetch(api, {
     method: 'post',
+    credentials: 'include',
     headers: {
       Accept: 'application/json',
       'Content-Type': 'application/json',
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -79,7 +79,7 @@ function graphQLFetcher(graphQLParams) {
   const isDev = !window.location.hostname.match(
     /(^|\.)netlify\.com$|(^|\.)graphql\.org$/,
   );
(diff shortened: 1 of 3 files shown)


packages/graphiql/resources/renderExample.js
```diff
@@ -82,6 +82,7 @@ function graphQLFetcher(graphQLParams) {
   const api = isDev ? '/graphql' : 'https://swapi.graph.cool/';
```
acao: hey thanks for adding these!

as a final measure to confirm it works end to end, can you set the fetch URL to be fixed to the netlify URL in the other examples here? then this will close the reference ecosystem, as you will see the demo using that schema url in the deploy preview for this PR, the final piece of the puzzle if you will :)

acao: we would run the tests against the remote schema, but it proved too flaky :/

qqvm:
Seems like graphql instance is not running on test build

acao:
@qqvm close! just set it to the same url that https://graphql.org/swapi-graphql uses.
the netlify function is hosted in that repository

acao:
ah my bad @qqvm , it looks like you have formatting fixes you need to do. just run `yarn pretty` locally and add those changes to a commit

qqvm:
> 
> 
> ah my bad @qqvm , it looks like you have formatting fixes you need to do. just run `yarn pretty` locally and add those changes to a commit

I am not a NodeJS developer. Just suggested a small improvement.

acao:
@qqvm thanks for this, I've created a new PR that makes a slightly different adjustment, as our lambda implementation requires a different `credentials` configuration.

acao:
@qqvm i changed this to `credentials : omit` by default in the example and that seems to work fine with consuming our own schema. it's easily customizeable 👍

PR #1138, opened 2019-12-19
convert LSP Server to Typescript, remove watchman

acao:
0 compile errors! now just tell me what i did wrong 😂

- [x] - convert `graphql-language-service-server`  to typescript
- [x] - convert `graphql-language-service` to typescript
- [x] - stumbled upon, fixed bug in `graphql-language-service-interface` with missing deprecated fields in autocomplete? https://github.com/graphql/graphiql/pull/1138/files#diff-564f41e59f16fb85571ebd58e501253eR106
- [x] - two `@ts-ignore` lines in server still
- [x] - still some tsc compile issues in CLI client lib?
- [x] - we need more tests for LSP server badly, still coming in between 30-40% it looks like


packages/graphql-language-service-types/src/index.ts
```diff
@@ -277,28 +282,26 @@ export type ObjectTypeInfo = {
...
-export type Diagnostic = {
-  range: Range;
-  severity?: number;
-  code?: number | string;
-  source?: string;
-  message: string;
-};
+export type Diagnostic = DiagnosticType;
 
-export type CompletionItem = {
```
acao: hmm... i think i fixed this in terms of how `isDeprecated` matches up with `deprecated` in the built-in vscode interface, but that should be mapped to our built in fields, and I think we should extend `vscode-languageserver-protocol` `CompletionItem` to handle both.

acao:
@divyenduz would you like to give this a review?

acao:
oops forgot to remove the `WIP` prefix last week

divyenduz:
@acao Sure, I will do that, is the public interface completely same? Apart from review, I can also integrate this with the GraphQL extension maybe to test.

acao:
@divyenduz indeed! that would be great. that's the goal, the interface should be completely the same, though slightly adopted to use more built in vscode types.

acao:
I plan on keeping the watcher in for this go around, though it's been a lot of trouble. Is the GraphQL extension using the watcher or a vscode extension client watcher?

divyenduz:
> Is the GraphQL extension using the watcher or a vscode extension client watcher?

It uses the watcher from this project but that has been troublesome, feel free to remove it, the extension can be mended later :)

acao:
I think I will keep it for this PR, to keep it as close to a refactor as possible. going to start creating PRs off of this one

acao:
@divyenduz what do you think? should we just remove it in this PR? it seems that removing it will add its own complexity, so possibly in a successive PR? that way I can make sure the tests here match up

also i noticed that i had actually broken `isDeprecated` in terms of the filtering logic. I changed it back to extend `CompletionItem` for `isDeprecated` in addition to the vscode expected `deprecated` flag. it will map to the official `deprecated` flag but still filters values. 

The possiblity is that via a configuration option, the deprecated fields could still appear in completion but with the official `deprecation` notice that will allow vscode to show the deprecation notices if needed. This would be relatively easy to implement as it would only require removing some `filter()` statements sprinkled throughout interface, as these manual filters are implemented upstream from codemirror or server implementations.

divyenduz:
@acao Keeping it in this PR, and getting rid of it in a successive PR sounds like a good plan.

acao:
@divyenduz exciting update

- it works great in `vscode-graphql` (your debug runner setup is very helpful!)
- I removed the watchman!

divyenduz:
@acao Exciting 🎉

I would like to use this and cut a release in the VSCode extension (after testing it a lot manually), as watchman was a big pain point there. But how do I grab a hold of the alpha? Is there an automatic alpha release? 

I know that [releasing master](https://github.com/graphql/graphiql/blob/master/RELEASING.md) is a manual process right now but was wondering if I can get a bite of this code in alpha or something?

acao:
I'm going to go ahead and cut an alpha release of everything. thats a good idea, of cutting an alpha automatically

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -202,19 +203,22 @@ function getSuggestionsForFieldNames(
 function getSuggestionsForInputValues(
   token: ContextToken,
   typeInfo: AllTypeInfo,
-): Array<CompletionItem> {
+): CompletionItem[] {
   const namedInputType = getNamedType(typeInfo.inputType as GraphQLType);
   if (namedInputType instanceof GraphQLEnumType) {
-    const values = namedInputType.getValues();
+    const values: GraphQLEnumValues[] = namedInputType.getValues();
```
yoshiakis: `GraphQLEnumValues[]` is not defined in this file, and without defining a type here, `values` is implied as `GraphQLEnumValue[]` from `namedInputType.getValues()`. I'm wondering if it's just a mistake or is there any reason to define a type for `values` here?

packages/graphql-language-service-types/src/index.ts
```diff
@@ -281,23 +281,11 @@ export type CustomValidationRule = (
...
-  kind?: number;
-  detail?: string;
-  sortText?: string;
-  documentation?: string | null | undefined;
-  // GraphQL Deprecation information
-  isDeprecated?: boolean | null | undefined;
-  deprecationReason?: string | null | undefined;
+export type CompletionItem = CompletionItemType & {
+  isDeprecated?: boolean;
+  deprecationReason?: string;
```
yoshiakis: Isn't it `string | null | undefined`? Because [deprecationReason: Maybe\<string>;](https://github.com/graphql/graphql-js/blob/b4bff0ba9c15c9d7245dd68556e754c41f263289/tstypes/type/definition.d.ts#L729) in `GraphQLEnumValue` interface and [deprecationReason?: Maybe\<string>;](https://github.com/graphql/graphql-js/blob/b4bff0ba9c15c9d7245dd68556e754c41f263289/tstypes/type/definition.d.ts#L510) in `GraphQLField` interface.

acao:
good catch! string | undefined will have to do for now

packages/graphql-language-service-types/src/index.ts
acao: good catch! string | undefined will have to do for now. ill fix this in the symbols support PR

PR #1144, opened 2019-12-21
feat: replace use of enzyme with react-testing-library

ryan-m-walker:
- added `@testing-library/react` to graphiql package
- added `@testing-libary/jest-dom` to graphiql package
- removed usage of `ENZYME` env variable in test script in graphiql packages
- removed enzyme adapter setup from `enzyme.config.js` file and renamed it
to `test.config.js`
- removed use of enzyme and replaced with react-testing-library in all
graphiql package test files and updated tests to use
react-testing-library patterns
- moved mockStorage to own file so it can be reused in other test files
and added other methods to it
- added a CodeMirror mock to be used in tests
- mocked out codemirror addon modules

completes: https://github.com/graphql/graphiql/projects/4#card-30693199

This PR gets rid off all uses of Enzyme in the GraphiQL package and replaces them with @testing-libary/react.


acao:
@ryan-m-walker ok so i merge the react 16 upgrade PR so you'll just need to rebase it looks like?

ryan-m-walker:
@acao ok, I rebased

acao:
@ryan-m-walker nice that was quick, thank you!

acao:
oh drat, i just merged the storybooks PR and so another conflict here

ryan-m-walker:
@acao no worries, i'll try to fix later tonight

ryan-m-walker:
@acao there's a minor change I want to make anyways

ryan-m-walker:
@acao should be good now

packages/graphiql/src/components/__tests__/helpers/codeMirror.js
```diff
@@ -0,0 +1,81 @@
+export function MockCodeMirror(node, { value, ...options }) {
+  let _eventListeners = {};
+  const mockTextArea = document.createElement('textarea');
```
acao: I don't think it matters now but it may eventually, but our instantiation pattern for GraphiQL query editor is a bit different from the generic Codemirror example that tends to use`<textarea />`.

We actually end up using a `<section />`:

[Codemirror Invocation](https://github.com/graphql/graphiql/blob/master/packages/graphiql/src/components/QueryEditor.js#L78)
[`this_node` attached via old ref pattern from `<section />` element](https://github.com/graphql/graphiql/blob/master/packages/graphiql/src/components/QueryEditor.js#L219)

ryan-m-walker: we target a `<section />` to instantiate CodeMirror but code mirror renders some more elements including a `<textarea />` inside that root node:

image

they handle the typing and presentation in a pretty complex way it seems, where they use the type area as a temporary moving cursor and then render out typed text in some presentational elements. i figured for our testing purposes it'd just be easier to use a base text area and the abstracted mocked internals of CodeMirror for our testing purposes and keep the API our code interacts with the same. If you have a better idea let me know though.

acao: indeed yes! just pointing out there will be a subtle difference in the DOM hierarchy, but I think the only delta would be caused by container targets for full-height container

acao: either way this is really great, and we can use this mock for codemirror-graphql I hope, which we still need to convert from mocha to jest

packages/graphiql/package.json
```diff
@@ -39,7 +39,7 @@
     "cypress-open": "yarn e2e-server 'cypress open'",
     "e2e": "yarn e2e-server 'cypress run'",
     "e2e-server": "start-server-and-test 'cross-env PORT=8080 node test/e2e-server' 'http-get://localhost:8080/graphql?query={test { id }}'",
-    "test": "cross-env ENZYME=true node ../../resources/runTests",
+    "test": "cross-env node ../../resources/runTests",
```
acao: we don't need `cross-env` here anymore if we aren't passing the enzyme env flag

packages/graphiql/src/components/DocExplorer/__tests__/FieldDoc.spec.js
```diff
@@ -35,47 +36,58 @@ const exampleObject = new GraphQLObjectType({
...
+    const { container, rerender } = render(
       <FieldDoc
         field={exampleObject.getFields().string}
         onClickType={jest.fn()}
       />,
     );
-    expect(W.find('MarkdownContent').text()).toEqual('No Description\n');
-    expect(W.find('TypeLink').text()).toEqual('String');
-    expect(W.find('Argument').length).toEqual(0);
+    expect(container.querySelector('.doc-type-description')).toHaveTextContent(
```
acao: I'm worried about how stable it will be using classes instead of component names, especially since we are about to change how layout/theming works and use theme-ui and @emotion

packages/graphiql/src/components/__tests__/helpers/codeMirror.js
```diff
@@ -0,0 +1,81 @@
+export function MockCodeMirror(node, { value, ...options }) {
```
acao: also the jest convention is to use __mocks__ directory I think:
https://jestjs.io/docs/en/manual-mocks#mocking-user-modules

however the `codeMirrorModules` const export maybe belongs in helpers?

packages/graphiql/src/components/__tests__/helpers/codeMirror.js
ryan-m-walker: gotcha, yeah this might need some changes as we go. I already added a few more elements to the mock CodeMirror DOM in another PR that uses this branch as its base because it need to listen for click drag events on the gutter bar.

packages/graphiql/src/components/DocExplorer/__tests__/FieldDoc.spec.js
ryan-m-walker: hmm yeah, I wasn't to keen on using querySelector but I was trying to avoid making any changes to non-test code. Do you think it's worth adding either labels or testids to the components? We mostly just use a lot of classes in these components and react-testing-library doesn't let you target react implementation code

Issue #1146, opened 2019-12-21
Use vscode workspaces, default to client watcher pattern ala LSP


acao:
We have a nice custom watcher built into our server using `fb-watchman`, however vscode and other LSP oriented IDE clients would expect to be able to provide their own, and for this server to respond to workspace watch/file change notifications sent from the IDE client.

Thus, we need to decouple this watcher from the server such that it doesn't bring many additional dependencies, and the `fb-watchman` dependency itself would become an `optionalDependencies` entry that we can suggest to the user if they are working with an IDE that expects a server, and they aren't using an LSP client compatibility layer with a built-in watcher for their IDE (which is probably common).

We could also ship it as an additional wrapper package, of the server bundled with fb-watchman. no harm in adding more packages to the monorepo!

acao:
@zth do you want to weigh in here as well? it seems that LSP reccomends against server side watchers, so we can move that logic to a vscode client lib of some kind

zth:
I think this approach sounds good! I actually think it'll be fine to just remove watchman/the server side watcher entirely, and then wait for the need for a server side watcher to actually arise before building one. What do you think about that?

acao:
@zth thats what i'm hoping for, yeah! we can just reccomend that they use an lsp-oriented client implementation, like vim-lsp, emacs-lsp, sublime-lsp, etc. if we need to revive the watcher by popular demand we can in a contrib space

acao:
https://github.com/sublimelsp/LSP i actually really want to create an example extension with this, it looks very mature and would be a great example.  as for the multi language regex parser, i think we could export that as a shared module to use for this interface https://code.visualstudio.com/api/references/vscode-api#2532

PR #1147, opened 2019-12-22
[Language Server] Support more languages

zth:
This PR introduces support for more source languages in the Language Server than `js` and raw `graphql` files. It does the following:
1. Adds support for `jsx`, `ts`, `tsx` and `re` (ReasonML)
2. Move all source extraction to use regex's, removing the previous `findGraphQLTags` that used `babel` to parse and extract tags from the `js` documents.
3. Adds _bail early_ file filters for all languages, in an attempt to ensure some performance.

Adding support for other languages should be fairly trivial as long as extracting sources from that language's files lends itself well to using regex's. For example, the Apollo Language Server (which started out as a fork from this LS), uses the regex-based approach, and supports Python, Dart and probably some more languages as well, in a rather straight forward way by using regex's.

### Is using regex's a good idea?
The rationale for using regex's is that extracting GraphQL sources from tags in documents is pretty straight forward. While there's great support for parsing and extracting sources from JavaScript and TypeScript using an AST-based approach, this probably won't be the case for other languages (think Python, ReasonML etc).

Extracting sources feel straight forward and easy enough that a regex based approach can make sense. But this is just my opinion, what do you feel?

### Should we default to including all languages or should it be opt in?
Currently with this PR, all supported files/languages are included when traversing the project and extracting GraphQL operations, but one could make the case supporting other things than `js` and `graphql` should be opt in. I've added a file filter for each language to ensure we can bail early on files that don't contain any operations, but I'm interested in hearing what you think about including all supported languages by default.


acao:
I'm a bit confused, we already have support for js and jsx? Did you add support via something other than babel?

zth:
@acao Hmm, like I said in the PR comment, I added support via regex's instead of babel. Now there's no reason babel couldn't still be used for `js`/`jsx`, but supporting `ts`/`tsx` would need some other mechanism to work (using the TS parser, or Babel with TS support, which may produce another type of AST that might or might not work with the current `findGraphQLTags`). Since the regex based approach works the same for js/ts (which the babel approach does not necessarily do), I moved both of them to the regex approach. That can of course be reverted back to using babel again if that's desirable.

Re jsx support: I haven't been able to make that work locally, and I think it's because the current code treats any file that doesn't end in exactly `.js` like a raw `.graphql` file: https://github.com/graphql/graphiql/blob/e90f2437dc6761ec72acb090be908eab14c1f857/packages/graphql-language-service-server/src/MessageProcessor.js#L687-L708

acao:
@zth would you be interested in replacing the babel parser as an experiment in how far you can take this new approach?

zth:
@acao I believe that's what I've already done in this PR, or maybe I'm misunderstanding what you mean? This PR removes the use of the Babel parser to extract GraphQL operation from JS files and replaces it with using regex to extract the source instead.

acao:
@zth I'm just confused because you didnt remove `findGraphQLTags` but i see now, that this totally replaces it, and doesn't remove the dependencies. its a great POC! 

for the full on PR, we will do this in typescript and add a bunch more tests to ease out edgecases, for the many ways template tags can be done in javacsript or typescript for example. maybe add a couple more languages if it's not too difficult. and then in that or a later pass, the ability to provide your own regexes as we were discussing!

so excited.

now remember, we will be moving all of this to the client and removing watchman from the server completely. so all of this work will come in super handy for creating a language service client abstraction!

https://code.visualstudio.com/api/references/vscode-api#workspace

zth:
`findGraphQLTags` is actually removed: https://github.com/graphql/graphiql/pull/1147/files#diff-94b4e244f58f1b38bdc5d1c4e2606d0f

Doing it in TS and adding more tests sounds great! I guess we need https://github.com/graphql/graphiql/pull/1138 to land before that can happen though, right?

acao:
@zth yes indeed! #1138 will happen, so i think we can abstract this into a module for #1146

acao:
@zth why don't you introduce this to it's own library? I might be doing the same for our own parsing logic, moving it to utils package I guess?

PR #1188, opened 2020-01-03
chore(examples): switch to using our own swapi-api

acao:
lets use our own swapi-api example as a reference example. because we use a wildcard we need to omit credentials in the example.

### DIFF
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -76,10 +76,10 @@ function graphQLFetcher(graphQLParams) {
   // When working locally, the example expects a GraphQL server at the path /graphql.
   // In a PR preview, it connects to the Star Wars API externally.
   // Change this to point wherever you host your GraphQL server.
-  const isDev = !window.location.hostname.match(
-    /(^|\.)netlify\.com$|(^|\.)graphql\.org$/,
-  );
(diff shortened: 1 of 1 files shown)


acao:
the netlify deploy preview under the list of PR checks is our confirmation that these requests are working now, by omitting `credentials` from the request headers, as reccomended here in MDN docs:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials#What_went_wrong

packages/graphiql/resources/renderExample.js
```diff
@@ -76,18 +76,18 @@ function graphQLFetcher(graphQLParams) {
...
+    : 'https://swapi-graphql.netlify.com/.netlify/functions/index';
   return fetch(api, {
     method: 'post',
     headers: {
       Accept: 'application/json',
       'Content-Type': 'application/json',
     },
     body: JSON.stringify(graphQLParams),
-    credentials: 'include',
+    credentials: 'omit',
```
benjie: @acao Prefer `credentials: 'same-origin'` which should Just Work :tm:. Here omitting is correct, but if someone uses this example code and changes the address to their own one then it might not work with their own stack because it's not sharing the cookies. Using `'same-origin'`  by default could reduce our support burden for issues like this.

PR #1198, opened 2020-01-06
[new-ui] Tabs & Tabbars

walaura:
Screenshot 2020-01-06 at 16 41 44

Adds `<TabBar/>`

- We got some mockups with buttons next to the tabs or on the opposite side - my idea for those is to create a `<Toolbar />` component (basically a wrapper around flexbox) that can wrap things like a `<Tabbar/>` or a `<ButtonBar/>` or anything. 

- I also removed an old `Layout.js` that was unused and lingering around


acao:
spectabular!

walaura:
this is ready to go on my end! i wanna give toolbars their own pr branching out from here!

PR #1199, opened 2020-01-06
docs: fix README.md relative links

jpaquim:
Remove single quotes and leading `.`, which were making the links lead to GitHub 404 pages.


acao:
thanks @jpaquim, good find!

PR #1203, opened 2020-01-08
[new ui] Toolbar component

walaura:
Add a `<Toolbar />` component to wrap elements like tabbars

[🌈 Demo 🌈](https://deploy-preview-1203--graphiql-test.netlify.com/storybook/?path=/story/toolbar--toolbar-with-tabs)

## Screenshots

Screen Shot 2020-01-08 at 4 55 52 pm
Screen Shot 2020-01-08 at 4 55 56 pm


acao:
ah! i see what you've done here. looks very nice

packages/graphiql/src/new-components/Toolbar/Tabs.js
```diff
@@ -0,0 +1,48 @@
...
+Tab.propTypes = { active: PropTypes.bool };
+
+const Tabs = ({ tabs, active, onChange }) => {
+  return (
+    <WithDividers>
+      {tabs.map((tab, index) => (
+        <Tab
+          key={index}
+          active={active === index}
+          onClick={() => onChange(index)}>
```
acao: eventually we might want a readable ID here, but i think that would be up to the particular reducer action that fires when `onChange` happens, not this business-agnostic component.

packages/graphiql/src/new-components/Toolbar/support/WithDividers.js
```diff
@@ -0,0 +1,65 @@
...
+      width: '1px',
+    }}
+  />
+);
+Divider.propTypes = { innerSx: PropTypes.object };
+
+const WithDividers = ({ children, padding = false, ...props }) => {
+  return (
+    <ul
+      data-contains-divider
```
acao: I didn't know you could pass attributes this way in JSX?

packages/graphiql/src/new-components/List/List.js
```diff
@@ -1,14 +1,14 @@
 /** @jsx jsx */
 import { jsx } from 'theme-ui';
-
 import PropTypes from 'prop-types';
 
-const ListRow = ({ children, flex = false, padding = true }) => (
+const ListRow = ({ children, flex = false, padding = false }) => (
```
acao: i like making `padding` explicit, excellent idea

walaura: i'm a bit conflicted bc ideally itll always be the children of listrow that have padding but it comes in super handy for storybook bc it's mostly text blocks if that makes sense?

walaura:
hold up why are the tabs listed twice on storybook will have a look

packages/graphiql/src/new-components/List/List.js
acao: yeah that does make sense

Issue #1191, opened 2020-01-06
Support `graphql@15.0.0` release


IvanGoncharov:
@acao @benjie Took a long long time than I initially anticipated but I finally released `graphql@15.0.0-rc1`.
It includes a number of small breaking changes:
https://github.com/graphql/graphql-js/releases/tag/v15.0.0-alpha.1
https://github.com/graphql/graphql-js/releases/tag/v15.0.0-alpha.2
https://github.com/graphql/graphql-js/releases/tag/v15.0.0-rc.1

Can you please test it with packages in this repo so I can release `15.0.0` without worrying about breaking GraphiQL?

Note: It doesn't yet include two-stage introspection that I currently working on and will add in the next release since it's not a breaking change.

acao:
awesome work @IvanGoncharov ! i tested it with first alpha before, looking forward to giving it a spin!

acao:
@IvanGoncharov good news, only a few failing tests around error message formatting, and the minified build is looking solid with some manual and automated tests. I'll push a PR  tonight, and hopefully I can check in on some of the breaking changes in more depth, as well as test out the LSP server

acao:
yeah just that one failing test related to graphql-config for the dev server. everything else is looking great overall. I'll keep it open and bump to the next RC if/when that comes.

IvanGoncharov:
@acao @benjie Just release RC2 can you please test it?
https://github.com/graphql/graphql-js/releases/tag/v15.0.0-rc.2

acao:
yes, can do!

acao:
@IvanGoncharov as noted on the PR we merged, we need to remove peer resolutions up to graphql 14 across the board, correct?

IvanGoncharov:
@acao Yes, it would be great!

acao:
@IvanGoncharov i did a patch release where we merged the #1204 bump, and will follow up with peer resolutions!

acao:
then we can follow up with seperate issues for adopting each new feature, i can create a whole project on here for it

acao:
@IvanGoncharov going to close this, we pre-released with 15 support a while ago, and i'm going to open new issues to adopt new features, as we are likely about to rebuild the doc explorer from scratch

PR #1204, opened 2020-01-08
feat: upgrade to graphql@15.0.0 for #1191

acao:
Have `graphql@15.0.0` working as a baseline, without necessarily adding support for new language features yet. Consequently, it requires an upgrade to `graphql-config-rc.2`. This is why we are still in alpha! Further PRs will follow that address things like interfaces implementing interfaces, etc

- [x] use `instrospectionQuery()` rather than constant (deprecated)
- [x] bump resolutions for 15
- [x] minor validation error formating issues
- [x] ensure GraphiQL works end to end (view deploy preview, cypress suite)
- [x] issue with `graphql-config` and server tests? potentially `graphql-config` is not working with 15.
- [x] check server for regressions, use with a vscode extension
- [x] manually check for more specific regressions in GraphiQL, other LSP implementations that are related to breaking changes.


acao:
@urigo so we are using `graphql-config@2.2.1` still, and it seems theres a bug with the introspection result in this test:

https://github.com/graphql/graphiql/blob/chore/graphql-15-rc/packages/graphql-language-service-server/src/__tests__/GraphQLCache-test.js#L46

when i print out the result of the introspection request, i get:
`Error: Must provide Source. Received: undefined.`

acao:
only bug left is that this type is deprecated in graphql 15:
image

@ardatan @urigo

maybe i can open an issue?

acao:
https://github.com/ardatan/graphql-tools/pull/1 i think this is the right place?

Urigo:
I think this was already fixed on `graphql-toolkit`'s master.

tomorrow we'll make sure to have updated versions of it published

acao:
@Urigo everything looks good now, thank you!

acao:
oh, tests were fine, but build not so much, spoke too soon there
we just updated to the latest `graphql-config`, maybe @ardatan knows whats up?

image

acao:
@IvanGoncharov it's finally ready for you to review, what do you think?

acao:
@IvanGoncharov I think the only thing we missed here as a baseline is removing the peer dependencies on lower versions of graphql. We should only support 14 and 15 now correct?

PR #1061, opened 2019-12-04
chore: improve documentation & examples

acao:
Addresses #1009, overhauling docs for recent build changes, and to improve the developer experience overall

- simplified  the cdn example to honor its true name, and have no external files so it's 100% clear what it does. it's cloned to `graphiql/cdn.html` for netlify
- the webpack example copies its result to the base of `graphiql` subpackage on `build-demo`. this means we can access it at https://graphiql-test.netlify.com/webpack for master, and the equivalent for each deploy preview. It also means adding it to the e2e suite could be relatively easy even? (note: the e2e suite currently tests graphiql.min.js which is also built with webpack. but the /webpack basepath gives us the ability to test webpack _implementations_
- updated docs to show demos that are made from this PR, so the links won't work til merge, but try them in the deploy previews!
  - /webpack
  - /dev
  - /cdn
  - /analysis
- in terms of the docs, i tried to cover as many of the most common issues upfront. an FAQ should be great to fill in the rest. this will eventually be a gatsby docsite, as in a couple weeks eventually!

PR #1212, opened 2020-01-10
docs: fix CDN example fetcher text fallback

benjie:
Missed out from this review comment: https://github.com/graphql/graphiql/pull/1061/files#r356005286

(`response.json()` [returns a promise](https://developer.mozilla.org/en-US/docs/Web/API/Body/json)) so it'll never throw synchronously assuming `response` is a response object, so the `try/catch` wasn't doing anything.


acao:
oops missed this one, thanks! is this fixed for `packages/graphiql/renderExample` and `examples/graphiql-webpack` as well?

benjie:
The others are all fine :+1:

PR #1214, opened 2020-01-12
chore: add typescript tooling to graphiql package

ryan-m-walker:
* added tsconfig files for esm and cjs
* updated scripts to use tsc instead of babel
* converted `find.js` and `find.spec.js` to typescript
* added `@babel/preset-typescript` to `babel.config`
* updated webpack to use babel-loader on `.ts` and `.tsx` files
* added `fork-ts-checker-webpack-plugin` to type check typescript files when using webpack
* added source maps to webpack builds


packages/graphiql/package.json
```diff
@@ -25,17 +25,16 @@
   "scripts": {
     "webpack": "webpack --config resources/webpack.config.js",
     "analyze-bundle": "cross-env NODE_ENV=production CDN=1 ANALYZE=1 yarn webpack -p",
-    "build": "yarn build-clean && yarn build-cjs && yarn build-esm && yarn build-flow",
-    "build-cjs": "babel src --ignore '**/__tests__/**,**/__mocks__/**,*.spec.*' --out-dir dist/",
-    "build-esm": "cross-env ESM=1 babel src --ignore '**/__tests__/**,**/__mocks__/**,*.spec.*' --out-dir esm/",
+    "build": "yarn build-clean && yarn build-cjs && yarn build-esm",
```
acao: im considering removing these commands since they arent used in CI

packages/graphiql/src/utility/find.ts
```diff
@@ -3,15 +3,12 @@
...
- *
- * @flow
- *
  */
 
 export default function find<T>(
   list: Array<T>,
   predicate: (item: T) => boolean,
-): ?T {
+): T | undefined {
```
acao: ope, you found the only flow in GraphiQL itself!

packages/graphiql/src/utility/__tests__/find.spec.ts
```diff
@@ -12,7 +12,7 @@ describe('Find', () => {
     expect(find([1, 2, 3], num => num === 2)).toEqual(2);
   });
 
-  it('should return undefined if element which returns true for predicate dont exist in the array', () => {
+  it("should return undefined if element which returns true for predicate doesn't exist in the array", () => {
```
acao: shouldnt this be single quotes? weird that prettier didnt pick up on it

acao:
oops looks like we're building graphiql twice now!

the root `build-js` script is almost gone thankfully, between this PR and another, so just remove `graphiql` from there for this PR

acao:
i love it! so

if you:

`yarn build-clean` - delete all of the build dist
and
`yarn workspace graphiql dev`
it still builds the whole project tree! this is gonna be amazing!

now we can split out CI tasks more readily as github actions (such as cypress), our CI is going to be so fast

packages/graphiql/src/utility/find.ts
acao: oh and this should be `T | void`

acao:
going to merge this as #1215 with a few of these missing pieces, awesome work!

PR #1225, opened 2020-01-14
WIP: Use new GraphQL Config

ardatan:
- [x] Use new GraphQL Config
- [x] Remove extra endpoint extension

### About endpoint
In new GraphQL Config, there is no extension called endpoint so all sources of schema are defined as `schema` in GraphQL Config file. 

### Related https://github.com/graphql/graphiql/pull/1204#issuecomment-572195548 and https://github.com/graphql/graphiql/issues/1191
This alpha version of GraphQL Config supports GraphQL-js v15. We're testing our graphql-toolkit which has core logic in the CLI for both v14 and v15 in the CI.
https://github.com/ardatan/graphql-toolkit/pull/403/checks?check_run_id=387927623

### Related https://github.com/graphql/graphiql/pull/920
Also user can add their own custom schema loaders in two different ways;
```yml
schema:
    - your-own-schema-pointer:
         loader: my-loader#loaderFn
```

or the developer can add their loaders in the library code;
```ts
const MyOwnExtension: GraphQLExtensionDeclaration = api => {
  // Schema
  api.loaders.schema.register(new CodeFileLoader());
  api.loaders.schema.register(new GitLoader());
  api.loaders.schema.register(new GithubLoader());
  api.loaders.schema.register(new ApolloEngineLoader());
  api.loaders.schema.register(new PrismaLoader());

  // Documents
  api.loaders.schema.register(new CodeFileLoader());

  return {
    name: 'my-own-extension',
  };
};
const config = await loadConfig({ extensions: [MyOwnExtension]});

```
More info; https://graphql-config.com/docs/loaders'

### Related https://github.com/graphql/graphiql/pull/1118

Let's say you have some custom extensions in your configuration file;
```yml
schema: some-schema-pointer.graphql
extensions:
   my-own-extension:
       myOwnField: true
```

You can get like below
```ts
const { myOwnField } = config.get('my-own-extension');
```


acao:
this is great! i was just about to merge a PR that refactors all of this code to Typescript in the next day or two. we also have a PR open for graphql 15 support. keep in mind that we need the LanguageService class in interface to work in the browser for monaco support here in the next month or so.

acao:
ah nice i see how several things are addressed here.

i think we can merge this before the typescript refactor. do we think this is ready? 

will  this break for existing 2.x.y compatible config?

also this PR is a great opportunity to update docs further and show which graphql config options are available and how to use them? @divyenduz ping

divyenduz:
@ardatan Thanks for the PR :) 🎉

This would probably break for the VSCode GraphQL extension as it uses the endpoint extension for this [in editor query](https://github.com/prisma-labs/vscode-graphql/releases/tag/v0.1.4) feature. Thanks for the heads up. I would still not block this merges/release and make changes (by using an alternate way of finding the endpoint OR removing that feature for now) in the VSCode GraphQL extension because then we can move more of the ecosystem faster together. 

@acao What's the current way of testing PRs? Do they get released as alpha or do I set it up locally and link it to the code (using `yarn link`? maybe) where I want to test it? 🎉

ardatan:
@divyenduz we can directly pass URL in schema field in new GraphQL Config.
```yaml
schema: http://localhost:3000/graphql
```

So in that case, GraphQL Config's getSchema method would give you an executable schema as GraphQLSchema which can be used with graphqljs directly.

About migration process and breaking changes, we can encourage people to create a new GraphQL Config file using new GraphQL CLI from scratch. Existing GraphQL Config users can migrate to new API easily by calling a single command `graphql init`

Also new GraphQL Config also supports Prisma as well like below;
```yaml
schema: ./prisma.yml
```

What do you think? Should we follow another path or do all these work for you? I'd love to hear your thoughts.

divyenduz:
@ardatan This would work for me (I can release the new extension with semver and you can always revert the extension if you don't want to change your GraphQL config today). 

Is there a link that I can refer to in the VSCode extension docs though? That provides a migration guide to the new GraphQL config format?

zth:
@ardatan this is great! Any chance this could be rebased to latest master? Quite a lot has happened with conversion to TS etc.

acao:
i think we should wait til graphql config 2 is stable OR better yet help get graphql config to a stable state? i mean @urigo has even offered maintainership in the past, if thats even needed. once i go full time in a week i could definately see myself diving in deep helping @ardatan move things along possibly?

acao:
@ardatan lets meet next week with @urigo next tuesday to finalize things for the stable release? is that too long to wait?

Urigo:
@acao no problem, let's do that, we would love to wait for you to get everything integrated according to your needs

acao:
@ardatan re-created this PR as #1342 with the typescript migration via a cherry-pick, but you can rebase/etc as you please here if you prefer that. Just need to remove anything watchman related and a few other upstream changes

acao:
@ardatan @Urigo thanks so much for all your help on this so far, everything is looking great, also the graphql 15 rc 2 PR is working great with this new implementation, thank you!

PR #1238, opened 2020-01-16
Use Object.assign instead of spread to enable MS Edge

mfix22:
Context: https://github.com/carbon-app/carbon/pull/932

### DIFF
--- a/packages/graphql-language-service-parser/src/onlineParser.ts
+++ b/packages/graphql-language-service-parser/src/onlineParser.ts
@@ -235,7 +235,7 @@ function pushRule(
   if (!rules[ruleKind]) {
     throw new TypeError('Unknown rule: ' + ruleKind);
   }
-  state.prevState = { ...state };
+  state.prevState = Object.assign({}, state);
   state.kind = ruleKind;
   state.name = null;
   state.type = null;


acao:
hey @mfix22! if you want Edge support, you'll need to use a bundler like webpack or rollup -or- you can use the CDN version if your needs are simple.

mfix22:
@acao I see, thanks! 

I'm fairly sure this is the only thing that is preventing this library from supporting Edge, but I understand if you want to be explicit about not supporting it.

acao:
the idea is that GraphiQL supports whatever browsers your configure it to support. So if you include a browserslist in your project you’ll be set!

maxr96:
> the idea is that GraphiQL supports whatever browsers you configure it to support. So if you include a browserslist in your project you’ll be set!

Can you explain this idea a bit more? How I am supposed to tell `graphql-language-service-parse` package to build with Edge support? If I add a `browserslist` to my project it still doesn't fix the problem.

benjie:
@maxr96 Check out our custom webpack example; you can configure this with whatever browser support you need: https://github.com/graphql/graphiql/tree/4803aec94cc723bbcaa8cfa3c5f811fbf6cd3d20/examples/graphiql-webpack

PR #1263, opened 2020-01-23
feat: add support for more languages, starting with ReasonML

zth:
(I redid https://github.com/graphql/graphiql/pull/1147 here)

This PR introduces support for more source languages in the Language Server than `js` and raw `graphql` files. It does the following:

1. Adds support for `jsx`, `ts`, `tsx` and `re` (ReasonML)
2. Move all source extraction to use regex's, removing the previous `findGraphQLTags` that used `babel` to parse and extract tags from the js documents.
3. Adds bail early file filters for all languages, in an attempt to ensure some performance.
4. Adding support for other languages should be fairly trivial as long as extracting sources from that language's files lends itself well to using regex's. For example, the Apollo Language Server (which started out as a fork from this LS), uses the regex-based approach, and supports Python, Dart and probably some more languages as well, in a rather straight forward way by using regex's.

Ping @acao


packages/graphql-language-service-server/src/__tests__/MessageProcessor-test.ts
```diff
@@ -208,56 +208,120 @@ describe('MessageProcessor', () => {
...
-query Test {
-  test {
-    value
-    ...FragmentsComment
-  }
-}
-`);
-  });
+  describe('getQueryRange', () => {
+    describe('js/ts', () => {
```
acao: so this test is marked as a js/ts test, but i see a flow js file? can we have a test for a flow .js, normal es6 .js, and a .ts file? just to be certain everything is hunky dory

packages/graphql-language-service-server/src/languageUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+  let i = 0;
+
+  function rangeContains(range: StrRange, index: number) {
+    return range.start <= index && index < range.end;
+  }
+
+  function getLocation(range: StrRange, index: number): Location {
+    return {
+      line: offsetLine + range.line,
+      column: offsetColumn + index - range.start,
```
acao: from what i understand, you could change `offsetColumn` here to 0 and remove line 22 and it would not change anything? is this or `offsetLine` meant for something? or is line 21/22 there for configurability/visibility? just curious more than anything

packages/graphql-language-service-server/src/languages.ts
```diff
@@ -0,0 +1,66 @@
...
+const jsExtractor = makeExtractTagsFromSource(jsGraphQLTagsRegexp);
+
+const reasonFileFilterRegexp = new RegExp(/(\[%(graphql|relay\.))/);
+const reasonGraphQLTagsRegexp = new RegExp(
+  /(?<=\[%(graphql|relay\.\w*)[\s\S]*{\|)[.\s\S]+?(?=\|})/g,
+);
+const reasonExtractor = makeExtractTagsFromSource(reasonGraphQLTagsRegexp);
+
+const jsLanguages: LanguageConfig = {
+  extensions: ['js', 'jsx', 'ts', 'tsx'],
```
acao: does this work for `.esm.js`, etc? also `.mjs` we need as well i think?

packages/graphql-language-service-server/src/languages.ts
```diff
@@ -0,0 +1,66 @@
...
+  fileFilter: (content: string) => jsFileFilter.test(content),
+  extractSources: (content: string) =>
+    jsExtractor(content).map(source => ({
+      ...source,
+      template: source.template.replace(jsRemoveInterpolationRegexp, ''),
+    })),
+};
+
+const graphql: LanguageConfig = {
+  extensions: ['graphql'],
```
acao: folks will want to add the other extension i think. on second thought from what i've said before, its fine to add `.gql` here as a catch all, as the spec never refers to it and always refers to .graphql files, so officially we do not endorse that extension but its there for the language server ok fine, haha

packages/graphql-language-service-server/src/languages.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const reasonML: LanguageConfig = {
+  extensions: ['re'],
+  fileFilter: (content: string) => reasonFileFilterRegexp.test(content),
+  extractSources: reasonExtractor,
+};
+
+export const languages: ReadonlyArray<LanguageConfig> = [
+  jsLanguages,
```
acao: so instead of readonly, why not export a function that pushes to this const array in memory, so someone could register some languages by calling this globally even, when instantiating the server seperately from the server instantiating config?
```ts
function registerLanguage(lang: LanguageConfig) {
     languages.push(lang)
} 
```
then, whenever `getLanguageFromExtension` is invoked the new languages are there

packages/graphql-language-service-server/src/__tests__/MessageProcessor-test.ts
zth: Yeah, I'll add tests for more extensions :+1: 
Since it's regexp based now we don't really need to care about any other content than the actual graphql tag, which means it's only really the extension that's interesting to test, at least in my view.

packages/graphql-language-service-server/src/languages.ts
zth: Adding them to the tests!

packages/graphql-language-service-server/src/languages.ts
zth: I'll add `.gql` + tests :+1:

packages/graphql-language-service-server/src/languages.ts
zth: Great idea! That'll make it very flexible. Adding it! :)

packages/graphql-language-service-server/src/languageUtils.ts
zth: This is actually copied/vendored from https://github.com/Rich-Harris/locate-character so I don't really dare change anything unless absolutely necessary :grin:

acao:
we also want to remove @babel/parser and types in this PR, and other now unused dependencies

zth:
> we also want to remove @babel/parser and types in this PR, and other now unused dependencies

That should be the only one unused, right?

packages/graphql-language-service-server/src/__tests__/MessageProcessor-test.ts
```diff
@@ -208,56 +208,154 @@ describe('MessageProcessor', () => {
...
+    
+    export function Example(arg: string) {}`;
+
+            const contents = getQueryAndRange(text, 'test.' + ext);
+            expect(contents.length).toEqual(0);
+          });
+        });
+      }
+
+      ['js', 'ts', 'jsx', 'tsx', 'es6', 'mjs', 'esm.js'].forEach(testExtension);
```
acao: nicely done!

for future reference. it.each() is great for this

also for future reference, I just wanted to test a single ts file to make sure we were set up to parse typescript as well, since this example still doesnt show a plain javascript example

acao:
@zth yeah and `@babel/types`, just double check the` findGraphQLTags` file you deleted but I think those are the only two

acao:
ok gonna give this a manual local test with `vscode-graphql` and then if all looks good we should be able to merge!

packages/graphql-language-service-server/src/languageUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+          new Position(end.line, end.column),
+        ),
+      });
+    }
+  }
+
+  // Reset RegExp
+  regexp.lastIndex = 0;
+
+  return [...sources];
```
captbaritone: Why clone this array?

zth: It's to make TypeScript happy, turning a regular array into a ReadonlyArray.

PR #1264, opened 2020-01-24
feat: export more functions

yoshiakis:
Notice: I'll PR this change once #1265 has completed since I want to rebase the branch at this PR on master branch after #1265 has finished.

This change exports more functions from `graphql-language-service-interface` so `codemirror-graphql` can use them, too (#902 ).


acao:
hey thanks! sure this works, though ideally soon we should be instantiating the GraphQLLanguageService in codemirror instead as we will be doing with monaco, so we can leverage the class instance for a bunch of useful things

acao:
also, with lerna's conventional commits implementation, you dont need to indicate the package for scope, as this will automatically generate a change to only the interface changelog anyways

acao:
I take that back, I see now that codemirror uses the `ContextToken` argument which the LanguageService does not expose.

this should be fine then!

packages/graphql-language-service-types/src/index.ts
```diff
@@ -247,6 +247,16 @@ export type ContextToken = {
   style: string;
 };
 
+export type ContextTokenForCodeMirror = {
```
acao: `ContextToken` is only used by Codemirror, so lets combine this into one type

packages/graphql-language-service-types/src/index.ts
```diff
@@ -283,11 +293,24 @@ export type CustomValidationRule = (
...
+  label: string;
+  isDeprecated?: boolean;
+};
+
 export type CompletionItem = CompletionItemType & {
   isDeprecated?: boolean;
   deprecationReason?: string;
 };
 
+export type CompletionItemForCodeMirror = {
```
acao: what is diffferent about this type? do you just need to make it `deprecationReason?: string | null`

packages/graphql-language-service-types/src/index.ts
yoshiakis: Isn't `ContextToken` used inside `graphql-language-service-interface` even if you do with vscode?
https://github.com/graphql/graphiql/blob/917c3931634cf27eb31a0383ff7c15ffd902d221/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L394

packages/graphql-language-service-types/src/index.ts
yoshiakis: In [a successive PR](#902), `getAutocompleteSuggestions` for CodeMirror returns an object that only contains properties `label`, `type`, `documentation`, `isDeprecated`, `deprecationReason`. It doesn't contain many of properties of `CompletionItemType`, so I thought using a type designated for it, which is  `CompletionItemForCodeMirror`, it is easier to understand what `getAutocompleteSuggestions` for CodeMirror returns.

acao:
@yoshiakis coming back to this and realizing that a) our flow types are again woefully out of date and b) the monaco PR is about to lead to some LSP interface changes.

yoshiakis:
@acao Thank you for the information! This PR was one of PRs to be needed to fix #868. However, GraphiQL and GraphQL Playground seem to use monaco instead of codemirror-graphql in the near future so I'm thinking to give up to fix it.

acao:
@yoshiakis oh my bad, i think i was mistaken. let me double check this PR

acao:
looks good to me!

PR #1265, opened 2020-01-24
feat(gls-types): add more type aliases

yoshiakis:
This change adds and exports more type aliases so `graphql-language-service-interface` and `codemirror-graphql` can use them (#902 , #1264 ).

### DIFF
--- a/packages/graphql-language-service-types/src/index.js.flow
+++ b/packages/graphql-language-service-types/src/index.js.flow
@@ -222,6 +222,16 @@ export type ContextToken = {
   style: string,
 };
 
+export type ContextTokenForCodeMirror = {
+  start: number,
+  end: number,
+  string: string,
+  type: string | null,
+  state: State,
+};
+
+export type ContextTokenUnion = ContextToken | ContextTokenForCodeMirror;
+
 export type TypeInfo = {
   type: ?GraphQLType,
   parentType: ?GraphQLType,
@@ -269,6 +279,19 @@ export type CompletionItem = {
   documentation?: ?string,
   // GraphQL Deprecation information
   isDeprecated?: ?boolean,
+  deprecationReason?: ?string,
+};
+
+export type CompletionItemBase = {
+  label: string,
+  isDeprecated?: boolean,
+};
+
+export type CompletionItemForCodeMirror = {
+  label: string,
+  type?: GraphQLType,
+  documentation?: ?string,
+  isDeprecated?: boolean,
   deprecationReason?: ?string,
 };
 
--- a/packages/graphql-language-service-types/src/index.ts
+++ b/packages/graphql-language-service-types/src/index.ts
@@ -247,6 +247,16 @@ export type ContextToken = {
   style: string;
 };
 
+export type ContextTokenForCodeMirror = {
+  start: number;
+  end: number;
+  string: string;
+  type: string | null;
+  state: State;
+};
+
+export type ContextTokenUnion = ContextToken | ContextTokenForCodeMirror;
+
 export type AllTypeInfo = {
   type: GraphQLType | null | undefined;
   parentType: GraphQLType | null | undefined;
@@ -283,9 +293,22 @@ export type CustomValidationRule = (
 
 export type Diagnostic = DiagnosticType;
 
+export type CompletionItemBase = {
+  label: string;
+  isDeprecated?: boolean;
+};
+
 export type CompletionItem = CompletionItemType & {
   isDeprecated?: boolean;
   deprecationReason?: string;
+};
+
+export type CompletionItemForCodeMirror = {
+  label: string;
+  type?: GraphQLType;
+  documentation: string | null | undefined;
+  isDeprecated?: boolean;
+  deprecationReason: string | null | undefined;
 };
 
 // Below are basically a copy-paste from Nuclide rpc types for definitions.


packages/graphql-language-service-types/src/index.ts
```diff
@@ -283,11 +293,24 @@ export type CustomValidationRule = (
...
+  label: string;
+  isDeprecated?: boolean;
+};
+
 export type CompletionItem = CompletionItemType & {
   isDeprecated?: boolean;
   deprecationReason?: string;
 };
 
+export type CompletionItemForCodeMirror = {
```
acao: where is this used?

packages/graphql-language-service-types/src/index.ts
```diff
@@ -283,11 +293,24 @@ export type CustomValidationRule = (
 
 export type Diagnostic = DiagnosticType;
 
+export type CompletionItemBase = {
```
acao: and where is this used?

acao:
thanks!

we are about to replace `ContextToken` with the vscode LSP protocol `ContextToken` in another PR, fyi. the goal is to replace most of the types with vscode's LSP protocol types

acao:
can you go ahead and do that in this PR actually? import `ContextToken as ContextTokenType` using the pattern at the top with the other vscode interfaces, and set our `ContextToken` to that? theres another PR for that but it has some other issues

packages/graphql-language-service-types/src/index.ts
acao: ohhh i see, this will be used by a successive codemirror PR?

acao:
Found these, they might be useful for us to just pull in as well:
https://github.com/flowtype/flow-for-vscode/blob/master/flow-libs/vscode-languageserver-protocol.js.flow
thanks @orta ! any way we can help publish these standalone somewhere?

packages/graphql-language-service-types/src/index.js.flow
```diff
@@ -222,6 +222,16 @@ export type ContextToken = {
   style: string,
 };
 
+export type ContextTokenForCodeMirror = {
```
acao: looking at this more, `ContextToken` is only used by codemirror, and does not seem to be used by vscode/monaco at all, which is interesting.

This means that we don't need a different type for codemirror

This also explains why you need to consume the methods directly instead of instantiating the class

acao: https://github.com/graphql/graphiql/blob/de4005b0fd486d5d81b50dcdc226f990c181ec4b/packages/graphql-language-service-interface/src/GraphQLLanguageService.ts#L198

both here and with hover, we dont pass the 4th context token argument
this explains a lot!

packages/graphql-language-service-types/src/index.js.flow
```diff
@@ -272,6 +282,19 @@ export type CompletionItem = {
   deprecationReason?: ?string,
 };
 
+export type CompletionItemBase = {
```
acao: there should be just one `CompletionItem` that ideally extends the vscode typescript definition, or mirrors what we have.

packages/graphql-language-service-types/src/index.ts
```diff
@@ -283,11 +293,24 @@ export type CustomValidationRule = (
...
+  isDeprecated?: boolean;
+};
+
 export type CompletionItem = CompletionItemType & {
   isDeprecated?: boolean;
   deprecationReason?: string;
 };
 
+export type CompletionItemForCodeMirror = {
+  label: string;
```
acao: following above, let's combine completion items.

packages/graphql-language-service-types/src/index.ts
```diff
@@ -283,11 +293,24 @@ export type CustomValidationRule = (
...
+  type?: GraphQLType;
+  documentation: string | null | undefined;
```
acao: please use optionals `?:` instead of `| undefined`

packages/graphql-language-service-types/src/index.js.flow
yoshiakis: Thank you for your feedback! Should I change the code to make function `getTokenAtPosition` return a token which has the same type as `ContextToken` for codemirror? Because it currently returns a token which has a different type.
https://github.com/graphql/graphiql/blob/917c3931634cf27eb31a0383ff7c15ffd902d221/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L391

packages/graphql-language-service-types/src/index.ts
yoshiakis: Sure!

packages/graphql-language-service-types/src/index.ts
yoshiakis: Hmm. apparently, the completion items for vscode/monaco is kind of different than the  completion items for codemirror. I thought if we combined these items, it would make it difficult to see which items are for vscode/monaco, for codemirror, or for both. Should we combine them?

yoshiakis: Or, instead of exporting these type aliases and functions I'm trying to export in other [PR](#1264), how about having these in `codemirror`. (Added functions is the same as the ones in `graphql-language-service-interface` but accept different type.)

acao:
@yoshiakis i think this has been resolved already? by combining and extending types? want to reopen this?

yoshiakis:
@acao It's not resolved yet but It's ok to keep this PR closed! I think there is a more suitable solution for this and if I have time to do that I'll open another PR.

PR #1267, opened 2020-01-26
chore: switch to importing validation rules from graphql's public API

IvanGoncharov:
I'm looking into cleaning up how validation rules are exported in
`15.0.0`. So I decided to refactor this code and switch rules imports to
use `*Rule` that is public API instead of importing internal files
directly. Sadly we forget to expose `ExecutableDefinitions` rule in
public API but I will fix it in `15.0.0` and backport into `14.x.x`

### DIFF
--- a/packages/graphql-language-service-utils/src/validateWithCustomRules.ts
+++ b/packages/graphql-language-service-utils/src/validateWithCustomRules.ts
@@ -7,12 +7,19 @@
  *
  */
 
-import { DocumentNode, TypeDefinitionNode } from 'graphql/language';
-import { GraphQLError } from 'graphql/error';
-import { GraphQLSchema } from 'graphql/type';
 import { CustomValidationRule } from 'graphql-language-service-types';
 
-import { specifiedRules, TypeInfo, validate } from 'graphql';
+import {
+  DocumentNode,
+  Kind,
+  specifiedRules,
+  validate,
+  GraphQLError,
+  GraphQLSchema,
+  NoUnusedFragmentsRule,
+  KnownFragmentNamesRule,
+} from 'graphql';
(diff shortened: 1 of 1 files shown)


packages/graphql-language-service-utils/src/validateWithCustomRules.ts
```diff
@@ -7,12 +7,19 @@
...
+  DocumentNode,
+  Kind,
+  specifiedRules,
+  validate,
+  GraphQLError,
+  GraphQLSchema,
+  NoUnusedFragmentsRule,
+  KnownFragmentNamesRule,
+} from 'graphql';
+import { ExecutableDefinitions } from 'graphql/validation/rules/ExecutableDefinitions';
```
IvanGoncharov: Here are the main changes, rest is just refactoring.

acao: awesome, thank you @IvanGoncharov ! this will come in very handy

packages/graphql-language-service-utils/src/validateWithCustomRules.ts
```diff
@@ -23,54 +30,34 @@ export function validateWithCustomRules(
...
-      if (error.message.indexOf('Unknown directive') === -1) {
-        return true;
+  const errors = validate(schema, ast, rules);
+  return errors.filter(error => {
+    if (error.message.indexOf('Unknown directive') !== -1 && error.nodes) {
+      const node = error.nodes[0];
+      if (node && node.kind === Kind.DIRECTIVE) {
+        const name = node.name.value;
+        if (name === 'arguments' || name === 'argumentDefinitions') {
+          return false;
```
acao: much better 👍

PR #1272, opened 2020-01-26
Programatic graphql config

zth:
This is a draft of adding a way to supply a pre-existing `graphql-config` to the LS. The idea is that external consumers, like a vscode extension, should be able to provide defaults and extensions of the existing config (or generate it programatically all together) in a way that can be supplied directly to the LS.

Feedback much welcome!


packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -54,8 +54,11 @@ const {
   DIRECTIVE_DEFINITION,
 } = Kind;
 
-export async function getGraphQLCache(configDir: Uri): Promise<GraphQLCache> {
-  const graphQLConfig = await getGraphQLConfig(configDir);
+export async function getGraphQLCache(
+  configDir: Uri,
+  config?: GraphQLConfig,
+): Promise<GraphQLCache> {
+  const graphQLConfig = config || (await getGraphQLConfig(configDir));
```
acao: what i was thinking is that if you can pass a config instance here, you might be able to create the graphql-config programatically from an object literal, especially with the new 3.0.0 graphql-config once thats stable

zth: Not sure I follow what you mean. Could you expand a bit?

acao: my bad, i'm seeing here where 2.0.0 better suits this need, i was mistaken and thought that 1.2.x supported this too, but i see that this is the only option for now

acao:
@zth we merged the update to 3.0.0 rc 16 so have at it! it will be stable by the time we are. there are a few fantastic new options!

acao:
@zth i think this looks ready, and should work seamlessly with the new graphql-config

zth:
@acao do you want me to resolve the last merge conflict, and then it's good to go for merging?

acao:
@zth yeah if you could, that would be awesome. would love to pull this into the monaco service effort

acao:
@zth ah, we merged another upstream PR, so a bit more to rebase.

zth:
https://github.com/graphql/graphiql/pull/1432 supersedes this.

PR #1305, opened 2020-02-04
feat(definition): get scalar definition

divyenduz:
This PR makes it possible to jump to the definition of a named scalar type. 

How this would look like once integrated into VSCode for example: 
CleanShot-2020-02-04-at-213317


acao:
nice work!

PR #1320, opened 2020-02-06
fix: show a correct error message for LSP CLI

### LINKED ISSUE #1319: [graphql-language-service]: TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be one of type string or Buffer. Received type object
`graph-language-service` doesn't return an error message that shows what you need to do.

## how to reproduce
Run one of the following commands in the root of package `graphql-language-service`, then you got the same error message as the title.
`bin/graphql.js autocomplete --text foo --schemaPath ...`
`bin/graphql.js outline --text foo`
`bin/graphql.js validate --text foo --schemaPath ...`
(Before running these scripts, you need to run `yarn build` in advance.)

## cause
`error` given at https://github.com/graphql/graphiql/blob/917c3931634cf27eb31a0383ff7c15ffd902d221/packages/graphql-language-service/src/client.ts#L105
or
https://github.com/graphql/graphiql/blob/917c3931634cf27eb31a0383ff7c15ffd902d221/packages/graphql-language-service/src/client.ts#L134
or
https://github.com/graphql/graphiql/blob/917c3931634cf27eb31a0383ff7c15ffd902d221/packages/graphql-language-service/src/client.ts#L148
is actually type object, though `process.stderr.write` method doesn't allow for type object. Therefore, the method throws an error.

yoshiakis:
fixes #1319.

### DIFF
--- a/packages/graphql-language-service/src/client.ts
+++ b/packages/graphql-language-service/src/client.ts
@@ -102,7 +102,7 @@ function _getAutocompleteSuggestions(
     process.stdout.write(JSON.stringify(resultObject, null, 2));
     return GRAPHQL_SUCCESS_CODE;
   } catch (error) {
-    process.stderr.write(error.stack + '\n');
+    process.stderr.write((error?.stack ?? String(error)) + '\n');
     return GRAPHQL_FAILURE_CODE;
   }
 }
--- a/packages/graphql-language-service/src/client.ts
+++ b/packages/graphql-language-service/src/client.ts
@@ -102,7 +102,7 @@ function _getAutocompleteSuggestions(
     process.stdout.write(JSON.stringify(resultObject, null, 2));
     return GRAPHQL_SUCCESS_CODE;
   } catch (error) {
-    process.stderr.write(error);
+    process.stderr.write(error.stack + '\n');
     return GRAPHQL_FAILURE_CODE;
   }
 }
@@ -131,7 +131,7 @@ function _getDiagnostics(
     process.stdout.write(JSON.stringify(resultObject, null, 2));
     return GRAPHQL_SUCCESS_CODE;
   } catch (error) {
-    process.stderr.write(error);
+    process.stderr.write(error.stack + '\n');
     return GRAPHQL_FAILURE_CODE;
   }
 }
@@ -145,7 +145,7 @@ function _getOutline(queryText: string): EXIT_CODE {
       throw Error('Error parsing or no outline tree found');
     }
   } catch (error) {
-    process.stderr.write(error);
+    process.stderr.write(error.stack + '\n');
     return GRAPHQL_FAILURE_CODE;
   }
   return GRAPHQL_SUCCESS_CODE;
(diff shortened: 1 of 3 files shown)


packages/graphql-language-service/src/client.ts
```diff
@@ -102,7 +102,7 @@ function _getAutocompleteSuggestions(
     process.stdout.write(JSON.stringify(resultObject, null, 2));
     return GRAPHQL_SUCCESS_CODE;
   } catch (error) {
-    process.stderr.write(error);
+    process.stderr.write(error.stack + '\n');
```
benjie: Since we're already getting errors here, should we be more cautious? Something like:

```suggestion
    process.stderr.write((error?.stack ?? String(error)) + '\n');
```

This helps protect against someone throwing `undefined` (or, less likely, `null`):

```
> try { throw undefined; } catch (error) { process.stderr.write(error.stack + '\n'); }
Thrown:
TypeError: Cannot read property 'stack' of undefined
```

acao:
we’re having trouble with error messages in server, so please be sure to address the issue there rather than downstram in the cli

packages/graphql-language-service/src/client.ts
acao: @yoshiakis what do you think?

acao:
@yoshiakis thanks for this! just the one suggestion from @benjie to consider

packages/graphql-language-service/src/client.ts
yoshiakis: Sorry for late. Yes @benjie, I like your idea!

How about these? (Please tell me if the error expression looks weird 'cause I'm not proud of my writing skill in English.)
```TypeScript
if (error) {
  process.stderr.write(error.stack + '\n');
} else {
  process.stderr.write('falsy value such as undefined was thrown' + '\n');
}
```

acao: looks good to me!

acao: actually now that i think about it, i think benjie's suggestion is most suitable. we want to use String(error) to allow different error types to stringify themselves, which includes printing the call stack

yoshiakis: Unfortunately, String(error) may not include the call stack. You can check this by putting `process.stderr.out(String(error) + '\n');` below `process.stderr.write(error.stack + '\n');`.

benjie: `error?.stack ?? String(error)` should handle this: if error is set, and it has a stack property which is non-null and non-undefined, use this (i.e. the stack); otherwise fall back to `String(error)`

yoshiakis: @benjie You are right! I misunderstood what nullish coalescing operator (??) does.

I agree that benjie's suggestion is better than mine.

yoshiakis:
Probably we need to change settings so that Jest understand optional chaining. Test failed in my local environment.

acao:
closing for #1381

PR #1323, opened 2020-02-12
fix: prevent exceptions for unparse-able anon ops

### LINKED ISSUE #1308: GraphiQL throws "Uncaught TypeError: Cannot read property 'loc' of undefined" when there are more that one unnamed queries/mutations
GraphiQL has this behavior since the 0.14.0 version. The 0.13.x is fine

acao:
you'll see now that the graphql parser still fails on multiple anonymous operations in one query, however we're at least able to parse the query and not break the application with keys or undefined paths, and the dropdown still works

Fixes #1308


acao:
ended up bringing this into the typescript migration to fix a bug, so consider this fix merged :D

PR #1346, opened 2020-02-17
chore: refactor introspection utility for easier maintenance

acao:
this should be all we need for introspection queries, especially as they are becoming increasingly dynamic with GraphQL 15

would be nice to allow users to pass settings to `getIntrospectionQuery()` via props or plugins eventually

### DIFF
--- a/packages/graphiql/src/utility/introspectionQueries.ts
+++ b/packages/graphiql/src/utility/introspectionQueries.ts
@@ -5,9 +5,9 @@
  *  LICENSE file in the root directory of this source tree.
  */
 
-import { getOperationAST, parse, introspectionQuery } from 'graphql';
+import { getIntrospectionQuery, getOperationAST, parse } from 'graphql';
 
-export { introspectionQuery } from 'graphql';
+export const introspectionQuery = getIntrospectionQuery();
 
 export const staticName = 'introspectionQuery';
 const operationDocument = getOperationAST(parse(introspectionQuery), null);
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -6107,6 +6107,14 @@ cross-env@^7.0.0:
...
     cross-spawn "^7.0.1"
 
+cross-fetch@2.2.2:
+  version "2.2.2"
+  resolved "https://registry.yarnpkg.com/cross-fetch/-/cross-fetch-2.2.2.tgz#a47ff4f7fc712daba8f6a695a11c948440d45723"
+  integrity sha1-pH/09/xxLauo9qaVoRyUhEDUVyM=
+  dependencies:
+    node-fetch "2.1.2"
+    whatwg-fetch "2.0.4"
+
```
IvanGoncharov: @acao Why do you need these changes?

acao: my bad, a mistake from a rebase

packages/graphiql/src/utility/introspectionQueries.ts
```diff
@@ -5,109 +5,18 @@
...
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-  }
-`;
+export const introspectionQuerySansSubscriptions = introspectionQuery.replace(
```
acao: @IvanGoncharov is this suitable do you think? i want to eventually expose the parameters for `getIntrospectionQuery` to user configuration

packages/graphiql/src/utility/introspectionQueries.ts
```diff
@@ -5,109 +5,18 @@
...
 
-import { getOperationAST, parse, introspectionQuery } from 'graphql';
+import { getIntrospectionQuery } from 'graphql';
 
-export { introspectionQuery } from 'graphql';
+export const introspectionQuery = getIntrospectionQuery();
 
-export const staticName = 'introspectionQuery';
-const operationDocument = getOperationAST(parse(introspectionQuery), null);
+export const staticName = 'IntrospectionQuery';
```
acao: also I noticed that this changed?

acao:
@IvanGoncharov is this good for now, or should we expose the `getIntrospectionQuery()` options to the user via props somehow?

packages/graphiql/src/utility/introspectionQueries.ts
IvanGoncharov: @acao It's a normal temporary workaround 👍

acao:
@IvanGoncharov thank you!

PR #1361, opened 2020-02-23
improvement: default import for codemirror

ttzztztz:
Since the "codemirror" package have a default export, we can use it without using the `import * as CM from 'codemirror'`, we have a better, tidier way to import it.


acao:
just seems like a style preference, but thanks though!

PR #1362, opened 2020-02-24
fix: a warning caused by an infinity value

ttzztztz:
Reproduce:
Start the GraphiQL dev server, go to the localhost:8080, drag the divide bar to the right-most position, if the position `rightSize === 0`, the React State `EditorFlex` will be `Infinity`, and when the state dispatched to the CSS style property, CSS doesn't recognize `Infinity` and will throw a warning in the console.

My Solution:
Set a very big number (10000), if `rightSize === 0`, we should return the big number and avoid producing `Infinity` value in Javascript.

Safari Console Error:
Screen Shot 2020-02-24 at 19 49 19

Chrome Console Error:
Screen Shot 2020-02-24 at 19 59 00


acao:
we are re-implementing all of this now, but thanks!

PR #1374, opened 2020-02-25
fix: type check

Gumichocopengin8:
I got the following errors in packages/graphql-language-service-server/src/__tests__/GraphQLCache-test.ts
```
Argument of type 'Map<unknown, unknown>' is not assignable to parameter of type 'Map<string, ObjectTypeInfo>'.
  Type 'unknown' is not assignable to type 'string'
```

So added types to receive benefits of typescript


acao:
great work! thank you

PR #1377, opened 2020-02-26
chore(graphiql): minor refactoring based on object destructuring

pranshuchittora:
Trying to make code look cleaner and elegant. 

Would love to make more enhancements :+1:


acao:
we are in the midst of rewriting this component! thanks though!

PR #1379, opened 2020-02-28
test(parser): test for the parser-> Rules

pranshuchittora:
Would like to have some suggestions regarding the testing.

Q. Is the current approach of writing test is correct?
Q. Shall I write unit test or e2e test for parser?
Q. Any best practice.

cc @IvanGoncharov @acao


IvanGoncharov:
@pranshuchittora Thanks for the effort 👍 
But covering parser with tests is a part of the GSoC project and I don't think it's a good idea to start working on it ahead of GSoC.
Please focus on your "GSoC proposal" instead, where you can include your proposals on how to better test this code.

Issue #1165, opened 2019-12-26
GraphiQL Discussion: Breaking API Changes & New Interfaces 💥 -> 🚀


acao:
## 1. `GraphQL.state`
 (e.g. `GraphiQL.state.query`) is being accessed directly, and is advertised somewhat as a public interface in the Readme and has been for some time. However, our move to context means that providers will contain state. Also there will be multiple queries with multple tabs! Also if I'm not mistaken it's not really reccomended to access Component state directly like that?

**replacement proposal:**

```ts
const { query } = GraphiQL.state
```
becomes

```ts
const = { operation } = useOperationContext({ tabId: 0 })
```
and this one which will require the parent workspace context to work
```ts
const = { operation } = useOperationContext({ tabId: 'current' })
```
also:
```ts
<OperationProvider tabId={1}>
  <OperationContext.Consumer>{({ operation }) => ()}</OperationContext.Consumer>
</OperationProvider>
```

## 2. `GraphiQL.methods()`

There are also Component methods that are used to access values, submit queries programatically, and for many other purposes in the wild.

These also will move to context provider, possibly just dispatch actions with types that pass args mapped to some of the same and some new method names.

then you can readily dispatch actions via a similar interface:

```ts
const = { operation, setOperation } = useOperationContext({ tabId: 'current' })

return (
  <textarea onChange></textarea>
)
```
also:
```ts
<OperationProvider tabId={1}>
  <OperationContext.Consumer>{({ setOperation }) => ()}</OperationContext.Consumer>
</OperationProvider>
```

*a better approach with useGraphiQLReducers()*

```ts
import { useGraphiQLReducers, Actions } from 'graphiql'

function ASTReducer(action, state) {
  switch(action.type) {
    case [Actions.onOperationChange]: {
       state.operationAST = parse(state.operation)
       return state
     }
     default {
       return state
     }
  }
}

useGraphiQLReducers({
   init: () => { operation: 'mutation SendMsg{ send(msg:"hello!"){response} }',
   reducers: [ASTReducer]
})
```

the GraphiQL.methods() interfaces are also accessible outside of a bundler for react JSX (Parcel, Webpack, Rollup, etc). this new proposal however would require one of these bundlers to achieve these kinds of capabilities. You can also use `React.createElement` directly with the contexts we will export to avoid that, but that might grow cumbersome

## 3: `GraphiQL.Logo` and friends

[...]

PR #1380, opened 2020-02-28
WIP: feat: Add migration context to enable GraphiQL refactoring

zephraph:
This PR in essence wraps the entirety of GraphiQL in an aggregated
context provider to which other context providers can be attached. 

The short term goal of this is to separate out what is now GraphiQL state
into separate, smaller contexts which can be hooked into my plugins, etc

Being as we must ensure the provider is wrapped around GraphiQL I had to make
a facade component to do the wrapping and proxy props to the actual class.

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -47,6 +47,10 @@ import {
   introspectionQueryName,
   introspectionQuerySansSubscriptions,
 } from '../utility/introspectionQueries';
+import {
+  MigrationContextProvider,
+  MigrationContext,
+} from 'src/state/MigrationContext';
 
 const DEFAULT_DOC_EXPLORER_WIDTH = 350;
 
@@ -63,7 +67,7 @@ if (majorVersion < 16) {
 }
 
 declare namespace global {
-  export let g: GraphiQL;
+  export let g: GraphiQLInternals;
 }
 
 export type Maybe<T> = T | null | undefined;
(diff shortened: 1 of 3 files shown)


packages/graphiql/src/state/GraphiQLSchemaProvider.tsx
```diff
@@ -113,7 +113,7 @@ export const schemaReducer: SchemaReducer = (
  */
 
 export type SchemaProviderProps = {
-  config: SchemaConfig;
+  config?: SchemaConfig;
```
zephraph: This already provided a default so it seems it'd be fine to have as optional.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -134,21 +138,70 @@ type GraphiQLState = {
...
-    return JSON.stringify(result, null, 2);
-  }
+export const GraphiQL: React.FC<GraphiQLProps> &
+  GraphiQLStaticProperties = props => {
+  return (
+    <MigrationContextProvider>
+      <GraphiQLInternals {...props} />
+    </MigrationContextProvider>
+  );
+};
```
zephraph: I know this and the changes that follow are a bit ugly. I'm trying to preserve the api without requiring an extra wrapper which required some... creativity.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -134,21 +138,70 @@ type GraphiQLState = {
...
+  Footer: typeof GraphiQLFooter;
+  QueryEditor: typeof QueryEditor;
+  VariableEditor: typeof VariableEditor;
+  ResultViewer: typeof ResultViewer;
+  Button: typeof ToolbarButton;
+  ToolbarButton: typeof ToolbarButton;
+  Group: typeof ToolbarGroup;
+  Menu: typeof ToolbarMenu;
+  MenuItem: typeof ToolbarMenuItem;
+}
```
zephraph: Gotta be a better way to do this.

zephraph:
Note that the context isn't as of yet actually being used. To access the schema provider you'd use `this.context.schema` inside the `GraphiQLInternals` class component.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1266,6 +1296,8 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
   };
 }
 
+GraphiQLInternals.contextType = MigrationContext;
```
zephraph: This is the magic that enables `this.context.schema` inside the class component.

packages/graphiql/src/components/GraphiQL.tsx
acao: we're about to drop this interface for 1.0.0 so don't worry about it too much:
https://github.com/graphql/graphiql/issues/1165

packages/graphiql/src/components/GraphiQL.tsx
acao: oh yes, i remember it used to work this way with the old context API too, amazing how little the class component interface changed

PR #1384, opened 2020-02-28
chore: improve the way to print to stderr

yoshiakis:
print to stderr even if falsy value is thrown or thrown error does not have property `stack`. it was intended to implement by #1381


acao:
thanks, my bad, thought that change had been committed

PR #1386, opened 2020-02-29
chore: improved the CLI's help output

pranshuchittora:
The current `--help` output of the `graphql-language-service` CLI misses the possible <commands>.

### DIFF
--- a/packages/graphql-language-service/package.json
+++ b/packages/graphql-language-service/package.json
@@ -32,6 +32,6 @@
     "graphql-language-service-interface": "^2.4.0-alpha.1",
     "graphql-language-service-server": "^2.4.0-alpha.1",
     "graphql-language-service-utils": "^2.4.0-alpha.1",
-    "yargs": "^3.32.0 || ^7.0.0"
+    "yargs": "^15.2.0"
   }
 }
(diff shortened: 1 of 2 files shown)


pranshuchittora:
What's the roadmap for updating the deps.


https://github.com/graphql/graphiql/blob/bd4627e15d9224151b8c5f8fbb57c38187a2e1b2/packages/graphql-language-service/package.json#L35


https://github.com/yargs/yargs/blob/8aae3332251d09fa136db17ef4a40d83fa052bc4/package.json#L3

acao:
@pranshuchittora thanks for this PR! very helpful contribution.

the upgrade to yargs 7 would be a great contribution, will require a minor refactor likely, but should not be seen as a breaking change for the CLI

pranshuchittora:
> the upgrade to yargs 7 would be a great contribution, will require a minor refactor likely, but should not be seen as a breaking change for the CLI

Would love to contribute :+1: 
I think v7 might look outdated as the latest version of `yargs` is v15.

What's the project deps updating/upgrading strategy?

acao:
@pranshuchittora oh in that case, yes please update it to the latest, and if that means no breaking changes for the user, then we can prefix the squash commit with `deps:` or `fix:` to incur a patch, which will already happen next release due to the bugfix we just merged for the cli

pranshuchittora:
@acao pls review the new changes :smile:

acao:
thanks @pranshuchittora ! another thing to note is that we don't often need scope, because lerna automatically detects package scope for us

PR #1399, opened 2020-03-03
Mark prop-types peerDependency as optional

felixfbecker:
This avoids the log spam of unmet peer dependency warnings by yarn on every install for users that don't use prop-types.


acao:
we are in the midst of rewriting GraphiQL and there will be no prop types

felixfbecker:
Oh, nice. When do you think that will be released?

felixfbecker:
Does it hurt to merge this in the meantime?

Issue #1388, opened 2020-03-02
"Clear" button for Query Editor


harshithpabbati:
Clearing queries in the Query Editor could make use of a "Clear" button in the user interface of the Query Editor to make it easier to clear all content present in it quickly, just as with the "Copy" button (with a possible keyboard shortcut - maybe `Shift-Ctrl-L` or similar?) 

![Screenshot 2020-03-02 at 8 58 29 PM](https://user-images.githubusercontent.com/43822585/75690304-9a4a5a00-5cc8-11ea-956e-2c4b377d3de7.png)

harshithpabbati:
I would like to work on this

acao:
we are switching out state management, so a clear command will exist soon!

beerose:
Hi @acao! Any ETA on this redesign?

acao:
it is well underway! see pinned issues

beerose:
Yeah, I did. Also checked the milestone, but there's no due date, so I wanted to ask if you maybe have some rough estimation on the 2.0 beta.

PR #1400, opened 2020-03-03
feat: Add clear option

harshithpabbati:
Closes: #1388 

On click clears the query editor and variable editor

@acao can you review this


acao:
we will be replacing this entire toolbar, see the pinned ticket about the redesign! thank you though!

PR #1409, opened 2020-03-07
[reference] renderGraphiql Package

### LINKED ISSUE #1117: Discussion: `renderGraphiQL` and middlewares
What if we had a `renderGraphiQL` package that:

- wrapped react dom render and createElement for cdn users
- provides a simple function with a single configuration object to provide settings, some day plugins
- allows you to generate your own fetcher by providing URI, headers, etc
- is designed for the greater GraphQL community who isn't ready to use webpack to build a custom instance yet. 500K cdn hits a month!

then a bunch of middlewares for different node servers to mount it statically, or to serve it with a lambda

iFlameing:
Fixes #1117. This will help the CDN users to easily implement graphiql IDE without using any build tools like Webpack. They can perform all the actions they will intend to do without build tools.


packages/graphiql-renderGraphiQL-middlewares/renderGraphiQL.js
```diff
@@ -0,0 +1,105 @@
...
+  function onEditVariables(newVariables) {
+    parameters.variables = newVariables;
+    updateURL();
+  }
+  function updateURL() {
+    history.replaceState(null, null, locationQuery(parameters));
+  }
+
+  return ReactDOM.render(
+    React.createElement(GraphiQL, {
```
acao: this is looking great so far!

let's return ReactDOM.render() but use a normal component, rather than React.createElement(). For an example, you can [see my original concept](https://github.com/graphql/graphiql/blob/94d64b19f5fd4534a7ee0aa0d1f56b980a94401b/packages/graphiql/src/renderGraphiQL.js):

as far as the logic for handling query parameters, let's extract that into a seperate module as well, in this same package.

iFlameing: Ok!!!

iFlameing:
@acao I moved the functions in a separate module. can you take a look at it, please!!!!

acao:
@iFlameing looking better, now lets expose the options for generating the fetcher in my example i linked to. lets port over that logic and the typescript definitions

acao:
@iFlameing like this:
https://github.com/graphql/graphiql/blob/94d64b19f5fd4534a7ee0aa0d1f56b980a94401b/packages/graphiql/src/renderGraphiQL.js#L12

iFlameing:
@acao I provided the options for generating the fetcher and also added typescript. Please! take a look!!

acao:
@iFlameing looking excellent! I’ll take a closer look tonight. Do you want to create a render-graphiql-example as well? also lets make sure directory names are lower case, for better linux support

iFlameing:
@acao Yes! I want to create an example. Can you give me some starter points regarding how I can use it in an example?

iFlameing:
@acao I have created a simple example in javascript https://github.com/iFlameing/graphiqlCDN. I don't know how I can build this package so that I can use it in the HTML file.

acao:
@iFlameing my bad, i meant like in the `/examples` repository, so the example would be in this PR.

please just re-create from the webpack example, and use the encapsulation function instead

acao:
@iFlameing hey so, lets get this one revived! I think it's close, can you remove the package-lock.json? I'll see if i can fix the typescript issue you were having

acao:
@iFlameing can you give me permission to commit to this PR?

iFlameing:
@acao I am happy to do any help :) You can commit to this PR. If you need any help please ping me here or in discord :)

iFlameing:
@acao wait a sec I will do a commit for removing package-lock.json :)

acao:
It looks like there are still some `package-lock.json` files if you can clean those up!

iFlameing:
@acao Removed the package-lock.json :)

acao:
Nice! i've been working on it for a few hours, and I almost have the typescript working, and the type interfaces defined. there was a lot missing, because there was no tsconfig.json, the typescript was not working for the package at all - in vscode, etc. many of the functions didn't work at all, so I might add some basic unit tests as well?

acao:
so, theres just still a lot of work that needs to be done to get this ready, still working through linting issues, and i've stashed my work and will come back!

sibelius:
what is missing from this one?
I'd like to use the latest GraphiQL in my Koa app

acao:
@sibelius a lot of things, i spent a lot of time just trying to get it to pass the linter or compile.

most importantly, it has a major security vulnerability that we just fixed in playground

we just need to re-do this entire PR

iFlameing:
@acao can you push your changes so that I can take a look and learn the things which I am missing in this PR

acao:
@iFlameing first you will need to fix the original linting failures. you can see the list of fixes you need to make when you use a linter.

there are many more considerations but we will be going in a different direction

packages/graphiql-rendergraphiql-middlewares/renderGraphiQL.tsx
```diff
@@ -0,0 +1,53 @@
...
+  // Collect the URL parameters
+  let parameters = queryParameters();
+  let otherParams = otherParameters(parameters);
+  options.url = url + locationQuery(otherParams).toString();
+
+  if (!options.fetcher) {
+    options.fetcher = await getFetcher(options);
+  }
+
+  return ReactDOM.render(
```
acao, line 42: I think we need to render to a string here? it's hard to decide. or possibly we have a shared component, and have one version of this function that renders to DOM, and the other to string?

this is what I was thinking of using to render to string:
https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup

acao:
@sibelius if you're interested in helping to get this where it needs to be, let's start with opening a PR to fix the linting failures. @iFlameing you are also welcome to fix the linting errors here so we can move forward on this PR

then, let's evaluate the best method to do this. the playground middlewares suggest one route, but we have other options.  express-graphql has a nice approach as well. let's make sure whatever route we choose works offline instead of relying on CDN urls, as well

examples/graphiql-rendermiddleware/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
```
myselfop143, line 6: packages/graphiql-renderGraphiQL-middlewares/renderGraphiQL.js

myselfop143 (approved):
packages/graphiql-renderGraphiQL-middlewares/renderGraphiQL.js

dimaMachina:
closing due to obsolescence

PR #1417, opened 2020-03-12
feat: capture unknown commands making use of the inhouse s…

### LINKED ISSUE #1416: [CLI]: Handle unrecognized commands
`yargs` arg-parser library comes shipped with a `strict()` method as part of the API to handle unrecognized commands. We can leverage it to show up an appropriate message if the user passes in an unknown command.

jamesgeorge007:
fixes #1416


acao:
thank you @jamesgeorge007 !

PR #1420, opened 2020-03-13
feat(cli): recommend matching commmands

### LINKED ISSUE #1419: [CLI]: Recommend matching commands
`yargs` arg-parser library provides `recommendCommands()` function as part of the API to suggest the closest match to an unknown command passed in.

jamesgeorge007:
closes #1419


acao:
thank you!

PR #1427, opened 2020-03-14
feat: typescript, tsx, jsx support for LSP server using babel

acao:
- introduces typescript, tsx and jsx support via babel parser plugins
- #1263 by @zth is an alternative to solving this problem which also supports reasonml, and potentially other  a non-ast branch by @zth using regular expressions should  be reconsidered.
-tests for typecript and tsx added
- `getQueryAndRange` changed to `parseDocument`, and made override-able as `parse` by `startServer`.

### DIFF
--- a/packages/graphql-language-service-server/README.md
+++ b/packages/graphql-language-service-server/README.md
@@ -80,8 +80,9 @@ For each transport, there is a slight difference in JSON message format, especia
 | ---------------: | ---------------------------- | ------------------------------------------- |
 |      Diagnostics | `getDiagnostics`             | `textDocument/publishDiagnostics`           |
 |   Autocompletion | `getAutocompleteSuggestions` | `textDocument/completion`                   |
-|          Outline | `getOutline`                 | Not supported yet                           |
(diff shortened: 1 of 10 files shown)


divyenduz:
Does this do the same job as https://github.com/Quramy/ts-graphql-plugin? 
If yes, are there any usage docs that I can refer to? 🤔

acao:
yes! the only difference now is that it will handle ts/tsx files when you open them. still need to get vscode workspaces configured correctly to read all the embedded graphql

acao:
to add, the typescript language plugin is still valuable for a lot of cases.

zth:
This looks good to me! This will allow people who want to support other languages than the mainstream ones (like ReasonML for me) to do that through replacing `parseDocuments`.  👍

acao:
@zth last commit here fixes that bug i saw in discord, otherwise looks like everything is working fine! I've had some issues with setting up our example extension, but i think i understand what I'm missing after working with yours and @divyenduz 's more carefully.

PR #1421, opened 2020-03-13
Monaco Mode - Phase 1

acao:
For the first phase, we will only support completion and maybe diagnostics. Summarized more completely in #1445

**Note**: I will be pulling all the LSP changes out of this PR and creating seperate PRs to introduce those, so that this PR will soon be ~20 new files, a few changes and thats it, ideally.

Everything that isn't complete is partially underway. The only issue now is that monaco can't find the AMD id of our custom language worker.

# Outcomes
- introduces `packages/monaco-graphql` and `examples/monaco-webpack`
- Diagnostics (working), Completion Support (todo)
- Required baseline: schema loading, primitive config loading

# Todo
- [x] GraphQL Mode & contribution
- [x] GraphQL Worker Manager
- [ ] GraphQL Worker - written but worker manager will not load it
- [ ] GraphQL Language features
  - [ ] completion - mode providers work, worker doesnt
  - [ ] Diagnostics - mode providers work, worker doesnt
- [ ] Create graphql-config loaders for browser config load by default
- [ ] programatically provide graphql-config ala #1272?
- [ ] handle schema loading/changing schema via monaco service

Scoped to *only* completion for now. Other features will be iterative

# How to Develop/Test
1. check out this branch
2. from the root: `yarn --force`
3. from the root: (watcher) `yarn tsc --watch` (one time) `yarn build-ts`
4. in another terminal tab, from `examples/monaco-graphql`: `yarn start`

PR #1428, opened 2020-03-15
feat: Monaco graphql Completion

rebornix:
Re https://github.com/graphql/graphiql/pull/1421 

Had a good discussion offline with @acao about how to incrementally support GraphQL LSP in Monaco and we agreed to implement Completion first, with which we can test the runtime agnostic GraphQL LS and get a good understanding of integration between Monaco and GraphQL LS.

The code is still not demo-ready due to three blockers, which can be resolved later:

* runtime agnostic GraphQL LS
* `GraphQLCache` initialization
* Convert AST token kind to `CompletionItemKind` https://github.com/graphql/graphiql/pull/1428/files#diff-6d88bd46b01fdd44682f45d55e5202bdR59

### DIFF
--- a/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
@@ -15,6 +15,7 @@ import {
   GraphQLType,
   GraphQLCompositeType,
   GraphQLEnumValue,
+  Kind,
 } from 'graphql';
 
 import {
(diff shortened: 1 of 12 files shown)


acao:
awesome work, @rebornix ! thank you so much!

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -52,6 +53,34 @@ import {
   objectValues,
 } from './autocompleteUtils';
 
+// TODO@acao,rebornix
```
acao: ah i see, yes i dont think we need to take `step` into account, as this context isn't needed in this case

acao:
going to go ahead and merge this to the working branch! thanks a bunch

PR #1430, opened 2020-03-16
Feat: Added XML support to Graphiql

alinauroz:
**Added Support for XML Results**
Now graphiql can handle json as well as xml results. You have to choose response type from topbar using Handle JSON or Handle XML buttons.

**Using express-graphql-multimode**
Added options to send XML as well as json results to express-graphql and published it as express-graphql-multimode. It gets json and sends xml or json to client.


acao:
this looks great!

we have a feature freeze at the moment, as we are re-architecting GraphiQL entirely.  you'll be able to provide this as a plugin later. thanks for this!

alinauroz:
> this looks great!
> 
> we have a feature freeze at the moment, as we are re-architecting GraphiQL entirely. you'll be able to provide this as a plugin later. thanks for this!

The new Graphiql is using MonacoEditor instead of CodeMirror. So this plugin also needs some changes.

acao:
@alinauroz in fact, there are no plugins yet, so it will need a lot of changes! Stay posted. Thanks for this though!

alinauroz:
@acao you are right. This will need a lot of changes. While working on this feature, we also need to assure that all other things are right. For example, when graphiql gets schema in XML form, DOC Explorer fails to work. I wrote a code (using recursive approach) to convert xml schema to json so that doc explorer can work fine but this solution failed to work because schema is a large file (a monolith) and due to recursive approach the stack overflows.

To overcome this problem, i did some changes in express-graphql-multimode (an npm package that i published, by doing some changes in express-graphql, to test xml result view in graphiql). If data is schema, express-graphql-multimode sends result as json otherwise as XML. Check this pseduo code


````
if data has __schema
 sendAsJSON(data)
else if result_mode is XML
 sendAsXML(data)
else
 sendAsJSON(data)
````

alinauroz:
@acao Check this [proposal](https://github.com/alinauroz/gsoc-proposal/blob/master/proposal.md#project). It has a plan to add support for different kinds of formats to graphiql with timeline. Create issue to give feedback or comment [here](https://github.com/alinauroz/gsoc-proposal/issues/2)

PR #1438, opened 2020-03-22
fix: make sure that custom parser is used if passed to process

zth:
This ensures that a custom parser is used if passed to the server.

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLCache.ts
+++ b/packages/graphql-language-service-server/src/GraphQLCache.ts
@@ -56,6 +56,7 @@ export async function getGraphQLCache(
   configDir: Uri,
   extensions?: Array<(config: GraphQLConfig) => GraphQLConfig>,
   config?: GraphQLConfig,
+  parser?: typeof parseDocument,
 ): Promise<GraphQLCacheInterface> {
   let graphQLConfig =
     config ?? ((await loadConfig({ rootDir: configDir })) as GraphQLConfig);
@@ -64,7 +65,7 @@ export async function getGraphQLCache(
       graphQLConfig = await extension(graphQLConfig);
     }
   }
-  return new GraphQLCache(configDir, graphQLConfig);
+  return new GraphQLCache(configDir, graphQLConfig, parser);
 }
 
 export class GraphQLCache implements GraphQLCacheInterface {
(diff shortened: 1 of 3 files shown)


acao:
@zth looks great!
maybe we should add a test for this?

packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -75,15 +76,21 @@ export class GraphQLCache implements GraphQLCacheInterface {
...
+    parser?: typeof parseDocument,
+  ) {
     this._configDir = configDir;
     this._graphQLConfig = graphQLConfig;
     this._graphQLFileListCache = new Map();
     this._schemaMap = new Map();
     this._fragmentDefinitionsCache = new Map();
     this._typeDefinitionsCache = new Map();
     this._typeExtensionMap = new Map();
+    this._parser = parser || parseDocument;
```
acao: we should probably only supply the default in one place here, and remove the || in `MessageProcessor`. Also i think we should use `??` here.

zth: Fixed!

zth:
I think tests would be great for sure, but there are no tests at that level right now from what I can see, and I think it'd be a bit involved to add it. Seeing as this fixes a feature that's currently broken, could we have this merged before tests are added?

packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -90,7 +90,7 @@ export class GraphQLCache implements GraphQLCacheInterface {
     this._fragmentDefinitionsCache = new Map();
     this._typeDefinitionsCache = new Map();
     this._typeExtensionMap = new Map();
-    this._parser = parser || parseDocument;
+    this._parser = parser;
```
acao: I was thinking that the default should be set here, rather than in `MessageProcessor`

zth: Looking at it, I'm not sure that makes sense. `addHandlers` is instantiating `MessageProcessor`, which in turn is instantiating the `GraphQLCache`. But `MessageProcessor` is also using the parser, so it needs a value set, which means it needs to set a default, or force providing a parser.

Issue #1439, opened 2020-03-24
GraphiQL `onEditQuery` is called without query argument after TypeScript migration


cshaver:
I'm updating our installation of graphiql to `1.0.0-alpha.1`, and thanks to the new TypeScript types (🎉) I was was really quickly made aware that the type signature of the function passed to `GraphiQL` as the `onEditQuery` prop has changed:

![image](https://user-images.githubusercontent.com/2136754/77381603-947dec80-6d3b-11ea-885f-2249f8cef858.png)

This breaks our usage of graphiql-explorer. (see example setup  which uses `onEditQuery` https://github.com/OneGraph/graphiql-explorer-example/blob/master/src/App.js#L189)

It looks like this change was made in the [recent TypeScript migration](https://github.com/ryan-m-walker/graphiql/commit/2de0f44d3851a76b070417a48c69ff4050f8674b#diff-d9161d32c36caed395ae91f2856c023eL948).

I understand that this prop is likely going to be deprecated in the new API per #1165, but it looks to me like this change was unintentional (`onEditQuery` is unchanged in the README, and per https://github.com/graphql/graphiql/pull/1242#issue-364380842 that this should have minimal production changes).

acao:
this might be a good candidate for a hotfix to backport

acao:
or wait, so this is only an issue in alpha, as 0.17.5 was before the typescript migration

cshaver:
@acao right! I had the issue in 1.0.0-alpha.1/3, so no need to backport

acao:
great! thanks for this, we can get another alpha out here shortly. I'll close this once it's been released. glad the new typescript types are working out for you :smile:

acao:
pre-release is out, find 1.0.0-alpha.4 :)

acao:
use latest alpha btw! @cshaver lemme know if you have any issues

PR #1440, opened 2020-03-24
fix: fix query argument missing from onEditQuery call

cshaver:
Fix for https://github.com/graphql/graphiql/issues/1439


acao:
@cshaver thank you!

PR #1441, opened 2020-03-27
feat: make sure @ triggers directive completion automatically

zth:
This adds `@` as an explicit trigger character for code completion in the LS. What this does is that it brings up the autocomplete list as soon as you type "@", rather than you needing to manually activate autocomplete (ctrl+space) to autocomplete directives.


acao:
nice! we will have to figure out how to do this in monaco as well. i wonder if we could achieve this in getAutocmpleteSuggestions in the language service itself?

PR #1449, opened 2020-04-02
chore: upgrade graphql-config

zth:
This upgrades `graphql-config` to their first `rc` version, which contains a feature I really want to use 👼


acao:
@zth awesome thanks! does everything still work? did you test this PR with your local config?

PR #1450, opened 2020-04-02
build: add missing include lists

amcasey:
`rootDir` doesn't imply `include`, so output files were being incorrectly included as inputs.  Besides being less efficient, this made it necessary to clean before each TypeScript build.

### DIFF
--- a/packages/graphql-language-service-interface/tsconfig.esm.json
+++ b/packages/graphql-language-service-interface/tsconfig.esm.json
@@ -16,5 +16,6 @@
       "path": "../graphql-language-service-utils"
     }
   ],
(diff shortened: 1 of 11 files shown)


acao:
thanks for this @amcasey ! I had addressed this in another unrelated in progress PR for monaco, but this is an even better way to fix it. it'll be so nice to not have to clean between builds, thanks!

amcasey:
@acao I didn't want to mess with it, but you might also consider updating your build script: https://github.com/graphql/graphiql/blob/master/package.json#L24.

acao:
yep, thats what i was referring to in my last comment

PR #1459, opened 2020-04-04
feat: Monaco Mode - Phase 2 - Mode & Worker

acao:
For the second phase, we will implement a monaco mode and worker, and we only support completion and maybe diagnostics. Summarized more completely in #1445

# Outcomes
- introduces `packages/monaco-graphql` and `examples/monaco-webpack`
- Diagnostics (working), Completion Support (todo)
- Required baseline: schema loading, primitive config loading

# Todo
- [x] GraphQL Mode & contribution
- [x] GraphQL Worker Manager
- [x] GraphQL Worker - written but worker manager will not load it
- [x] GraphQL Language features
  - [x] completion - mode providers work
  - [x] Diagnostics - mode providers work

# Another PR?
- [ ] Create graphql-config loaders for browser config load by default
- [ ] programatically provide graphql-config ala #1272?
- [ ] handle schema loading/changing schema via monaco service

Scoped to *only* completion for now. Other features will be iterative

# How to Develop/Test
1. check out this branch
2. from the root: `yarn --force`
3. from the root: (watcher) `yarn tsc --watch` (one time) `yarn build-ts`
4. in another terminal tab, from `examples/monaco-graphql`: `yarn start`

PR #1460, opened 2020-04-04
Hookup Completions with Monaco Editor

rebornix:
The completions items from the language service are now returned back to the editor.


acao:
already in the new PR! #1459

PR #1462, opened 2020-04-05
feat: schema provider and operation session context using context/useReducer

ryan-m-walker:


packages/graphiql/src/state/actions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type OperationSucceededAction = {
+  type: ActionTypes.OperationSucceeded;
+  payload: SuccessPayload;
+};
+
+export type OperationErroredAction = {
+  type: ActionTypes.OperationErrored;
+  payload: ErrorPayload;
+};
```
cshaver, line 54: I like that the action types are defined explicitly here! i tend to do this [indirectly with actions defined as functions](https://github.com/graphql/graphiql/pull/1463/files#diff-89da5d5b80d79f61b86bc31ea55c5555R17-R22) but this reads nicer

ryan-m-walker: haha i was thinking I liked the way you did it because you don't have to write them all out and it can be a bit more concise, but I guess it's also helpful to have them explicitly defined so if you go inspect the the definition you actually show up at a type instead of a function or something

packages/graphiql/src/state/sessionActions.ts
```diff
@@ -0,0 +1,104 @@
...
+  type: SessionActionTypes.OperationRequested,
+});
+
+export type EditorLoadedAction = {
+  type: SessionActionTypes.EditorLoaded;
+  payload: {
+    context: EditorContexts;
+    editor: CodeMirror.Editor;
+  };
+};
```
cshaver, line 38: re: https://github.com/graphql/graphiql/pull/1462#discussion_r403643878

how 'bout something like this?
```suggestion
export type EditorLoadedAction = ReturnType<typeof editorLoadedAction>;
```

ryan-m-walker: @cshaver yeah i do think i like that better

PR #1359, opened 2020-02-21
feat:  schema provider and operation session context using context/useReducer

acao:
please take this and run with it tomorrow however you see fit!
it introduces:

- `useReducers` which is meant to expose interfaces (to plugins eventually down the line, but for now we will supply them as props via the provider). glad to field suggestions on how to re-architect this!
- a few `useReducer` related utilities inspired from previous redux workflows
- a test utility for RTL, for testing provider output more readily
- `SchemaProvider`, `SchemaContext`, etc. feel free to combine this provider into a migration provider like you were discussing earlier @zephraph 

typescript types are not happy yet because I'm still wrapping my head around typescript + redux style patterns

note:

some features are temporarily disabled by this effort in the interest of further iterations before we release another alpha:

- subscriptions support
- multiple operations per query
- persisted operations (though your previous query history works in the left sidebar, soon to be deprecated)

PR #1463, opened 2020-04-05
[wip] feat: schema provider and operation session context using context/useReducer

cshaver:


packages/graphiql/src/state/GraphiQLSessionProvider.tsx
```diff
@@ -90,120 +98,117 @@ export const SessionContext = React.createContext<
...
+  const changeOperation = useDispatchedCallback(
+    dispatch,
+    changeOperationAction,
+  );
+  const changeVariables = useDispatchedCallback(
+    dispatch,
+    changeVariablesAction,
+  );
+
+  const executeOperation = React.useCallback(
```
ryan-m-walker, line 188: i'm not as strong with the `useCallback` hooks but would it be possible to instead have the `sessionState` argument being passed in here, we could just have the useCallback watch the props it cares from from the state in this scope? see my comment here on the base branch for more context on what i'm trying to get at:

https://github.com/graphql/graphiql/pull/1359/files/97f347f5946126e75961568b663f918ddb6530ef#r403618312

ryan-m-walker: like could we do something like

``` TypeScript
const executeOperation = React.useCallback(
   async (operationName?: string) => {
    // ...
}, [state.operation, state.variables])
```

packages/graphiql/src/state/GraphiQLSessionProvider.tsx
```diff
@@ -90,120 +98,117 @@ export const SessionContext = React.createContext<
...
 
 export type SessionProviderProps = {
   sessionId: number;
   fetcher?: Fetcher;
   session?: SessionState;
-  children: any;
+  children: React.ReactNode;
 };
 
 export function SessionProvider({
```
acao, line 164: for `SessionProvider`, wondering if we should (eventually) consume the three editor instances, as we will use them for handling change events that way with monaco. each editor has `editor.onDidChangeContent` or some such, so we can use those when we export the handlers for operation/variables change and such?

packages/graphiql/src/state/EditorProvider.tsx
```diff
@@ -0,0 +1,87 @@
...
+      return state;
+    }
+  }
+}
+
+type EditorContextType =
+  | (EditorState & {
+      context: EditorContexts;
+      onLoadEditor: (...args: Parameters<typeof editorLoadedAction>) => void;
+      onChange: (...args: Parameters<typeof editorChangeAction>) => void;
```
acao, line 57: both this and `onLoadEditor` would be required, as in the case of results we have a "controlled component" type situation, where we don't care about `onChange` because it's static. possibly we can have a `controlled={false}` by default for the `EditorProvider`?

packages/graphiql/src/state/GraphiQLSessionProvider.tsx
cshaver: totally agree! I hadn't looked at the signature of this function yet, good call

packages/graphiql/src/state/GraphiQLSessionProvider.tsx
cshaver: absolutely! I'll tinker on consuming these today

packages/graphiql/src/state/EditorProvider.tsx
cshaver: would a `readOnly` prop cover that you think?

cshaver:
Closing this one to reopen after rebasing onto #1462

PR #1475, opened 2020-04-10
Feat: Add URI Prop

ryan-m-walker:
This PR adds a `uri` prop to `GraphiQL`. The user can now either pass just a string URI, in which case GraphiQL will use the default fetcher, or they can provide their own fetcher if they want more control. If neither are provided GraphiQL will throw

This PR also gets rid of the `schemaLoader` function an just uses the fetcher now instead


acao:
@ryan-m-walker give yourself a good lockfile update and yarn run if you can! maybe the feature branch has an out of date lockfile? looks like thats what failing in CI

PR #1481, opened 2020-04-10
fix: hover log

divyenduz:
If there was an error in schema parsing, the hover handler silently
fails.

The aim of this PR is to raise the issue, the fix might be added at
another level (and maybe even needed at other places).

For this PR, we re-throw the error.

After this PR, invalid schema yields: 
CleanShot 2020-04-11 at 03 14 10@2x

P.S. the tests are much faster now! All of them took like 18s on my machine 🙌

### DIFF
--- a/packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
@@ -243,7 +243,9 @@ export class GraphQLLanguageService {
     const projectConfig = this.getConfigForURI(filePath);
     const schema = await this._graphQLCache
       .getSchema(projectConfig.name)
-      .catch(() => null);
+      .catch(e => {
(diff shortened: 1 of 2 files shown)


acao:
@divyenduz we should wrap all the errors at the top, then! 

the schema error being trapped was there from before so I kept it, it would make so much sense to remove, but i noticed it during a refactor, so I wasn't able to remove it then.

this is why i added a try/catch to startServer() but i removed it. let's remove the try/catch from hover, and bubble it up to where this function is called, so we can trap all the errors from LSP methods in one place.

acao:
@divyenduz this approach should work across the board:
https://github.com/graphql/graphiql/pull/1482/files

acao:
closing for #1482! confirmed it fixes this bug and many others. thanks @divyenduz !

PR #1482, opened 2020-04-11
fix: repair CLI, handle all schema and LSP errors

acao:
This should do what @divyenduz set out to do in #1481 more universally. Any exception in any of the handlers bubbles up to the GLS output channel now.

It also repairs the CLI. It was targeting `esnext` by default, thus a use of nullish coalescing was breaking the CLI (not supported natively in node yet). 

So, I set it to `target: es2017` for node 10 support.

Additionally, CLI should default method to `node` as advertised.

Tested with both the CLI using validate command, and `vscode-graphql` - the only issue there is the custom prisma extension is throwing an error. the extension works great if you remove that extension. I think it just needs to be upgraded for the new graphql-config?

### DIFF
--- a/packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-interface/src/GraphQLLanguageService.ts
@@ -125,6 +125,9 @@ export class GraphQLLanguageService {
     // schema/fragment definitions, even the project configuration.
     let queryHasExtensions = false;
     const projectConfig = this.getConfigForURI(uri);
+    if (!projectConfig) {
+      return [];
+    }
     const { schema: schemaPath, name: projectName, extensions } = projectConfig;
 
     try {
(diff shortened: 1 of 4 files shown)


acao:
@divyenduz should probably add a few tests as well?

acao:
@huv1k I saw you were interested in the first PR, do you want to give this one a test run? 

This should solve the problem for all handlers and other errors (hover, completion, diagnostics, etc). I added this to a PR a while back where we were trying to add more error handling, because apparently there wasn't much before, but there weren't enough people to review the PR at the time so I removed it to make it simpler.

To help me out, could someone open a PR to add tests to this? If not I can work on it some more tomorrow. Have a lot of GraphiQL and Monaco things going on, making lots of changes to the LSP!

acao:
have a few more improvements! dang the LSP server has some issues :(

acao:
@divyenduz this takes care of a few things now!

PR #1494, opened 2020-04-14
improvement: don't trap these other schema loading errors

acao:
addresses some outstanding exception trapping i spotted, related to #1481 and #1482. @divyenduz @zth what do you think?


zth:
@acao excellent! 👍

PR #1498, opened 2020-04-14
refactor: friendlier to read diagnostic codes

acao:
and lets use a constant too, make refactoring easier!

### DIFF
--- a/packages/graphql-language-service-interface/src/getDiagnostics.ts
+++ b/packages/graphql-language-service-interface/src/getDiagnostics.ts
@@ -30,11 +30,24 @@ import {
 
 import { DiagnosticSeverity, Diagnostic } from 'vscode-languageserver-types';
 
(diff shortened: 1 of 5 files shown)


packages/graphql-language-service-interface/src/getDiagnostics.ts
```diff
@@ -31,10 +31,21 @@ import {
 import { DiagnosticSeverity, Diagnostic } from 'vscode-languageserver-types';
 
 export const SEVERITY = {
-  ERROR: 1 as DiagnosticSeverity,
-  WARNING: 2 as DiagnosticSeverity,
-  INFORMATION: 3 as DiagnosticSeverity,
-  HINT: 4 as DiagnosticSeverity,
+  Error: 'Error' as 'Error',
```
acao, line 34: @rebornix i think this was even more along the lines of what you were thinking of?

Issue #1501, opened 2020-04-15
docs: re-introduce typedoc deploy using netlify/actions github actions


acao:
we used to publish at our deploy preview/lsp, however I think just publishing on each merge to master is fine for the time being. we were having issues that were slowing down the netlify build. i would like to break out the netlify config into seperate actions using the cli action. the root package.json script is just `build-docs`

daemon1024:
Hey @acao , I would like to help with this issue

> we used to publish at our deploy preview/lsp, however I think just publishing on each merge to master is fine for the time being. 

`deploy-preview` generates a deploy from a pull request or merge request so we are already doing that i guess based on https://github.com/graphql/graphiql/blob/91df0841d381803527b7755f493984c532d6d3e5/netlify.toml#L6

So it is unclear to me , what i need to do, can explain it to me.

acao:
awesome! glad to have help with this

yes, previously we were doing this, however we would like to take a new approach now. we don't want to slow down the deploy previews any further, as they were ~8-10m with typedoc, possibly because of configuration issues that have been addressed, but it still adds ~2-3m to the build and breaks CI stability which as we all know is really helpful for highly collaborative projects. now they seem closer to 6m which is better.

as you can see from recent action logs (which should be public, are they?), moving from jenkins to github actions allowed us to parralelize CI tasks better and increase overall speed down to 2 minutes.  

we are hoping to move more of this to github actions, as not all of what happens on netlify build for build preview

so, for these things that don't need to be generated on every commit for the deploy-preview, we can use [netlify's own action](https://github.com/netlify/actions/tree/master/cli) to be able to use netlify cli commands and actions to publish the typedoc and other things that have been a part of deploy preview to parallel actions.

the other advantage is that we can deploy the API docs to a seperate site than the deploy preview, and we will probably create even more sites that are generated by this repo soon.

daemon1024:
@acao Got it. 

> as you can see from recent action logs (which should be public, are they?), moving from jenkins to github actions allowed us to parralelize CI tasks better and increase overall speed down to 2 minutes.

Yup they are public and I can see the build times being 2-3 min.

> we are hoping to move more of this to github actions, as not all of what happens on netlify build for build preview
> 
[...]

PR #1506, opened 2020-04-18
Add netlify action to github-workflows

daemon1024:
Ref: #1501

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: bahmutov/npm-install@v1
-      - run: yarn build && yarn build-bundles && yarn build-demo
+      - run: yarn build-ts && yarn build-demo
 
       - name: Deploy to netlify
         uses: netlify/actions/cli@master
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -1,0 +1,21 @@
+name: typedoc deploy to netlify
+
+on: pull_request
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: bahmutov/npm-install@v1
+      - run: yarn build && yarn build-bundles && yarn build-demo
+
+      - name: Deploy to netlify
+        uses: netlify/actions/cli@master
+        env:
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.STAGING_NETLIFY_SITE_ID }}
+        with:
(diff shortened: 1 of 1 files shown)


acao:
awesome work! there's no need for `yarn build-bundles` or possibly even `yarn build` if I understand how typedoc works correctly? either way, `yarn build-ts && yarn build-demo` is probably all we need for a clean build, as `yarn build` also builds codemirror which is needed for demos/bundles, but not for typedoc

daemon1024:
> there's no need for `yarn build-bundles` or possibly even `yarn build` 

I just based it on the current `netlify.toml`
> `yarn build-ts && yarn build-demo` is probably all we need for a clean build

Got it. Will change it. 

Also should I remove the `netlify.toml` is it needed anymore?

acao:
@daemon1024 the netlify.toml should remain the same for now, as it's used for the deploy preview and master build. after this PR is when we would revisit the `netlify.toml` deploy preview steps and see what we can improve there.

this action will fail on merge because the environment secrets, however I can just add those myself, do I have commit rights on this branch?

daemon1024:
> after this PR is when we would revisit the `netlify.toml` deploy preview steps and see what we can improve there.

:+1: 
> this action will fail on merge because the environment secrets, however I can just add those myself

You can set secrets for your repository by going to [`Setting -> Secrets`](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) .

> do I have commit rights on this branch? 

I guess not, i can provide you with the access if needed?

acao:
yes please, and then I can complete this PR and we will be on our way!

acao:
oh you don't need to invite me to the whole repo, just check the box in this PR in the bottom right to allow maintainers access

daemon1024:
> oh you don't need to invite me to the whole repo, just check the box in this PR in the bottom right to allow maintainers access

you already have that access then :)

daemon1024:
Something more is needed with `yarn build-ts` ? I see the workflow check failing , after I added that!

acao:
oh and we have to change the `on` condition too, oops. now its gonna deploy to the root subdomain on each commit, haha yikes

daemon1024:
only deploy on master branch? 

      on:
          pull_request:
            branches:
              - master

acao:
yes. theres an issue and I'll have to come back to this later. thanks for all your help!

daemon1024:
> yes. 

will add it then.
>theres an issue and I'll have to come back to this later. 

No problem
>thanks for all your help!

Happy to help. Let me know if I can help with something else.

daemon1024:
> there's no need for `yarn build-bundles` or possibly even `yarn build` if I understand how typedoc works correctly? either way, `yarn build-ts && yarn build-demo` is probably all we need for a clean build, as `yarn build` also builds codemirror which is needed for demos/bundles, but not for typedoc

i guess the run command should be ` yarn build-ts && yarn build-docs `

PR #1508, opened 2020-04-18
ci: add danger file to ci

acao:
This will be to check for common issues folks have when creating PRs


.github/workflows/ci.yml
```diff
@@ -4,10 +4,13 @@ jobs:
   test:
     name: Lint & Test
     runs-on: ubuntu-16.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v1
       - uses: bahmutov/npm-install@v1
       - run: yarn ci && node_modules/.bin/codecov
+      - uses: danger/danger-js@9.1.8
```
orta, line 13: I'd recommend against the GitHub action and use yarn, maybe I should just outright remove the action - I think it's a bad abstraction and probably the reason why you have weird errors

acao: good to know! good idea too, yeah. an action really doesn't even make sense at this, I hadn't thought about collecting test coverage when it started. I'll just run it after codecov. thanks!

acao:
hello @manh-vv! these are changes to CI, so the changes are reflected in the execution of the CI checks you see right here in this PR. as you can see, I'm adding the `danger` package script to the end of the `ci` package script we execute for the first CI github action, "lint and test".

acao:
if you would like to help work on the GraphiQL react app, we have an open PR for the rewrit and there are lists of things for folks to work on there!

acao:
@orta finally got to some relevant errors, it seems theres an issue where it cant find @typescript-eslint plugin although thats working fine for us for yarn lint. thoughts?

orta:
I've never used the plugin, so unsure - I generally recommend running the tools lint separately and reading the results JSON

PR #1468, opened 2020-04-06
feat: [RFC] Rewrite for monaco editor, react context and the beginnings of the redesign

acao:
moved to a non forked branch, #1523

PR #1521, opened 2020-05-11
feat: monaco-graphql docs, api, improvements

acao:
These are most of the monaco and LSP changes pulled from #1468. This will make #1468 a lighter PR and allow folks to keep moving this language mode forward as we deliberate on the new GraphiQL RFC

- configure/change schema options using API (currently static and non configurable in master)
- add more detailed `monaco-graphql` docs
- move language worker to it's own file
- new `monaco.languages.graphql.api` class
- configure formatting options (prettier config!) using API
- RuleKinds implemented further, improving parser DX
- improve language service's schema loading mechanisms and API
- restores `typedoc` for deploy preview/master build. just append `/typedoc` to preview url
- add schema url bar to monaco example for switching schemas

## caveats

- this is for only a single schema per language service instance. to follow the example of [`monaco-yaml`](https://github.com/pengx17/monaco-yaml/blob/d64a8d55382aeccede336313e9d4134a5f149aa7/src/languageservice/services/yamlSchemaService.ts), we would need to be able to retrieve schema per file URI in the workspace, similar to how it's done with graphql config in the LSP server.
- yes the example is ugly. make it prettier if you want! i'm just using it to test the APIs.

## links
- [new `monaco-graphql` readme](https://github.com/acao/graphiql/blob/feat/monaco-docs-api/packages/monaco-graphql/README.md)
- [live demo](https://deploy-preview-1521--graphiql-test.netlify.app/monaco/)
- [typedocs for `monaco-graphql`](https://deploy-preview-1521--graphiql-test.netlify.app/typedoc/modules/monaco_graphql.html)

(API patterns are following the patterns used by other monaco modes, such as `monaco-json`, `monaco-typescript`, `monaco-html`, etc)

monaco-schema-change

How to Demo:

1. go to `examples/monaco-graphql-webpack`
2. run `yarn start`

OR

deploy preview url plus `/monaco`

### DIFF
--- a/packages/monaco-graphql/src/api.ts
+++ b/packages/monaco-graphql/src/api.ts
@@ -1,0 +1,19 @@
+import * as languageFeatures from './languageFeatures';
+import type { DocumentNode } from 'graphql';
+import { SchemaResponse } from 'graphql-languageservice';
+
+export class MonacoGraphQLApi {
+  private worker: languageFeatures.WorkerAccessor;
+  constructor({ accessor }: { accessor: languageFeatures.WorkerAccessor }) {
+    this.worker = accessor;
+  }
+
+  async getSchema(): Promise<SchemaResponse> {
+    const langWorker = await this.worker();
+    return langWorker.getSchemaResponse();
+  }
+  async parse(graphqlString: string): Promise<DocumentNode> {
+    const langWorker = await this.worker();
+    return langWorker.doParse(graphqlString);
+  }
+}
--- a/packages/monaco-graphql/src/api.ts
+++ b/packages/monaco-graphql/src/api.ts
@@ -1,19 +1,16 @@
 import * as languageFeatures from './languageFeatures';
-import type { DocumentNode } from 'graphql';
-import { SchemaResponse } from 'graphql-languageservice';
-
 export class MonacoGraphQLApi {
   private worker: languageFeatures.WorkerAccessor;
   constructor({ accessor }: { accessor: languageFeatures.WorkerAccessor }) {
     this.worker = accessor;
   }
 
-  async getSchema(): Promise<SchemaResponse> {
+  getSchema = async () => {
     const langWorker = await this.worker();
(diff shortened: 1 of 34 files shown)


acao:
@rebornix what do you think of the `monaco.languages.graphql` interfaces I added?

packages/monaco-graphql/src/api.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { DocumentNode } from 'graphql';
+import { SchemaResponse } from 'graphql-languageservice';
+
+export class MonacoGraphQLApi {
+  private worker: languageFeatures.WorkerAccessor;
+  constructor({ accessor }: { accessor: languageFeatures.WorkerAccessor }) {
+    this.worker = accessor;
+  }
+
+  async getSchema(): Promise<SchemaResponse> {
```
cshaver, line 11: I don't think `this` is properly bound here (got an error when I had it linked)
```suggestion
  getSchema = async () => {
```

packages/monaco-graphql/src/api.ts
```diff
@@ -0,0 +1,19 @@
...
+    const langWorker = await this.worker();
+    return langWorker.getSchemaResponse();
+  }
+  async parse(graphqlString: string): Promise<DocumentNode> {
```
cshaver, line 15:
```suggestion
  parse = async (graphqlString: string) => {
```

packages/monaco-graphql/src/api.ts
cshaver: Oh! I just remembered, that was because I called it from a global variable in the dev console 😆 Consider this a nit, haha

acao: ah! it's still a good pattern

acao:
@cshaver can you confirm for posterity that you've given this a whirl locally and it all works? i recall you referring to something as such

PR #977, opened 2019-10-08
chore: Full support i18n (i18next) excluding localization services

Xambey:
I implemented full localization support based on the i18next plugin. Two languages are currently supported:
- English
- Russian

P.S For verification, don't forget to clear localstorage

PR #1526, opened 2020-05-13
feat: Add support for i18n

harshithpabbati:
Screenshot 2020-05-08 at 5 35 59 PM

Thanks @Xambey for your [PR](https://github.com/graphql/graphiql/pull/977).


harshithpabbati:
@acao can you review this ?

packages/graphiql/package.json
```diff
@@ -16,6 +16,7 @@
   "module": "esm/index.js",
   "files": [
     "dist",
+    "src/locales",
```
acao, line 19: hmm, is there a reason these aren't being built I wonder? maybe we should configure webpack to copy them so that we can ensure relative resolutions work. overall I'm averse to the idea of publishing src to npm unless its an un-compiled package

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -120,26 +123,28 @@ export type GraphiQLState = {
  */
 export const GraphiQL: React.FC<GraphiQLProps> = props => {
   if (!props.fetcher && !props.uri) {
-    throw Error('fetcher or uri property are required');
+    throw Error(i18n.t('Errors:GraphiQL requires a fetcher function'));
```
acao, line 126: this should be 'fetcher or uri property are required' for now. we can leave out that locale string from the `ru` translations for now until someone can give it another pass

packages/graphiql/src/locales/ru/Errors.json
```diff
@@ -0,0 +1,7 @@
+{
+  "GraphiQL requires a fetcher function": "GraphiQL запрашивает функцию выборки",
```
acao, line 2: and we can just remove this line, or make the righthand value the english for the new error until we get a new translation for this locale string

packages/graphiql/tsconfig.esm.json
```diff
@@ -11,7 +11,7 @@
       "src/*": ["*"]
     }
   },
-  "include": ["src"],
+  "include": ["src", "src/**/*.json"],
```
acao, line 14: `.json` files should always be included if `resolveJSONModules` is enabled, right?

resources/tsconfig.base.cjs.json
```diff
@@ -5,6 +5,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "esModuleInterop": true,
+    "resolveJsonModule": true,
```
acao, line 8: :+1:

acao:
overall, this is looking fantastic! thanks so much @harshithpabbati

harshithpabbati:
ok cool! I will make the required changes.

packages/graphiql/tsconfig.esm.json
harshithpabbati: AFAIK it works even if we don't include it

harshithpabbati:
@acao updated the PR

PR #1536, opened 2020-05-19
[RFC, Not ready for merge] Refactored Language Service API to be transport agnostic

ncthbrt:
This is a first pass at improving the LanguageService API. 

My main intention here was to:
- Add the ability to execute queries from within the language service
- Improve type safety 
- Remove conflating concerns from the Language Service. For example the URI and request options were baked into the Language Server config but would only apply to a subset of possible schema fetchers. 

Configuration here is performed on the schema fetcher and executors using a closure, instead of passing the config each time. 

This PR is not ready for merge as it'd break the GraphQL worker. Additional work will be required to bring the above benefits to configuration of the worker.

Would love to get some feedback on these preliminary changes.

### DIFF
--- a/packages/graphql-languageservice/src/LanguageService.ts
+++ b/packages/graphql-languageservice/src/LanguageService.ts
@@ -4,7 +4,14 @@
  *  This source code is licensed under the MIT license found in the
  *  LICENSE file in the root directory of this source tree.
  */
-import { parse, GraphQLSchema, ParseOptions, ValidationRule } from 'graphql';
+import {
+  parse,
+  GraphQLSchema,
+  ParseOptions,
+  ValidationRule,
+  BuildSchemaOptions,
+  IntrospectionOptions,
+} from 'graphql';
 import type { Position } from 'graphql-language-service-types';
 import {
   getAutocompleteSuggestions,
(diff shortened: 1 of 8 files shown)


packages/graphql-languageservice/src/LanguageService.ts
```diff
@@ -106,12 +111,7 @@ export class LanguageService {
     _uri: string,
     documentText: string,
     customRules?: ValidationRule[],
-  ) => {
-    if (!documentText || documentText.length < 1) {
-      return [];
-    }
-    return getDiagnostics(documentText, await this.getSchema(), customRules);
-  };
+  ) => getDiagnostics(documentText, await this.getSchema(), customRules);
```
ncthbrt, line 114: Not sure why this was taken out. I'll add it back

packages/graphql-languageservice/src/LanguageService.ts
```diff
@@ -13,45 +20,41 @@ import {
...
+    buildSchemaOptions,
+  }: GraphQLLspConfig) {
     if (parser) {
       this._parser = parser;
     }
-    if (schemaLoader) {
-      this._schemaLoader = schemaLoader;
-    }
-    if (schemaBuilder) {
-      this._schemaBuilder = schemaBuilder;
```
ncthbrt, line 53: Not sure why this is gone either. Will also add it back.

packages/graphql-languageservice/src/LanguageService.ts
acao: @ncthbrt maybe try a rebase? probably because you started this PR before I merged an upstream change, oops!

packages/graphql-languageservice/src/LanguageService.ts
```diff
@@ -73,24 +76,26 @@ export class LanguageService {
...
-      this._schemaConfig,
-    )) as SchemaResponse;
+    this._schemaResponse = (await this._schemaLoader()) as SchemaResponse;
     return this._schemaResponse;
   }
 
   public async loadSchema() {
     const schemaResponse = await this.loadSchemaResponse();
-    this._schema = this._schemaBuilder(
+    this._schema = buildSchemaFromResponse(
```
acao, line 85: need to change this back to `this.schemaBuilder`

packages/graphql-languageservice/src/http/HttpLanguageService.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  if (endpointConfig) {
+    return createHttpQueryExecutor({
+      uri: endpointConfig.uri,
+      requestOpts: endpointConfig.requestOpts,
+    });
+  }
+}
+
+export class HttpFetcherLanguageService extends LanguageService {
```
acao, line 55: huh. not a huge fan of extending classes at all. also, we should only have a single language service class

ncthbrt: We could probably then move it back, just with more sophisticated typings of the config? @acao

acao: sure, though I'm not sure what further sophistication is needed, it's hard to see what you are referring to?

acao:
@ncthbrt awesome work all around!

PR #1538, opened 2020-05-19
improvement: add resizer component

harshithpabbati:
This PR adds a new Resizer component. So we will be able to resize the editors. We can simply wrap the editor with resizer component like this

```ts
<Resizer
  border="bottom"
  handlerStyle={{ backgroundColor: 'rgba(0, 0, 0, 0.2)' }}
>
  <main>{`Main content`}</main>
</Resizer>
```

This above example will add a resizer option below the `main` tag and we can also pass the required styles to it.

@acao can you review this

### DIFF
--- a/packages/graphiql/src/components/common/Resizer/ResizeHandler.tsx
+++ b/packages/graphiql/src/components/common/Resizer/ResizeHandler.tsx
@@ -1,0 +1,84 @@
+/** @jsx jsx */
+import { jsx } from 'theme-ui';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { ResizeHandlerData, ResizeHandlerProps } from './types';
+
+export const ResizeHandler: React.FC<ResizeHandlerProps> = props => {
+  const { dir, onStart, onEnd, onUpdate, className, style, children } = props;
+
+  const [isDragging, setIsDragging] = useState(false);
+  const listenersRef = useRef<ResizeHandlerData['listenersRef']>(null);
+
+  const handleMouseMove = useCallback(
+    e => {
+      onUpdate(e);
+    },
+    [onUpdate],
+  );
+
+  const cleanMouseListeners = useCallback(() => {
+    const oldRef = listenersRef.current;
+    if (oldRef) {
+      window.removeEventListener('mousemove', oldRef.handleMouseMove);
+      window.removeEventListener('touchmove', oldRef.handleMouseMove);
+      window.removeEventListener('mouseup', oldRef.handleMouseUp);
+      window.removeEventListener('touchend', oldRef.handleMouseUp);
+    }
+  }, []);
+
+  const handleMouseUp = useCallback(
+    e => {
+      setIsDragging(false);
+      cleanMouseListeners();
+      onEnd(e);
+    },
+    [cleanMouseListeners, onEnd],
+  );
+
+  const handleMouseDown = useCallback(
+    e => {
+      setIsDragging(true);
+      cleanMouseListeners();
+
+      listenersRef.current = {
+        handleMouseMove,
+        handleMouseUp,
+      };
+      window.addEventListener('mousemove', handleMouseMove);
+      window.addEventListener('touchmove', handleMouseMove);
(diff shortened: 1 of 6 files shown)


acao:
great!

I've configured netlify actions for `feat/use-context-hooks`, so if you push another commit, it should trigger a netlify deploy preview!

acao:
https://deploy-preview-1538--graphiql-test.netlify.app/storybook check it out!

harshithpabbati:
> https://deploy-preview-1538--graphiql-test.netlify.app/storybook check it out!

Yeah, I saw it. Can you review the PR @acao

acao:
@harshithpabbati still reviewing!

packages/graphiql/src/components/common/Resizer/ResizeHandler.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, []);
+
+  return (
+    <div
+      onMouseDown={handleMouseDown}
+      onTouchStart={handleMouseDown}
+      className={className}
+      style={{
+        cursor: `${dir}-resize`,
+        ...style,
```
acao, line 71: a) should we use `style` here or `_css` or `sm`?
b) would it make more sense to just spread rest props instead of a style prop?
c) should we be using a different component if we want to allow theme-ui props?

@jxnblk

packages/graphiql/src/components/common/Resizer/ResizeHandler.tsx
```diff
@@ -0,0 +1,84 @@
+/** @jsx jsx */
+import { jsx } from 'theme-ui';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { ResizeHandlerData, ResizeHandlerProps } from './types';
+
+export const ResizeHandler: React.FC<ResizeHandlerProps> = props => {
+  const { dir, onStart, onEnd, onUpdate, className, style, children } = props;
+
+  const [isDragging, setIsDragging] = useState(false);
+  const listenersRef = useRef<ResizeHandlerData['listenersRef']>(null);
```
acao, line 10: would we want to use `forwardRef` here to allow users to pass `ref`? or do we only need it internally?

packages/graphiql/src/components/common/Resizer/Resizer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  );
+
+  return (
+    <div
+      ref={boxRef}
+      id={id}
+      className={className}
+      style={{
+        position: 'relative',
+        ...containerStyle,
```
acao, line 114: same here, maybe shouldn't spread style props instead of just spreading props

packages/graphiql/src/components/common/Resizer/ResizeHandler.tsx
jxnblk: Generally, the `style` attribute is good to use for dynamic styles, but it doesn't look like that's really the case here, unless `dir` changes frequently. If `dir` is changing based on mouse events, etc., then `style` is fine to use here. For passing through `style` as props, it's sort of the same thing, but I haven't looked at where that's coming from. Does that make sense?

acao: yes! ok good, just wanted to make sure we weren't baking in anything that would "break" theme-ui implementations or user extensibility

acao:
@harshithpabbati most of my review questions are related to extensibility. so for now, I think this looks great overall for the firs tpass. as we begin to implement it in our own components, we will probably find ways we need to extend it for our purposes and for the purposes of plugin developers

(this includes my comments about forwardRef, etc)

harshithpabbati:
> @harshithpabbati most of my review questions are related to extensibility. so for now, I think this looks great overall for the firs tpass. as we begin to implement it in our own components, we will probably find ways we need to extend it for our purposes and for the purposes of plugin developers
> 
> (this includes my comments about forwardRef, etc)

Yeah sure, we can extend it when we begin implementing that into our components.

PR #1541, opened 2020-05-21
fix: remove @ts-nocheck and fix type errors

stonexer:
I got an type error when using `graphql-language-service-server`.

```
server/node_modules/graphql-language-service-interface/dist/getAutocompleteSuggestions.d.ts:2:26 - error TS2305: Module '"../../graphql-language-service-types/dist"' has no exported member 'ContextToken'.

server/node_modules/graphql-language-service-interface/dist/getAutocompleteSuggestions.d.ts:2:40 - error TS2305: Module '"../../graphql-language-service-types/dist"' has no exported member 'State'.
```

It seems that `ContextToken` and `State` are defined in `graphql-language-service-parser instead` of `graphql-language-service-types`. So I removed the `@ts-nocheck` flag and fix the type errors.

@acao can you help me to review this

### DIFF
--- a/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 /**
  *  Copyright (c) 2019 GraphQL Contributors
  *  All rights reserved.
@@ -16,15 +15,12 @@ import {
   GraphQLType,
   GraphQLCompositeType,
   GraphQLEnumValue,
-  Kind,
-  KindEnum,
   GraphQLField,
+  GraphQLFieldMap,
 } from 'graphql';
 
 import {
   CompletionItem,
-  ContextToken,
-  State,
   AllTypeInfo,
   Position,
 } from 'graphql-language-service-types';
@@ -49,7 +45,8 @@ import {
 import {
   CharacterStream,
   onlineParser,
-  RuleKind,
+  ContextToken,
+  State,
   RuleKinds,
 } from 'graphql-language-service-parser';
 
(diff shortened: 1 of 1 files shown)


packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -229,7 +230,7 @@ function getSuggestionsForFieldNames(
         sortText: String(index) + field.name,
         label: field.name,
         detail: String(field.type),
-        documentation: field.description,
+        documentation: field.description || undefined,
```
acao, line 233: we have a `Maybe` that might be more consistent to use here?

acao:
@stonexer thank you, this is a good catch!

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
stonexer: Yes but I didn't find a better approach ...

`CompletionItem` is defined in `vscode-languageserver-types`

```ts
CompletionItem.documentation?: string | MarkupContent | undefined
```

and `field.description` is defined in `graphql`

```ts
GraphQLEnumValue.description: Maybe<string>
```

stonexer: `?? undefined` seems to be better than `|| undefined`...

@acao do you have some advice here?

acao: @stonexer agreed! this rewrite predatates nullish coalescing, so now that it's available, that would be a good addition

acao: oh! yes then `| undefined` is fine. i thought we had a `type Maybe<T> = T | undefined` in our types

PR #1542, opened 2020-05-21
feat(merge): make fragment merging type-aware

benjie:
Some massive improvements to `mergeAsts`:

- Doesn't inline fields/fragments if they use directives
- Doesn't drop following selection sets using same field name (merges them instead)
- Factors in types if optional schema is provided, allowing for deeper merging (i.e. no more `... on User {...}` when you're already referring to the User type)
- Stronger types throughout
- Tighter conversion to `InlineFragment` (doesn't add extraneous attributes)
- Still manages this on a single `visit`

### DIFF
--- a/packages/graphiql/jest.config.js
+++ b/packages/graphiql/jest.config.js
@@ -3,6 +3,8 @@ const base = require('../../jest.config.base')(__dirname);
 module.exports = {
   ...base,
 
-  // All tests disabled currently
-  testMatch: ['nothing'],
+  testMatch: [
+    // All other tests disabled currently
+    '**/src/utility/__tests__/*[-.](spec|test).[jt]s?(x)',
+  ],
 };
--- a/packages/graphiql/src/utility/__tests__/mergeAst-fixture.ts
+++ b/packages/graphiql/src/utility/__tests__/mergeAst-fixture.ts
@@ -16,6 +16,10 @@ export const fixtures = [
       query Test {
         id
       }`,
+    mergedQueryWithSchema: `
+      query Test {
+        id
+      }`,
   },
   {
     desc: 'inlines simple nested fragment',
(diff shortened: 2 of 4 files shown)


acao:
hey @benjie ! do you want to change this to target `1.0.0` for now, and then we can port it over to the 2.0.0 rewrite when it's released for 1.0.0 users?

acao:
looking great @benjie!
unfortunately there isn't a way to test this yet since the merge capability was amongst what was temporarily disabled
https://deploy-preview-1542--graphiql-test.netlify.app/bundle/dist/

do we need more test cases for this, possibly? potentially some to make sure we see directives left alone, etc?

acao:
maybe also a test to confirm that it preserves comments? that's been a highly requested feature

benjie:
> unfortunately there isn't a way to test this yet since the merge capability was amongst what was temporarily disabled

That's okay; we still pass the unit tests (which I've re-enabled and expanded), so we can be pretty confident that it'll still work at least as well as the old version did:

From CI: `PASS graphiql packages/graphiql/src/utility/__tests__/mergeAst.spec.ts`

> maybe also a test to confirm that it preserves comments? that's been a highly requested feature

It _does not_ preserve comments; neither does the old version. Yes this would be nice; but it's an entirely different body of work out of scope of this PR (AFAIK the AST doesn't include comments, so we'd need to grab these from the source and then some how re-apply them at the end like prettier does; this is quite an undertaking). Also it poses its own challenges about how comments should be merged.

acao:
ah! this explains a lot. i wonder if it's planned for comments to be available in AST?

acao:
@benjie let's call this good to merge, looks great to me locally, and then I can backport it to 1.0.x for a nice treat :D

benjie:
Awesome; merge at will. Most ASTs don’t contain comments (they’re generally skipped by the lexer as if they were whitespace) but it’d certainly be a welcome addition.

Issue #59, opened 2015-11-17
How to use Graphiql when /graphql protected by JWT token (authorization header)


sulliwane:
My **/graphql** route is protected by **JWT token**, so every HTTP request needs to set:

```
 headers: {
      Authorization: 'Bearer ' + token
    }
```

To get through the authentication middleware and hit **/graphql**.

How to manage this authentication step when using **graphiql** ? graphiql is so convenient, it's a pity to not use it :(
thanks for any suggestion!

KyleAMathews:
Download @skevy's https://github.com/skevy/graphiql-app and your header. I have the same setup as you and use this daily.

sulliwane:
@KyleAMathews awesome, I will give it a shot (my main computer is linux distro though...). thanks

skevy:
@sulliwane - I'm sure it would compile on Linux - just haven't tried.

sulliwane:
I just tried ```npm install && npm start``` under linux (ubuntu 14.04) and graphiql-app-V0.3.1, but got blank screen, see here: https://github.com/skevy/graphiql-app/issues/10

I close this issue as I have the answer now.

cancan101:
Any reason to close this issue and opposed to adding the header editor to the (main) graphiql?

sulliwane:
No reason, indeed it would be nice if we could set the header directly into graphiql app. (I'm using graphiql-app right now.)

leebyron:
GraphiQL is generic and doesn't know anything about the network! This is great because you can configure it to use whatever your GraphQL server uses. This is how @skevy was able to build a standalone app version of GraphiQL.

I suggest taking a look at the [example](https://github.com/graphql/graphiql/blob/master/example/index.html) - where you see `graphQLFetcher` is where you can configure your network fetching however you like, in your case by including any authentication details that may be necessary.

Often, when your GraphQL endpoint is behind a authorization feature like this, it is common to expose the GraphiQL tool via a different endpoint on your server.

MarkAndrewJohnson:
@leebyron I would agree being generic is a reason not to automate everything.  I would think, though, that being generic is exactly why it DOES makes sense to have a field allowing custom headers to be set manually by the user!

I'm also unclear as to your last statement.  It doesn't avoid the need for auth tokens to be pushed around.

eugenehp:
[...]

PR #1543, opened 2020-05-22
feat: Add Headers Editor to GraphiQL

connorshea:
This adds a header editor, for submitting HTTP headers when executing a query, as described in #59.

image

This is especially useful when you want to send an authentication token with your request, for example.

This adds an optional `headers` parameter to the `fetcher` which can then be merged into any other headers that the application sets itself.

```js
import React from 'react';
import ReactDOM from 'react-dom';
import GraphiQL from 'graphiql';
import fetch from 'isomorphic-fetch';

function graphQLFetcher(graphQLParams, headers) {
  return fetch(window.location.origin + '/graphql', {
    method: 'post',
    headers: Object.assign(headers, { 'Content-Type': 'application/json' }),
  }).then(response => response.json());
}

ReactDOM.render(<GraphiQL fetcher={graphQLFetcher} />, document.body);
```

TODO:
- [x] Add tests.
- [x] Make sure this isn't a breaking change, or if it is we need to warn users and make sure upgrading is as easy as possible.
- [x] Add the ability to disable the headers editor entirely.
- [x] Make the tab selection style less ugly.
- [x] Rewrite my commit messages to pass the linter.
- [x] Figure out how to get the development application to handle the headers as headers (probably by modifying the fetcher it's rendered with).
- [x] Probably add some way to set a default value for the headers editor?

### DIFF
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -91,7 +91,7 @@ function updateURL() {
 // Defines a GraphQL fetcher using the fetch API. You're not required to
 // use fetch, and could instead implement graphQLFetcher however you like,
 // as long as it returns a Promise or Observable.
-function graphQLFetcher(graphQLParams, headers = {}) {
+function graphQLFetcher(graphQLParams, opts = { headers: {} }) {
   // When working locally, the example expects a GraphQL server at the path /graphql.
   // In a PR preview, it connects to the Star Wars API externally.
   // Change this to point wherever you host your GraphQL server.
@@ -101,16 +101,19 @@ function graphQLFetcher(graphQLParams, headers = {}) {
     : 'https://swapi-graphql.netlify.app/.netlify/functions/index';
 
   // Convert headers to an object.
-  if (typeof headers === 'string') {
-    headers = JSON.parse(headers);
+  if (typeof opts.headers === 'string') {
+    headers = JSON.parse(opts.headers);
   }
 
   return fetch(api, {
     method: 'post',
-    headers: Object.assign({
-      Accept: 'application/json',
-      'Content-Type': 'application/json',
-    }, headers),
+    headers: Object.assign(
+      {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+      headers,
+    ),
     body: JSON.stringify(graphQLParams),
     credentials: 'omit',
   })
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -146,6 +146,7 @@ ReactDOM.render(
     onEditHeaders: onEditHeaders,
     defaultVariableEditorOpen: true,
     onEditOperationName: onEditOperationName,
+    headerEditorEnabled: true,
   }),
   document.getElementById('graphiql'),
 );
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -100,8 +100,9 @@ function graphQLFetcher(graphQLParams, opts = { headers: {} }) {
     ? '/graphql'
     : 'https://swapi-graphql.netlify.app/.netlify/functions/index';
 
+  let headers = opts.headers;
   // Convert headers to an object.
-  if (typeof opts.headers === 'string') {
+  if (typeof headers === 'string') {
     headers = JSON.parse(opts.headers);
   }
 
--- a/packages/graphiql/resources/renderExample.js
+++ b/packages/graphiql/resources/renderExample.js
@@ -36,6 +36,20 @@ if (parameters.variables) {
   }
 }
 
+// If headers was provided, try to format it.
+if (parameters.headers) {
+  try {
+    parameters.headers = JSON.stringify(
+      JSON.parse(parameters.headers),
+      null,
+      2,
+    );
+  } catch (e) {
+    // Do nothing, we want to display the invalid JSON as a string, rather
+    // than present an error.
+  }
+}
+
 // When the query and variables string is edited, update the URL bar so
 // that it can be easily shared.
 function onEditQuery(newQuery) {
@@ -45,6 +59,11 @@ function onEditQuery(newQuery) {
 
 function onEditVariables(newVariables) {
   parameters.variables = newVariables;
+  updateURL();
+}
+
+function onEditHeaders(newHeaders) {
+  parameters.headers = newHeaders;
   updateURL();
 }
 
@@ -72,7 +91,7 @@ function updateURL() {
 // Defines a GraphQL fetcher using the fetch API. You're not required to
 // use fetch, and could instead implement graphQLFetcher however you like,
 // as long as it returns a Promise or Observable.
-function graphQLFetcher(graphQLParams) {
+function graphQLFetcher(graphQLParams, headers = {}) {
   // When working locally, the example expects a GraphQL server at the path /graphql.
   // In a PR preview, it connects to the Star Wars API externally.
   // Change this to point wherever you host your GraphQL server.
@@ -80,12 +99,18 @@ function graphQLFetcher(graphQLParams) {
   const api = isDev
     ? '/graphql'
     : 'https://swapi-graphql.netlify.app/.netlify/functions/index';
+
+  // Convert headers to an object.
+  if (typeof headers === 'string') {
+    headers = JSON.parse(headers);
+  }
+
   return fetch(api, {
     method: 'post',
-    headers: {
+    headers: Object.assign({
       Accept: 'application/json',
       'Content-Type': 'application/json',
-    },
+    }, headers),
     body: JSON.stringify(graphQLParams),
     credentials: 'omit',
   })
@@ -110,9 +135,11 @@ ReactDOM.render(
     fetcher: graphQLFetcher,
     query: parameters.query,
     variables: parameters.variables,
+    headers: parameters.headers,
     operationName: parameters.operationName,
     onEditQuery: onEditQuery,
     onEditVariables: onEditVariables,
+    onEditHeaders: onEditHeaders,
     defaultVariableEditorOpen: true,
     onEditOperationName: onEditOperationName,
   }),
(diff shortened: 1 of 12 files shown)


connorshea:
I'll also need to create a compatibility layer for the `defaultVariableEditorOpen` prop, since I've renamed it (there are multiple tabs now, so the name `defaultVariableEditorOpen` doesn't make sense).

I don't really like the new name `defaultExtraEditorOpen`, so I'm open to suggestions if anyone has anything. Maybe `defaultSecondaryEditorOpen`?

acao:
@connorshea this is looking great!

let's keep `defaultVariableEditorOpen` prop named the same, and add a new prop that makes more sense like you're suggesting, `defaultSecondaryEditorOpen`.

and then we can alias the old prop to the new one. this is how we can *extend* the props interface without *breaking* properties.

packages/graphiql/src/css/app.css
```diff
@@ -161,14 +161,14 @@
   position: relative;
 }
 
-.graphiql-container .variable-editor {
+.graphiql-container .extra-editor {
```
acao, line 164: let's use the same class names as before to keep parity, and add new ones as needed

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -510,32 +549,69 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
+                  <div
+                    style={{
+                      cursor: 'pointer',
+                      color: headerEditorActive ? 'red' : '#000',
+                      display: 'inline-block',
+                      marginLeft: '20px',
+                    }}
+                    onClick={this.handleOpenHeaderEditorTab}
+                    onMouseDown={this.handleTabClickPropogation}>
+                    {'Query Headers'}
```
ncthbrt, line 581: This might be nitpicking, but `Query Headers` feels a bit confusing to me. I think `Request Headers` may be more accurate, as the headers are part of the transport mechanism, not the graphql spec itself, if that makes sense.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -116,12 +123,15 @@ export type GraphiQLState = {
...
   query?: string;
   variables?: string;
+  headers?: string;
   operationName?: string;
   docExplorerOpen: boolean;
   response?: string;
   editorFlex: number;
-  variableEditorOpen: boolean;
-  variableEditorHeight: number;
+  extraEditorOpen: boolean;
```
ncthbrt, line 131: Saw it was in your todo, so not worried about it, but think it's important to have an easy way to disable the headers tab

packages/graphiql/src/components/GraphiQL.tsx
acao: agreed!

packages/graphiql/src/components/GraphiQL.tsx
acao: yes, `variableEditorOpen` prop should not be getting removed here, i thought it was only going to be deprecated?

acao:
fyi your code already passes the linter and prettier check, only thing failing now is unit/cypress tests

packages/graphiql/src/components/GraphiQL.tsx
connorshea: Updated 👍

packages/graphiql/src/css/app.css
connorshea: @acao as in have both `variable-editor` and `extra-editor` here, or just use `variable-editor`?

acao: we should make sure the actual dom elements have *both* classes, and you can use the new classes here. in the new 2.0.0 RFC, we don't use CSS classes

packages/graphiql/src/components/GraphiQL.tsx
connorshea: Do we need to leave the `variableEditorOpen` prop in the GraphiQLState type? I assumed we only needed to leave the `defaultVariableEditorOpen` prop for the GraphiQLProps type.

packages/graphiql/src/css/app.css
connorshea: 👍 updated

packages/graphiql/src/components/GraphiQL.tsx
acao: ah, ok, in that case we're fine. i should have time to actually check this out locally tomorrow. been trying to finish up a big PR today

harshithpabbati:
@connorshea great work!

I found two issues in this
video

In the above GIF you can see that when I click on the tab the editor is not being opened.

harshithpabbati:
And also, when I change the tab to query variables then the value is not visible when we come back to request headers.

video1

packages/graphiql/src/components/HeaderEditor.tsx
```diff
@@ -0,0 +1,234 @@
...
+  private _node: HTMLElement | null = null;
+  ignoreChangeEvent: boolean;
+  constructor(props: HeaderEditorProps) {
+    super(props);
+
+    // Keep a cached version of the value, this cache will be updated when the
+    // editor is updated, which can later be used to protect the editor from
+    // unnecessary updates during the update lifecycle.
+    this.cachedValue = props.value || '';
+    this.ignoreChangeEvent = true;
```
harshithpabbati, line 59: Initialize `ignoreChangeEvent` to false. Else it will not be triggered on the first render

harshithpabbati: [This](https://github.com/graphql/graphiql/commit/e54e1a8691483f1d336231314130d9822481b3be) is the recent fix in Query Variables

connorshea:
@harshithpabbati I think I've fixed it so the editor contents persists now. It only came at the expense of a terrible hack to get the editor gutters to render properly when the page thought there were two editors rendered :)

packages/graphiql/src/components/GraphiQL.tsx
connorshea: Should the prop be `headerEditorEnabled` or `headerEditorDisabled`? And should it be enabled by default or not?

connorshea:
Should I add functions for `toggleVariableEditor` and `toggleHeadersEditor` that can be overridden by the user? Docs and History already have them, so I guess it'd make sense?:

```typescript
handleToggleDocs = () => {
  if (typeof this.props.onToggleDocs === 'function') {
    this.props.onToggleDocs(!this.state.docExplorerOpen);
  }
  this.setState({ docExplorerOpen: !this.state.docExplorerOpen });
};

handleToggleHistory = () => {
  if (typeof this.props.onToggleHistory === 'function') {
    this.props.onToggleHistory(!this.state.historyPaneOpen);
  }
  this.setState({ historyPaneOpen: !this.state.historyPaneOpen });
};
```

connorshea:
Unrelated to this PR, but I just noticed onToggleHistory isn't documented in the `README`.

connorshea:
Tests are passing and everything seems to work :)

I still have a few small questions:
- Should I add the `onOpenVariableEditor` and `onOpenHeaderEditor` props ([original comment here](https://github.com/graphql/graphiql/pull/1543#issuecomment-633148283))
- Is it fine for the prop to be named `headerEditorEnabled`, and is defaulting to enabled acceptable? ([original comment here](https://github.com/graphql/graphiql/pull/1543#discussion_r429581096))

And if anyone has suggestions on how to make the header/variable tab buttons a bit less ugly, I'd be happy to hear them :P

packages/graphiql/resources/renderExample.js
```diff
@@ -72,20 +91,26 @@ function updateURL() {
 // Defines a GraphQL fetcher using the fetch API. You're not required to
 // use fetch, and could instead implement graphQLFetcher however you like,
 // as long as it returns a Promise or Observable.
-function graphQLFetcher(graphQLParams) {
+function graphQLFetcher(graphQLParams, headers = {}) {
```
ncthbrt, line 94: Hmm. I think we should make the second arg an object with the following type ` { headers: {[key:string]: any}`

ncthbrt: The reason being that not all fetchers will care about headers and maybe instead would want other kinds of properties.

harshithpabbati:
Hi @connorshea,

I just have a suggestion to have `Request Headers` tab like this. 

Screenshot 2020-05-25 at 4 51 51 PM
 
What do you say @acao?

packages/graphiql/resources/renderExample.js
acao: @connorshea do you think you could add this to the PR? makes perfect sense, allows us to pass in other values, if we continue with this pattern in 2.0.0

acao:
@harshithpabbati maybe someone could do a layout like that in 2.0.0 but this design is more consistent

acao:
@connorshea another tweak would be to use grayscale instead of red for the accent/activity color, otherwise it's looking pretty good!

acao:
@connorshea are you able to make that final change?

acao:
I will just go ahead and merge, and adjust the colors in another PR. thanks for all this work!

connorshea:
@acao the active/inactive tabs are now black and gray and I've added the FetcherOpts change.

acao:
@connorshea thank you! sorry if it's kinda boring, just wanted to make sure it's neutral for now. the more exciting design is for 2.0.0!

acao:
ah hm, something about the cypress suite executing queries, and JSON parsing, interesting. you should be able to run `yarn e2e` from the root if you like. let me know if you have any issues

connorshea:
@acao no problem, I was fully aware it didn't look very good in the initial implementation :)

I'll take a look at the cypress tests.

connorshea:
@acao should be good to go now :)

acao:
@connorshea oops! so my suggestion to change the text color actually introduced an accessibility debt, my bad.

I'm going to make another PR to address a few other issues I discovered in a cursory, automated runtime audit

justinmchase:
Which version of graphql would this have landed in?

I just went and got the latest from npm `"graphql": "^15.5.0"` and have set `graphiql: true` but this feature appears to be missing.

harshithpabbati:
Hi, you need to pass in `headerEditorEnabled` prop, to get the header editor tab. It's default to false.

You can look at the props [here](https://github.com/graphql/graphiql/tree/main/packages/graphiql#props).

justinmchase:
@harshithpabbati 

Awesome, I see it now. I changed it to `graphiql: { headerEditorEnabled: true }` and that did it. Probably seems like a setting that should be true by default but thanks for now.

PR #1544, opened 2020-05-22
fix: Move all componentWillUnMount functionality to respective events

### LINKED ISSUE #1515: Warning: Can't perform a React state update on an unmounted component
I am using `GraphiQL` in the form as a textarea field and every time I hit submit and navigate away from the page I get this warning

`
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
`
   `in GraphiQL (created by GraphQlTextArea)`

Can someone please look into this?

harshithpabbati:
Closes #1515 
Move all functionality from `componentWillUnMount` lifecycle

@acao can you review this PR

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -1001,7 +1001,7 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
 
   handleEditVariables = (value: string) => {
     this.setState({ variables: value });
-    this._storage.set('variables', value);
+    debounce(500, () => this._storage.set('variables', value));
     if (this.props.onEditVariables) {
       this.props.onEditVariables(value);
     }
@@ -1049,9 +1049,11 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
               this.docExplorerComponent.showDoc(type);
             }
           });
-          this._storage.set(
-            'docExplorerOpen',
-            JSON.stringify(this.state.docExplorerOpen),
+          debounce(500, () =>
+            this._storage.set(
+              'docExplorerOpen',
+              JSON.stringify(this.state.docExplorerOpen),
+            ),
           );
         }
       }
@@ -1114,7 +1116,9 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
       const leftSize = moveEvent.clientX - getLeft(editorBar) - offset;
       const rightSize = editorBar.clientWidth - leftSize;
       this.setState({ editorFlex: leftSize / rightSize });
-      this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex));
+      debounce(500, () =>
+        this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex)),
+      );
     };
 
     let onMouseUp: OnMouseUpFn = () => {
@@ -1178,9 +1182,11 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
           docExplorerOpen: true,
           docExplorerWidth: Math.min(docsSize, 650),
         });
-        this._storage.set(
-          'docExplorerWidth',
-          JSON.stringify(this.state.docExplorerWidth),
+        debounce(500, () =>
+          this._storage.set(
+            'docExplorerWidth',
+            JSON.stringify(this.state.docExplorerWidth),
+          ),
         );
       }
       this._storage.set(
@@ -1192,9 +1198,11 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     let onMouseUp: OnMouseUpFn = () => {
       if (!this.state.docExplorerOpen) {
         this.setState({ docExplorerWidth: hadWidth });
-        this._storage.set(
-          'docExplorerWidth',
-          JSON.stringify(this.state.docExplorerWidth),
+        debounce(500, () =>
+          this._storage.set(
+            'docExplorerWidth',
+            JSON.stringify(this.state.docExplorerWidth),
+          ),
         );
       }
 
@@ -1212,9 +1220,11 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     this.setState({
       docExplorerWidth: DEFAULT_DOC_EXPLORER_WIDTH,
     });
-    this._storage.set(
-      'docExplorerWidth',
-      JSON.stringify(this.state.docExplorerWidth),
+    debounce(500, () =>
+      this._storage.set(
+        'docExplorerWidth',
+        JSON.stringify(this.state.docExplorerWidth),
+      ),
     );
   };
 
@@ -1249,9 +1259,11 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
           variableEditorHeight: bottomSize,
         });
       }
-      this._storage.set(
-        'variableEditorHeight',
-        JSON.stringify(this.state.variableEditorHeight),
+      debounce(500, () =>
+        this._storage.set(
+          'variableEditorHeight',
+          JSON.stringify(this.state.variableEditorHeight),
+        ),
       );
     };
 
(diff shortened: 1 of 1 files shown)


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1133,6 +1114,7 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
       const leftSize = moveEvent.clientX - getLeft(editorBar) - offset;
       const rightSize = editorBar.clientWidth - leftSize;
       this.setState({ editorFlex: leftSize / rightSize });
+      this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex));
```
acao, line 1117: let's debounce this one, using `utility/debounce`.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1080,6 +1049,10 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
               this.docExplorerComponent.showDoc(type);
             }
           });
+          this._storage.set(
```
acao, line 1052: let's debounce this one as well

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1217,6 +1212,10 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     this.setState({
       docExplorerWidth: DEFAULT_DOC_EXPLORER_WIDTH,
     });
+    this._storage.set(
```
acao, line 1215: debounce this one

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1250,6 +1249,10 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
           variableEditorHeight: bottomSize,
         });
       }
+      this._storage.set(
```
acao, line 1252: debounce as well

acao:
awesome work @harshithpabbati ! |

we just need to make sure to debounce the more high-throughput events to avoid performance/render-blocking issues, as invoking lots of storage.set() events can sometimes lead to issues like that

as for the delay, 500-1000ms should be fine. these values only need to be persisted in case the user navigates away or refreshes. this will essentially delay the write until the user has finished, or if there's lots of dragging or variables editing going on, it will only write to ls every 500ms-1000ms, rather than writing to localstorage 60 times a second or more

harshithpabbati:
@acao should I debounce the whole function ?

acao:
@harshithpabbati no, only the storage write events

harshithpabbati:
> @harshithpabbati no, only the storage write events

ok, I will update that

acao:
if you debounce the setState() events, we get weird issues with resizing panes as you can see

harshithpabbati:
> if you debounce the setState() events, we get weird issues with resizing panes as you can see

oh, yeah true 👍

packages/graphiql/src/components/GraphiQL.tsx
harshithpabbati: If I make it like this ` debounce(500, this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex)));`

then this error is showing up
```
Argument of type '{ isQuotaError: boolean; error: any; }' is not assignable to parameter of type '(...args: any[]) => any'.
  Type '{ isQuotaError: boolean; error: any; }' provides no match for the signature '(...args: any[]): any'.ts(2345)
```
Should I make a function for storage set and then pass the variables ?

@acao can you help me with this?

acao:
@harshithpabbati why did you close this PR?

packages/graphiql/src/components/GraphiQL.tsx
acao: @harshithpabbati did not see this before, sorry

harshithpabbati:
> @harshithpabbati why did you close this PR?

That's because we found a better solution than this.

acao:
@harshithpabbati i'm not sure that's the case, we just found a better way to reproduce the error

harshithpabbati:
Oh is it ? Then cool!

harshithpabbati:
> @harshithpabbati i'm not sure that's the case, we just found a better way to reproduce the error. is there a conversation that i did't see where you and Mateo decided his solution was better?

No, we didn't have any discussion

acao:
@harshithpabbati yes, as i mentioned in discord, we will wait until both solutions are ready, and make sure we can reproduce the error with both and see if one doesn't solve it better than the other. both proposals are welcome.

harshithpabbati:
Okay, so there is an error when I use denounce

If I make it like this debounce(500, this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex)));

then this error is showing up

Argument of type '{ isQuotaError: boolean; error: any; }' is not assignable to parameter of type '(...args: any[]) => any'.
  Type '{ isQuotaError: boolean; error: any; }' provides no match for the signature '(...args: any[]): any'.ts(2345)
Should I make a function for storage set and then pass the variables ?

@acao can you help me with this?

packages/graphiql/src/components/GraphiQL.tsx
acao: @harshithpabbati that argument is supposed to be a function it looks like, so I think you'll want to call:

```ts
debounce(500, () => this._storage.set('editorFlex', JSON.stringify(this.state.editorFlex)));
```

harshithpabbati: Okay, cool thanks @acao 

I will update the PR

harshithpabbati:
Done updated @acao :)

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1195,12 +1182,24 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
           docExplorerOpen: true,
           docExplorerWidth: Math.min(docsSize, 650),
         });
+        this._storage.set(
+          'docExplorerWidth',
+          JSON.stringify(this.state.docExplorerWidth),
```
acao, line 1187: ah, another one that needs debounced. anything that has to do with high throughput events

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1195,12 +1182,24 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
+        JSON.stringify(this.state.docExplorerOpen),
+      );
     };
 
     let onMouseUp: OnMouseUpFn = () => {
       if (!this.state.docExplorerOpen) {
         this.setState({ docExplorerWidth: hadWidth });
+        this._storage.set(
+          'docExplorerWidth',
+          JSON.stringify(this.state.docExplorerWidth),
```
acao, line 1201: same here!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1033,6 +1001,7 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
 
   handleEditVariables = (value: string) => {
     this.setState({ variables: value });
+    this._storage.set('variables', value);
```
acao, line 1004: also high throughput. this event is triggered on every key strike

acao:
looking good @harshithpabbati !

as a rule of thumb, any event that has its events triggered many many times a second, such as resize or keyboard stroke, should debounce all the `localStorage.setItem()` calls (via `storage.set()`)

harshithpabbati:
> looking good @harshithpabbati !
> 
> as a rule of thumb, any event that has its events triggered many many times a second, such as resize or keyboard stroke, should debounce all the `localStorage.setItem()` calls (via `storage.set()`)

Sure, I will update it soon

harshithpabbati:
@acao is this PR helpful ?

acao:
@harshithpabbati it might still be?

acao:
lol oops! tried merging via UI, fixing it now. gonna try and see if this resolves our user's issue

PR #1545, opened 2020-05-22
fix: trigger edit variables on first render

### LINKED ISSUE #1532: onEditVariables is not triggered on first render
I'm trying to use `<GraphiQL>` as a controlled component, with a global redux store.

- In `graphiql@1.0.0-alpha.8`, `onEditVariables` is not triggered on the first render. It works fine in `graphiql@0.17.5`. 

I noticed that in 1.x, VariableEditor ignores change events till the first `componentDidUpdate` https://github.com/graphql/graphiql/blob/ce926bb7ab1e6fbf978ec53a0449c33869073c7d/packages/graphiql/src/components/VariableEditor.tsx#L62  I guess `ignoreChangeEvent` should be initialized to false

- Also both versions do not debounce `onEditVariables`, whereas `onEditQuery` seems to be debounced. 

Happy to contribute/raise a PR for this.

harshithpabbati:
Closes #1532 

Initialised `ignoreChangeEvent` to `false` in Variables Editor

@acao can you review this

### DIFF
--- a/packages/graphiql/src/components/VariableEditor.tsx
+++ b/packages/graphiql/src/components/VariableEditor.tsx
@@ -51,7 +51,7 @@ export class VariableEditor extends React.Component<VariableEditorProps> {
   editor: (CM.Editor & { options: any }) | null = null;
   cachedValue: string;
   private _node: HTMLElement | null = null;
-  ignoreChangeEvent: boolean;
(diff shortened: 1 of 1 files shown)


acao:
nice, thank you!

PR #1550, opened 2020-05-25
feat: [RFC] Introduce Query Params

harshithpabbati:
This PR introduces Query Params

* [x] Pass query, variables, operationName through URI
* [ ] Parse through the URI to get query, variables, operationName and format variables


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -110,6 +111,7 @@ export type GraphiQLState = {
   subscription?: Unsubscribable | null;
   variableToType?: VariableToType;
   operations?: OperationDefinitionNode[];
+  parameters?: any;
```
acao, line 114: we are not adding state here, the goal is to eliminate state in GraphiQL component. this logic belongs in SessionProvider

acao:
@harshithpabbati going to close this PR because it needs to be re-created. please lets focus on your existing PRs before opening new ones.

PR #1558, opened 2020-05-29
Persistent context

secmohammed:
- [x]  setup storage adapter.

- [x] setup context boilerplate that interacts with storage to persist data.

- [x] decide whether to persist data or not when mutating context's state.

### DIFF
--- a/packages/graphiql/src/storage/index.ts
+++ b/packages/graphiql/src/storage/index.ts
@@ -5,6 +5,7 @@ import {
   OnFinishType,
   OnErrorType,
 } from './storage.interface';
+
 export class CustomStorage implements Storage {
   private static storage: BaseStorage;
   private static instance: CustomStorage;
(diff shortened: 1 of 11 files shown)


packages/graphiql/src/storage/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (onFinish) {
+        onFinish();
+      }
+    } catch (error) {
+      if (onError) {
+        onError(error);
+      }
+    }
+  }
+  getItem(key: string, onError?: OnErrorType): string | undefined {
```
acao, line 29: these need to be async, so we can do remote storage (such as via http)

acao:
@secmohammed remember, the main reason for the context provider is to instantiate, and make the storage instance available through.

for example, a user could provide an in memory interface to Map, and then in components call `mapFromContext.getItem()` which calls `map.get()` or `map.set()`

secmohammed:
> @secmohammed remember, the main reason for the context provider is to instantiate the storage instance, and make the storage instance available to any components.
@acao  It's already available to whatever implements the BaseStorage interface. Map can be used throughout the storage api as well, but we will have to create a custom map that includes the following methods:

```
getItem(key: string): string | null;
clear(): void;
setItem(key: string, value: string): void;
removeItem(key: string): void;
```

PR #1559, opened 2020-05-29
fix: remove reducer dispatcher from useCallback dependencies

secmohammed:
It's referenced here that the returned dispatcher from useReducer doesn't need to be passed as a dependency to useCallback because the reducer is evaluated directly in the render phase.
https://github.com/facebook/react/issues/14099#issuecomment-436647120


acao:
thanks @secmohammed !

PR #1571, opened 2020-06-05
fix: call debounce statements as they are functions

harshithpabbati:
@acao this PR will fix the storage issue


acao:
@harshithpabbati lets fix the functionality in your last PR. These events should be persisting state as the events are fired, and this behavior should not be needed.

PR #1575, opened 2020-06-10
feat: standalone monaco API

acao:
- replace `monaco.languages.graphql` with an `api` export object
- combine api and defaults into one interface
- get rid of `.d.ts` typings and module extension which caused issues for normal .ts typings
- update documentation to reflect API changes
- add `setSchema()`


acao:
heads up @Nishchit14 - changing this API so it's easier to use!  thanks @rebornix for the advice on this. works much better this way. no more typescript mess @cshaver !

PR #1578, opened 2020-06-11
fix: Set headers to localstorage

harshithpabbati:
This PR is to add headers to local storage.


acao:
looks good! let's make this an opt-in option via props though, perhaps?

harshithpabbati:
Sure @acao, I will update it

acao:
thank you!

the reason I wanted to do this is so that the security risk could be opt-in.

acao:
lets make it `shouldPersistHeaders` maybe?

harshithpabbati:
done @acao 👍

acao:
awesome, thanks @harshithpabbati !

PR #1579, opened 2020-06-11
feat: [RFC] Introduce Query Params

harshithpabbati:
This PR introduces Query Params

Todo
* [x] Pass query, variables through URI
* [x] Parse through the URI to get query, variables


packages/graphiql/src/api/providers/GraphiQLBrowserProvider.tsx
```diff
@@ -0,0 +1,88 @@
...
+  );
+
+  const changeQueryStringParams = React.useCallback(
+    (parameter: string, value: string) =>
+      dispatch(queryStringParamsChangedAction(parameter, value)),
+    [dispatch],
+  );
+
+  React.useEffect(() => {
+    updateQueryStringURL(state.queryStringParams);
```
acao, line 75: instead of doing the parsing here, the parsing should happen in the reducer. just to keep all the processing happening in one place, on a per-action basis

harshithpabbati:
Also, @acao looks like `main` branch is not configured in netlify for preview URL's

acao:
@harshithpabbati fixed! this was supposed to be changed programatically in netlify.toml but it didn't seem to work. should be good now!

packages/graphiql/src/components/VariableEditor.tsx
```diff
@@ -89,6 +92,7 @@ export function VariableEditor(props: VariableEditorProps) {
       if (!ignoreChangeEvent) {
         cachedValueRef.current = editor.getValue();
         session.changeVariables(cachedValueRef.current);
+        browser.changeQueryStringParams('variables', cachedValueRef.current);
```
acao, line 95: exactly what i was hoping for! *excellent* work @harshithpabbati !

harshithpabbati:
sure, will do 👍

acao:
@harshithpabbati thanks for implementing this in the RFC as well!

packages/graphiql-2-rfc-context/src/api/providers/GraphiQLBrowserProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export const initialBrowserState: BrowserState = {
+  queryStringParams: {
+    operation:
+      parseQueryStringURL(window.location.search).operation ||
+      parseQueryStringURL(window.location.search).query,
+    variables: parseQueryStringURL(window.location.search).variables,
+    operationName: parseQueryStringURL(window.location.search).operationName,
+  },
+  queryStringParamsURL: '',
```
imolorhe, line 25: I'm not sure I understand why we need to have both `queryStringParams` and `queryStringParamsURL`. If I understand it correctly, they both contain the same data, just formatted differently.

packages/graphiql-2-rfc-context/src/components/VariableEditor.tsx
```diff
@@ -89,6 +92,7 @@ export function VariableEditor(props: VariableEditorProps) {
       if (!ignoreChangeEvent) {
         cachedValueRef.current = editor.getValue();
         session.changeVariables(cachedValueRef.current);
+        browser.changeQueryStringParams('variables', cachedValueRef.current);
```
imolorhe, line 95: @acao question: should the `browser` context util be separated from the `session` context util? From what I can see, we seem to also be setting `session` whenever we are changing the query string. So perhaps the `session` context util can simply update the browser context as well? Or is there another reason why we have them separate that I'm not aware of?

acao: Session here refers to the current tab, incidentally haha. Maybe EditorSession would be more explicit?

acao: either way, all of this will be replaced. excellent points to consider in the next state management iteration.

acao:
@thomasheyenbrock @jonathanawesome  I think this one is really important to address in 2.0.0! 

Users have long requested a configurable and safe way to handle query params, but most in their implementations are left up to them. The expectation of `query`, `variables` and a few other safe parameters (http request url is safe now) is quite standard now

Issue #983, opened 2019-10-19
GraphQL LSP, Codemirror & Monaco Roadmap [WIP]


acao:
# GraphQL IDE Ecosystem Roadmap

![Screen Shot 2019-11-25 at 1 19 22 PM](https://user-images.githubusercontent.com/49038/69566927-3821e080-0f86-11ea-94d1-629ba9648d6c.png)

This is a WIP. Hoping to add more items. But it shows the trajectory of where everything needs to go in our biggest undertaking, to adopt an LSP plugin system, a react enabled plugin system for GraphiQL that layers over that, and of course monaco. New features will need to fit at the end of this roadmap, essentially.

## GraphiQL

We want to take the current GraphiQL - make it solid, modern and be prepared for the re-design.

`0.x.y`

- [X] - #977 - full i18next!
- [X] - #967 - full accessibility!
- [ ] - e2e harness with Cypress and a nice suite to start with
- [ ] - React 16! Use deprecated class methods for now
- [ ] - Get smarter with a core react context, provide `useGraphQLContext` hook, maybe with parsing and visitors/ast?
- [ ] - Expose events interface to use internally and via plugins via `useReducer`, etc?
- [ ] - Convert several smaller components to implement the hooks interface, use for existing functional components
- [ ] - Begin with a 'plugins' array prop that renders Components that use context/reducer events and/or FCs that use hooks interface

Then we want to switch the editor to monaco:

`1.0.0`
- [ ] - Adopt new monaco language mode, stabilize
- [ ] - keymap/command customization to allow previous (sublime based) keymap, or new vscode-oriented keymap, or per-shortcut customization. Default keymap should have parity with last 0.x.y release?

## GraphQL Language Service Protocol (LSP)
- `graphql-language-interface`
- `graphql-language-parser`

With monaco, we can start working with a Language Server Protocol, which can be consumed by everyone on the web and in their IDE.

- [X] - convert to TS
- [ ] - Symbols
- [ ] - Go To Reference
- [ ] - Plugins in the interface?
- [ ] - Adopt upcoming 15.0.x parser capabilities
- [ ] - Replace `@babel/parser` with a language agnostic parser for `gql` template tags and other configurable patterns.
- [ ] - Universal file system
- [ ] - Universal config, graphql-config?

## GraphQL Service 
- `graphql-language-service`
- `graphql-language-service-server`

[...]

PR #1590, opened 2020-06-21
Update the header editor to work in the 2.0 version of GraphiQL

connorshea:
Surprisingly, I think this mostly works. I originally intended to just try this because I wanted to mess with the 2.0 editor, but this actually seems to work surprisingly well.

I'm not sure if we wanted the headers editor as a plugin for 2.0, or something else?

I didn't spend that long on this, so if this isn't the route we want to go down that's fine by me. Like I said, I didn't originally intend for this to end up becoming anything.

image

### DIFF
--- a/packages/graphiql/src/api/providers/GraphiQLSessionProvider.tsx
+++ b/packages/graphiql/src/api/providers/GraphiQLSessionProvider.tsx
@@ -12,6 +12,7 @@ import {
   operationRequestAction,
   operationSucceededAction,
   variableChangedAction,
+  headerChangedAction,
   operationChangedAction,
   operationErroredAction,
   tabChangedAction,
@@ -24,6 +25,7 @@ export type SessionReducer = React.Reducer<SessionState, SessionAction>;
 export interface SessionHandlers {
   changeOperation: (operation: string) => void;
   changeVariables: (variables: string) => void;
+  changeHeaders: (headers: string) => void;
   changeTab: (pane: string, tabId: number) => void;
   executeOperation: (operationName?: string) => Promise<void>;
   operationError: (error: Error) => void;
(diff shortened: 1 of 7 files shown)


packages/graphiql/src/api/providers/GraphiQLSessionProvider.tsx
```diff
@@ -34,6 +36,7 @@ export const initialState: SessionState = {
   sessionId: 0,
   operation: { uri: 'graphql://graphiql/operations/0.graphql' },
   variables: { uri: 'graphql://graphiql/variables/0.graphql' },
+  headers: { uri: 'graphql://graphiql/variables/1.graphql' },
```
connorshea, line 39: I think this is what we want here? The problem would be that the session state would be trying to pull from whatever the second tab is regardless of whether the headers editor is enabled.

acao:
hey, i'm terribly sorry but this was supposed to be a plugin. we need plugin API first!

acao:
2.0.0 alpha still has massive restructuring/rewriting to do before we can start introducing more features. this will be implemented as a plugin using the plugin API. See the pinned issue https://github.com/graphql/graphiql/issues/983

connorshea:
Fine by me, like I said I just did this on a whim :)

acao:
re-opened for reference, because I'm going to be borrowing most of this for the plugin!

the only difference is the imports really:
```
import { jsx } from 'theme-ui';

import React from 'react';

import { EditorWrapper } from '@graphiql/sdk';

import { useEditorsContext } from '@graphiql/sdk';
import { useSessionContext } from '@graphiql/sdk';

import type { EditorOptions } from '@graphiql/sdk';
```

and a plugin definition

PR #1593, opened 2020-06-23
fix: headers tab - highlighting and schema fetch

### LINKED ISSUE #1592: bug: headers editor does not add headers to schema fetch
oops!

headers editor does not add headers to the schema fetch.

currently you need to manually add them to your `fetcher` function.

this was something I overlooked in PR review, my bad!

acao:
- headers tab does not re-fetch schema with same headers when fetcher prop changes
- codemirror mode for json, `graphql-headers` is a non-existant mode
- resolves #1592

### DIFF
--- a/packages/graphiql/src/components/HeaderEditor.tsx
+++ b/packages/graphiql/src/components/HeaderEditor.tsx
@@ -71,13 +71,14 @@ export class HeaderEditor extends React.Component<HeaderEditorProps> {
     require('codemirror/addon/search/searchcursor');
     require('codemirror/addon/search/jump-to-line');
     require('codemirror/addon/dialog/dialog');
+    require('codemirror/mode/javascript/javascript');
     require('codemirror/keymap/sublime');
 
     const editor = (this.editor = this.CodeMirror(this._node, {
       value: this.props.value || '',
       lineNumbers: true,
       tabSize: 2,
-      mode: 'graphql-headers',
+      mode: { name: 'javascript', json: true },
       theme: this.props.editorTheme || 'graphiql',
       keyMap: 'sublime',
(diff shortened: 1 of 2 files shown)


packages/graphiql/src/components/HeaderEditor.tsx
```diff
@@ -71,13 +71,14 @@ export class HeaderEditor extends React.Component<HeaderEditorProps> {
...
     require('codemirror/addon/dialog/dialog');
+    require('codemirror/mode/javascript/javascript');
     require('codemirror/keymap/sublime');
 
     const editor = (this.editor = this.CodeMirror(this._node, {
       value: this.props.value || '',
       lineNumbers: true,
       tabSize: 2,
-      mode: 'graphql-headers',
+      mode: { name: 'javascript', json: true },
```
connorshea, line 81: @acao This was done because the variable editor was set to `graphql-variables`, is that an actual mode or should we change that as well?

https://github.com/graphql/graphiql/blob/1124834c0751fbd560c7909b498c605fb5d0f65d/packages/graphiql/src/components/VariableEditor.tsx#L88

acao: `graphql-variables` is an actual mode in the `codemirror-graphql` package in this repo, as you can see in the `require()` statements
https://github.com/graphql/graphiql/blob/1124834c0751fbd560c7909b498c605fb5d0f65d/packages/graphiql/src/components/VariableEditor.tsx#L82

acao: codemirror is not the simplest to wrap your head around, so I can totally relate with this choice! haha

PR #1594, opened 2020-06-24
chore(test): set up for monorepo test running

benjie:
Whilst working on #1542 I noticed it's a bit of a pain to run tests right now. This PR sets the repo up for easy jest test running; each package can have it's own configuration, optionally based on the base configuration. The root jest.config.js is just responsible for loading all these other projects.

You'll notice that I've replaced `<rootDir>` with `${__dirname}` in various places, because `<rootDir>` now refers to each of the package roots rather than the repo root.

You'll also note that I removed the `testPathIgnorePatterns: ['codemirror-graphql','packages/graphiql']` in favour of having those projects configure their own test matches (in this case, matching `nothing`).

I have _NOT_ checked that code coverage still works; that's the one thing that may need some more scrutiny. Other than this, I think this change is completely uncontroversion and highly beneficial :)


acao:
awesome! we disabled subpackage test scripts though, the goal is to run tests from the root. trying to avoid subpackage scripts entirely so that scripts are always executed at the root. `Development.md` has tips for executing tests for specific directories from the root.

acao:
I feel like we tried this approach at some point and there was a reason we chose against it. I will check it out later, as I'm knee deep in plugin API rewrite

benjie:
To be clear; this approach allows you to run the tests from the root and, should you desire, also from the package. I've only ran them from the root so far, and it works great:

Screenshot_20200624_163652

benjie:
The main advantage is allowing per-project configuration _when desired_. In particular you currently have graphiql and codemirror-graphql entirely disabled in the root config which makes it incredibly difficult to run their tests; with this approach you can instead configure those projects to run only the tests you wish to allow without making your configuration complex for other projects.

acao:
cool! whatever works.  the handy part about this approach, from proprietary projects, is that it's easier to override configs and add different test renderers and mocks to specific subpackages.

also note that many tests are disabled, as most of whats in `graphiql` package is still in the process of being rewritten, reorganized or deprecated.

acao:
tried pinging you on discord, do you have time for a quick call?

Issue #1599, opened 2020-06-26
Types that lack an "isDeprecated" are included in the deprecation list by default


pleunv:
As of v1 the `isDeprecated` field defaults to `true` if it is not found or defined, which can be seen [here](https://github.com/graphql/graphiql/blob/main/packages/graphiql/src/components/DocExplorer/TypeDoc.tsx#L112). This behavior changed from the v0.x behavior where fields without an `isDeprecated` field [are simply being ignored](https://github.com/graphql/graphiql/blob/v0.13.2/packages/graphiql/src/components/DocExplorer/TypeDoc.js), which is also what you'd expect.

The issue we are running into is that our graphql server (graphql-dotnet) omits the `isDeprecated` field for input type fields that are not deprecated, which is now causing the "Show deprecated fields" button to pop up and contain all fields when clicked.

![image](https://user-images.githubusercontent.com/182021/85838248-9374b500-b799-11ea-99d9-ce529e060793.png)

Was this change done deliberately, and would you be ok with a PR that reverts this to the v0.x behavior?

acao:
woah good catch! seems this is a regression from the typescript migration
can you open a PR against 1.0.0 branch for that? thank you!

harshithpabbati:
can I make a PR for this @acao?

acao:
we have to restructure the repo a bit before we can make the next release to graphiql 1.x, so, sometime in the coming weeks when i get a chance! stay tuned!

acao:
this was released last week in 1.0.4, forgot to update here! enjoy!

PR #1604, opened 2020-06-27
chore: add only the fields which are deprecated to deprecation list

harshithpabbati:
Fixes: #1599,

This PR is to add only the deprecated fields to the deprecated list.


acao:
works great, thanks @harshithpabbati !

PR #1610, opened 2020-06-30
feat: dynamic import codemirror

### LINKED ISSUE #1606: GraphiQL depends on require()
I'm trying to use GraphiQL with snowpack and this highlighted that GraphiQL uses `require` internally.  Now that there's a standard for a browser implementation of dynamic loading, can this be refactored to use dynamic import instead? Having the library be dependent upon webpack or browserify feels like the wrong (or at least an out of date) solution.

acao:
remove require for snowpack, resolves #1606

### DIFF
--- a/packages/graphiql/src/components/HeaderEditor.tsx
+++ b/packages/graphiql/src/components/HeaderEditor.tsx
@@ -44,7 +44,7 @@ type HeaderEditorProps = {
  *
  */
 export class HeaderEditor extends React.Component<HeaderEditorProps> {
-  CodeMirror: any;
+  CodeMirror: typeof CM.default | null = null;
   editor: (CM.Editor & { options: any }) | null = null;
   cachedValue: string;
   private _node: HTMLElement | null = null;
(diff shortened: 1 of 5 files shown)


FredKSchott:
Based on how Webpack/Rollup work, delaying your `require()` call probably isn't actually improving performance for you.

Because `require()` is synchronous, Webpack is already bundle all of the `codemirror` package code that you use into the same bundle as that file. So as written on master, it's not actually any different from importing them statically at the top of the file via `import 'codemirror'`. 

The one nice thing about `require` is that you're deferring the execution time of each addon, but by looking briefly at the CodeMirror codebase these addons don't seem that expensive to initialize (most of their work is hooked up to events in the editor, and doesn't run when you first load it). 

If you'd like, I'd be happy to create a quick PR that just moves these `require()` calls up to top-level `import` calls in the files that use it, so that you can see how the bundle size isn't affected.

acao:
@FredKSchott unfortuantely we cannot. thanks for the explanation but i already know all of this.  we made this change a few years ago to support server side rendering.

FredKSchott:
Sounds good, that's a shame to hear but I'm glad you're on top of it 👍

PR #1612, opened 2020-07-05
[Parser] add benchmarks

hereisnaman:
Screenshot 2020-07-05 at 9 54 41 AM


acao:
looking good @hereisnaman ! at first glance, there are these issues:
- package-lock.json file appears in the PR, which doesn't work because we are using yarn workspaces
- formatting errors occur in CI, you need to run `yarn format` locally first

hereisnaman:
@acao Thanks for the review. I have updated the PR.

packages/graphql-language-service-parser/src/benchmark/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  printResult(results, name, schema);
+};
+
+const runWholeTest = (name: string, schema: string) => {
+  const suite = new Benchmark.Suite('', {
+    onComplete: (e: any) => printResult(e.target.stats, name, schema),
+  });
+
+  const parser = onlineParser();
+  const state: any = parser.startState();
```
acao, line 88: you can use `State` from `graphql-language-service` or `graphql-language-service-parser` here if you want

packages/graphql-language-service-parser/src/benchmark/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  },
+);
+const githubSchema = fs.readFileSync(
+  path.resolve(__dirname, './fixtures/github.graphql'),
+  {
+    encoding: 'utf8',
+  },
+);
+const githubGraphqlJsSchema = fs.readFileSync(
+  path.resolve(__dirname, './fixtures/github-graphqljs.graphql'),
```
acao, line 139: can you clarify the difference between these last two schemas? when I get a chance I could manually diff the two files, but just looking at both SDL files and filesizes i don't see the difference between them

hereisnaman: The difference is the file size only. We are performing two tests here:
1) parsing whole schema at once.
2) parsing line by line

For the 1st test we are testing with short and big files. For the 2nd we are only testing with the short one.

acao: image
maybe i'm mistaken, but both of these files look to be the same size? (kb on left)

hereisnaman: Oh, yes thanks. I missed it. Will fix.

PR #1616, opened 2020-07-15
feat: graphql-config@3 support in lsp server

acao:
- [x] load `documents` instead of `includes`
- [x] fix/adjust tests for `graphql-config@3.03` conversion
- [x] upgrade extensions implementation

## Bonus
- [x] pre-load documents and schema cache on init
- [x] pre-load schema and type definitions cache on definition lookup (solves issues with "lazy definitions" in vscode-graphql)
- [x] add support for `WorkspaceSymbolDefinitions`
- [x] fix regression with variable definition completions #1622

## Todo

This needs more tests!


packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -296,11 +292,7 @@ export class GraphQLCache implements GraphQLCacheInterface {
     }
     const filesFromInputDirs = await this._readFilesFromInputDirs(
```
acao, line 293: @ardatan i had some questions about `await project.loadDocuments()`, as I think it could replace a lot of functionality we have in `GraphQLCache`. Our implementation has a different structure, where there is a collection item per file, and thus potentially multiple asts and other extracted data, whereas with `Source[]` it seems to be a collection of ASTs/strings with potentially entries with the same `location` file reference. I think we can at least get rid of our redundant logic to glob, read and parse files seperately here. There may be a bit of data missing for this to be accomplished, for which I can open a PR to `graphql-config` to add that data to these interfaces?

ardatan: @acao Sure! You can open a PR for that :)

acao: Looking into this further and I don't know if I'll need to actually!

paired with this PR, all of this is working:
https://github.com/prisma-labs/vscode-graphql/pull/184/files

however, there is some redunant logic between our GraphQLCache and MessageProcessor now, as of this PR, and I think I need to rewrite some of this. Either way, going to see if making it more `graphql-config` based in file loading will help! will definately make it more extensible!

acao: @ardatan the way that we cache some of the information about files, we need to store the range. it would be nice if there was a `source.range` in `getDocuments()` from which to compute offsets

acao: it seems i can address this issue with custom loader extensions as well. i need to be able to set specific parser configs as well. probably will be the solution in the long run

acao:
I still want to add tests to this, and whats more, it's had no peer reviews or testing, and has a major impact on most of the GraphQL IDE ecosystem! the greatest impact is on the server, but there are some wide-reaching and important bugfixes, some of which I turned into some small but critical PRs here that I have merged. So I may release these and a few other patches as hotfix patch releases across the ecosystem before merging this PR as a minor version bump to everything but `interface` ideally - `langauge server`, `lsp-cli`, and thus `vscode-graphql`

PR #1626, opened 2020-07-30
Show field directives inside of FieldDoc.

folone:


acao:
hey thanks for this! can you make the PR against 1.1.1 branch, instead of 2.x, as we are rewriting the docs component?

folone:
Right, sorry, will do!

PR #1630, opened 2020-08-01
Update `codemirror-graphql` to use Jest and remove other testing deps

codecog:
This PR updates the unit tests container within `codemirror-graphql` to use Jest instead of mocha & chai. Further work needs to be conducted to make the tests pass and fix the time out errors that are occuring but that is outside the scope of this work.

Other additions:
- Add `'**/__{tests,mocks}__/*.{js,jsx,ts,tsx}’` to eslint override files
- Update @types/jest to 26.0.8

Signed-off-by: Josh Kelly <josh@codecog.dev>


acao:
thanks @codecog ! can you make it depend on the same test helpers and codemirror mocks as the other repos?

I've been a little further along than this myself, for example when i was converting the mocha suites to Jest last year, and always got stuck at those test timeouts with codemirror, to the point that I just left it unconverted.

unfortunately we can't merge a PR with broken tests, so once this PR has all tests passing we can merge!

codecog:
No problem will get those last couple of pieces fixed up in that case

acao:
awesome! lemme know on discord if you have any questions. iirc, part of the issue was also converting to jest's signature for asserting promise errors/resolutions, which is different from chai-as-promised

packages/codemirror-graphql/package.json
```diff
@@ -33,17 +33,14 @@
...
-  },
   "scripts": {
     "build": "yarn build-clean && yarn build-js && yarn build-esm && yarn build-flow .",
     "build-js": "babel src --root-mode upward --ignore **/__tests__/**,**/__mocks__/** --out-dir .",
     "build-esm": "cross-env ESM=true babel src --root-mode upward  --ignore **/__tests__/**,**/__mocks__/** --out-dir esm && node ../../resources/renameFileExtensions.js './esm/{**,!**/__tests__/}/*.js' . .esm.js",
     "build-clean": "rimraf {mode,hint,info,jump,lint}.{js,esm.js,js.flow} && rimraf esm results utils variables coverage __tests__",
     "build-flow": "node ../../resources/buildFlow.js",
     "watch": "babel --optional runtime resources/watch.js | node",
-    "test": "nyc --require @babel/polyfill --reporter lcov --reporter text mocha $npm_package_options_mocha"
+    "test": "nyc --require @babel/polyfill --reporter lcov --reporter text jest"
```
gagoar, line 43: Just doesn't need istanbul at all. it comes already there. so if you just do --coverage when running tests it will get you the reports you need.

acao: yes! also no need for `@babel/polyfill` now either, since we are using `jest-ts`

gagoar: an example of this can be found [here](https://github.com/gagoar/ts-node-template/blob/main/jest.config.js)

acao: in this case, the `codemirror-graphql` `jest.config.js` should look the same as all the others, where it just resolves the shared config which takes care of this. we're using the `projects` option in the root config. like other packages, this PR should provide a single line for `jest.config.js` and should only execute `jest` for `test` script. all the jest configuration is already taken care of for the whole monorepo. may need to tweak things a bit for the codemirror mock though

imolorhe:
@codecog any updates on this one? If you don't mind I'd like to create a new PR branching off this one with the resolved conflicts so we can have the changes merged (and with the co-authored attribution of course 🙂).

codecog:
No updates I'm afraid, feel free to take this one over. Thanks for your effort in helping.

acao:
and thank you so much @codecog ! wish I'd been more available to help with this effort before

imolorhe:
This can be closed now @acao

PR #1631, opened 2020-08-04
improvement: Show directives inside of FieldDoc.

folone:
Directives are not exposed in the introspection schema (discussion here: https://github.com/graphql/graphql-spec/issues/300). Nevertheless, if GraphiQL is supplied with an SDL schema, it can access AST directives for fields. It would be useful to show field directives in Doc Explorer for that case.


acao:
@folone they are exposed in introspection schema in graphql 15! this is fantastic thank you

Issue #1638, opened 2020-08-11
[1.0.x] fetcher Observable is expected to have different subscribe method arguments in _fetchQuery and fetchSchema


n1ru4l:
**From Discord:**
> For GraphiQL 1.0.3 I noticed that I get a polymorphic input for an observable subscribe function returned by the fetcher function. For the introspection query that is three functions as the arguments (next, error, complete) for all the other operation sent however it is a single object argument with the functions next, error and complete on it.
See:
https://github.com/n1ru4l/graphql-live-queries/blob/2766455876b4c8e3d90bedf47e62d34da50509b6/packages/example/src/static-html.ts#L79

So the issue I see here is that observables are subscribed to with two different methods, but maybe I am misunderstanding it or it has some legacy context that I am not aware of.
https://github.com/graphql/graphiql/blob/2a7b9c13f93833205fb1fc3712aadd6165deea66/packages/graphiql/src/components/GraphiQL.tsx#L1618-L1627 which is used here https://github.com/graphql/graphiql/blob/2a7b9c13f93833205fb1fc3712aadd6165deea66/packages/graphiql/src/components/GraphiQL.tsx#L800-L808

and

https://github.com/graphql/graphiql/blob/2a7b9c13f93833205fb1fc3712aadd6165deea66/packages/graphiql/src/components/GraphiQL.tsx#L926-L945

It seems to me that the `fetchSchema` method does support `fetcher` to return and handle an observable (by converting it to a promise) while expecting the arguments of the subscribe method/function to be different than how the   `_fetchQuery` treats it (which also utilizes `fetcher` for receiving an Observable (or Promise)).

**Possible Workarounds:**

- Check whether the first parameter passed to subscribe is a function (seems rather like a hacky solution). (This is what I am currently doing here: https://github.com/n1ru4l/graphql-live-queries/blob/2766455876b4c8e3d90bedf47e62d34da50509b6/packages/example/src/static-html.ts#L74-L101 **Note:** I built my own fetcher that uses socket.io).

- Use an Observable library that supports both "signatures" and picks the correct one based on the amount/type of arguments it is called with (requires adding another dependency or writing it yourself, which brings you back to the first workaround).

- Return a Promise from fetcher (instead of Observable) if it receives an introspection query.

**Possible Solution:**

Change the `observableToPromise` function to accept an observable with the same subscribe argument signature as inside `_fetchQuery`.

acao:
thanks for looking into this @n1ru4l ! great research.

[...]

PR #1639, opened 2020-08-11
fix: unify how observables are consumed in different places

n1ru4l:
Closes https://github.com/graphql/graphiql/issues/1638

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -1615,16 +1615,16 @@ function observableToPromise<T>(
     return observable;
   }
   return new Promise((resolve, reject) => {
-    const subscription = observable.subscribe(
-      v => {
+    const subscription = observable.subscribe({
+      next: v => {
         resolve(v);
         subscription.unsubscribe();
       },
-      reject,
-      () => {
+      error: reject,
+      complete: () => {
         reject(new Error('no value resolved'));
       },
-    );
+    });
   });
 }
 


acao:
thanks for this! most of the implementation out there are [like this](https://github.com/apollographql/GraphiQL-Subscriptions-Fetcher/blob/master/src/fetcher.ts), so can we introduce backwards compatibility somehow?

n1ru4l:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length

We could make an assumption based on the number of arguments, that however would fail if the observable implementation uses rest or default parameters.

Other than that I would not know any solution despite dealing with it in username in the fetcher function.

n1ru4l:
This change would not break subscriptions-transport-ws (https://github.com/graphql/graphiql/issues/1638#issuecomment-673376897)

n1ru4l:
Hey @acao, I guess you closed this due to lacking feedback from me. We already talked about AsyncIterables being the future for the fetcher prop but I think we should still address/tackle this issue. It is now back on my radar after having seen https://github.com/enisdenjo/graphql-ws/issues/93

So the conclusion was that all major observable implementation support both the

**A**
```tsx
type Observable {
  subscribe(next: function, complete: function, error: function):an any
}
```

and 

**B**
```tsx
type Observable {
  subscribe(args: {
    next: function;
    complete: function;
   error: function;
  }): any
}
```

signatures. I only see concerns for users that do not use any of those libraries that are ALSO mix and match both call signatures in their custom fetcher implementation. So IMHO it is pretty safe to simply unify the call signatures within GraphiQL to either use **A** and **B** in order to avoid workarounds like the following: https://github.com/enisdenjo/graphql-ws/issues/93#issuecomment-758625106

acao:
it wont let me reopen this PR for some reason?

acao:
oh because theres no 1.0.0 branch anymore that’s why

n1ru4l:
I could send a new pr

PR #1652, opened 2020-08-26
chore: fix handle history pane and handle documentation explorer state

### LINKED ISSUE #1650: historyPaneOpen is not working properly
When open and close the history panel, the flag `historyPaneOpen` does not respect the current visual state of it. The live demo has the same behaviour.

harshithpabbati:
Fixes #1650 
This PR is to fix the history pane and documentation explorer pane state


acao:
thank you @harshithpabbati ! looks like it works

PR #1654, opened 2020-08-30
Parser/tests

hereisnaman:
This PR includes tests for the graphql parser packages with 100% code coverage. Both Query and Type Definition tests are included.


packages/graphql-language-service-parser/src/CharacterStream.ts
```diff
@@ -152,7 +152,7 @@ export default class CharacterStream implements CharacterStreamInterface {
   indentation = (): number => {
     const match = this._sourceText.match(/\s*/);
     let indent = 0;
-    if (match && match.length === 0) {
+    if (match && match.length !== 0) {
```
acao, line 155: this is a significant bugfix, if thats what i gather it is?

acao: can you explain what this does? is there a bug with handling indentation?

acao: @hereisnaman if you have a chance to answer this, I would love to merge this PR once I know better what this change entails. I can look at the PR more closely soon to gain a better understanding myself

acao: @yoshiakis what do you make of this change? this would be a clean tests-only PR to merge, however this one programattic change has me wondering if this is a bugfix or a mistake?

yoshiakis: @acao can you give me a few days to see if this is a bugfix or not?

jonsevoich:
Welcome to Gboard clipboard, any copy will be saved here.
https://GMAILaccounts.google.com
 Yoshi S. Newbie on git hub.
 If interested helping  figure out method for response to check/_ values
/payment billingaccount _
Please email at isafevendors@gmail.com

Gmail mail account
Jon

packages/graphql-language-service-parser/src/CharacterStream.ts
hereisnaman: Hey @acao this PR increases the test coverage for the online parser to 100%.

acao: yes that is good news! but also this particular line I'm referring to here is potentially a breaking change, and we want to evaluate what that change does and if it is desired?

acao:
@yoshiakis don't know what was with that, whether it was spam or an autoresponse or what, haha! reported

packages/graphql-language-service-parser/src/CharacterStream.ts
acao: @hereisnaman we just need to identify whether this line change is a bugfix, or whether it introduces a bug

yoshiakis: It's a bugfix since method `indentation` always returns 0 without this fix. It's useless. I don't know if it should be fixed though.

`codemirror-graphql` needs `state.indentation` or `state.levels` to know how many indentations are put. To change the value of `state.indentation`, `indentation` method is required. `state.indentation` and `state.levels` are interdependent. So, `indentation` method is needed, though`codemirror-graphql` isn't using `indentation` method from `graphql-lanaguage-service-parser`. Actually it's using [the one from codemirror](https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/util/StringStream.js#L55).

How about any other modules? I believe `indentation` method from `graphql-lanaguage-service-parser` isn't used by any other modules under `graphiql` package. `monaco-graphql` looks like it determines how many indentations are put in a different way. (maybe by the feature of prettier?)

I searched for the word of `indent` or `indentation` in issues and PRs to look for open issues or PRs to know someone want this bug to be fixed, though I haven't found anything about it. (Or I overlooked.)

So again, I don't know if it should be fixed or not.

acao: hmm, yes this is what i was thinking! with monaco and vscode we can configure that, the language server uses this parser for autocomplete so, i dont think this is used there either!

PR #1688, opened 2020-10-02
fix: Fixed build issues

anoopmd:
I was trying to run the monaco example. I ran into a couple of issues. This PR addresses them. Have also attached screenshots of error that I had enocuntered.
monaco-error

acorn-error

### DIFF
--- a/examples/monaco-graphql-webpack/babel.config.js
+++ b/examples/monaco-graphql-webpack/babel.config.js
@@ -1,14 +1,1 @@
-module.exports = {
-  sourceMaps: true,
-  presets: [
-    [
-      require.resolve('@babel/preset-env'),
-      {
-        targets: { browsers: ['last 2 chrome versions'] },
-        bugfixes: true,
-      },
-    ],
-    [require.resolve('@babel/preset-typescript'), {}],
-  ],
-  plugins: [require.resolve('@babel/plugin-proposal-class-properties')],
-};
+module.exports = require('../../babel.config');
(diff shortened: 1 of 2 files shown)


examples/monaco-graphql-webpack/babel.config.js
```diff
@@ -1,14 +1 @@
-module.exports = {
```
anoopmd, line 1: We are using the root babel config for the `graphiql` package. This removal should be safe.
Plus it fixes the lint error.

examples/monaco-graphql-webpack/package.json
```diff
@@ -12,17 +12,19 @@
...
-    "monaco-graphql": "file:../../packages/graphiql",
+    "monaco-graphql": "file:../../packages/monaco-graphql",
     "prettier": "^2.0.2"
   },
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "7.8.3",
     "@babel/plugin-syntax-dynamic-import": "7.8.3",
     "@babel/preset-env": "7.9.5",
     "@types/prettier": "^2.0.0",
+    "acorn": "^8.0.3",
```
anoopmd, line 23: I ran `yarn` both in the root as well as the monaco example folder. I got the error saying that `acorn` path is missing and that it needs to be installed.

anoopmd:
@acao Please take a look when you find some time. Thanks :)

acao:
@anoopmd thanks for addressing this!

i wonder if this could be solved by adding the webpack monaco middleware package to the example devDependencies

i think this happened when i removed examples from the workspace

are you running `yarn` inside the example first before running `yarn start`?

acao:
@anoopmd I resolved this with #1696, thanks for the heads up!

PR #1695, opened 2020-10-19
fix(codemirror-graphql): give interface field name suggestions

ardatan:
Currently `codemirror-graphql` plugin ignores the fields of an interface type and doesn't give suggestions for those fields as hints because `getSuggestionsForFieldNames` uses `isObjectType` that returns `false` for interface types.

### DIFF
--- a/packages/codemirror-graphql/src/hint.js
+++ b/packages/codemirror-graphql/src/hint.js
@@ -34,6 +34,7 @@ import {
   SchemaMetaFieldDef,
   TypeMetaFieldDef,
   TypeNameMetaFieldDef,
+  isInterfaceType,
 } from 'graphql';
 import type {
   GraphQLField,
@@ -225,7 +226,7 @@ function getSuggestionsForFieldNames(
   if (typeInfo.parentType) {
     const parentType = typeInfo.parentType;
     const fields: GraphQLField<*, *>[] =
-      isObjectType(parentType) && 'getFields' in parentType
+      isObjectType(parentType) || isInterfaceType(parentType)
         ? objectValues((parentType.getFields(): any))
         : [];
     if (isCompositeType(parentType)) {
--- a/packages/codemirror-graphql/src/__tests__/hint-test.js
+++ b/packages/codemirror-graphql/src/__tests__/hint-test.js
@@ -998,4 +998,22 @@ describe('graphql-hint', () => {
     );
     expect(suggestions7.list).to.deep.equal(expectedSuggestions);
   });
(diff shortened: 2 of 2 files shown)


packages/codemirror-graphql/src/hint.js
```diff
@@ -225,7 +225,7 @@ function getSuggestionsForFieldNames(
   if (typeInfo.parentType) {
     const parentType = typeInfo.parentType;
     const fields: GraphQLField<*, *>[] =
-      isObjectType(parentType) && 'getFields' in parentType
+      'getFields' in parentType
```
yoshiakis, line 228: It seems Flow doesn't understand `'getFields' in parentType` so errors is occurring in the next line by flow check. I'm thinking `isObjectType(parentType) || isInterfaceType(parentType)` without `getFields in parentType` is a suitable solution here because [GraphQLObjectType](https://github.com/graphql/graphql-js/blob/v15.0.0/src/type/definition.js#L701) and [GraphQLInterfaceType](https://github.com/graphql/graphql-js/blob/v15.0.0/src/type/definition.js#L1011) **must** have `getFields()` method so `getFields in parentType` is never reached. What do you think?

yoshiakis (changes requested):
Thanks for willing to fix this issue! I believe this bug has been introduced since merging [this PR](https://github.com/graphql/graphiql/pull/902) which I had created. I left a comment for you.

packages/codemirror-graphql/src/hint.js
ardatan: That makes sense! Done!

acao:
looking good @ardatan , and thanks for the test. @yoshiakis do you feel this has everything it needs?

acao:
also, @ardatan for future reference the scope is automatically gleaned from the changed paths in the commit, so you can just prefix with `fix:`

ardatan:
Thank you @acao 🙏

acao:
release was cut:

https://github.com/graphql/graphiql/blob/96a387b4b205780d11add31a6dbe720cd534a933/packages/codemirror-graphql/CHANGELOG.md#bug-fixes

`graphiql@1.0.6` contains this fix

yoshiakis:
@ardatan I ran the test for codemirror-graphql on my local environment, and the test failed. In the test code you've added, `expectedSuggestions` should also include `__typename` meta-field.

yoshiakis:
@acao It seems like the test for codemirror-graphql doesn't currently run on GitHub action. Is it intended to do that?

PR #1698, opened 2020-10-23
chore: move typescript to devDependencies

cshaver:
This should reduce the size of node_modules when using graphiql by quite a bit!


acao:
thanks! how did it even end up in dependencies to begin with?!

PR #1701, opened 2020-10-28
feat(graphiql): add prop for adding toolbar content while preserving …

n1ru4l:
…the default buttons

### DIFF
--- a/packages/graphiql/README.md
+++ b/packages/graphiql/README.md
@@ -203,7 +203,7 @@ GraphiQL supports customization in UI and behavior by accepting React props and 
 
 - `shouldPersistHeaders`: an optional boolean which enables to persist headers to storage when `true`. Defaults to `false`
 
-- `additionalToolbarContent`: if you want to keep the default toolbar content (Prettify, Merge Button etc.) while adding additional items pass any react element to this property. E.g. you could add multiple GraphiQL.Button elements by wrapping them inside a react fragment. Defaults to `null`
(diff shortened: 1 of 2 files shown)


acao:
one thing i want to think about, as we add new props for adding additional content, is to think about consistency. since we've indicated that in 2.x we will be deprecating `GraphiQL.Toolbar` etc for obvious reasons, we have the opportunity to build new interfaces, and I'd really like to improve the props situation.

for example, if we make the sidebar tabs configurable, eventually in-editor tabs, is there a need for some consistency in the naming of the props here and in that future case?

I was thinking of even consolidating it to a few props, so that in 2.x we might have something like:
```js
theme={myCustomTheme} // custom styles will go here, even for specific components. "plugin" components below will be build with 2.x theme-ui components which inherit this theme
toolbar={{ additionalContent: [<MyButton />, <MyButton />], display: "auto-hide" }}
sidebarLeft={{ replace: ({ schema, operation, options }) => [<GraphiQLExplorer schema={schema} currentOperation={operation}  />, <NetworkTab currentOptions={options} />], display: "collapse"}}
sidebarRight={{ replace: ({ schema, operation }) => [<CodeExporter schema={schema} operation={operation} />, <SchemaExplorer schema={schema} />], display: "collapse"}}
```

acao:
in addition to that comment, I'd like to start adding an implementation of almost every config option in the graphiql-webpack-example, so it can become a kitchen sink style example. A PR just to set that up may be a pre-requesite, however

n1ru4l:
I like that idea! Before dragging that too long, would you be fine with me implementing that initially for the toolbar with only the `additionalContent` option? Also instead of an array we can also use a jsx fragment.

acao:
oh yes that's what i meant, to start a pattern that we can grow into, I just didn't want to introduce a new prop we would need to replace

PR #1703, opened 2020-10-30
feat: deeper fragment merging

benjie:
Follow up to #1542; this passes the schema through to mergeAST so that the query can be further simplified in a type-aware way (i.e. spreading a fragment of type `Test` inside a scope that can only be type `Test` will result in the fragment being fully inlined).

Given this query:

```graphql
{
  test {
    id
    ...T
  }
}

fragment T on Test {
  isTest
}
```

Pressing "Merge" previously simplified to:

```graphql
{
  test {
    id
    ... on Test {
      isTest
    }
  }
}
```

But with this change, it now simplifies to:

```graphql
{
  test {
    id
    isTest
  }
}
```


acao:
thank you @benjie !

PR #1710, opened 2020-11-04
fix: crash on receiving an LSP message in "stream" mode

### LINKED ISSUE #1708: LSP server broken when using stream method
Run server with:
```
./node_modules/graphql-language-service-cli/bin/graphql.js server --method stream
```
and send the `initialize` message to it.

The server will just crash in a non-crashy way (just exit with 0 exit code).

This issue was introduced by https://github.com/graphql/graphiql/pull/1651/files where it has changed the logger to post to `stdout` for non-Error messages:
```js
const processSt = severity === DIAGNOSTIC_SEVERITY.Error ? process.stderr : process.stdout;
```

This doesn't work for `stream` method as `stdio` is reserved for communicating with the LSP client.

@acao

rchl:
Don't write log messages to stdout in "stream" mode. stdout is reserved
for communicating with the LSP client in the "stream" mode. Write all
types of log messages to stderr.

Fixes LSP server crashing on the first message it receives (or maybe
LSP client closing it because it received an invalid response).

Also:
 - the callback to socket.write() was always logging the "err" parameter
   even if there was undefined which caused spurious lines in the
   console. Only actually log that if "err" is truthy.
 - Avoid too many newlines at the end of the log message as that also
   causes spurious empty lines in the log and the console.
 - Gracefully handle the case when the response for "workspace/configuration"
   returns a null value. Those configuration options have sane defaults
   and don't need to be present.
 - The Logger._stream property was removed as it was unused.

Resolves #1708

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Our commit messages are linted by `commitlint` following the angular changelog c
 
 You will need to include a type prefix for all commit messages. For example:
 
-`git commit -m 'fix: fix window undefined error in result viewer'
+`git commit -m 'fix: fix window undefined error in result viewer'`
 
 ### Type Prefixes
 
--- a/packages/graphql-language-service-server/src/Logger.ts
+++ b/packages/graphql-language-service-server/src/Logger.ts
@@ -8,21 +8,24 @@
  */
 
 import { Logger as VSCodeLogger } from 'vscode-jsonrpc';
+import { DiagnosticSeverity } from 'vscode-languageserver';
 
 import * as fs from 'fs';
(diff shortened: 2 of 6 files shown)


acao:
fantastic, thank you!

rchl:
There might be an issue with `trim()` call that I've added. Let me do a small fixup.

rchl:
Fixed with the latest commit. Feel free to squash all commits on merging.

(Also noticed that when schema server is not running, that will trigger an unhandled exception but I'll handle that in a separate PR).

acao:
fabulous work looks good to merge. thanks for this!

PR #1714, opened 2020-11-11
chore: upgrades for graphql@15.4.0

acao:


packages/graphql-language-service-interface/src/getDiagnostics.ts
```diff
@@ -90,20 +91,20 @@ export function validateQuery(
...
   );
 
-  // Note: findDeprecatedUsages was added in graphql@0.9.0, but we want to
-  // support older versions of graphql-js.
-  const deprecationWarningAnnotations = !findDeprecatedUsages
-    ? []
-    : mapCat(findDeprecatedUsages(schema, ast), error =>
-        annotations(error, DIAGNOSTIC_SEVERITY.Warning, 'Deprecation'),
-      );
+  // Note: we still want to support older versions that don't have NoDeprecated
```
acao, line 94: @IvanGoncharov notice the change here, where we were previously supported an old approach to `findDeprecatedUsages`, which seems to now be deprecated for something more like our `mapCat` behavior?

acao:
ah. and flow check errors for `codemirror-graphql`. this will not be an easy one

PR #1715, opened 2020-11-12
fix: languageserver filepath on Windows

lanwin:
The current impl of MessageProcessor.ts produces invalid file pathes on windows and so file operations did not work.

### Problem

This is cause `new URL(uri).pathname` did will not return a valid path from an fileURL on Windows.

`new URL("file:///c%3A/test/test.txt").pathname` returns `/c%3A/test/test.txt` witch has two problems.

1. In Windows the first / need to be omitted, since most pathes needs to begin with driver letters.
2. The path is url encoded witch is not supported in Windows file system.

### Solution

The best solution would be to use nodes `url.fileURLToPath` instead of `new URL(uri).pathname`. But this is not supported before node v10.12.0 and since I have no idea what the minimal supported node version is, i simply added a minimal fileURLToPath impl at the top.

`fileURLToPath("file:///c%3A/test/test.txt")` should now return 'c:/test/test.txt' on Windows and works like before on other platforms (except it always url decodes, witch should also be a good thing on other platforms).


packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -1030,7 +1043,7 @@ export class MessageProcessor {
 
     await this._graphQLCache.updateObjectTypeDefinition(
       rootDir,
-      new URL(uri).pathname,
+      fileURLToPath(new URL(uri)),
```
acao, line 1046: it would make more sense to add this in the `GraphQLCache` layer, IMO, since these functions are invoked many more times throughout this file.

packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -68,6 +69,16 @@ import { GraphQLExtensionDeclaration } from 'graphql-config';
 import type { LoadConfigOptions } from './types';
 import { promisify } from 'util';
 
+// It would be far better to use nodes native url.fileURLToPath but
+// its not supported before node v10.12.0.
```
acao, line 73: we support node v10.17+ because dynamic import

packages/graphql-language-service-server/src/MessageProcessor.ts
lanwin: So you mean passing the uri directly to the cache layer instead of the file Path and resolve the file path there only once?

lanwin: I can try to change that. Whats the fastest way to test if the lang server works?

packages/graphql-language-service-server/src/MessageProcessor.ts
lanwin: Only to be sure. That means the minimal node version is node v10.17. Would it be ok if I would add the minimal version to package.json?

acao (changes requested):
Just need to abstract this better, ideally build it into the existing logic for building file URIs in GraphQLCache. this should be taken care of across OSes by `path.resolve()` by default, and using `url.fileURLToPath1 as you suggest. any improvements to switching between file URIs (what vscode API uses) and file paths would be welcome.

Sadly, my plans to refactor most of the `MessageProcessor` logic into a `TextDocumentCache` and `TextDocument` class have been dashed by changed priorities at work. I'd rather just be able to use getters for `document.path` or `document.url` for this behavior, but I opted to add features and convert to typescript when i inherited the project.

packages/graphql-language-service-server/src/MessageProcessor.ts
acao: we had `engines` defined for some packages before, but it caused too many downstream issues

packages/graphql-language-service-server/src/MessageProcessor.ts
acao: yarn watch and executing the CLI is one way

to test in vscode, just yarn link the server workspace to vscode-graphql repo, and then run the vscode launcher for vscode-grapqhl. and you can run `yarn watch` in the monorepo, though the vscode extension debug launcher still requires refreshes since it compiles at startup time

acao: my plan to pull vscode-graphql into this monorepo, and thus a one click DX for the language server for VScode or Vim is a long way off given all the other priorities

packages/graphql-language-service-server/src/MessageProcessor.ts
lanwin: I switched the impl to use the native fileURLToPath.

packages/graphql-language-service-server/src/MessageProcessor.ts
lanwin: So I tested it and it works.

lanwin: Whats todo for integration vscode-graphql into this repo? Maybe I can send a pullrequest for that.

lanwin:
First, ive tested my change and it works. So I would suggest to merge this pullrequest and release the vscode extension to make it work on Windows again.

> Just need to abstract this better, ideally build it into the existing logic for building file URIs in GraphQLCache. this should be taken care of across OSes by `path.resolve()` by default, and using `url.fileURLToPath1 as you suggest. any improvements to switching between file URIs (what vscode API uses) and file paths would be welcome.

Yes from what I can see is the API of GraphQLCache misleading at the moment. In example filePath is from type Uri. But it uses real file pathes. So I would suggest to make the cache only work with file Uris and use fileURLToPath only within the cache when the real files are getting resolved. If that is what you want, I can see if I can get a pullrequest ready for it.

> Sadly, my plans to refactor most of the `MessageProcessor` logic into a `TextDocumentCache` and `TextDocument` class have been dashed by changed priorities at work. I'd rather just be able to use getters for `document.path` or `document.url` for this behavior, but I opted to add features and convert to typescript when i inherited the project.

That would make sens, but since I have nearly no Idea how the whole language server works, this change would be heavy for me.

acao:
@lanwin can you just implement this everywhere we call `updateFragmentDefinitionCache` or `updateObjectTypeDefinition`? It's likely you just aren't invoking all of the language features or cases that would trigger this bug. For example, have you tried symbol lookup? is outline working for you?

lanwin:
Ive did the changed and made "Go to Definition" work with it.

What I noticed is that "Go to Definition" in large schema files (like Github) works but takes several seconds to compleat. 

Currently the linter fails, but that seems related to some other change in master.

lanwin:
So problem is fixed and tests are running now. What do you think?

packages/graphql-language-service-server/src/__tests__/MessageProcessor-test.ts
```diff
@@ -295,7 +296,7 @@ describe('MessageProcessor', () => {
     };
 
     const result = await messageProcessor.handleDefinitionRequest(test);
-    await expect(result[0].uri).toEqual(`file://${queryDir}/test3.graphql`);
```
acao, line 298: hmmm... I don't think LSP will like this. LSP (and thus vscode, and other LSP implementations) need this result to be a URL

lanwin: I had changed the value of queryDir to be an uri. To reflect that butter Ive now renamed that variable to queryPathUri instead.

acao:
@lanwin was trying this PR out, and I noticed I can't resolve the conflict without merging because I don't have write access to PR branch

acao:
everything looks to be working great in a posix environment for me using this branch. thanks for all the help!

PR #1724, opened 2020-12-01
feat: add AsyncIterable support to fetcher function

n1ru4l:


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -958,8 +964,32 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
       });
 
       return subscription;
+    } else if (isAsyncIterable(fetch)) {
+      (async () => {
+        try {
+          for await (const result of fetch) {
```
n1ru4l, line 970: https://caniuse.com/mdn-javascript_statements_for_await_of

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1654,6 +1679,39 @@ function isObservable<T>(value: any): value is Observable<T> {
   );
 }
 
+function isAsyncIterable(input: unknown): input is AsyncIterable<unknown> {
+  return (
+    typeof input === 'object' && input !== null && Symbol.asyncIterator in input
```
n1ru4l, line 1684: https://caniuse.com/mdn-javascript_builtins_symbol_asynciterator

n1ru4l: Should be all fine given that we don't support IE11 :)

packages/graphiql/src/components/GraphiQL.tsx
n1ru4l: Should be all fine given that we don't support IE11 :)

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -88,10 +88,15 @@ export type FetcherResult =
   | string
   | { data: any };
 
+export type FetcherReturnType =
+  | Promise<FetcherResult>
+  | Observable<FetcherResult>
+  | AsyncIterable<FetcherResult>;
```
n1ru4l, line 94: Still not 100% sure which typing is the best `AsyncIterable` vs `AsyncIterableIterator`.

Had the same discussion on the PR for defer and stream. `graphql`, `execute` and `subscribe` return `AsyncIterableIterator`, so I would go with that.

n1ru4l: With `AsyncIterable` we can potentially be more flexible with other libraries (as it is also compatible with `AsyncIterableIterator`).

The only difference for the fetcher implementation is for cleanup:

**`AsyncIterable`**: `asyncIterable[Symbol.asyncIterator]().return.?()`
**`AsyncIterableIterator`**: `asyncIterable.return?.()`

acao:
this looks great! im just wondering how we can best demonstrate this feature at this point? maybe just a quick entry in the readme showing the two alternatives to fetcher? 

or a docs/fetcher.md for some more in depth examples? been thinking of this as a way to break down the super long readme and to prepare for a docsite

eventually, i’d like to have a docsite where we could show many graphiql props permutations. we could use storybook or play room or use some other strategy.

n1ru4l:
For starters, we could just add another example to the readme directly calls `execute`/`subscribe` imported from the `graphql` package. I can take care of that.

n1ru4l:
Right now I tested the implementation by starting the GraphiQL dev server and pasting the following into **`packages\graphiql\resources\index.html.ejs`** :D

```html
<script>
      window.process = globalThis.process = {
        env: {
          NODE_ENV: "development"
        }
      }
    </script>
    <script type="module">
      import * as graphql from "https://unpkg.com/graphql@15.4.0/index.mjs?module"

      const schema = new graphql.GraphQLSchema({
        mutation: new graphql.GraphQLObjectType({
          name: "Mutation",
          fields: {
            ping: {
              type: graphql.GraphQLBoolean,
              resolve: () => true
            }
          }
        }),
        query: new graphql.GraphQLObjectType({
          name: "Query",
          fields: {
            ping: {
              type: graphql.GraphQLBoolean,
              resolve: () => true
            }
          }
        }),
        subscription: new graphql.GraphQLObjectType({
          name: "Subscription",
          fields: {
            count: {
              type: graphql.GraphQLInt,
              subscribe: async function * () {
                console.log("KEK")
                for (let counter = 0; counter < 10; counter++) {
                  yield { count: counter };
                  await new Promise(res => setTimeout(res, 1000))
                }
              }
            }
          }
        }),
      });

      const fetcher = (graphQLParams) => {
        const document = graphql.parse(graphQLParams.query);
        const operation = graphql.getOperationAST(document, graphQLParams.operationName);
        if (operation.operation === "subscription") {
          // AsyncIterable
          return graphql.subscribe(
            schema,
            document,
            {},
            {},
            graphQLParams.variables,
            graphQLParams.operationName
          )
        }

        // Promise or AsyncIterable
        return graphql.execute(
          schema,
          document,
          {},
          {},
          graphQLParams.variables,
            graphQLParams.operationName
        )
      }

      window.schemaFetcherContext = {
        schema,
        fetcher
      }

    </script>
```

You will also have to adjust **`packages\graphiql\resources\renderExample.js`** to use `window.schemaFetcherContext.fetcher`

acao:
thats nice to see it working, but we will need to keep that example as is for now.

lets just keep it to that readme entry for now, and we will come up with a way to show more detailed examples soon!

n1ru4l:
Yeah, I just want some way for you to properly test before shipping a broken GraphiQL :D I am currently working on the readme section. In the meantime, I also fixed some issues that I could identify with the custom fetcher.

acao:
@n1ru4l perhaps for now let's implement it in one of the examples, and then we can cross-reference to that from the readme?

packages/graphiql/src/components/GraphiQL.tsx
acao: all of this will be passed through browserslist and babel anyways

also to note re: `for await of`: afaik this will gaurantee serial execution, this is desired?

acao:
two more things:

1. just one unit test
2. implement an example [`here`](https://github.com/graphql/graphiql/blob/main/examples/graphiql-webpack/src/index.jsx), and then link to it from the `packages/graphiql/Readme`

acao:
now available in `graphiql@1.2.0`!

Issue #656, opened 2018-01-31
Pull root schema description from documentation comments in schema definition


jedahan:
It would be nice if this was overrideable:

https://github.com/graphql/graphiql/blob/master/src/components/DocExplorer/SchemaDoc.js#L38

kevinsimper:
Yes, that is a good idea!

PR #1731, opened 2020-12-11
fix: display schema description if available

pmantica1:
Resolves: https://github.com/graphql/graphiql/issues/656


pmantica1:
I am a novice with javascript so I would appreciate some help linting and test this.

PR #1739, opened 2020-12-18
feat: also support fetcher functions that return Promise<Observable> or Promise

n1ru4l:
…or Promise<AsyncIterator>

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -88,10 +88,11 @@ export type FetcherResult =
   | string
   | { data: any };
 
-export type FetcherReturnType =
-  | Promise<FetcherResult>
-  | Observable<FetcherResult>
-  | AsyncIterable<FetcherResult>;
+type MaybePromise<T> = T | Promise<T>;
+
+export type FetcherReturnType = MaybePromise<
+  FetcherResult | Observable<FetcherResult> | AsyncIterable<FetcherResult>
+>;
 
 export type Fetcher = (
   graphQLParams: FetcherParams,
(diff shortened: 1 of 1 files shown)


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -88,10 +88,11 @@ export type FetcherResult =
...
   | { data: any };
 
-export type FetcherReturnType =
-  | Promise<FetcherResult>
-  | Observable<FetcherResult>
-  | AsyncIterable<FetcherResult>;
+type MaybePromise<T> = T | Promise<T>;
+
+export type FetcherReturnType = MaybePromise<
+  FetcherResult | Observable<FetcherResult> | AsyncIterable<FetcherResult>
```
acao, line 94: can we have a type for or even export this union, and then re-use it below on line 938?

n1ru4l: Like https://github.com/graphql/graphiql/pull/1739/commits/cd7061a7204e687dcdbd129725f196fb6f07ba5a?

acao: yes, ty!

acao:
just need to run `yarn format`. i think there was a bug when we upgraded `lint-staged`

n1ru4l:
@acao I fixed the formatting and rebased upon `main`.

acao:
thanks! the next release I'll try to bundle more - is it cool if it's a week or two on this? if not, i can pre-release? the changelog is getting kinda mechanical and empty, and "release often" means something different in commmercial vs open source software... haha!

acao:
these improvements you've been making have been subtle and powerful! the multipart/wss helper i've been meaning to publish should probably use `AsyncInterator`, don't you agree? since we want to treat it as the "first class" implementation?

n1ru4l:
@acao I am a fan of smaller and frequent releases, but do as you like :) +1 for making AsyncIterator the "first class"!

PR #1741, opened 2020-12-21
use cheap invariant

### LINKED ISSUE #1740: BUG: assert is still a dependency
looking at #1719, I think the goal of the lib is to remove assert, but there are still a few cases of this lib using invariant from it:
- https://github.com/graphql/graphiql/blob/39f5e45a9d2bb469260b62ff8057cc988d853470/packages/graphql-language-service-cli/src/client.ts#L12
- https://github.com/graphql/graphiql/blob/8f0a4ac3331785994e4fe2d98637af0359314359/packages/graphql-language-service-interface/src/getDiagnostics.ts#L20

this causes breaks in webpack v5.

if the goal is a small bundle, i think keeping invaraint (from fbjs or other) may still be useful, but replacing it with a simple throw could be good, too. Let me know what direction yall wanna take it.

cc @danielrearden @acao

mattkrick:
Fix #1740 

assert is a node package that webpack 4 included automatically. 
Webpack 5 does not, so any dependency on it will bork the package.

With only a handful of uses, it doesn't make sense to add the invariant package that would minify away the messages, so we just do a quick & dirty to get it working again. can test this out by building it with webpack 5.

@acao let me know if this is what you had in mind.

### DIFF
--- a/packages/graphql-language-service-interface/src/getDiagnostics.ts
+++ b/packages/graphql-language-service-interface/src/getDiagnostics.ts
@@ -52,7 +52,7 @@ export const DIAGNOSTIC_SEVERITY = {
 const invariant = (
   condition: any,
   message: string,
-): asserts condition => {
+) => {
   if (!condition) {
     throw new Error(message);
   }
--- a/packages/graphql-language-service-interface/src/getDiagnostics.ts
+++ b/packages/graphql-language-service-interface/src/getDiagnostics.ts
@@ -51,7 +51,7 @@ export const DIAGNOSTIC_SEVERITY = {
 
 const invariant = (
   condition: any,
-  message: string | number,
+  message: string,
 ): asserts condition => {
   if (!condition) {
     throw new Error(message);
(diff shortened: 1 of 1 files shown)


mattkrick:
@acao looks like the build is already breaking the lint rules in other files. Would it be possible to get this merged in so webpack 5 folks can use the latest version?

acao:
you have both typescript and prettier errors, can you fix those?

acao:
thank you! sorry there seems to still be an issue with `lint-staged` config. 

this looks about good to merge. we used the same strategy in the last PR where we overlooked this one, including creating a similar invariant-style function 😆 or maybe i just threw when the condition failed inline? anyways, we can to shore this all up after this hotfix, thank you

after this, I'd like to update our examples repo for webpack 5, which I haven't worked with much yet. maybe keep the 4.x version and show a 5.x version as well, same for monaco until graphiql implies monaco

I have another quick PR to fix the examples. I removed the examples directory from workspaces because it made the yarn cache faster, but then they don't resolve correctly at all and it's a problem.

acao:
ah one last thing, if you can, run `yarn format`

acao:
image

PR #1742, opened 2020-12-22
feat: interfaces implementing interfaces for the LSP

acao:
# Features

This introduces a few new language features for `implements` syntax, including the new(ish) `&` operator, both for interfaces & types.

## Parsing & Highlighting

This implements support for & in the parser, which means highlighting support for codemirror graphql

## Autocompletion

Previously there was no autocompletion for any of this

```graphql
type Interface implements So▌
```
and
```graphql
type Interface implements Something & So▌
```

monaco-autocomplete-interfaces

now we have 

## Inline Autocompletion

it will also autocomplete for inline interfaces that may not be in the schema yet, for writing SDL files for example.

```graphql
interface Example { id: String }
type Interface implements Ex▌
```

## Where this feature works

- parser & interface
- `graphql-language-service-server`
- `vscode-graphql`
- `monaco-graphql`
- `codemirror-graphql` highlighting for & syntax works because of the parser implementation in this PR

## HTD

(netlify)
1. open the monaco netlify demo linked in CI jobs below
5. provide github personal access token to demo (no scopes needed, delete when finished from localstorage! oauth demo coming soon)
6. begin building types and interfaces that implement the github interfaces and eachother
7. confirm autocompletion works everywhere between the keyword “implements” and the first directive or { bracket

(manual)
1. clone repo
2. yarn
3. yarn build
4. yarn start-monaco
5. follow from step 2 above to the end

To see how this will impact graphiql, run yarn start-graphiql, and you’ll see the highlighting/parsing work for & syntax now.

## What's Next
- `codemirror-graphql` has an additional step for implementing these new autocompletion features. @yoshiakis and I found that codemirror-graphql’s hint.js is mostly a dupe of `getAutocompleteSuggestions`. So in the next PR, we will be able to replace most of whats in hint.js for codemirror with `getAutocompleteSuggestions`. This will make feature parity between monaco-graphql and codemirror-graphql much easier to maintain
- autocompletion for interface fields when extending an interface
- consider adding more validation rules for interfaces or object types (spec decision)

### DIFF
--- a/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
@@ -21,6 +21,8 @@ import {
   visit,
   VariableDefinitionNode,
   GraphQLNamedType,
+  isInterfaceType,
+  GraphQLInterfaceType,
 } from 'graphql';
 
 import {
@@ -84,6 +86,7 @@ export function getAutocompleteSuggestions(
 
   const kind = state.kind;
   const step = state.step;
+
   const typeInfo = getTypeInfo(schema, token.state);
   // Definition kinds
   if (kind === RuleKinds.DOCUMENT) {
@@ -96,13 +99,17 @@ export function getAutocompleteSuggestions(
     ]);
   }
 
+  if (kind === RuleKinds.IMPLEMENTS) {
+    return getSuggestionsForImplements(token, state, schema, queryText);
+  }
+
   // Field names
   if (
     kind === RuleKinds.SELECTION_SET ||
     kind === RuleKinds.FIELD ||
     kind === RuleKinds.ALIASED_FIELD
   ) {
-    return getSuggestionsForFieldNames(token, typeInfo, schema, kind);
+    return getSuggestionsForFieldNames(token, typeInfo, schema);
   }
 
   // Argument names
@@ -154,7 +161,7 @@ export function getAutocompleteSuggestions(
     (kind === RuleKinds.OBJECT_FIELD && step === 2) ||
     (kind === RuleKinds.ARGUMENT && step === 2)
   ) {
-    return getSuggestionsForInputValues(token, typeInfo, kind as string);
+    return getSuggestionsForInputValues(token, typeInfo);
   }
 
   // complete for all variables available in the query
@@ -240,8 +247,6 @@ function getSuggestionsForFieldNames(
   token: ContextToken,
   typeInfo: AllTypeInfo,
   schema: GraphQLSchema,
-  // kind: RuleKind.SelectionSet | RuleKind.Field | RuleKind.AliasedField,
-  _kind: string,
 ): Array<CompletionItem> {
   if (typeInfo.parentType) {
     const parentType = typeInfo.parentType;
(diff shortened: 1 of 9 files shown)


packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -320,6 +324,56 @@ function getSuggestionsForInputValues(
   return [];
 }
 
+function getSuggestionsForImplements(
```
acao, line 327: @IvanGoncharov here is the working implementation so far!

acao:
@yoshiakis so I've gotten this working in `codemirror-graphql`... by replacing `hint.js` almost entirely 🤯. only one failing test left, but that doesn't mean there won't be other regressions

I decided to do what we discussed a while ago, and made `type` be passed by `getAutocompleteSuggestions` in interface itself.

so far, just this is working for all but one test:

```js

CodeMirror.registerHelper('hint', 'graphql', (editor, options) => {
  const schema = options.schema;
  if (!schema) {
    return;
  }

  const cur = editor.getCursor();
  const token = editor.getTokenAt(cur);
  const rawResults = getAutocompleteSuggestions(
    schema,
    editor.getValue(),
    token,
  );
  const tokenStart =
    token.type !== null && /"|\w/.test(token.string[0])
      ? token.start
      : token.end;
  const results = {
    list: rawResults.map(item => ({
      text: item.label,
      type: item.type ? String(item.type) : String(item.detail),
      description: item.documentation,
      isDeprecated: item.isDeprecated,
      deprecationReason: item.deprecationReason,
    })),
    from: { line: cur.line, column: tokenStart },
    to: { line: cur.line, column: token.end },
  };

  if (results && results.list && results.list.length > 0) {
    results.from = CodeMirror.Pos(results.from.line, results.from.column);
    results.to = CodeMirror.Pos(results.to.line, results.to.column);
    CodeMirror.signal(editor, 'hasCompletion', editor, results, token);
  }

  return results;
});
```

acao:
this would significantly reduce the effort of introducing new language features to `codemirror-graphql`

yoshiakis:
I got it! It makes sense to do so. Though in that case, I think we should change [this part](https://github.com/graphql/graphiql/blob/d81db0c67e5c9fd197a5f0aac4a581947ceccf04/packages/graphql-language-service-server/src/MessageProcessor.ts#L489) so that `graphql-language-service-server` doesn't send information on type when is requested auto complete suggestions list in order to reduce the amount of data sent.

yoshiakis:
@acao Why do you want to change the type of `item.type` to a string type?

acao:
> @acao Why do you want to change the type of `item.type` to a string type?

this is what caused certain tests to pass, it seems to be that the tests expect type to be a string

acao:
`@yoshiakis I will introduce all of that as a follow up PR. The only impact this PR has on codemirror-graphql` is at the parsing level (& syntax highlighting will work now, for example).
THen, in the next PR we can add autocomplete, and simplify the codemirror mode in the process hopefully!

acao:
merging this *without* `codemirror-graphql` support, important to note again. that is coming in a follow up PR

PR #1747, opened 2020-12-28
feat: merge completion logic (for implements &, variables)

acao:
in order to introduce new language features across language server and browser ecosystems, here we combine the autocomplete logic in `codemirror-graphql` and in the `graphql-language-service-interface`.

# `&` and `implements`
- introduces #1742 parsing and autocomplete logic for `&` syntax to `codemirror-graphql` and thus `graphiql`
- address several edge cases present in #1742 at the LSP level
- `implements` logic now covers:
	- filtering out interface suggestions which are already in the `implements &` list
	- filtering out the interface suggestion for the name of the interface you're extending
	- show both schema interfaces and interfaces in the file
	- show interface name suggestions as you type
	- (these bugs were introduced because we decided to include interfaces which aren't part of `schema` yet, but are inline in the file, to avoid confusion)

# completion for variables
- because we are combining autocomplete logic, the variables completion logic used by the LSP server was causing issues in GraphiQL
- to ameliorate this, we remove usage of `parse` for variable completion that I introduced because it was causing EOF errors. use token parser instead, like the rest of the `getAutocompleteSuggestions` library. then we are able to parse incomplete graphql operations and SDL
- only show variables when they are compatible with the argument
- show variable types (previously it was just name)
- show available variables before you start typing, including alongside enum or boolean completion options

cc: @yoshiakis 

notes:

- yes, we are adding `type` to `CompletionItem` results, but for the language server side, it will be converted to a string or set to `undefined` by the time the server is issuing the response to the LSP client (since it transmits over IPC, streams or sockets)


acao:
ah, the last commit fixes SO many bugs and adds so many improvements but i forgot a flow type fix and will need to come back to this

acao:
@yoshiakis another huge series of improvements in the last commit!

the variables completion was using graphql parse which just broke the heck out of codemirror-graphql. despite being new functionality, the EOF parse bug kept happening for incomplete queries, so i replaced it with `runOnlineParser`

several other edge cases fixed, all summarized in the commit message

last step is to test it all thoroughly with the LSP server! seems to be working working great in codemirror-graphql, graphiql and monaco-graphql with manual testing

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -815,6 +890,13 @@ export function getTypeInfo(
         if (state.name) {
           type = schema.getType(state.name);
         }
+        // TODO: collect already extended interfaces of the type/interface we're extending
+        //  here to eliminate them from the completion list
+        // because "type A extends B & C &" should not show completion options for B & C still.
+
+        // if(state.prevState?.kind === RuleKinds.IMPLEMENTS) {
+        //   interfaceDef?.addInterface(new GraphQLInterfaceType({ name: state.name}))
```
acao, line 898: @IvanGoncharov this is the area I was talking about, not sure how best to do this other than cloning the interfaceDef instead of being able to mutate it?

acao: @danielrearden this is where I need to clone and re-create the interface type, but I also need to determine when the interface definition is complete. I think this is when state.kind === `INTERFACE_DEF` AND step === 2? so then i can do interfaceDef = null

PR #1750, opened 2021-01-05
feat: implied or external fragments, resolves #612

### LINKED ISSUE #612: Plugin: Gatsby-style implicit fragments
I'm [adding support to Gatsby for components/plugins to ship with fragments](https://github.com/gatsbyjs/gatsby/pull/2320) that can be used out-of-the-box when building sites.

It'd be great if there was a way to add these to GraphiQL. Right now there's no way to add a fragment other than changing the default text and that still wouldn't work as then graphiql would fail as there'd be unused fragments.

If it's possible to add fragments to the schema and I just missed that, would love a pointer there!

acao:
allows external, implied fragments in autocompletion, validation, and operation execution. resolves #612 

- adds this capability to `monaco-graphql`, `codemirror-graphql`, and `graphiql`
- provide an array of fragment definitions or a string containing fragment definitions
- only fragments you use are added in operation exec, thus in fetcher params as well
- still only shows valid autocomplete entries for fragment spread
- also exposes a documentAST of the current operation in `onEditQuery` and in the `fetcher`
- concatenate fragments in codemirror for linting

todo:

- [x] code completion for GraphiQL, `monaco-graphql`, `codemirror-graphql`
- [x] operation execution
- [x] validation
- [x] test coverage


acao:
@sgrove note this change introduces use of internally and exposes a single AST document, that only changes on editor or props-driven query change. it's been added to `onEditQuery` and to the fetcher params. anywhere else that something like this would be handy for your plugin needs?

PR #1753, opened 2021-01-07
fix: add missing nullthrows dependency to utils

Rugvip:
Fixes the issue seen in https://github.com/backstage/backstage/issues/3961

Projects that don't use `graphql-language-service-server` may run into this issue, since that's the only package that declares the existing `nullthrows` dependency.


acao:
oh good catch, thank you!

Rugvip:
No problem :grin:

Can recommend https://www.npmjs.com/package/eslint-plugin-import with the `no-extraneous-dependencies` rule for avoiding this issue btw!

PR #1758, opened 2021-01-10
[WIP] feat: field and value def autocomplete

stonexer:


packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
```diff
@@ -476,4 +476,51 @@ query name {
       { label: 'Character' },
       { label: 'TestInterface' },
     ]));
+
+  it('provides correct types suggestions when typing a typename after a field', () => {
+    expect(testSuggestions('type Book { id: I', new Position(0, 18))).toEqual([
+      { label: '__DirectiveLocation' },
+      { label: '__TypeKind' },
+      { label: 'Episode' },
+      { label: 'InputType' },
```
stonexer, line 485: This is probably an existing bug, as the `InputType` should probably not appear here

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
```diff
@@ -246,6 +248,16 @@ export function getAutocompleteSuggestions(
     return getSuggestionsForVariableDefinition(token, schema, kind);
   }
 
+  // Input value definition
+  if (kind === RuleKinds.INPUT_VALUE_DEF && step === 2) {
```
stonexer, line 252: I'm not particularly sure what step means here, but I just found it to be representative of the current state. Hope to get some advice :)

acao: yep, that's it! essentially steps for a parser rule to advance.

here's where the magic happens in parser

https://github.com/graphql/graphiql/blob/c4cba85e869105f5363ebcb3826b4cf7b7e62c00/packages/graphql-language-service-parser/src/onlineParser.ts#L256-L256

this is probably the most straightforward implementation example. with ARGUMENT, step 1 is when we want argument name, and step 2 is when we want an input value. 

https://github.com/graphql/graphiql/blob/c4cba85e869105f5363ebcb3826b4cf7b7e62c00/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts#L161

so, if i wanted to implement autocompletion for field names for object types that extended interfaces, i would do that on `OBJECT_TYPE` step 1.

packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
acao: there's probably some simple logic we could add here for that if you want

image

to access the GraphQLObjectType in this case, I think you want `state.prevState.type`. you should be able to getFields() and then match the fieldname up to check the type

acao:
awesome work so far! works with object type fields. would you like me to look into a way to make it so that the autocomplete can begin a space after the `fieldName:`, rather than immediately? we have the same issues where i improved this for arugment variables

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
acao: oh and i guess you found one that has three steps! :D

packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
stonexer: Thanks! But I don't seem to understand the role of field name here. 

In my current thinking, since `state.prevState.type`(type Book) is a GraphQLObjectType, and `InputType` is a GraphQLInputObjectType, so it should not appear in the hintList.

acao: my bad, misunderstood you before. i understand this bug better now. i’ll circle back after some sleep

stonexer:
> would you like me to look into a way to make it so that the autocomplete can begin a space after the `fieldName:`, rather than immediately?

Sure, that's what I want too.

I noticed that in the typescript file, when I type ...

```ts
interface Author {
  name: string;
}

interface Book {
  author:  ...
}
```

I don't get an active suggestion after I type `:` or `:` with a space. But then if I trigger ctrl+space manually, I can also have the full suggestion.

The difference is that when I type the first letter here, like `A`, there is an active suggestion directly.

I think it would be nice to have a similar logic here.

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
imolorhe: Just thinking aloud.. is it possible for the steps to have constants for referencing them? In order to give the steps meaning beyond just `2` for example? 🤔

acao: good question! they have a different meaning in the context of each rule. we could have `Step.First` but that wouldn't be much more readable, or would it?

imolorhe: Not really, it would pretty much be the same thing 😄 I'll think about it a bit more.

acao: we could also specify enums for each rule that has multiple steps that are referenced, such as `ObjectTypeSteps.Name = 1` and `ObjectTypeSteps.Field = 2`, and then `FragmentDefintionSteps.Name = 1`, etc

acao:
@stonexer that's so odd, because it works in graphiql and monaco-graphql:

image

packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
imolorhe: Maybe that would make sense, since the meaning of each step is defined by the GraphQL type being worked on. My only concern is that, since the values overlap, there might be an issue where they are mixed up. Typescript wouldn't complain since they are all numbers (except they are typed with the right enum type 🤔 ) but the behavior wouldn't be the expected behavior.

For example, `InputObjectTypeSteps.Field` might be equal to `RuleKindStep.Name`. Maybe my concern is a non-issue though. Perhaps if there is a way to type guard `step` to one of the enums, that would make `step` have just one meaning and remove ambiguity.

stonexer:
@acao Haha，it's not a bug here. Actually I mean the experience of using monaco to edit typescript files.

imolorhe (approved):
LGTM. Only concern I see would be the dropped test coverage. @acao is there a requirement for PRs not to reduce the test coverage?

acao:
@imolorhe that is definitely something to consider! I disabled that once because a PR dropped coverage because of a refactor (I deleted so much redundant code with good coverage that it caused overall coverage to go down). I evaluate it on a case-by-case basis as a maintainer, but for this PR I think we should have additive coverage, though we're adding three cases.

I don't think this PR is ready myself because of the issue with spaces we discussed in discord, unfortunately we didn't surface that discussion here so let me summarize:

**desired behavior:**

autocompletion for variables should be triggered on `: `, non-breaking space included. ideally, the mode/ide implementation should be configured to advance the cursor the extra space automatically.

**current behavior:**
autocomplete behavior is triggered immediately following a `:` character, however we want there to be an additional space. if the user adds a space after the `:` character, then the completion options vanish. this is essentially the opposite behaviour.

imolorhe:
Okay got it. Will un-approve and wait for that fix then 😄

stonexer:
> autocompletion for variables should be triggered on `: `, non-breaking space included. ideally, the mode/ide implementation should be configured to advance the cursor the extra space automatically.

In https://github.com/graphql/graphiql/blob/main/packages/monaco-graphql/src/languageFeatures.ts#L196

Are there other situations where we need `':'` to trigger? Can we simply remove the `':'` from them

acao:
good question! we can experiment with that

stonexer:
> good question! we can experiment with that

Tried it, I think it's not bad

acao:
this adjustment looks better for monaco editor, but it still doesn't fix the issue on the codemirror side. gonna keep looking into this!

acao:
so sad i didn't merge this sooner! this is incredible

acao:
@stonexer this is sooo close - maybe we can configure `codemirror-graphql` to handle the `insertText` param ourselves, since that's at this point only monaco specific?

acao:
image

i would like to have this case covered, as I think it's one of the most important ones for the LSP. Many tools implement their own autocompletion hack on top of this to achieve it here

acao:
@dwwoelfel this is another good place for the tokenizer

stonexer:
@acao I'm back haha :), sorry for disappearing for a long time before for personal reasons. I added `\s` in `AUTO_COMPLETE_AFTER_KEY`, and suggestions are auto triggered on typing space. Could you help to have a review again?

acao:
@stonexer so glad you are back! definitely can review again soon, yes!

acao:
@stonexer I'd love to revive this effort soon! it's so close!

acao:
@stonexer I've been looking at [chevrotain's grammar](https://github.com/Chevrotain/chevrotain/tree/master/examples/grammars/graphql) for graphql since last year, and it's enticing

I think it's worth looking at replacing the language parser with this entirely moving forward.

PR #1770, opened 2021-01-19
feat: @defer, @stream, websockets support in a createGraphiQLFetcher utility

acao:
This introduces a backwards compatible client wrapper that provides `graphql-ws` subscriptions specification, as well as support for `@defer` and `@stream` [`IncrementalDelivery`](https://github.com/graphql/graphql-over-http/blob/master/rfcs/IncrementalDelivery.md) spec.

It wraps clients such as `fetch-multipart-graphql`, `graphql-ws` and even legacy support for `subscriptions-transport-ws`

new npm packages:

- `@graphiql/toolkit` will become a set of types and utilities for building tools like GraphiQL. it will here prevent a minor cyclic dependency issue between the fetcher module and `graphiql` itself, which impacts typescript project references. breaking out `graphiql` into more component packages as such will enable other architectural changes for 2.0
- `@graphiql/build-fetcher` is the standalone utility for generating a `Fetcher` instance that for this iteration will support http multipart IncrementalDelivery as well as graphql-ws transport

important changes:

- `graphiql` `Fetcher` prop now supports an array of incremental responses.

### Todo
- [x] new packages and utility and some integration tests
- [x] demo implementation
- [x] working server implementation of both `graphql-ws` & multipart with `express-graphql` defer stream release tag
- [ ] more unit & integration test coverage 
- [ ] more documentation
- [x] end to end tests

### how to demo locally

from the repo root:

1. `yarn --force && yarn build`
1. `yarn start-graphiql`

then this query should be enough to get started:
```graphql 
{
  streamable @stream(initialCount: 2) {
    text
  }
}
```

### demo in netlify?

https://deploy-preview-1770--graphiql-test.netlify.app/?query=%7B%0A%20%20isTest%0A%20%20stream%3A%20streamable(delay%3A%20500)%20%40stream(initialCount%3A%202)%20%7B%0A%20%20%20%20text%0A%20%20%7D%0A%20%20anotherStream%3A%20streamable(delay%3A%20200)%20%40stream(initialCount%3A%200)%20%7B%0A%20%20%20%20text%0A%20%20%7D%0A%7D%0A

### feedback welcome!

I'm interested in feedback on all working features, including the new utility, and the GraphiQL `IncrementalDelivery` implementation. (anything except documentation, tests, still some polish needed there, haha!)

### Credit

`@defer` & `@stream` support has been a HUGE effort across the GraphQL community.
Spec development and libraries by @robrichard @lilianammmatos and @enisdenjo
Prior Art from @urigo and @n1ru4l


acao:
cc @n1ru4l @enisdenjo @robrichard @lilianammmatos @urigo: you all might find this PR interesting as I'm using your all's code.  I figured you might know of some important common configuration settings I might be overlooking for the first iteration?

packages/graphiql-build-fetcher/src/lib.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @returns {Client | null}
+ */
+export const createWebsocketsClient = (
+  options: BuildFetcherOptions,
+): Client | null => {
+  let wsClient: Client | null = null;
+  try {
+    try {
+      // TODO: defaults?
+      wsClient = createClient({
```
acao, line 53: @enisdenjo so, this is the default pattern, if only `subscriptionsUrl` is provided. the user can optionally provide their own instance of a client, which would bypass this behavior entirely. 

the idea of a `wsClientOptions` option came to mind, however a `wsClient` option itself seemed even more powerful. do you think a `wsClientOptions` with `graphql-ws`'s `ClientOptions` should be offered as well, here?

enisdenjo: Other runtime options can be configured and I would recommend allowing the user to do so. You may have a custom server configuration and might need to fine tune the client so that they work in coherence.

But, this can be achieved by providing your own client instance, so, `subscriptionsUrl` + `wsClient` would be just enough.

packages/graphiql-build-fetcher/package.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "license": "MIT",
+  "main": "dist/index.js",
+  "module": "esm/index.js",
+  "scripts": {
+    "server": "ts-node test/server.ts"
+  },
+  "dependencies": {
+    "graphql-ws": "^4.1.0",
+    "graphql-transport-ws": "^1.9.0",
```
enisdenjo, line 21: Why are you using the deprecated [`graphql-transport-ws`](https://www.npmjs.com/package/graphql-transport-ws) lib? Its just an old name for `graphql-ws`, the two are fundamentally the same. [`graphql-ws@1.9.1`](https://github.com/enisdenjo/graphql-ws/releases/tag/v1.9.1) = [`graphql-transport-ws@1.9.0`](https://github.com/enisdenjo/graphql-ws/releases/tag/v1.9.0)

I wouldnt advocate the usage of the deprecated library at all. The full focus should be on `graphql-ws`.

```suggestion
```

packages/graphiql-build-fetcher/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+const url = 'https://myschema.com/graphql';
+
+const subscriptionsUrl = 'wss://myschema.com/graphql';
+
+const fetcher = buildGraphiQLFetcher({
+  url,
+  wsClient: createClient({
+    url: subscriptionsUrl,
+    keepAlive: 2000,
```
enisdenjo, line 117: I think that the `keepAlive` is superfluous here. Its mainly useful in applications which have a lot of moving parts where you dont want to waste resources (a lot of connects/disonnects to a WS server).

I would, however, recommend using the `lazy: false` option. This way GraphiQL would immediately connect to a server and the user can know right away if WS works (instead of being forced to issue a subscription operation just to check the basics).

[Something like what PostGraphile's PostGraphiQL does: shows a status badge on the top right indicating the state of the WS connection.](https://github.com/graphile/postgraphile/blob/af39c3dcea40646c9217785fa93a9601612c2b83/postgraphiql/src/components/PostGraphiQL.js#L666-L725)

```suggestion
```

packages/graphiql-build-fetcher/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export interface BuildFetcherOptions {
+  /**
+   * url for HTTP(S) requests. required!
+   */
+  url: string;
+  /**
+   * url for websocket subscription requests
+   */
+  subscriptionsUrl?: string;
```
enisdenjo, line 14: The WebSocket transport supports _all_ GraphQL operations, including streaming queries (like the `@live` or `@deferred` directive).

I guess WS would be used only for subscriptions, which is why the prop makes sense; but, ultimately you can communicate with GQL only through WS if you wish.

Just a thought.

packages/graphiql/package.json
```diff
@@ -62,15 +63,19 @@
...
     "cross-env": "^7.0.0",
     "css-loader": "3.4.2",
     "cssnano": "^4.1.10",
     "express": "4.17.1",
-    "express-graphql": "0.9.0",
+    "express-graphql": "experimental-stream-defer",
     "fork-ts-checker-webpack-plugin": "4.1.3",
     "graphql": "15.4.0",
+    "graphql-ws": "^4.1.0",
+    "graphql-transport-ws": "^1.9.0",
```
enisdenjo, line 77: Same as above, `graphql-transport-ws` is deprecated. I dont recommend using it!

enisdenjo (changes requested):

packages/graphiql-build-fetcher/src/lib.ts
enisdenjo: Thinking about it a bit more, I'd expose the `connectionParams` too. I see it being used a lot because it is the "headers" of the WS connection.

packages/graphiql-build-fetcher/package.json
acao: we want to have backwards compatibility with the old transport spec since it's so widely used?

packages/graphiql-build-fetcher/src/types.ts
acao: yes we're familiar with this pattern, however the intention of this library is to support the more widely adopted transport spec implementations, the "90%+ case" if you will.

packages/graphiql-build-fetcher/README.md
acao: oh yeah i was just adding an arbitrary setting so people could see that more than just URL is available. `lazy: false` sounds like a good idea, I can add that to the default implementation

packages/graphiql-build-fetcher/package.json
enisdenjo: Thats [`subscriptions-transport-ws`](https://github.com/apollostack/subscriptions-transport-ws).

acao: is `graphql-ws` backwards compatible with older client implementations?

packages/graphiql-build-fetcher/src/types.ts
acao: if someone wants to ship a fetcher for that pattern in a standalone package like this one, they are welcome to

acao:
@enisdenjo another issue I had is that I couldn't get the `graphql-ws` implementation working with `express-graphql`. I'm focusing on the multipart side right now since I was able to get that working

packages/graphiql-build-fetcher/package.json
enisdenjo: [`graphql-ws` implements the new Protocol exclusively.](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) It is not interoperable with [Apollo's Protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md).

A WebSocket can use **just one subprotocol**; therefore, built-in interoperability is simply not feasible. Read more about this topic [here](https://github.com/enisdenjo/graphql-ws/issues/83) and [here](https://github.com/graphql/graphql-over-http/pull/140#issuecomment-739823000).

If you want to support both the deprecated Apollo's lib and the new one, [you have to configure a connection delegator on the server side](https://github.com/enisdenjo/graphql-ws#ws-backwards-compat).

enisdenjo:
[If you're using `express-graphql` out of the box, it has `subscriptions-transport-ws` built-in.](https://github.com/graphql/express-graphql/blob/0fa7ace4f94fc25708a9f1e68a29775ceb623a25/package.json#L98) The two libraries are not interchangeable. I think @junminstorage is cooking something up in this regards.

In the meantime, [you have to configure `express-graphql` following this recipe](https://github.com/enisdenjo/graphql-ws#express) in order to support the new Protocol.

packages/graphiql-build-fetcher/package.json
acao: @benjie what is your take on this? from what I understand, we can just attempt a connection with `graphql-ws` protocol and if it fails, fail over to apollo's protocol. that's what we are doing here in this PR. @enisdenjo the reality here is that most of our users using websockets will probably still be using the old protocol for now and not want to change just for graphiql, even though it's being deprecated. we could release with only `graphql-ws` support, but it will probably lead to a flurry of bug reports and confusion even if we document this. 

that said, pointing out that the apollo protocol is deprecated might be enough, but we have users using the old apollo protocol in many other language implementations still.

ad-hoc standards are so confusing, good thing you're finalizing all of this into an adoptable standard. maybe the hard luck approach is best, and if they need backwards compatibility

packages/graphiql-build-fetcher/src/types.ts
n1ru4l: Would it be possible to create a simple option for using WebSocket for everything exclusively or an optional function configuration option that receives the operation payload and then depending on the returned value of that function (e.g. `http` or `ws`) the corresponding protocol will be used?

acao:
> [If you're using `express-graphql` out of the box, it has `subscriptions-transport-ws` built-in.](https://github.com/graphql/express-graphql/blob/0fa7ace4f94fc25708a9f1e68a29775ceb623a25/package.json#L98) The two libraries are not interchangeable. I think @junminstorage is cooking something up in this regards.
> 
> In the meantime, [you have to configure `express-graphql` following this recipe](https://github.com/enisdenjo/graphql-ws#express) in order to support the new Protocol.

i already tried the recipe and it didn't work. i’m not referring to the graphiql implementation in `express-graphql` obviously, since this new module is what will be used to replace that

enisdenjo:
Ahh, alright then. Please share the problem you faced and I will investigate.

packages/graphiql-build-fetcher/src/types.ts
acao: folks are welcome to create their own standalone fetcher packages like these for other patterns! we are only interested in adopting widely used, advanced transport specs for this effort

packages/graphiql-build-fetcher/package.json
enisdenjo: Yes, thats correct. We try using `graphql-ws` and if the socket closes with the close event: `1002: Protocol Error` its very likely that the subprotocol is denied ([thats how Apollo's WS server denies invalid subprotocols](https://github.com/apollographql/subscriptions-transport-ws/blob/3590e076f549744af717dd3257fce78a659b7176/src/server.ts#L148)). In this case, fall back to `subscriptions-transport-ws`.

acao:
@enisdenjo i fixed it! i've been battling with `graphql` version clashes though, since we need this in `experimental` branch

packages/graphiql-build-fetcher/src/lib.ts
acao: ah, fantastic idea!

packages/graphiql-build-fetcher/package.json
benjie: Specifically this line of code (`"graphql-transport-ws": "^1.9.0"`) should be removed, @enisdenjo authored both and renamed `graphql-transport-ws` to simply `graphql-ws` :+1: 

---

On the broader subject of `subscriptions-transport-ws` vs `graphql-ws` support; I also agree with @enisdenjo - the best approach is to try connecting with `graphql-ws` and if that connection fails then try again using `subscriptions-transport-ws`. Even though probably 95% of GraphQL APIs that support websockets currently don't support `graphql-ws`, in future this should shift since Apollo's protocol is officially unmaintained; we should skate to where the puck is going not where it has been. It's possible for servers to support both protocols (as the next release of PostGraphile does), in that case `graphql-ws` should win.

acao: @benjie awesome! thats what we're doing in this PR already, just convincing @endisdenjo that it makese sense

acao: I think what I meant to add here is this:

https://www.npmjs.com/package/subscriptions-transport-ws

got `graphql-transport-ws` and `subscriptions-transport-ws` mixed up sorry!

acao:
A bunch more changes are coming! including an implementation in netlify

packages/graphiql-build-fetcher/package.json
acao: @enisdenjo my apologies for taking so long to realize this 😆  I knew something was wrong when I couldn't import the `SubscriptionsClient` class I was used to!

enisdenjo: Haha, thats ok! Glad you figured it out. 😄

junminstorage:
> [If you're using `express-graphql` out of the box, it has `subscriptions-transport-ws` built-in.](https://github.com/graphql/express-graphql/blob/0fa7ace4f94fc25708a9f1e68a29775ceb623a25/package.json#L98) The two libraries are not interchangeable. I think @junminstorage is cooking something up in this regards.
> 
> In the meantime, [you have to configure `express-graphql` following this recipe](https://github.com/enisdenjo/graphql-ws#express) in order to support the new Protocol.

I am also building a fetcher support for both graphql-ws and subscriptions-transport-ws on express-graphql's graphiql IDE, it looks like this fetcher can be shared and reused. Will a umd version being published?

acao:
> > [If you're using `express-graphql` out of the box, it has `subscriptions-transport-ws` built-in.](https://github.com/graphql/express-graphql/blob/0fa7ace4f94fc25708a9f1e68a29775ceb623a25/package.json#L98) The two libraries are not interchangeable. I think @junminstorage is cooking something up in this regards.
> > In the meantime, [you have to configure `express-graphql` following this recipe](https://github.com/enisdenjo/graphql-ws#express) in order to support the new Protocol.
> 
> I am also building a fetcher support for both graphql-ws and subscriptions-transport-ws on express-graphql's graphiql IDE, it looks like this fetcher can be shared and reused? Will a umd version being published?

hey great question! this library will ship as `GraphiQL.buildFetcher()` only in the umd bundle for graphiql. a standalone umd would be cool, but i need to overhaul the build tooling yet again to do that properly haha

junminstorage:
That could be just fine 😄   express-graphql's graphiql IDE already has dependency on graphiql umd bundle. 

About this buildFetcher() method's signature , I can confirm passing ```subscriptionsUrl```  and ```wsClient``` either in graphql-ws or subscriptions-transport-ws will just be fine. Or even better, if we can pass in another string to tell this method which wsClient is being passed, in stead of  "try connecting with graphql-ws and if that connection fails then try again using subscriptions-transport-ws", which is consistent with with graphile/postgraphile#1338 as well.

acao:
good news! you might want to try a field called `streamable` in the graphiql netlify preview 😎

acao:
graphiql-defer-stream

acao:
oh dang, the deploy preview doesn't use the local schema.. :/ forgot, fixing that

acao:
graphiql-defer-stream

maraisr:
May I suggest we wrap [`meros`](https://github.com/maraisr/meros) instead of fmg. A feather light runtime, that's faster, and also runs on the server. So for isomorphic apps, there is no need to wrap 2 clients.

acao:
@maraisr i was just pondering that a little bit ago! i saw it linked in the spec docs and was gonna check it out tomorrow. would you like to create a PR against this PR to replace it?

acao:
I think we're good here! Any necessary improvements can come in later iterations.

PR #1774, opened 2021-01-26
improvement: move incremental client from fmg to meros

maraisr:
With this PR I aim to replace fmg with meros, for better isomorphic nature and ease of use.

---

relates: #1770


acao:
ah, yeah you have to run `yarn build` at the root first!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1077,38 +1075,10 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
-                isWaitingForResponse: false,
-                response: GraphiQL.formatResult(response),
-              });
-            } else {
-              this.setState({
-                isWaitingForResponse: false,
-                response: GraphiQL.formatResult(result),
-              });
-            }
+            this.setState({
```
acao, line 1078: without aggregating `totalResponse` this won't work, I don't thinK?

maraisr: ya. haha correcting.

acao:
these test failures are strange!

packages/graphiql-create-fetcher/src/lib.ts
```diff
@@ -124,24 +127,40 @@ export const createLegacyWebsocketsFetcher = (
...
         ...fetcherOpts?.headers,
       },
-      onNext: parts => {
-        // @ts-ignore
-        sink.next(parts);
-      },
-      onError: sink.error,
-      onComplete: sink.complete,
+    }).then(response => {
+      // TODO: Can we make this payload type-safe?
```
acao, line 145: I think we would expect `FetcherResult` here right?

maraisr: ya, right!

acao:
ah, still the unrelated test failures that shouldn't be happening. ill take a look!

acao:
@maraisr you need to update the tests because you change how `createMultipartFetcher` is invoked!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1077,31 +1088,35 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
-                if (increment.path) {
-                  const [path, index] = increment.path;
-                  const data = result?.data[path] || [];
-                  // place them at the exact index. this matters a lot
-                  data[index] = increment.data;
-                  result.data = { ...result?.data, [path]: data };
-                } else {
-                  result.data = { ...result?.data, ...increment.data };
+            if (isIncrementalDelivery(result)) {
+              // @ts-ignore
```
acao, line 1092: shouldn't need to ts ignore this? i didn't locally?

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1066,7 +1070,14 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
         );
       }
 
-      const totalResponse: FetcherResult = { data: {} };
+      const isIncrementalDelivery = (
+        result: FetcherResult,
+      ): result is IncrementalDeliveryResult => {
+        // @ts-ignore
```
acao, line 1076: again, no `@ts-ignore` needed here

packages/graphiql-toolkit/src/types.ts
```diff
@@ -35,20 +35,23 @@ export type FetcherOpts = {
...
 
-export type FetcherResult =
+export type IncrementalDeliveryResult = {
+  data?: any;
+  errors?: any[];
+  path?: [string, number];
+  hasNext: boolean;
+};
+
+export type FetcherResultObject =
```
acao, line 45: how about `FetcherResultPayload` instead?

packages/graphiql/src/components/GraphiQL.tsx
acao: again, here's the code I had, no ts ignore needed:

```ts

 const totalResponse: FetcherResult = { data: {} };

      // _fetchQuery may return a subscription.
      const subscription = await this._fetchQuery(
        editedQuery as string,
        variables as string,
        headers as string,
        operationName as string,
        shouldPersistHeaders as boolean,
        (result: FetcherResult) => {
          if (queryID === this._editorQueryID) {
            if (typeof result !== 'string' && result.hasNext) {
              if (result.errors) {
                totalResponse.errors = [
                  ...(totalResponse?.errors || []),
                  ...result?.errors,
                ];
              }
              if ('path' in result) {
                const [path, index] = result.path;
                const data = totalResponse?.data[path] || [];
                // place them at the exact index. this matters a lot
                data[index] = result.data;
                totalResponse.data = { ...totalResponse?.data, [path]: data };
              } else {
                totalResponse.data = {
                  ...totalResponse?.data,
                  ...result?.data,
                };
              }
            }
            this.setState({
              isWaitingForResponse: false,
              response: GraphiQL.formatResult(totalResponse),
            });
          } else {
            this.setState({
              isWaitingForResponse: false,
              response: GraphiQL.formatResult(result),
            });
          }
        },
      );

      this.setState({ subscription });
      
  ```
and 
```ts
export type FetcherResult =
  | {
      data: IntrospectionQuery;
      errors?: Array<any>;
      hasNext?: boolean
    }
  | string
  | { data?: any; errors?: Array<any>; hasNext?: boolean }
  // for IncrementalDelivery via mereos
  | {
      data: any;
      errors?: any[];
      path: [string, number];
      hasNext: boolean;
    };

```

acao: the type gaurd is a nice pattern in theory but something is just not right if you need these ignores. if you try bringing that condition i wrote back inline, you'll see the inference engine picks it up, or it should at least?

packages/graphiql-toolkit/src/types.ts
```diff
@@ -35,20 +35,23 @@ export type FetcherOpts = {
...
   | {
       data: IntrospectionQuery;
       errors?: Array<any>;
```
acao, line 48: you just need to make `hasNext` optional here, and then you won't need to ts-ignore every time `hasNext` is referenced.

this was all I needed for the refactor of fetchQuery:

```ts
export type FetcherResult =
  | {
      data: IntrospectionQuery;
      errors?: Array<any>;
      hasNext?: boolean
    }
  | string
  | { data?: any; errors?: Array<any>; hasNext?: boolean }
  // for IncrementalDelivery via mereos
  | {
      data: any;
      errors?: any[];
      path: [string, number];
      hasNext: boolean;
    };
```

packages/graphiql-toolkit/src/types.ts
```diff
@@ -35,20 +35,23 @@ export type FetcherOpts = {
...
   // for IncrementalDelivery
-  | Array<{
-      data?: any;
-      errors?: any[];
-      path?: [string, number];
-      hasNext: boolean;
-    }>;
+  | IncrementalDeliveryResult;
+
+export type FetcherResult = FetcherResultObject | string;
```
acao, line 54: I don't think we need to compose them into distinct types to use seperately? `FetcherResult` infers the right type if you use it properly! were there issues with that for you?

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1066,7 +1070,14 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
+        return typeof result === 'object' && 'hasNext' in result;
```
acao, line 1077: `typeof n === 'object'` is super risky. for example: `const n = null; typeof n === 'object' // true`

packages/graphiql/src/components/GraphiQL.tsx
maraisr: Ya, I'll give it another check. For some reason I think the language server in vscode didn't complain, but did in IntelliJ.

I'll confirm, but when I did the type narrowing to object (instead of not string), it didn't like that "hasNext" was likely not in the object, seeing as the union finds the lowest common denominator. So in the typeguard I think it's fine. 

The rest should be removed though. 

Cheers! 🥂

packages/graphiql-toolkit/src/types.ts
maraisr: I think what I wanted here was to have a concrete type, as in to target a specific element in that union when i did my type guard.

But ya, think we can ditch that now. with the `hasNext?: boolean` thingo.

PR #1777, opened 2021-02-01
[RFC] fix: fix block string parsing in language parser

dwwoelfel:
This is an attempt to fix block string support in GraphiQL.

The goal is to make a query like the following parse properly in GraphiQL:

```graphql
{
  hasArgs(
    string: """      
      hello
      world
    """
  )
}
```

It seems like there shouldn't be logic about how block strings are constructed in `onlineParser.ts`, but I'm having a hard time keeping that knowledge solely in `Rules.ts`.  I'm finding it very tricky to write any logic that requires information from the previous line.


acao:
awesome, thank you for starting on this!

i’ll have to take a closer look later. this parser is designed to handle inline and multiline graphql strings the same. in terms of tracking data within rules, steps are the best way to do that. 

the change you made to `runOnlineParser` does seem out of place, however some change to the parser might be necessary

imolorhe:
62 files changed. Are all the changes in this PR related to fixing the block string parsing?

acao:
@imolorhe it's my bad, I had to do some last few force pushes to main branch before we switched to `changesets` and were able to eliminate local publishing.

@dwwoelfel I can take care of rebasing it for you if you like?

imolorhe:
@acao Oh okay, that explains it.

dwwoelfel:
I'll rebase against master and force push.

dwwoelfel:
The changes have been rebased.

acao:
thank you for taking care of that so quickly @dwwoelfel ! hoping to take a look later

acao:
@dwwoelfel on closer inspection, everything appears to be working as designed! excellent work

you are correct that this is a hack in terms of design, albeit clever - although if it were a more unique situation, a change to `onlineParser` would have made sense, we can easily implement all of this in `Rules.ts`

1. add a `BlockString` or `MultilineString` to `LexRules`, and take the regular expression for `String`, and seperate it into two expressions

2. and you can add this below `StringValue` in `ParseRules`
```ts
MultiStringValue: [p('"""'), 'StringValue', p('"""')]
```
3. then, we should look for anywhere in `graphql-language-service-interface` where `RuleKinds.STRING` might be referenced

4. add a test to `OnlineParser-test.ts` i suppose?

dwwoelfel:
@acao I don't think that approach can work for multiple lines in a block string. The approach works for rules like `FragmentSpread` because the multiple lines are separated by whitespace and onlineParser has special handling for whitespace that will advance the stream so you can get to the next line.

With a multiline block string, there is no way for the logic in Rules.js to consume the stream.`onlineParser` tries to tokenize the string content as if it were the next part of the graphql document and you eventually end up with an invalid document.

Looking at the [python codemirror mode](https://github.com/codemirror/CodeMirror/blob/9d0f9d19de70abe817e8b8e161034fbd3f907030/mode/python/python.js#L150), which also has multiline strings delimited by triple-quotes, they set a `tokenizer` in the state to handle multiple lines. Is that something you think could work here?

acao:
@dwwoelfel that sounds like a great idea!

ajbouh:
@dwwoelfel would you recommend your change as-is to fix block string parsing?

dwwoelfel:
@ajbouh The patch was fully functional when I created the PR. 

I think it's fine as-is, but an approach that uses a `tokenizer` is probably better for testing and maintainability.

PR #1783, opened 2021-02-02
feat: upgrades meros to not eagerly yield

maraisr:
Upgrades `meros ` which now support multiple payloads per yield for a more optimized experience.

### DIFF
--- a/packages/graphiql-create-fetcher/package.json
+++ b/packages/graphiql-create-fetcher/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "graphql-ws": "^4.1.0",
     "subscriptions-transport-ws": "^0.9.18",
-    "meros": "^1.1.0-beta.1",
+    "meros": "^1.1.0-beta.2",
     "@n1ru4l/push-pull-async-iterable-iterator": "^2.0.1",
     "@graphiql/toolkit": "^0.0.1"
   },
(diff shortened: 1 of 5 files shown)


acao:
@maraisr awesome! can you do a quick prettier run

maraisr:
Thanks for the review @acao this has been done; as well as added tri-state support. 🚀

acao:
fabulous!

acao:
@maraisr do we know why the `IncrementalDelivery` suite is failing? the payloads don't seem to match up with the ones we had before, but the diff is hard to read

acao:
ah i see!
https://github.com/graphql/graphiql/pull/1783/checks?check_run_id=1819298097

it's returning payloads with `errors: [null,null,null,null]`

maraisr:
Sorry was MIA for a while, yeah haha my bad, fixed this hopefully..

maraisr:
@acao  ill need to spend more time on it, ill give this another gander in the morning. Not sure why its failing in the CI but works on my pc.

maraisr:
@acao think we should be good now, gee glad we had e2e's... There was actually a small bug in meros with this one. Definitely going to push for unit tests before I release 1.1.0. But ya in the mean time, this should be gucci. I do apologize about this.

PR #1776, opened 2021-01-26
fix: search icon misalignment

iifawzi:
Closes #1773 

Since using `top` can't produce the same result in different screens due to many reasons, I changed the `search-box` display to flex, and centered the icons inside it. 

let me know if you do prefer using different approach.

PR #1768, opened 2021-01-18
fix: run onToggleDocs when setting docExplorerOpen to false

ChiragKasat:
Fix for issue #1717 

Fire onToggleDocs, when resizing the doc explorer to a small size and finally the docExplorerOpen state is set to false.

PR #1784, opened 2021-02-03
Adds missing changelog entries

maraisr:
We really should get in the habit of doing this in the PR's themselves, so we reference the PR author and number in the changelog correctly. So i've had to improvise in this scenario.

Thanks @acao


.changeset/config.json
```diff
@@ -1,6 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@1.5.0/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "changelog": [
+    "@changesets/changelog-github",
+    { "repo": "graphql/graphiql" }
+  ],
```
maraisr, line 6: This just formats the changelog entries a little nicer, puts in @<author> tags and PR numbers and other niceties. Most OSS projects that use Changesets use this.

cc @Andarist @mitchellhamilton

Andarist: Yep - this is a cool generator that I'm recommending to be used by default. 👍

acao:
yes we added the action but it didnt work retroactively

the issue with this PR is that there are package bumps. We just need the two md files. i can just re-create it after work!

acao:
also, this was for #1776 and #1768. I can just redo this, my goal was just to create the two MD files sorry!

.changeset/config.json
acao: sounds great! can you open another PR to add this?

PR #1788, opened 2021-02-06
Move @graphiql/create-fetcher to @graphiql/toolkit

harshithpabbati:
Moving create fetcher functionalites into toolkit


acao:
awesome work! can you add to the readme for create fetcher that it was inspired by graphql-subscriptions-fetcher, and thank @urigo?

harshithpabbati:
> @Urigo

acao:
oh hmm, I don't think we're going to be removing the package in this PR. remember we're going to leave it in with the readme pointing to toolkit for one last release?

acao:
i guess we can just delete it though? what do you think @imolorhe or @benjie ? should we just delete the package from npm or leave a waypoint? it has hundreds of downloads since yesterday but that's fine

imolorhe:
If the number of packages using it isn't a lot yet, then it should be fine to delete it? Else i think updating the readme that the functionality has been moved should be fine as well.

acao:
@imolorhe yes, graphiql is only in rc with it. ok sounds good!

wish we’d thought of this before i announced it

thanks so much @harshithpabbati !

acao:
@imolorhe I’ve been meaning to get to this one and its holding up the next RC. if you could take a look that would be awesome!

packages/graphiql-toolkit/src/__tests__/buildFetcher.spec.ts
```diff
@@ -13,7 +13,7 @@ import {
   createWebsocketsFetcherFromUrl,
   createMultipartFetcher,
   createSimpleFetcher,
-} from '../lib';
+} from '../../dist/lib';
```
imolorhe, line 16: This change should not be necessary

packages/graphiql-toolkit/src/__tests__/lib.spec.ts
```diff
@@ -1,5 +1,8 @@
 import { parse } from 'graphql';
-import { isSubscriptionWithName, createWebsocketsFetcherFromUrl } from '../lib';
+import {
+  isSubscriptionWithName,
+  createWebsocketsFetcherFromUrl,
+} from '../../dist/lib';
```
imolorhe, line 5: This change should not be necessary.

imolorhe (approved):

PR #1789, opened 2021-02-07
added regression test for state.rule bug

### LINKED ISSUE #899: Ensure we have a regression test for state.rule bug
As mentioned here, we'll want a regression test for this one: https://github.com/graphql/codemirror-graphql/pull/112#issuecomment-278198802

imolorhe:
Fixes #899

I couldn't reproduce the original error. It looks like there have been [more guards](https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service-parser/src/onlineParser.ts#L231) added to prevent that from happening. Added a regression test nonetheless.

### DIFF
--- a/packages/graphql-language-service-parser/src/__tests__/CharacterStream-test.ts
+++ b/packages/graphql-language-service-parser/src/__tests__/CharacterStream-test.ts
@@ -211,19 +211,19 @@ describe('CharacterStream', () => {
 
       stream.eatSpace();
 
-      expect(stream._pos).toEqual(2);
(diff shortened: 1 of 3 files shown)


imolorhe:
About changes in CharacterStream test file, we seem to be using private properties and methods of the CharacterStream class, which I don't think we should be doing. So tried to use the public methods available in the class where possible. There are still others but there is no direct public method for those so haven't touched them yet, but ideally we could look into it better. Perhaps we could use the closest equivalent public methods instead.

acao (approved):

acao:
awesome work @imolorhe!

PR #1790, opened 2021-02-08
Migrated codemirror-graphql to typescript

imolorhe:
Moving all codemirror-graphql logic to typescript.


packages/codemirror-graphql/package.json
```diff
@@ -13,34 +13,28 @@
...
-    "lint.js.flow",
-    "mode.esm.js",
-    "mode.js",
-    "mode.js.flow",
+    "dist",
     "utils",
     "variables",
-    "results"
+    "results",
+    "/*.js",
```
acao, line 23: this strategy works but then we end up with `babel.config.js` and then `jest.config.js`. is it possible to mix include/ignore arguments now with files param?

imolorhe: Excluded babel and jest configs

acao (approved):
looks great! lets merge it

acao:
oh crap @imolorhe we forgot the changeset ://

packages/codemirror-graphql/src/index.d.ts
```diff
@@ -0,0 +1,20 @@
+import * as CM from 'codemirror';
+import 'codemirror/addon/hint/show-hint';
+
+declare module 'codemirror' {
```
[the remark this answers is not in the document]
imolorhe: We should be able to rename it without issues

imolorhe:
Oh right! That can be another PR right?

PR #1799, opened 2021-02-10
feat: meros out of beta 🎉 

maraisr:


maraisr:
@acao I have opted to not add a changeset here, might want to wrap a few things in? Or think just add it, and collect in the versioning PR for a while?

acao:
@maraisr hey thanks for this! awesome work!

can you add the support to GraphiQL for `multiple: true` in this PR here, or a follow up?

eventually, I'd like to move most of the reusable business logic to `@graphiql/toolkit`

PR #1804, opened 2021-02-14
fix: `dset` doesnt deepmerge

maraisr:
Upon testing this feature further, I got aware that `dset` wouldn't "deep merge" the payloads. Which is only prevalent with the `@defer` directive. As up until now `@stream` wouldn't only append to an array—whereas `@defer` would merge into existing objects. 

This _would_ fix the issue, however still some work to be done on e2e's. So will leave in draft till then.

cc @acao


acao:
@maraisr thanks for this!

looks like theres a bug with the implementation according to the cypress suite, where only two payload items made it through, did you notice that? something seems to be broken

packages/graphiql/src/utility/dset-merger.ts
```diff
@@ -0,0 +1,52 @@
...
+
+@maraisr will be monitoring this.
+ */
+
+export function dset<T extends object, V>(
+  obj: T,
+  keys: string | ArrayLike<string | number>,
+  value: V,
+): void;
+export function dset(obj, keys, val) {
```
maraisr, line 15: @lukeed with your blessing I have in-lined this for now, once `dset/merge` if/when becomes a think I'll update this.

imolorhe:
I believe it is very important that this new utility comes with unit tests to cover all the use cases, especially the deep merge use case.

lukeed:
Agreed. At least the bare minimum for utility tests. When I polish this up as a standalone (soon) I'll have extensive coverage.

In summary, arrays and objects are merged. On any positional or key/index conflict, the newer value is taken. For non-object values, the new value is always given preference.

lukeed:
For transparency sake, @maraisr reached out to me about this and I provided the dset-merger implementation to be inlined since I don't have the bandwidth to make it it's own module atm. 

If there are any issues with it please let me know

maraisr:
> I believe it is very important that this new utility comes with unit tests to cover all the use cases, especially the deep merge use case.

Ya this is true. But the idea here isnt that external people consume this utility. Its analogous to using an npm package. The use-cases are covered in the e2e. This is a safe bet as it stands now, which once Luke's package drops we should be certain. 

Till this is is still 100x better than what's currently in mainline.

But happy for this PR to stagnate till this has dropped—because I still believe we need even more e2e's to cover all of the edge cases. But for this, I need to change the way its tested, and introduce a "unit test" for this multipart stuff, so that yeah we can fire fine-grained events and assert the state.

Waiting for your call on this one @acao

acao:
@imolorhe the create-fetcher has unit tests and end to end tests

acao:
I was looking for an alternative, and found `deep-set`, which also has the same issue `dset` has
https://github.com/jonschlinkert/set-value/pull/26

[`dot-prop`](https://www.npmjs.com/package/dot-prop) looks awesome, and the other utilities will provide very useful for working a playground esque settings API that plugins can extend, which was part of the 2.x roadmap

lukeed:
Deep-set, dset, set-value, lodash.set are all the effectively the same thing. They _set_/assign a value at a nested path.

The difference here is that you guys _dont_ want an assignment. You want to merge/collect values at a nested path. I know of no module that does this, hence the inlined fork of `dset` here.

lukeed:
Published [`dset@3.1.0`](https://github.com/lukeed/dset/releases/tag/v3.1.0) which contains the new `dset/merge` submodule.

maraisr:
@acao @imolorhe I have updated this branch with the new `dset` version. `dset` has a [comprehensive suite of tests](https://github.com/lukeed/dset/tree/master/test) focusing specifically on [solving the issues here](https://github.com/lukeed/dset/releases/tag/v3.1.0). 

This also means that graphiql will now _correctly_ articulate the results of an Incrementally Delivered response.

imolorhe (approved):

imolorhe:
Nice! Approved. Looks like changeset is missing?

maraisr:
@imolorhe I think there is a [changesets already](https://github.com/graphql/graphiql/pull/1804#issuecomment-778760306) in this pr. Feel free to assist me on the wording: https://github.com/graphql/graphiql/pull/1804/files#diff-10046eb3e404bb78fb2100699c84af52ebd1593ada28eea0318bd0a548c0cb60

imolorhe:
Oh okay I missed that. My bad.

PR #1816, opened 2021-03-28
fix: peer resolutions, limit example workspaces

acao:
this provides more targeted peer resolutions, based on known compatibility. `graphql@14.5.0` is when built-in `graphql` types are required. the last `14.4.x` release before that also added a new method that we use.

this also limits the scope of the workspaces to monaco, so that both start commands work, and the yarn lockfile can shrink in size and duplicate resolution complexity because of create-react-app/etc in other examples

prompts for review:

1. should we add `|| experimental-defer-stream`?
2. does not adding it create peer resolution problems for consumers? webpack example should be representative of that?
3. there is a project called `manypkg` that has a `fix` command that will normalize the dependency versions of all the workspaces, preventing issues with clashing versions. it will resolve to the highest version, and has a rule to automatically add the peer dependencies as a devDependency. we can include another commit to this PR with this fix, which would require some type fixes in language server because of babel versions in `findGraphQLTags` . this could also be another PR, but naturally makes sense as the next step after this PR

### DIFF
--- a/.changeset/young-bottles-taste.md
+++ b/.changeset/young-bottles-taste.md
@@ -12,4 +12,4 @@
 'monaco-graphql': patch
 ---
 
-improve peer resolutions for graphql 15. 14.5.0 minimum is for typescript types.
+improve peer resolutions for graphql 14 & 15. `14.5.0` minimum is for built-in typescript types, and another method only available in `14.4.0`
--- a/.changeset/young-bottles-taste.md
+++ b/.changeset/young-bottles-taste.md
@@ -1,0 +1,15 @@
+---
+'codemirror-graphql': patch
+'graphiql': patch
+'@graphiql/toolkit': patch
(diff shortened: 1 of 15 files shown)


imolorhe (approved):

imolorhe:
Welcome back 🙂

acao:
thanks @imolorhe !! excited to be back! I'll be contributing more some evenings this week and weekend. I really really appreciate your co-maintainership during my hiatus!

imolorhe (approved):

simkessy:
This is breaking builds

jay-deshmukh:
adding 

```json

"resolutions": {
    "graphql-language-service-parser": "1.9.0",
    "graphql-language-service-interface": "2.8.2"
  }
  
 ```
 
 to `package.json` fixed the issue for me

PR #1819, opened 2021-03-29
feat: createFetcher legacyClient option, simplify, docs

### LINKED ISSUE #1800: GraphiQL.createFetcher does not work for subscriptions-transport-ws
As shown in this PR on express-graphql: https://github.com/graphql/express-graphql/pull/739/files#diff-f851f2ce2fbcb0220aaca183def93d7ced5b533fead2016551cfd6d37b357ae5,  I started the graphql server with subscription using legacy subscriptions-transport-ws. 

Then on the client side I create fetcher like this:
```
window.GraphiQL.createFetcher({url: fetchURL, subscriptionUrl: ${safeSerialize(
            subscriptionEndpoint,
          )}});
```

Then I got error as shown in the below screenshot: 
Screen Shot 2021-02-10 at 9 29 46 PM

The same client side code works if I started a graphql server using graphql-ws though:
https://github.com/graphql/express-graphql/pull/739/files#diff-4f5f57981087dfef3de9e960e86656c7a1d1ef9f424fd0a75e32751163a286e1

acao:
hopefully fixes #1800 

- [x] `createGraphiQLFetcher` should have a `legacyClient` override option as has been noted by some users
- [x] `createGraphiQLFetcher` now only attempts an `graphql-ws` connection when only `subscriptionsUrl` is provided.
- [x] positive test coverage diff
- [x] document this new option
- [x] pass `connectionParams` with `subscriptionsUrl` 
- [ ] rename `createGraphiQLFetcher` to `createFetcher` as a breaking change?

### DIFF
--- a/.changeset/two-bulldogs-doubt.md
+++ b/.changeset/two-bulldogs-doubt.md
@@ -3,4 +3,6 @@
 'graphiql': patch
 ---
 
-`GraphiQL.createClient()` accepts custom `legacyClient`, exports typescript types, fixes #1800
+`GraphiQL.createClient()` accepts custom `legacyClient`, exports typescript types, fixes #1800. 
+
+`createGraphiQLFetcher` now only attempts an `graphql-ws` connection when only `subscriptionUrl` is provided. In order to use `graphql-transport-ws`, you'll need to provide the `legacyClient` option only, and no `subscriptionUrl` or `wsClient` option.
--- a/.changeset/two-bulldogs-doubt.md
+++ b/.changeset/two-bulldogs-doubt.md
@@ -1,0 +1,6 @@
+---
+'@graphiql/toolkit': minor
+'graphiql': patch
+---
+
+`GraphiQL.createClient()` accepts custom `legacyClient`, exports typescript types, fixes #1800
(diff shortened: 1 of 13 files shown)


imolorhe (approved):

acao:
@junminstorage would this unblock you?

junminstorage:
Yes, it does. having client to inject its own wsClient or legacyClient is great so createFetcher util can have the flexibility of not tied to any versions of clients if upgrades are needed.

acao:
@junminstorage should we specify `peerDependency` ranges and/or `optionalDependencies` do you think?

junminstorage:
I am not super familiar with the difference between those two, it looks like peerDependency is better fit since the create-fetcher's lib.ts imports the two clients, and the legacy client is only used for type checking, so is it devDependency?

acao:
> I am not super familiar with the difference between those two, it looks like peerDependency is better fit since the create-fetcher's lib.ts imports the two clients, and the legacy client is only used for type checking, so is it devDependency?  

yes good thinking. the legacy client should be a `devDependency`, and then `ws-client` should be a peer dependency with ranges we support

PR #1821, opened 2021-03-31
chore: render query history panel only when it's toggled

harshithpabbati:
Rendering the history panel only when it's toggled instead of using CSS to hide it.

### DIFF
--- a/.changeset/metal-cougars-fry.md
+++ b/.changeset/metal-cougars-fry.md
@@ -1,0 +1,5 @@
+---
+"graphiql": patch
+---
+
+fix: render query history panel only when it's toggled, instead of hiding with CSS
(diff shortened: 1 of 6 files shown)


acao:
ah! i think there's a bug in the new github actions workflow config where it doesn't work on PR create. if you push another commit it should test and deploy. I'll need to go fix that!

acao:
hmm! still no dice! strange. but the PR looks great, thank you!

acao:
this is a good move, thanks @harshithpabbati !

harshithpabbati:
hmm, test cases failed because the history panel is no more available in dom.

acao:
it should be once it's toggled open!

harshithpabbati:
> it should be once it's toggled open!

yup, did that!

acao:
thank you again @harshithpabbati !

acao:
@harshithpabbati just need to fix the formatting bug, may have been introduced by changesets. thus why its easier when you run `yarn changeset` locally yourself!

harshithpabbati:
@acao, fixed it!

PR #1840, opened 2021-04-19
fix(ws): use `wsConnectionParams` from fetcher options

enisdenjo:


acao:
Thanks for this! I thought I had fixed this but we must have had a bad rebase again.

PR #1841, opened 2021-04-19
fix(ws): Subscriptions async iterator completes and better error handling

### LINKED ISSUE #1827: How to unsubscribe from graphql-ws subscription
Hello,

When I do a subscription (using graphql-ws fetcher), it works, the results are well printed in the result panel but when I push the stop button, the results stop to be printed on the panel but it seems that nothing is done on the wsClient. GraphiQL does not tell to the wsClient to unsubscribe and events are still sent by the ws server.
If I repush the play button, a second subscription using a different UUID is sent to the server on the same websocket connection which if I'm not mistaken is illegal.
Is there a way to unsubscribe when the stop button is pressed ?
Thx

```javascript
const URL = "http://debian9acu:8000/graphql";
const URLws = "ws://debian9acu:8000/graphql-ws";
var fetcher = createGraphiQLFetcher( {url:URL,subscriptionUrl:URLws} );
const container = document.getElementById("graphiql");
const defaultQuery = `
{
}
`
render(
  <GraphiQL fetcher={fetcher} defaultQuery={defaultQuery} />,
  container
);
```

enisdenjo:
Closes #1827

### DIFF
--- a/packages/graphiql-toolkit/src/create-fetcher/lib.ts
+++ b/packages/graphiql-toolkit/src/create-fetcher/lib.ts
@@ -105,12 +105,13 @@ export const createWebsocketsFetcherFromClient = (wsClient: Client) => (
               `Socket closed with event ${err.code} ${err.reason || ''}`.trim(),
             ),
           );
+        } else {
+          sink.error(
+            new Error(
+              (err as GraphQLError[]).map(({ message }) => message).join(', '),
+            ),
+          );
         }
-        sink.error(
-          new Error(
-            (err as GraphQLError[]).map(({ message }) => message).join(', '),
-          ),
-        );
       },
     }),
   );
--- a/packages/graphiql-toolkit/src/create-fetcher/lib.ts
+++ b/packages/graphiql-toolkit/src/create-fetcher/lib.ts
@@ -92,15 +92,15 @@ export const createWebsocketsFetcherFromUrl = (
  */
 export const createWebsocketsFetcherFromClient = (wsClient: Client) => (
   graphQLParams: FetcherParams,
-) => {
+): AsyncIterableIterator<FetcherResult> => {
   let deferred: {
     resolve: (done: boolean) => void;
     reject: (err: unknown) => void;
   } | null = null;
-  const pending: any[] = [];
+  const pending: FetcherResult[] = [];
   let throwMe: unknown = null,
     done = false;
-  const dispose = wsClient.subscribe(graphQLParams, {
+  const dispose = wsClient.subscribe<FetcherResult>(graphQLParams, {
     next: data => {
       pending.push(data);
       deferred?.resolve(false);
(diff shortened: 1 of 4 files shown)


packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -92,10 +92,58 @@ export const createWebsocketsFetcherFromUrl = (
...
+      }
+      deferred?.reject(throwMe);
+    },
+    complete: () => {
+      done = true;
+      deferred?.resolve(true);
+    },
+  });
+  return {
+    [Symbol.asyncIterator]() {
```
n1ru4l, line 128: Note that Symbol.asyncIterator is not available in Safari 14

enisdenjo: Also, shouldn't the transpilation be handled by the user? If you'd like to support Safari 14 on iOS, simply polyfill the async iterator with Babel on build step. @acao what is your take on this?

n1ru4l: I have no strong feelings about this. For myself, I can say that I am shipping an app with Safari 14 support and try to avoid babel transforms where possible, that is why I created the lib. If GraphiQL does not need to rely on push-pull-async-iterable-iterator I am totally fine with it.

enisdenjo:
Fixed with [`@n1ru4l/push-pull-async-iterable-iterator@v2.1.3`](https://github.com/n1ru4l/push-pull-async-iterable-iterator/releases/tag/v2.1.3).

packages/graphiql-toolkit/src/create-fetcher/lib.ts
enisdenjo: Reverted back to using the push-pull-async-iterable-iterator following v2.1.3.

enisdenjo:
Note @n1ru4l, the iterator from `makeAsyncIterableIteratorFromSink` does not throw when calling `sink.error`.

n1ru4l:
https://github.com/n1ru4l/push-pull-async-iterable-iterator/releases/tag/v2.1.4 😰

enisdenjo:
Brilliant! Can confirm `v2.1.4` working @n1ru4l.

@acao this is ready to merge now.

n1ru4l:
Before merging please also reflect whether push-pull-async-iterable-iterator is a necessary dependency @acao 😇

acao:
Appreciate y’all putting so much into this already! Please ping me directly in the future if you have a PR ready for me as I have notifications disabled, and only tend to check on the weekends now that I’m still trying to take a break from maintaining

I don’t have time to look into whether or not that is a necessary dependency @n1ru4l , is it? Should we just merge and release this patch bugfix, and then drop it in a minor release as a follow up at least?

lastly can you... you actually created the changeset... 🥺😍 omg you guys are the best!

PR #1848, opened 2021-04-26
chore: upgrade cdn react versions

maraisr:
- upgrades to react 17
- removes unknown example js file

### DIFF
--- a/examples/graphiql-cdn/index.html
+++ b/examples/graphiql-cdn/index.html
@@ -30,11 +30,11 @@
     -->
     <script
       crossorigin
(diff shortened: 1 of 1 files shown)


acao:
@maraisr react 17 would take much more effort to support. Have you confirmed this works in a local build? The example you changed is not part of the preview

maraisr:
There are 2 changes here, happy to revert the react 17, but the `renderExample` script tag, is a reference to a file that doesnt exist at all—so removed it.

Really im gearing up to a browser capable build of toolkit.

PR #1879, opened 2021-05-26
fix: remove statement pending discussion

acao:
I made the choice of adding this to the readme w/o consulting colleauges

My co-maintainers are being forced to take sides and this is unfair

The technical steering commitee will review this decision, & decide what best reflects our interests together.

First and foremost, they will decide on a policy to make sure a rouge maintainer action like this cannot happen again without a decision making process involving the TSC.

I apologize for making this choice but not what I said, and will continue supporting an end to the apartheid in my own personal venues.

Please leave my co-maintainers and colleauges be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 <!-- @format -->
-> # Black Lives Matter 🖤 Free Palestine 🇵🇸
+
+> # Black Lives Matter 🖤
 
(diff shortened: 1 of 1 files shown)


Grish1:
Can you free Tibet instead?

PR #1889, opened 2021-06-15
feat: export ToolbarSelectOption and ToolbarMenuItem

henryqdineen:
It looks like `ToolbarSelect` is exported from the index module but `ToolbarSelectOption` isn't. We are upgrading from v0.14.2 to v1.4.2 and noticed that the `GraphiQL.Select` and `GraphiQL.SelectOption` properties are no longer available and appear to be [commented out](https://github.com/graphql/graphiql/blob/65aa64811ccd40c5e8ba99edec825a91115a3c8b/packages/graphiql/src/components/GraphiQL.tsx#L692-L694). When looking at moving over to `ToolbarSelect`  it seemed like it would be better if `ToolbarSelectOption` was also exported. `ToolbarMenu`/`ToolbarMenuItem` also appears to have the same issue. 

Thanks!

### DIFF
--- a/.changeset/strange-pugs-brush.md
+++ b/.changeset/strange-pugs-brush.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+feat: export ToolbarSelectOption and ToolbarMenuItem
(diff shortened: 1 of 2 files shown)


acao:
awesome thank you! those interfaces were dropped as noted here so we could upgrade react:
https://github.com/graphql/graphiql/issues/1165

Issue #1892, opened 2021-06-16
History is not saving when history pane is closed.


Rohitturbot:
after the version of graphiql `>1.4.1`  history is not updating when we run query or mutation but if the pane is open it update. 

![2021-06-16_19-06-06 (1)](https://user-images.githubusercontent.com/40739903/122229377-4d4ac100-ced6-11eb-9fd5-57228e52264a.gif)

Issue #1782, opened 2021-02-02
Store more than 20 queries in History?


alvaro1728:
Hi,

Is there a way to store more than 20 queries in the History pane?  I would prefer to store the last 50.

Thanks,
Alvaro

maraisr:
Hi there @alvaro1728 thank you for raising this, unfortunately there I is no way to store a higher count. You can see this value is hard-coded here: https://github.com/graphql/graphiql/blob/main/packages/graphiql/src/components/QueryHistory.tsx#L19

We have seen this issue come about in the past, and may look at this in the future. I'm only quite new to the project so would need @acao to make a decision with this one.

imolorhe:
Perhaps this can be an issue that will be addressed in v2.0 (making it configurable) instead of the current version. Although I think we can still have a reasonable cap on it. I assume 20 was chosen due to storage concerns.

acao:
`MAX_HISTORY_LENGTH` is set statically in this component. we could expose a prop in 1.0, but query history will be rewritten in 2.0 with a new UI and a hooks API

acao:
`maxHistoryLength` Is now the prop to use to configure this and override the default of 20!

PR #1914, opened 2021-06-27
fix: new util to manage historyStore outside of query history component

harshithpabbati:
fixes: #1892 

Since the history store is inside the query history panel, it's not getting updated until we open the panel, so instead, I created a new historyStore util to manage the store.

### DIFF
--- a/.changeset/short-mirrors-occur.md
+++ b/.changeset/short-mirrors-occur.md
@@ -1,0 +1,5 @@
+---
+'graphiql': minor
+---
+
+fix: history can now be saved even when query history panel is not opened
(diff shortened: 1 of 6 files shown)


harshithpabbati:
According to the discussion in #1782 

Created a new prop, which will allow passing in maxHistoryLength. To default, it is set to 20.

acao:
awesome, thank you @harshithpabbati !

acao:
@harshithpabbati I think this means we have several unused files to delete in `packages/graphiql/src/utils` now, can you remove the unused files?

harshithpabbati:
> @harshithpabbati I think this means we have several unused files to delete in `packages/graphiql/src/utils` now, can you remove the unused files?

I don't think there are any unused files.

acao:
@harshithpabbati I see now, after a closer reading, thank you!

PR #1923, opened 2021-07-09
Fix css theme override on result window

### LINKED ISSUE #1919: editorTheme option dont change the resultViewer background color
The theme dont apply on the result window due to this rule: 

https://github.com/graphql/graphiql/blob/65aa64811ccd40c5e8ba99edec825a91115a3c8b/packages/graphiql/src/css/app.css#L216

The selector `.result-window .CodeMirror` override the theme selector `.cm-s-monokai.CodeMirror`

cgarnier:
close #1919 

`editorTheme` option apply theme uniformly on all the codemirror instances so we can considerate that this background is specific to the graphiql theme. Other theme should be able to override correctly this color.

### DIFF
--- a/.changeset/three-ligers-tie.md
+++ b/.changeset/three-ligers-tie.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Fix result window theme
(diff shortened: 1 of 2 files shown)


acao:
Excellent fix, thank you!

PR #1925, opened 2021-07-21
chore: enable header editor by default

harshithpabbati:
This PR allows the latest version of the GraphiQL to persist and show the headers tab by default.


dotansimha:
@acao I think this can get in? Seems like it's passing. 
I'm not sure about changing the behaviour of the flags, because this might be a breaking change? (it's super minor, but still a change in behaviour)

acao:
I was not able to force push a rebase to this branch for some reason :(

harshithpabbati:
oh I see, I will rebase it just a min!

acao:
oh dang it was missing a changeset. I will add one for your commit!

PR #1926, opened 2021-07-22
docs: Fix usage of graphiql in umd

lisp719:
I tried UMD as per the README, but it did not work.
It seems that the arguments written in the README are wrong.


acao:
thank you!

PR #1934, opened 2021-08-04
fix(graphiql): add react 17, 18 in peerDependencies

tonyfromundefined:


acao:
I haven’t worked with react 17 or 18 yet, considering we are still using methods which are deprecated in 16, do we know if these versions are compatible with 1.x for certain, before the hooks rewrite?

PR #1943, opened 2021-08-31
Proposal: add @dotansimha as co-maintainer

acao:
I would like to add the prolific @dotansimha as a co-maintainer to the repository :)


dotansimha:
I would love to join! :)

acao:
I have permissions to do this, just want to honor the process if that is still needed

orta (approved):
Sounds good to me

PR #1947, opened 2021-09-12
docs: Update example URLs

PintTheDragon:
Some of the doc paths are incorrect under examples.


PintTheDragon:
I'm not signing a CLA for this.

PR #1951, opened 2021-09-17
Skip config updates when no custom filename is defined

GoodForOneFare:
Skip config reload when no custom filename is defined. Previously, any project without a `load.fileName` configuration would reload configuration on every change event. This is because `match(undefined)` always returns true.

```js
['graphql.config', 'graphqlrc', undefined].some(v => 'this-should-not-match'.match(v)?.length) // true
```


acao:
Thanks so much for opening a PR for this!

GoodForOneFare:
@acao I think this is ready for review 🙏  Please let me know if I should be tagging anyone else in!

Issue #1791, opened 2021-02-08
graphql-monaco is broken on monaco 0.21+


djpate:
Trying to install `graphql-mocaco` using the latest versions of monaco-editor will not work.

Tested broken on `0.22` and `0.21`. The version in the example works fine (`0.20`)

The rest call to the graphql endpoint to do the introspection query does not fire either.

I tried looking at their changelog but could not find something that would be a clear reason why.

acao:
yes, when the mode fails then the worker is not initialized

this is a known bug, thanks for reporting! i thought i had created a ticket this weekend

Nishchit14:
I am using `monaco-editor@0.27.0` and facing the same issue. @acao any plan to support it in near future? I really appreciate all the work done for `graphql-monaco`. If I can help anyway then please do let me know.

Nishchit14:
I am getting the error, `monaco is not defined`

![image](https://user-images.githubusercontent.com/5078921/133819856-fb3cfcc7-b182-4866-83e8-b6c3acdac219.png)

acao:
I'm hoping to get to it soon! There are some more urgent matters currently, but this was one of the next known issues to address.

So, here, I am using a pattern borrowed from the monaco examples:

https://github.com/microsoft/monaco-editor/search?q=CompletionItemKind

it must be that previously, monaco 0.20 and older were exporting this globally? I don't know how this worked before hahaha

I can't seem to figure out where to import this enum from now:

https://github.com/microsoft/monaco-editor/blob/a40145a4c9ee7f003e5b7dcbc98c567a8dadc158/typedoc/monaco.d.ts#L5700

here is a sandbox where I tried to import it:

https://codesandbox.io/s/async-flower-nl2kb?file=/src/index.ts

Nishchit14:
great, thanks for replying quickly @acao. I'll try to help here by debugging it with the above references.

Nishchit14:
I have cloned this [monaco-samples](https://github.com/microsoft/monaco-editor-samples) repo and try to debug apis of `monaco-editor@0.27.0`. 

https://github.com/microsoft/monaco-editor-samples/blob/main/browser-esm-webpack-typescript/src/index.ts#L1

```ts
import * as monaco from 'monaco-editor';

console.log(
	monaco.languages.CompletionItemKind,
	"monaco.languages.CompletionItemKind"
);
```
#### console output is here
[...]

PR #1952, opened 2021-09-20
feat: monaco-editor@0.27.0 support implemented

Nishchit14:
PR referring to this issue https://github.com/graphql/graphiql/issues/1791

### DIFF
--- a/packages/monaco-graphql/package.json
+++ b/packages/monaco-graphql/package.json
@@ -27,11 +27,11 @@
   },
   "devDependencies": {
     "graphql": "experimental-stream-defer",
-    "monaco-editor": "^0.20.0",
+    "monaco-editor": "^0.27.0",
     "vscode-languageserver-types": "^3.15.1"
   },
   "peerDependencies": {
     "graphql": ">= v14.5.0 <= 15.5.0",
-    "monaco-editor": "^0.20.0"
+    "monaco-editor": "^0.27.0"
   }
 }
--- a/examples/monaco-graphql-webpack/package.json
+++ b/examples/monaco-graphql-webpack/package.json
@@ -28,7 +28,7 @@
     "css-loader": "^3.5.1",
     "file-loader": "6.2.0",
     "html-webpack-plugin": "^4.2.0",
-    "monaco-editor": "^0.20.0",
+    "monaco-editor": "^0.27.0",
     "monaco-editor-webpack-plugin": "^1.9.0",
     "react-dom": "^16.12.0",
     "style-loader": "^1.1.3",
--- a/packages/graphiql-2-rfc-context/package.json
+++ b/packages/graphiql-2-rfc-context/package.json
@@ -51,7 +51,7 @@
     "i18next": "^19.4.4",
     "i18next-browser-languagedetector": "^4.1.1",
     "markdown-it": "^10.0.0",
-    "monaco-editor": "^0.20.0",
+    "monaco-editor": "^0.27.0",
     "monaco-graphql": "^0.4.2",
     "react-i18next": "^11.4.0",
     "theme-ui": "^0.3.1"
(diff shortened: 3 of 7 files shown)


acao:
@Nishchit14 do you feel this warrants a minor release since it's so deeply tied to this dependency?

Nishchit14:
Thanks, @acao, I am new to the changeset, please help me here the first time. I'll take care of it next time by myself.

and yes, it is more suitable with the minor release with I guess `v0.5.0`.

I suggest describing the version and its deps matrix as shown in this table
https://github.com/microsoft/monaco-editor-webpack-plugin#version-matrix
image


if someone is using `monaco-editor@0.20.0` then they would use `monaco-graphql@0.4.4` and so on.

Nishchit14:
please do not merge it, I am still researching some issues

acao:
Also noticing something has lapsed with our deploy previews… hmm!

YeomansIII:
Hi everyone, just following along with this PR and checking on the status. Is there anything else that needs to be resolved?

Nishchit14:
@YeomansIII It is good to go. If all tests are passing then we can merge it @acao

dotansimha:
> @YeomansIII It is good to go. If all tests are passing then we can merge it @acao

Thanks, that's great. Is there anything we need to do in terms of documentation? (maybe related to the version matrix?) 

Also, can you please add a changeset file? (Run `yarn changeset` and it will guide you)

acao:
@dotansimha a version matrix might be confusing, because this fix is still backwards compatible for older versions of monaco editor. we might want to figure out what the minimum version is for the `peerDependency` in the future.

acao:
@dotansimha we just need to add a changeset to this PR. it can be a minor version but for `monaco-graphql` only

packages/monaco-graphql/package.json
```diff
@@ -27,11 +27,11 @@
...
   "devDependencies": {
     "graphql": "experimental-stream-defer",
-    "monaco-editor": "^0.20.0",
+    "monaco-editor": "^0.27.0",
     "vscode-languageserver-types": "^3.15.1"
   },
   "peerDependencies": {
     "graphql": ">= v14.5.0 <= 15.5.0",
-    "monaco-editor": "^0.20.0"
+    "monaco-editor": "^0.27.0"
```
acao, line 35: this is the only thing that needs changed, my bad overlooked this before. this is still compatible with at least 0.20.0 if not earlier. we should test with a version matrix perhaps

acao (changes requested):
this is backwards compatible, so the `peerDependencies` entry should reflect that!

acao:
here is an example of monaco-editor using this interface in monaco-editor@0.17.0.

https://codesandbox.io/s/graphql-monaco-example-btx8s?file=/src/index.ts

we need to test this out properly, perhaps with a version matrix in github actions, and make sure peerDependendecies is correct, as it may be that our peer range is much earlier than 0.20.0 even, but we need to test against all features of monaco-editor that we use to be sure

acao:
@Nishchit14 or @ardatan this is high priority and ready to release once we revert the change to `peerDependencies` fyi!

packages/monaco-graphql/package.json
Nishchit14: you mean, we need to undo all other file changes, right?  @acao

Nishchit14: @acao I have reverted it. please review it.

acao: Fantastic, thank you!

acao (approved):

acao:
Ok last thing @Nishchit14 can you add the changeset entry or would you rather we add that? This will be an important fix to document!

Nishchit14:
sure @acao, If it is not urgent I'll add it tonight (in 10hrs).

Nishchit14:
@acao @dotansimha please review it, changeset has been created.

Issue #1963, opened 2021-10-04
Language server unknown fragment inconsistency


patrickszmucer:
When using a fragment that's not defined, the language server correctly reports the error:

![Screenshot 2021-10-04 at 18 25 58](https://user-images.githubusercontent.com/10088058/135896676-cb973c26-3135-48a7-a500-5969f735250c.png)

however, after saving the file, the error disappears:

![Screenshot 2021-10-04 at 18 26 05](https://user-images.githubusercontent.com/10088058/135896692-4cce329f-9cdc-4c43-9bf9-5905f4d9dc5f.png)

This seems to be due to the fact that when a file is modified, we call `getDiagnostics` with only two arguments (and `isRelayCompatMode` defaults to false):

https://github.com/graphql/graphiql/blob/94f73448acf215d9691f422eb4a17e4788b08427/packages/graphql-language-service-server/src/MessageProcessor.ts#L368

while when a file is saved, we call `getDiagnostics` with three arguments, and for some reason we invert the value of `isRelayCompatMode`:

https://github.com/graphql/graphiql/blob/94f73448acf215d9691f422eb4a17e4788b08427/packages/graphql-language-service-server/src/MessageProcessor.ts#L293-L297

patrickszmucer:
I opened a PR that I believe is a fix for this https://github.com/graphql/graphiql/pull/1964

ebsangam:
It is happening again.

PR #1964, opened 2021-10-04
Fix unknown fragment errors on save

patrickszmucer:
Fixes https://github.com/graphql/graphiql/issues/1963

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -293,7 +293,7 @@ export class MessageProcessor {
           const results = await this._languageService.getDiagnostics(
             query,
             uri,
-            this._isRelayCompatMode(query) ? false : true,
+            this._isRelayCompatMode(query),
           );
           if (results && results.length > 0) {
             diagnostics.push(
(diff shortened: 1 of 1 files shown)


acao:
Hmm thanks! I will have to check into the history of these function signatures and see why the legacy relay compat mode was originally inverted there, but if it seems to fix the issue that’s good. You’ve tested this fix locally?

patrickszmucer:
I've tested locally and it fixes the issue, but I haven't tested the relay compat mode (I'm not sure how to).

I git blamed this line and this is when it was introduced https://github.com/graphql/graphiql/commit/5d073d21c7b9dbab806e6956d01688feb8fc9a4e but I can't find any PR associated with it.

acao:
@patrickszmucer awesome to hear! the relay compat mode is for ancient relay 1 legacy support, I would consider it mostly deprecated and would be removed in the future.

acao:
@patrickszmucer are you able to sign the CLA? you only need to do it once for any Linux Foundation or GraphQL Foundation project :)

patrickszmucer:
@acao done! Again, sorry for the delay from my side.

acao:
np! @patrickszmucer really appreciate this!

PR #1979, opened 2021-10-13
fix: `monaco-graphql` worker issue, es module config should use `esNext `

iahu:
build es module code with `esNext` module type.


acao:
just for `monaco-graphql`?

acao:
@iahu thank you!

PR #1997, opened 2021-10-31
feat: monaco variables json, improved schema & worker loading & monaco api

acao:
a ton of monaco-graphql changes here, many many breaking changes. API is entirely rewritten.

## Changes
- adds variables JSON completion, validation, hover, etc using the JSON Schema support in `monaco-graphql`
  - you can now set variable diagnostics like this:
   ```ts
   monacoAPI.setDiagnosticSettings({ 
     // currently Record<string, string[]>, 
     // may opt for Record<string, { uris: string[], ...otherConfig }>
     variablesEditorValidation: {
       [myOperationModel.uri.toString()]: [variablesModel.uri.toString]
     },
    // same as json.jsonDefaults.setDiagnosticConfig() options
     variablesDiagnosticSettings: { 
       allowComments: true, // enables jsonc support
       validateSchema: 'error'
     }
   })
   ```
   - currently you can only overwrite this config all at once, but I may add a method to update it more atomically
   - this is an abstraction that inevitably leads to some encapsulation issues, so the readme will demonstrate how to add support without this method.
- uses `graphql` as the language ID finally. this defaults to microsoft's provided syntax support and language config (which you can find in the `monaco-editor` repository under `basicLangauges`).
  - important! this presumes that any improvements we need to make to the monarch grammar and language config will be contributed to `monaco-editor`, and we will serve as a full language worker implementation that extends their basic configuration.
- completely rewrote monaco graphql schema loading logic. schema is loaded synchronously in main process now, and passed off to the web worker. monaco does *not* do any business of fetching remote schema, ala `codemirror-graphql`. This is a feature as graphql userland has many ways to request remote schemas. See `graphql-tools` for example!
- accepts multiple schemas! using `schemas` ala `monaco-json` 🎉 
 	- for each `SchemaConfig`, you can provide either `schema`, `introspectionJSON`, `introspectionJSONString`, `documentString` and `documentAST` as schema input.
	- schema requires a `uri` that is not used for remote loading, but as a cache key and for the model uri. for definitions support, we will set an SDL string to a model at this URI that will be used for schema lookups. this can be a URL or a file URI, anything that  works with `monaco.URI.parse()`. this is a common pattern with monaco, and the file URIs can either be an internal implementation detail, or can be presented to users ala vscode/codesandbox/etc, that's up to you.
	- ala `monaco-json`, each entry in `schemas` has a `fileMatch` array of normal file paths or globs to match up graphql files to their schemas <3
	- normalized `graphql-config` output could likely be mapped to this!
	- the api exposes a function to updateSchemas() in the GraphQLWorker.
	- whenever the `schemas` change, the editor language features are reset
	- GraphQLWorker will keep an in-memory cache of schema information despite changes to config


dotansimha:
@acao I think having alpha releases will make testing for users easier. At the moment it's hard to test active PRs before they are being merged.

I started a draft PR here: https://github.com/graphql/graphiql/pull/2000 (based on changesets and release only affected packages)

acao:
@dotansimha sure! i'm just asking folks to review the deploy previews which seem to have vanished :(

acao:
@dotansimha we used alphas before and it was really messy :( I think your approach with using canaries for each PR will be much better

orta:
On TypeScript we generate playgrounds for people to test the PRs against by requesting them in the PR - feels like something similar could happen here?

acao:
@orta we generate three deploy previews for each PR currently

acao:
if you look at "deployments" below, they stopped deploying 3 days ago for some reason :/

acao:
we had users testing the deploy preview and I got one "hey this is awesome" comment on twitter so someone saw it at least

acao:
see:  #2000 has deploy previews for example.

i moved away from running the builds on netlify, and instead we push the builds directly from github actions using a 3rd party netlify upload action

acao:
@orta @dotansimha here's the preview fyi!

https://6183e02cae22a8ea2ad1df20--monaco-graphql.netlify.app/

acao:
So, one thing I really need feedback on is the schema loading mechanism. @orta your knowledge of monaco best practices would be super to have here, but I know you're busy.

If you notice, once the github schema finally loads, the variables mode loads language features far more quickly than our graphql mode. This latency in the graphql mode would be reflected every time a schema loads, and the larger the schema, the more dramatic it will be. luckily once a graphql language worker is instantiated, the schema persists, so this latency is not reflected in every webworker request thankfully

my assumption here is that printing the schema to an SDL string, handing it to the worker instance and asking the worker instance to convert an SDL string back to a GraphQLSchema is slowing us down. Obviously we can't just pass a GraphQLSchema primitive between the main and webworker processes.

a good modification to this schema loading refactor could be to just pass the introspection schema JSON off to the worker and do build client schema, which would probably reduce the CPU overhead dramatically, and may even impact the memory footprint at scale

speaking of, the memory heatmap for just the simple monaco demo is wild. something like 30MB just in strings in the buffer?! obviously the schema is the major factor here, as github's schema is sizeable enough for some testing of scaleability (this is why i moved to using the github api)

packages/graphql-language-service-utils/src/getVariablesJSONSchema.ts
```diff
@@ -0,0 +1,116 @@
...
+function getJSONSchemaFromGraphQLType(
+  type: GraphQLInputType,
+): { definition: JSONSchema6; required: boolean } {
+  let required = false;
+  let definition: JSONSchema6 = {};
+  if ('description' in type) {
+    definition.description = type.description as string;
+  }
+  if (isEnumType(type)) {
+    enumType(definition, type);
```
saihaj, line 77: imo rather than having these mutable functions which make it little magical these should just return normalized value here and then consume them here.

examples/monaco-graphql-webpack/src/index.ts
```diff
@@ -12,17 +12,21 @@ import JSONWorker from 'worker-loader!monaco-editor/esm/vs/language/json/json.wo
 // @ts-ignore
 import GraphQLWorker from 'worker-loader!monaco-graphql/esm/graphql.worker';
 
+import { LanguageServiceAPI } from 'monaco-graphql/src/api';
+
 const SCHEMA_URL = 'https://api.github.com/graphql';
 
 const SITE_ID = '46a6b3c8-992f-4623-9a76-f1bd5d40505c';
-let API_TOKEN = localStorage.getItem('ghapi') || '';
+let API_TOKEN = localStorage.getItem('ghapi') || null;
```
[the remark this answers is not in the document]
acao: This would break as API_TOKEN is mutated elsewhere

packages/graphql-language-service-utils/src/getVariablesJSONSchema.ts
acao: That would break recursion

orta:
I don't know much about how we stream the introspection query results, but If sending the results across webworkers is a pain maybe have each web worker make a request for it. Perhaps browsers are smart enough to share the request?

acao:
@orta sharing requests is a very interesting idea, though I think that is often more in the realm of service workers than web workers, right? 

I originally did have each webworker make their own schema request and this causes several issues:
- we still often need access to the schema in the main process which, for example, we can't pass the results of `getOperationFacts` across the webworker/main process boundary. we could change the function to return AST however.  passing the schema up from a webworker, instead of making yet another request from the main process, requires converting it to a string or AST. the github schema AST still threw an error last I tried but it could've been something else
-  it would cause extra request traffic because sometimes webworkers dissapear and come back (this may have been an issue with our implementation as well, but it seems webworkers are recreated, so keeping the schema primitive and string (better yet introspection string in the next commit) in a `Map` for now
- centralising the schema in the main process provides a single source of truth, which makes reloading, changing, cacheing, etc easier. 
- this allows people to pass their own API callbacks, primitives, etc for the purpose of loading schema, such as the fetcher pattern. this currently cannot be passed to a webworker without our pattern of bootstrapping the webworker in a separate file which is annoying and will be difficult to document and support across bundlers. I have thought of even using `createFetcher` in `monaco-graphql` `api` itself ?


also just as a side note, all of this needs some tests, both unit and cypress

packages/graphql-language-service-utils/src/getOperationFacts.ts
```diff
@@ -0,0 +1,67 @@
...
+    OperationDefinition(node) {
+      operations.push(node);
+    },
+  });
+
+  return { variableToType, operations, documentAST };
+}
+
+/**
+ * for backwards compatibility
```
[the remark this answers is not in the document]
acao: ah yes, we should put the `@deprecated` in the `graphiql` repo, with instructions to import it from `graphql-language-service`. 

I aim to propose merging `graphql-language-service-utils` with `graphql-language-service` as well, but perhaps the seperation makes sense if I aim to eventually use the `GraphQLLanguageService` class in the language server perhaps? Similar to how the official vscode language servers & monaco modes use the same underling class. 

It would make sense to put this in `@graphiql/toolkit` as well perhaps, because they are operation oriented, and we actually copied some of the other graphiql utilities into `vscode-graphql`

acao:
and it bears repeating, here's the deploy preview since we need to get those working properly again:
https://6183e02cae22a8ea2ad1df20--monaco-graphql.netlify.app/

packages/monaco-graphql/src/graphqlMode.ts
```diff
@@ -25,82 +28,87 @@ export function setupMode(defaults: LanguageServiceAPI): IDisposable {
...
-  function registerFormattingProvider(): void {
-    const { modeConfiguration } = defaults;
-    if (modeConfiguration.documentFormattingEdits) {
-      providers.push(
-        monaco.languages.registerDocumentFormattingEditProvider(
-          defaults.languageId,
-          new languageFeatures.DocumentFormattingAdapter(worker),
-        ),
-      );
+  defaults.onSchemaLoaded(_api => {
```
acao, line 35: this is probably one of the most important fundamental improvements to `monaco-graphql` in this PR. 

previously, the language service would load whether a schema was present or not (a behaviour I had inadvertently mirrored from `monaco-json`, which actually makes sense for them), and with webworkers, there was no single source of truth to know whether each webworker had independently requested and loaded it's own schema. this would lead to console errors and unnecessary churn for client implementations.

here, we use the event emitted `onSchemaLoaded` to wait for the schema to be present before loading any schema driven language features. note that schema-agnostic features such as language configuration, the tokens provider and the formatting provider ala `prettier/standalone` are registered above as immediately as possible, without waiting for `onSchemaLoaded`. 

as an argument to the `onSchemaLoaded` EventEmitter callback, the`api` instance is passed as an argument because why not? but the existing `GraphQLAPI` instance we have above is already sufficient to use for schema driven features at this point. `getSchema()` will read from the in memory cache, and `reloadSchema()` and `setSchema()` or `setSchemaConfig()` etc will trigger a new schema loading event, which will then re-load the language features as we expect

acao: I believe @cshaver and others have reccomended this idea before, which makes way more sense!

acao:
> [6183e02cae22a8ea2ad1df20--monaco-graphql.netlify.app](https://6183e02cae22a8ea2ad1df20--monaco-graphql.netlify.app/)

@china-bin what do you think of these improvements and API changes? i saw your docs mentioned `monaco-graphql` :)

The most troubling change i foresee is going from the api singleton to an async instantiation pattern. It guarantees some sanity in terms of loading schema-dependent language features when the schema is present, and I hope the `onSchemaLoaded` method is helpful as well

acao:
@dotansimha do you know why it's saying that there are no linked changesets for this PR? There were none originally, but they were added in a later commit. Hmm!

acao:
There's a bug in the lil demo app where the loading statuses don't show correctly, and it has to do with a race condition with the event emitters occurs when I switch to a schema from memory cache vs requesting it via network.  This is why you only see it the third time you switch a schema - the schema and language features change, but the status does not because these events are firing in the wrong order in the cache scenario. I think it's related to async vs sync event emitter handlers

Note that the JSONC part I mentioned on twitter lives entirely in the webpack example. just an example of how to integrate it.

I have ideas for adding operation execution to the language worker, but that can wait until after definition lookup, a proper file cache, and graphql-config style project config support.

I'm going to write up more in the PR description to explain the latest commits. They allow many features, for example, configuring the schema with a `schema` primitive.

acao:
@orta deploy previews are back again if you want to take a look!

acao:
wait what?! it's not even hitting the schema cache on netlify, making a request on every schema change!? it literally works different on my local 😆  😭  oh well there are plenty of other features to enjoy haha

acao:
funny enough, I tried serializing the JSONSchema response from webworkers, to hand off all the parsing and simple AST/business logic to the webworker, but it seemed a bit slower than doing it all in main process for a back and forth payload.

JSON.parse() is fast in the main process on the receiving end, but when the webworker does the work of generating the schema, it has to JSON.stringify() it to send it back to main (inevitably, the schema result needs to be consumed by a monaco editor API in main process)

Of course, if benchmarked it could go the opposite way for large operations with many complex variables. The JSON Schema generation is deeply recursive, so making a web worker handle the work in a way that doesn't block the main process is fabulous

acao:
@Nishchit14, @stonexer, @cshaver, @jensneuse anyone who might be interested!

acao:
- introduce api for diagnosticsOptions for validating variables json
- fileMatch schema config option allows multiple schemas
- generate jsonSchema in GraphQLWorker
- variables diagnostics happen alongside graphql diagnostics
- monaco-graphql no longer fetches remote schema
- update example to fetch schema
- move schema cache to language service
- initialize -> initializeMode, no longer async
- no more languageService in the main process

acao:
ah! I was worried this would happen.

the added behaviour for changing the worker schema is causing duplicate language workers to be generated :(

unless I can figure out this issue of the monaco webworker instantiation pattern, we will just have the mode re-create the GraphQLWorker between schema changes

acao:
once I push these doc changes tonight, hopefully folks can review the API changes and tell me what they think!

PR #1998, opened 2021-11-04
build: remove codemirror dependency from lsp

ferm10n:
`CodeMirror.Token` is identical to `ContextTokenForCodeMirror`

see https://github.com/graphql/vscode-graphql/pull/335#issuecomment-957665844

---

this will allow updating the vscode-graphql to update to the latest packages from this project without having to add codemirror as a dependency

### DIFF
--- a/.changeset/forty-apes-retire.md
+++ b/.changeset/forty-apes-retire.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-interface': patch
+---
+
+remove implicit codemirror dependency
--- a/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service-interface/src/getAutocompleteSuggestions.ts
@@ -7,7 +7,6 @@
  *
  */
 import { CompletionItemKind } from 'vscode-languageserver-types';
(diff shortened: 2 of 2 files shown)


acao:
I'm trying to recall exactly why we made this change before, but it doesn't seem like we need to import all of `@codemirror/types` just for one token type.  @yoshiakis is this going to be an issue for codemirror-graphql consumers, to keep the duplicate type in sync?

ferm10n:
Now that I say that though, _why_ does vscode-graphql's build type check this module anyway? Shouldn't node_modules get excluded?

imolorhe (approved):

imolorhe:
The only issue I see here would be keeping the types in sync, assuming the Token type in codemirror is updated, but this looks like a cleaner alternative, considering how the codemirror Token is probably the most used (and frequently referenced) type in the Codemirror package.

acao:
@ferm10n excellent point! i think we need to tweak some tsconfig on the vscode-graphql side at least...

@imolorhe another great point. I swear I found a way to automatically extract external definititions statically to the .d.ts files once, but after some time searching I couldn't remember how I accomplished it. Triple slash directives perhaps?

PR #2009, opened 2021-11-07
chore: deploy previews, remove examples workspaces

acao:
- hopefully fixes deploy previews
- remove examples from workspaces. makes yarn much faster. in preperation for yarn 3 upgrade


acao:
the deploy previews issue will be fixed by moving back to netlify builds, but i'm worried I'll be using up resources used by graphql.org

most of the examples workspaces have been removed!

PR #2010, opened 2021-11-07
Graphql 16 upgrade

acao:
Attempt to support GraphQL v16 (cloned from @saihaj's PR)

cc: @saihaj

Closes #1982

PR #2013, opened 2021-11-08
Add missing changesets

PabloSzx:
As mentioned in https://github.com/graphql/graphiql/pull/2010#issuecomment-962743181

- Added missing changeset for graphql-language-service-utils
- Added patch changeset for graphql-language-service-utils dependants

I also found that graphql-language-service-types was not published with the only change it had, which is adding graphql 16 in the peer dependencies:

- Added patch changeset for graphql-language-service-types

@acao

### DIFF
--- a/.changeset/large-points-add.md
+++ b/.changeset/large-points-add.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-utils': minor
+---
+
+upgrade to `graphql@16.0.0-experimental-stream-defer.5`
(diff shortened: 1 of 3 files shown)


saihaj (approved):

n1ru4l (approved):

acao:
@dotansimha do you know why changesets isn't auto-incremementing these releases like lerna publish did? are we missing config?

PR #2016, opened 2021-11-09
Configure changesets to automatically publish internal dependents

acao:
Configure changesets to automatically publish internal dependents, ala #2013 

it seems that lerna publish automatically handled this, whereas changesets allows you to change packages without publishing their changes. I think this option will *force* it, even if a user removes packages from the changeset


acao:
https://github.com/atlassian/changesets/pull/542

PR #2024, opened 2021-11-13
chore: fail netlify, consolidate graphiql webpack examples

acao:
The Netlify deploy previews are failing silently, so this at least exposes the failure

Experimenting to see if:

- the credentials failure is related to environments each having their own environment secrets?
- whether permissions errors mean PRs from forked repos aren't allowed to create deployments
- perhaps we should consolidate to one environment?


acao:
ok cool it worked that time I guess?

PR #2031, opened 2021-11-17
fix: Add graphql-language-service-utils to dependencies of codemirror-graphql

RDIL:
I made this change because it fixes codemirror-graphql with Yarn v2/3's plug-and-play.

Example error before this patch:

```js
./.yarn/__virtual__/codemirror-graphql-virtual-738f7121e6/0/cache/codemirror-graphql-npm-1.1.0-2d50dab87f-abfa711716.zip/node_modules/codemirror-graphql/hint.js
Module not found: codemirror-graphql tried to access graphql-language-service-utils, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

After, it should work just fine.

### DIFF
--- a/.changeset/beige-boxes-suffer.md
+++ b/.changeset/beige-boxes-suffer.md
@@ -1,0 +1,5 @@
+---
+'codemirror-graphql': patch
+---
+
+fix: Add graphql-language-service-utils to dependencies of codemirror-graphql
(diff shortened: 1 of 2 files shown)


acao:
thanks for this! #2037  will take care of this issue

Issue #2022, opened 2021-11-12
CM6 support for codemirror-graphql


sergeichestakov:
Hey there!

I was looking to add graphql support to our app which uses CodeMirror 6 and stumbled upon the `codemirror-graphql` package in this repo. However, it looks like that package only works for CodeMirror 5 and is therefore not compatible with CM6. I also couldn't find any alternatives that do work with CM6.

I saw that there was a way to use CM5 language packages in CM6 (e.g. the modes in this repo: https://github.com/codemirror/legacy-modes/) but only if they implement the `StreamParser` interface which I don't think is the case here.

Just wondering if there were any plans to add support for CM6 in the `codemirror-graphql` package. Happy to help out in this effort myself as well and open up a PR if this is something you guys want and no one else can pick it up.

Thanks in advanced!

DelysidBot:
;)

acao:
any support for that would be more than welcome!

i’m deep into preparing monaco-graphql for 2.0, but we have contributors and co-maintainers who  care a whole lot about codemirror-graphql, and considering they are both using the same underlying language service libraries, they will continue to improve one another

acao:
funny enough, I tried serializing the JSONSchema response from webworkers, to hand off all the parsing and simple AST/business logic to the webworker, it seemed a bit slower than in main for a fill back and forth payload. 

Of course, if benchmarked it could go the opposite way for large operations with many complex variables. The JSON Schema generation is deeply recursive, so making a web worker handle the work in a way that doesn't block the main process is fabulous

imolorhe:
Adding support for CM6 would be great. I can try to support any efforts on this but unfortunately can't commit to picking this up this week. I should be able to do so by next week though if it isn't already being implemented by then.

imolorhe:
I've added support for the stream parser. This only handles basic syntax highlighting.

imolorhe:
To have support for all the other features, we would need to implement a [Parser](https://lezer.codemirror.net/docs/ref/#common.Parser) class. The current parser in `graphql-language-service-parser` is not compatible with that class.

sergeichestakov:
Amazing thank you so much! Really appreciate the swift response.

Any plans to publish that package to NPM? Even just the basic syntax highlighting would be great.

acao:
released! 🥳  thanks @imolorhe !

sergeichestakov:
thanks so much! Super excited for this.

[...]

PR #2035, opened 2021-11-21
Added Codemirror 6 legacy mode support

imolorhe:
Implemented the codemirror.next legacy mode support for graphql mode.
This only covers syntax highlighting. Autocomplete, jump and info extensions
still need to be implemented to have feature parity with Codemirror v5.

From #2022

Before:
Screenshot 2021-11-21 at 4 52 55 PM

After:
Screenshot 2021-11-21 at 4 52 45 PM


acao:
Thank you so much! Will review asap, working on monaco as we speak.

acao:
@imolorhe also, if you add a changeset it will deploy a canary release :). Thanks for that @dotansimha !

acao (approved):
awesome work!  I think this is good to go.
let's go ahead and merge this, and then these steps will follow:

- a PR to change `codemirror-graphql` and `graphiql` to use `graphql-language-service` directly instead of interface, parser, and utilsl
- upgrade `graphiql` to use codemirror 6
- release!

imolorhe:
I like the ambition but I'm afraid we shouldn't update graphiql to use cm5 just yet 😄

Codemirror 6 is officially still in beta so not sure it would be wise for us to use it just yet.

Also the changes introduced here only provide support for syntax highlighting. All the other nice things we have (autocompletion, jump, linting, etc) in codemirror 5 aren't supported yet.

Adding support for those would require that we create a new graphql parser class that is supported by codemirror 6, which would likely be a major lift.

sergeichestakov:
FYI it looks like `@codemirror/*` packages should be specified as `peerDependencies` for published packages. I've opened a PR with more detail and a fix here: https://github.com/graphql/graphiql/pull/2369

pls let me know if that looks ok!

PR #2037, opened 2021-11-23
use graphql-lang-service in codemirror-graphql

acao:
use `graphql-language-service` instead of `graphql-language-service-utils`, etc - towards eventually combining them.

it makes it easier to manage dependency versions as well


willstott101:
Thanks for this! Unfortunately it looks like there is one left!

https://github.com/graphql/graphiql/blob/main/packages/codemirror-graphql/src/utils/mode-factory.ts#L16

I'm happy to open an MR

PR #2045, opened 2021-11-24
missing changeset, fix e2e, node 16

acao:
the node 16 upgrade seems to make yarn install about ~40s faster in github actions!


acao:
for some reason, these canaries still have different peer dependencies than we have in main... :( @dotansimha do you know why this would be? for example, installing graphql-language-service-server still shows the wrong peer dependencies for `graphql-language-service`, i'm not sure why

PR #2046, opened 2021-11-24
codemirror-graphql's mode-factory.ts to use graphql-language-service

willstott101:
Follow on from #2037


acao:
Oops good catch! I merged these in the wrong order it seems

PR #2047, opened 2021-11-24
Include source code in all packages to fix source maps.

willstott101:
Currently some build tools (**cough** react-scripts on webpack 5 **cough**) error out when trying to build source maps for the graphiql packages. This makes development much harder than it needs to be. By including the source code in the packages we build, the source maps are more likely to work...

PR #2048, opened 2021-11-24
Embed source code in source maps

willstott101:
N.B. This doesn't work with declarationMap - as that's only useful in an IDE - where the file existing is the only way most IDEs will show you the file... therefore #2047 is probably a better fix.


willstott101:
Closing due to #2047 being better.

Issue #1894, opened 2021-06-21
Poor trigger characters for the language server


rchl:
Currently the only registered trigger character for the language server is `@`:
https://github.com/graphql/graphiql/blob/65aa64811ccd40c5e8ba99edec825a91115a3c8b/packages/graphql-language-service-server/src/MessageProcessor.ts#L168

I think there are many cases when automatically triggering completions would be beneficial. For example when typing `{` to select fields of an object or when typing `(` after field name to specify the arguments to pass). None of those currently trigger auto-completion which means that the user has to trigger it manually with a shortcut.

rchl:
Also:
- `:` - when typing it in arguments list, for example `$arg: String`
- '$` - in `query($arg: String!)`, for example

acao:
This logic is built into getAutocompleteSelections! There is in fact autocomplete for these characters in language interface, which is used by both the language server and graphiql

rchl:
I don't quite understand what are you referencing here. :)

Is this is some sort of developer note on how this could be fixed?

acao:
Yes I will add notes. We need to add additionalTextItems or whatever they are called. I am planning a whole writeup on how to implement this for both language server and monaco and i think i have an issue already. The vscode-json-languageservice provides a great reference implementation

rchl:
`additionalTextEdits` is a property of the completion and can be used to used to perform additional edits on the document when a completion is committed. I'm not sure if that is strictly related to my request. The `triggerCharacters` is what controls when the completions are shown in the first place.

Do note that this needs to be done carefully to avoid triggering completions too eagerly. For example it might not be the best idea to trigger on typing `{` to open a block since the user might want to add a line break before triggering completions.

Unfortunately the LSP spec is not very flexible in this area since the triggers are only character-based. So it might not be that easy or even possible to define good triggers for completing fields, for example.

acao:
additionalTextEdits is the answer I assure you. Please review the completion support in vscode-json-languageservice to see how they use this feature

acao:
I have already gotten it working partially in experiments. Trigger characters setting does not do what you think. Additional text items is how we evaluate this on a per token basis

acao:
For example, in monaco graphql this weekend i had an experiment where any time a user selected an object or list stype field from completion, it would automatically expand braces and insert the cursor. That’s what you’re talking about in vscode right? see how the vscode-json-languageservice passes on these additional items in it’s completion to monaco-json and vscode differently.

[...]

PR #2070, opened 2021-11-29
fix: autocompletion for argument values, more trigger characters

### LINKED ISSUE #2068: Small bug when autocompleting variables
## Steps to Reproduce the Problem Or Description

Type
```
query ($test: String!) {
  hello (test: $[CURSOR HERE])
}
```

Press tab to autocomplete, select `$test`.

## Expected Behavior

```
query ($test: String!) {
  hello (test: $test)
}
```

## Actual Behavior

```
query ($test: String!) {
  hello (test: $$test)
}
```

## Specifications

  - GraphQL for VSCode Extension Version: v0.3.16
  - VSCode Version: 1.56.2
  - OS Name: Arch Linux
  - OS Version: 2021.05.01

acao:
fixes #2068 🎉  and #1894 (i think - @rchl do you agree?)

simplifies the code a bit. untangles a mess that I left when introducing argument value completion last year.

introduces more `triggerCharacters` to `monaco-graphql` as a precursor to #1997 

fixes this bug across the entire ecosystem by fixing the issue at the core `graphql-language-service-server` level

### DIFF
--- a/.changeset/green-ghosts-agree.md
+++ b/.changeset/green-ghosts-agree.md
@@ -1,0 +1,6 @@
+---
+'graphql-language-service-interface': patch
+'monaco-graphql': patch
+---
+
+Fix a bug with variable completion with or without `$` across the ecosytem. Introduce more triggerCharacters for fun!
--- a/packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
+++ b/packages/graphql-language-service-interface/src/__tests__/getAutocompleteSuggestions-test.ts
@@ -279,7 +279,7 @@ query name {
     ]);
   });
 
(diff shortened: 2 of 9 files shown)


acao:
@rchl introducing more `triggerCharacters` here just to get it out of the way, what do you think?

acao:
so, this doesn't impact `codemirror-graphql` to be clear, because it doesn't use the `command` or `insertText` options we've added to the autocompletion

rchl:
Sorry but to test this properly, I'd need this working in my editor. I'm not sure what to expect exactly from the codemirror behavior.

I'll check it out when it's out and report if I see any issues. :)

acao:
@rchl you should expect the monaco demo to complete variables and trigger completion. Codemirror is not changed by this PR

PR #2075, opened 2021-12-03
fix: fix `CompletionItemKind` import issue

iahu:
export JavaScript `CompletionItemKind` enum module from `vscode-languageserver-types`, but not only export a TypeScript type.


acao:
Thanks for opening this!

sorry there is a bug with deploy previews for forks

we need to not export this const from vscode-languageserver-types as importing it breaks the GraphQLWorker, and importing it in main process as we do now causes tree shaking/bundle bloat issues as well. Thus why we have two different toCompletion util functions. so, iterating over completions again & mapping the completion kinds in main process

iahu:
Sorry, I am not sure what is the tree shrinking issue you have described.
But we use it as an enum variable on [here](https://github.com/graphql/graphiql/blob/fe5bb0dca326ee307bb6a618536819c0b505f7c9/packages/monaco-graphql/src/languageFeatures.ts#L120), in that scene, we have to import it as a JavaScript variable, but not only a type, if not, ther should be a error.

acao:
I cannot seem to find where this is not working, can you elaborate/demonstrate? Is there a reason you made this change?

The monaco-graphql webpack example works great, and the const enum is inferred properly.

PR #2146, opened 2022-01-21
Adding Darkmode to graphiql

harshitkumar31:
Addresses https://github.com/graphql/graphiql/issues/129
Screenshot 2022-01-21 at 8 58 12 PM


acao:
also, can you add it to `GraphiQLProps` and document the property?  currently, it's only been added to `GraphiQLState`

packages/graphiql/src/css/app.css
```diff
@@ -592,3 +592,8 @@ div.CodeMirror-lint-tooltip > * + * {
 .cm-atom {
   color: #CA9800;
 }
+
+
+.graphiql-dark {
+  filter: invert(1) hue-rotate(180deg);
+}
```
acao, line 599: newline please!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -714,197 +721,205 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
-              this.editorBarComponent = n;
-            }}
-            className="editorBar"
-            onDoubleClick={this.handleResetResize}
-            onMouseDown={this.handleResizeStart}>
-            <div className="queryWrap" style={queryWrapStyle}>
-              <QueryEditor
-                ref={n => {
-                  this.queryEditorComponent = n;
+      <>
```
acao, line 724: I don't think we need to wrap the entire output in a fragment. I suspect this is why there is such a huge diff in this render!? would like to simplify this diff so I can understand these changes better

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -310,6 +310,7 @@ export type GraphiQLState = {
   operations?: OperationDefinitionNode[];
   documentAST?: DocumentNode;
   maxHistoryLength: number;
+  darkMode: boolean;
```
acao, line 313: do we want to provide a `defaultTheme`: "light" | "dark" prop so that implementers can default to dark mode for their users?

acao (changes requested):

acao:
looks great overall! again, thanks so much for this

acao:
damn, that was fast!

acao:
Actually you know what - we might need to re-evaluate this strategy for 1.0 because it will invert `GraphiQL.Logo` which is still supported until 2.0, as well as images in markdown docs. I didn't realize `filter` would work on images! Can we use a different strategy?

harshitkumar31:
@acao , Added a style to ensure images are inverted.
For the lack of a better example, I added an image inline to test it. This should address your concern w.r.t `GraphiQL.Logo`
Screenshot 2022-02-04 at 12 53 17 AM

acao:
thank you!

acao:
Hmmm... doc explorer font colors... I don't know... hmm!

pedroleinar:
Hi, how you doing ? what should be changed? Can I help you?

acao:
I have been working and sleeping? it is 7am here lol.

To be clear, I don’t think the css filter approach will work. Someone tried this PR years ago and I asked them to just override the css classes the more conventional way and they abandoned the PR.

This approach makes all the colors look weird, and now the doc explorer fails accessibility tests if dark mode is enabled because of the color contrast

There is a css file someone posted on gist in that thread that should be fine with some tweaking, but make sure it works for query history as well

in the new graphiql rfc as you can see we dont use CSS files, we use theme ui & emotion css, so you don’t have to worry about this working in 2.0 because we will use a different technique for accessing dark/light/solarized and other filtered theme modes, as well as user theme selection. User just provides a set of theme object literals.  This is why we were waiting for 2.x to do this but no one is working on graphiql or has time

harshitkumar31:
I see. I don't have enough cycles or knowledge to implement the suggested changes. I'll close the PR, thanks.

acao:
Well, you don’t need to close it in my opinion. I am glad to help you with this PR. I appreciate your initiative immensely, and a dark mode would be a great addition to wrap up 1.0

I completely forgot that a major part of this strategy should involve dynamically changing the codemirrir theme as well!

acao:
It could be that we also get filter: invert() working where the doc explorer items look funky now (try doc search as well). Probably just a few more css selectors is all we need

acao:
How about I create a new PR with your commit and my changes? That way you get credit for the original approach. I think we can make this work even if you don’t have time to finish. I think this will work out just fine!

harshitkumar31:
Sounds good. Thanks @acao

thomasheyenbrock:
As part of the designs for GraphiQL v2 we'll ship a dedicated dark mode implementation, for updates see #2328. I'm closing this PR in favor of the upcoming dark mode implementation. Thanks anyways! 🙂

PR #2154, opened 2022-02-04
fix: example uses the opts as headers

roderik:
After fighting all day with cors headers because 'shouldpersistheaders' was not an allowed key, I realized that the example is wrong 🤦🏼‍♂️


acao:
thanks for the fix!

Issue #2155, opened 2022-02-05
graphiql requires the "global" global variable to be defined


francisu:
This is not necessarily present in the browser:

https://developer.mozilla.org/en-US/docs/Glossary/Global_object

This causes a runtime error with using graphiql when bundling with the default configuration in either [esbuild]https://esbuild.github.io/ or [Vite]https://vitejs.dev/.

Fix is simple, PR will be provided shortly.

```
GraphiQL.js:661 Uncaught ReferenceError: global is not defined
at GraphiQL2.componentDidMount (GraphiQL.js:661:9)
at commitLifeCycles (react-dom.development.js:20663:24)
at commitLayoutEffects (react-dom.development.js:23426:7)
at HTMLUnknownElement.callCallback2 (react-dom.development.js:3945:14)
at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
at invokeGuardedCallback (react-dom.development.js:4056:31)
at commitRootImpl (react-dom.development.js:23151:9)
at unstable_runWithPriority (scheduler.development.js:468:12)
at runWithPriority$1 (react-dom.development.js:11276:10)
at commitRoot (react-dom.development.js:22990:3)
GraphiQL2.componentDidMount @ GraphiQL.js:661
commitLifeCycles @ react-dom.development.js:20663
commitLayoutEffects @ react-dom.development.js:23426
callCallback2 @ react-dom.development.js:3945
invokeGuardedCallbackDev @ react-dom.development.js:3994
invokeGuardedCallback @ react-dom.development.js:4056
commitRootImpl @ react-dom.development.js:23151
unstable_runWithPriority @ scheduler.development.js:468
runWithPriority$1 @ react-dom.development.js:11276
commitRoot @ react-dom.development.js:22990
performSyncWorkOnRoot @ react-dom.development.js:22329
(anonymous) @ react-dom.development.js:11327
unstable_runWithPriority @ scheduler.development.js:468
runWithPriority$1 @ react-dom.development.js:11276
flushSyncCallbackQueueImpl @ react-dom.development.js:11322
flushSyncCallbackQueue @ react-dom.development.js:11309


```

Ionshard:
I am also running into this error when building an application with GraphiQL via an Nx project

acao:
[...]

PR #2156, opened 2022-02-05
fix: should not depend on 'global' global variable (#2155)

francisu:


PabloSzx:
image

👀

acao:
Dammit, I said something about this in review but the review was still pending. My bad, will fix it

PR #2161, opened 2022-02-08
Improve the document cache miss error message

orta:
Helps folks get a bit more insight into what's going on when this error is thrown:

Screen Shot 2022-02-08 at 1 36 37 PM

### DIFF
--- a/.changeset/chilly-birds-poke.md
+++ b/.changeset/chilly-birds-poke.md
@@ -1,0 +1,6 @@
+---
+"graphql-language-service-server": patch
+"graphql-language-service": patch
+---
+
+Do not log errors when a JS/TS file has no embedded graphql tags
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -485,7 +485,7 @@ export class MessageProcessor {
 
     const cachedDocument = this._getCachedDocument(textDocument.uri);
     if (!cachedDocument) {
-      throw new Error(this._errorMessageForMissingDocument(textDocument.uri));
+      return [];
     }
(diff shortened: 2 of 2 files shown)


acao:
Thanks for this! Added to the pinned roadmap

In most cases this log message doesn't need to be reported at all.

For example, this error is shown when files are parsed but no documents are found, for example on save.  this is very common for .js/ts/etc files without graphql`` tags.

However, a more relevant message is clipped at the top, which indicates an error from `graphql-config`. This and other `graphql-config` error messages we should show in debug output, so that users can debug with `graphql-config` maintainers more easily.

The frequency of the message is a symptom of other issues with over-parsing that we need to ameliorate.

Again, thanks a bunch for this! This is a good step in the right direction for error logging in the LSP server

acao:
@orta can you add a changeset?

acao (changes requested):
I can't remember offhand if there's a log level setting - perhaps these can be `debug()` level output? Or removed. Do you think they are useful? I don't think so, as per the comment

acao:
So these messages are a great improvement to be sure, but instead of throwing an error, we can do this:

```ts
if (!cachedDocument) {
 	this.logger.debug(this._errorMessageForMissingDocument(textDocument.uri));
 	return
}
```

orta:
Yeah, agree, we should remove the error messages realistically because they're just not that useful

acao:
This was another one of those situations where I couldn’t understand why it was there in the first place. Previously almost nothing went to the output panel and all logging went to file, but I exposed some to output panel, and the logging to file is yet another thing to question, another artefact of the pre-LSP era of this server

acao:
@orta beautiful, thank you!

PR #2167, opened 2022-02-15
fix: global

acao:


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -531,7 +531,7 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     // Utility for keeping CodeMirror correctly sized.
     this.codeMirrorSizer = new CodeMirrorSizer();
 
-    if (global) {
+    if (typeof global !== undefined) {
```
PabloSzx, line 534: Typeof only returns strings, this conditional is always going to be true

PR #2187, opened 2022-02-20
integrate vscode-graphql with the repository

acao:
It's about time! This will save us and our users so much headache!

- [x]  integrate dependencies
- [x] get typescript project references build working
- [ ] fix graphql-js duplicate resolution error
- [ ] get tests working
- [ ] add `build-bundles` script for `esbuild` , to bundle/code split `vscode-graphql` when publishing
- [ ] fix `.vscode/launch.json`, confirm build watch is working

CC @divyenduz in case you're interested!


acao:
@brianwarner so uhh... hahahaha... this is what happens now when we try to merge repos

trying to include the history for vscode-graphql by using subtrees, but it seems like CLA bot doesn't like that many of the original contributors didn't sign the CLA because they contributed before it joined `graphql` org. What to do?

orta:
Good luck!

acao:
@orta blocking issue noted as #3 here if you're able to help:

https://github.com/graphql/graphiql/issues/2204

as you can see from the failures, we need to be on the latest version of graphql@16. when i tried to do that in this PR, the test suite failed for all packages.

Issue #2074, opened 2021-12-02
Variable editor have an infinity scroll when the user wanted to resize it


Zeuhan:
Hi Team,

When the user want to resize the variable editor and keep the bar at the top the container starts to increase without any kind of limit so that cause an inside infinite scroll.

Screenshot of the UI after scroll for a few seconds: 
![image](https://user-images.githubusercontent.com/48504276/144462299-03550ee5-e8c2-465a-863a-0aff82b78aab.png)

Screenshot of the amount of pixels that have in height: 
![image](https://user-images.githubusercontent.com/48504276/144462412-98c4aa26-202f-4329-9059-e3200ec57128.png)

Edit: I tried to put a max-height with the .CodeMirror-scroll class and I tried to figure out creating a div outside to limit the height but not results.

acao:
I can't reproduce it here yet:
https://graphiql-test.netlify.app/

perhaps you could provide a video or gif to demonstrate it?

Zeuhan:
Steps to reproduce:

1. Drag the variables menu to the top and maintain there for a little bit.
2. You will have an infinite scroll

Video from https://graphiql-test.netlify.app/ :
https://user-images.githubusercontent.com/48504276/145867868-5160131a-f6b7-4f95-ac1b-30d96223d46f.mov

acao:
thank you!

acao:
woah! that's a bug alright.

acao:
it must be a regression that I overlooked when we added the headers editor.

TheMightyPenguin:
@acao this seems to have been solved, should this issue be closed?

acao:
@TheMightyPenguin yes! thank you :)

PR #2188, opened 2022-02-21
fix(variable editor): limit height of editor

Shayshoon:
Added max height 100% to editor bar component

Relevant issue: https://github.com/graphql/graphiql/issues/2074


acao:
thank you @Shayshoon !

Shayshoon:
My pleasure!

PR #2189, opened 2022-02-22
feat: apply variable editor title text via CSS classes instead of inline styles

n1ru4l:
Apply variable editor title text styles via class `variable-editor-title-text` instead of using inline-styles. This allows better customization of styles. An active element also has the class `active`. This allows overriding the inactive state color using the selector `.graphiql-container .variable-editor-title-text` and overriding the active state color using the selector `.graphiql-container .variable-editor-title-text.active`.

### DIFF
--- a/packages/graphiql/src/css/app.css
+++ b/packages/graphiql/src/css/app.css
@@ -496,6 +496,16 @@ div.CodeMirror-lint-tooltip > * + * {
   margin-top: 12px;
 }
 
+.graphiql-container .variable-editor-title-text {
+  cursor: pointer;
+  display: inline-block;
+  color: gray;
+}
+
+.graphiql-container .variable-editor-title-text.active {
+  color: #000;
+}
+
 /* COLORS */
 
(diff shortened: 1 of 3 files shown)


packages/graphiql/src/css/app.css
```diff
@@ -496,6 +496,16 @@ div.CodeMirror-lint-tooltip > * + * {
   margin-top: 12px;
 }
 
+.graphiql-container .variable-editor-title-text {
+  cursor: pointer;
+  display: inline-block;
+  color: gray;
+}
+
+.graphiql-container .variable-editor-title-text.active {
```
n1ru4l, line 505: I could not identify a pattern for the naming of such selectors. Happy to replace `.active` with something else if desired!

acao: Oh, that’s should be fine!

acao:
Thanks @n1ru4l !

dotansimha (approved):

PR #2190, opened 2022-02-22
feat: onSchemaChange callback

n1ru4l:
This is a proposal for starting the discussion on such an API.

Currently, the only way of having the schema information outside of GraphiQL is to move the logic higher up in the tree and then pass the `GraphQLSchema` instance to `GraphiQL` via the `schema` property.

Having this convenient callback simplifies accessing the schema outside the `GraphiQL` component.

### DIFF
--- a/.changeset/seven-ties-march.md
+++ b/.changeset/seven-ties-march.md
@@ -1,0 +1,13 @@
+---
+'graphiql': minor
+---
+
+New callback property `onSchemaChange` for `GraphiQL`.
+
+The callback is invoked with the successfully fetched schema from the remote.
+
+**Usage example:**
+
+```tsx
+<GraphiQL onSchemaChange={schema => console.log(schema)} />
+```
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -284,6 +284,10 @@ export type GraphiQLProps = {
    * default: 20
    */
   maxHistoryLength?: number;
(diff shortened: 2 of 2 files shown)


dotansimha (approved):

acao:
Potentially we could do something like this for the 2.x rewrite, but if someone needs more control over the schema now, they can just pass the schema prop on schema change with their own schema fetch or by loading a schema file. I don’t think this change is necessary at the current moment

n1ru4l:
> they can just pass the schema prop on schema change with their own schema fetch or by loading a schema file.

Doing this is very verbose and not that intuitive, what are the potential issues with introducing this callback for GraphiQL 1.x.x?

acao:
We tried to do a feature freeze a while ago so folks would start contributing to 2.x. I think we need to introduce it again. No offense but I just spend all my time dealing with tweaks to graphiql@1 anymore. 

Please lets just focus on 2.x.  Half the code is there and the roadmap is there just please 😂

acao:
@dotansimha would you like to take over graphiql 1 maintenance so i can focus on 2.x? Otherwise I am inclined to feature freeze. The 2.x rewrite WG approved a feature freeze in 2020 😂

dotansimha:
> @dotansimha would you like to take over graphiql 1 maintenance so we can focus on 2.x? Otherwise I am inclined to feature freeze. The 2.x rewrite WG approved a feature freeze in 2020 😂

Sure, we would love to! :)

PR #2197, opened 2022-02-22
feat: tabs

n1ru4l:


acao:
@n1ru4l IIII seee what you're doing now... and it's freakin' fabulous :) 💯

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1052,11 +1227,18 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
+        fetcherOpts.headers = JSON.parse(this.state.headers);
+        // if state is not present, but props are
+      } else if (this.props.headers) {
+        fetcherOpts.headers = JSON.parse(this.props.headers);
+      }
+    } catch (err) {
+      this.setState({
+        response: 'Introspection failed as headers are invalid.',
+      });
+      return;
```
n1ru4l, line 1241: Note: we should have a refetch schema button

acao: I 100% agree, and a keyboard shortcut. We have a keyboard shortcut and a refresh button in the monaco editor demo already haha

acao:
part of the reason we wanted to rewrite the react componetry for this effort is `GraphiQL.tsx` itself

as this PR progresses, let's try to pull as much out of this monolith component, and into context providers & consumers as possible. otherwise, it will be very hard to expose useful APIs for user-defined components

acao:
@n1ru4l today I will try to restore deploy previews so that the community can easily see your work and give feedback. thanks for this!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -773,6 +881,134 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
...
+                        response: newActiveTab.response,
+                        tabs: { ...state.tabs, tabs, activeTabIndex: index },
+                      };
+                    }, this._persistTabsState);
+                  }}
+                  onClose={() => {
+                    if (this.state.tabs.activeTabIndex === index) {
+                      this.handleStopQuery();
+                    }
+                    this.setState(state => {
```
acao, line 935: as per the earlier request, I'd like to see if we can have some reusable functions external to this file for these state handlers.  Really don't want to increase complexity in this style, and most of these could be handled as pure reducer-like standalone functions within the arguments that can be later migrated into hooks

acao: Perhaps even the `Tabs` component could do this, and handle the tab state internally? and accept the required handlers as props for now?

acao (changes requested):
Let's see if we can architect this in a way that doesn't grow the GraphiQL.tsx component as much

acao:
@n1ru4l are you up for a discord discussion perhaps? or we could even record a pairing session?

packages/graphiql/src/components/GraphiQL.tsx
n1ru4l: The tabs component is not able of handling all the states as the state is also affected by various other event handlers. E.g. `handleEditQuery`.

I agree that we can refactor the whole state into a reducer in a follow-up pull request. I don't see why doing that would block advancing on this tabs feature.

In https://github.com/graphql/graphiql/pull/2197/commits/1c6ad348a35a04b3a1256760e8b28afc6f8abcc6 I moved the state update logic into reducer-like functions.

In general, I am not a big fan of separating state strictly from components and having to jump around too much for simple changes.

acao:
awesome! maybe they can live in a seperate file? would these be useful to people building other IDEs, and thus maybe live in a state management abstraction such as `@graphiql/toolkit`? if folks are using `immer` or similar it might not, but for many this could be quite useful, whether using vuex, redux, zustand, etc

packages/graphiql/src/components/GraphiQL.tsx
acao: I guess it's a common pattern for monolithic react library components to be one huge class component, essentially. People complain about the cognitive overhead, but you're correct, we can find ways to simplify it later. In the RFC effort two years ago we reduced this file to under 1000 lines, and moved a ton of the state into context. But we can still do that in later iterations, sure!

acao:
I still want to wait until deploy previews are working again so users can review changes like this. I have the day off because of my injury but I can wrap up that and a few other efforts hopefully

acao (approved):

acao:
@n1ru4l if you're able to make another commit, it should trigger a deploy preview now!

I will go through and update readme links later for the changed URLs, but this should work for now I think?

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -2084,3 +2283,179 @@ function isChildComponentType<T extends ComponentType>(
 
   return child.type === component;
 }
+
+function tabsStateEditHeadersReducer(
```
acao, line 2287: These are what i wanted to move into toolkit, but we can do that in a different PR

acao (approved):
This is ready to merge! Who wants to do the honors? Or shall I?

timsuchanek:
> This is ready to merge in my opinion. But I'd like it if someone else was able to review it!

I'll review this over the weekend!

timsuchanek:
Quick update: I'm checking out the PR now, working on a few styling adjustments that I'll propose shortly.
One question for @n1ru4l: Did you try out, if subscriptions are correctly working when switching back and forth between tabs with running subscriptions?

acao:
To note for 2.0.0: with `monaco-editor` we will need to be sure to not destroy the editor instance when changing tabs (by hiding the tab content with css) and instead to setModel() and change settings dynamically

timsuchanek:
This is a great start, awesome! I just worked on some styling adjustments, here the before and after:
### Before
https://user-images.githubusercontent.com/1094804/158071559-2aa350fc-9456-45aa-93d3-8ddf9072d27c.mp4
### After
https://user-images.githubusercontent.com/1094804/158071564-42a389a8-1e1b-43c9-8ad0-2ef8f4b0ebd7.mp4

### Changes
- If the tab title goes further than 100px, the tabs don't scale. In the playground, we just let the tabs grow with the title, that worked well so far, so that's also what I'm suggesting here
- The "X" (close button) is now using the same "X" unicode as the docs explorer and the styling is strongly inspired by how VSCode shows the tabs
- Instead of going for **bold** activated tabs, I suggest to just change the color for now, as the bold (especially for longer texts) makes the tabs jump around a littlebit
- Taking the style of the toolbar buttons for the "+" was a good idea, however in the row of tabs it feels a bit out of place. I turned that into a more minimal design that should fit better into the tabs.
- I adjusted the `.active`, `:active` and `:hover` background colors a bit - they seemed quite strong and I think the goal of showing which tab is active can be achieved with more smooth contrast

One thing I noticed, is that there's quite some delay between typing and the tab name updating. Not a big problem, just noticed that.
So once the subscriptions are tested and cleared, this can be merged from my side!

acao:
@timsuchanek looks like you need to fix a few of the unit tests

timsuchanek:
Thanks @acao, will do so in a bit!

packages/graphiql/src/components/Tabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+export function Tab(props: {
+  isActive: boolean;
+  title: string;
+  isCloseable: boolean;
+  onSelect: () => void;
+  onClose: () => void;
+}): React.ReactElement {
+  return (
+    <div
+      role="button"
```
acao, line 25: I think we need to use `tab` role here:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role

acao: it seems aria `tablist` spec also describes what the keyboard shortcuts should be, so that clears all of that up!

acao (changes requested):
We just need to use `tab`, `tablist` and `tabpanel` roles and other related aria attributes, for proper tab navigation

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role

this aria spec for tabs also determines which keyboard shortcuts we should use

update: switched to an approving review for now so as not to block, but with just a few attributes on a few components this could be made screen reader accessible

acao (approved):

packages/graphiql/src/components/Tabs.tsx
timsuchanek: Good points, I just added the 3!

timsuchanek:
I just fixed the tests and added your a11y recommendations @acao!
The last thing missing is checking out, if subscriptions work when switching tabs @n1ru4l, then from my side, we can ship it.

acao:
@timsuchanek awesome! Re: subscriptions tab, a great feature that should be on 2.x roadmap would be a special view for viewing subscriptions results. I think Altair, Apollo studio and/or Insomnia might have some nice implementations with more than what we do in playground, which is to just present a JSON array of results

n1ru4l:
I will have a look today!

timsuchanek:
That's a great point @acao! I'll also have a look at what they're doing right now.
Thanks @n1ru4l 🙏

n1ru4l:
So I noticed that the subscription implementation in the test schema was broken and also did not clean up timers properly. I fixed it and can verify that a subscription is properly stopped when switching tabs.

timsuchanek:
Thanks for looking into that @n1ru4l. One of the main benefits with tabs and subscriptions for me actually is, that I can start a subscription, switch into another tab, run a mutation and come back. That is something we support in the Playground.
I know, it's more work to implement, but in my view, that would make sense from a user perspective.
What's your view on that?
It's also fine for me to launch with the stopping subscriptions first and introduce it later.

n1ru4l:
@timsuchanek yeah that sounds like something we can add in a follow-up PR! I don't see it as something crucial that blocks shipping this improvement.

acao:
@n1ru4l sure another PR is fine, as long as we wait to release until subscriptions are fixed. The canary release for this PR is already available. There are already enough complaints about how subscriptions currently only show the latest payload, so a regression like this would lead to a lot more bug reports and confusion.

n1ru4l:
@acao I am not sure what you mean by "wait until subscriptions are fixed". Subscription work the same as before? I also implemented tabs as opt-in for now. I don't see anything blocking a release.

acao:
@n1ru4l my bad, I am mistaken, I see that you fixed it. it's very hard to review any of this from a phone!

timsuchanek:
Ok folks, so then let's merge this!

timsuchanek:
Before merging, one last question @acao - do you want to cut a release before we merge this?

timsuchanek (approved):

n1ru4l:
@timsuchanek This repository uses changesets. You commit changesets in your PRs (https://github.com/graphql/graphiql/pull/2197/files#diff-5fba9f77d0ead11d17bfe3bd4449bce1a9879ccc5a5243f359eaa4c0e55aae13).

Once merged to `main` a GitHub action automatically updates a PR for releasing the affected packages (https://github.com/graphql/graphiql/pull/2229)

As a maintainer, you simply have to merge that PR for releasing all packages (it updates changelogs, bumps versions, creates GitHub releases, and publishes to npm).

packages/graphiql/src/utility/fuzzyExtractOperationTitle.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export function fuzzyExtractOperationTitle(str: string): string {
+  const regex = /(query|subscription|mutation) ([a-zA-Z0-9]+)/;
+  const match = regex.exec(str);
+
+  if (match === null) {
+    return '<untitled>';
+  }
+
+  return match[2] as string;
```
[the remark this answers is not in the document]
acao: applied this fix!

packages/graphiql/src/components/Tabs.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
acao, line 1: Going to make a few more accessibility improvements here!

Some excellent prior art from our friends at codesandbox:

https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/components/FileTabs/index.tsx

This implementation in sandpack is a great example of an implementation that appears to be WCAG compliant

acao:
OK @timsuchanek @n1ru4l added some final a11y improvements, using the official wai-aria spec example and sandpack for reference

also made `tabs` prop optional, as only `fetcher` prop needs to be required (for now)

acao:
@timsuchanek i'm terribly sorry, i accidentally overwrote your commits when rebasing the upstream main... I probably missed some things that you had done with a11y and elsewise 😢

timsuchanek:
No problem @acao, can happen. I just re-applied the commits.
By switching to `button`, you introduced unintended side effects btw:
image
I just added a fix for that.

acao:
Thanks @timsuchanek ! I didn’t see that with the deploy preview, which is worriesome! Button element was reccomended by the wai-aria spec

acao:
@timsuchanek I think this last commit should fix the tests 🥳

timsuchanek:
That's awesome! How do you want to go about the release? I'm happy to give it a try to learn the changeset workflow

acao:
@timsuchanek would love for you to do the honors!

1. verify the changeset we added here has everything we want to have in it - for the changelog, it will append to this file 'thanks @timsuchanek!' with a markdown link to your profile, to the commit and the PR
2. once you merge this PR, the Version Packages PR will be updated, and you can see the new changelog diff there
3. then, you can merge the Version Packages PR (see the note about CLA, it's fine despite the Linux CLA bot not liking the changeset bot)
4. then we should probably the new feature on discord & twitter - would be great to automate this part eventually, but there is some nuance, since it's easy for us to fire off ~3-7 releases at once.

acao:
@timsuchanek just rebased, good to go!

acao:
@timsuchanek found another bug i introduced - tab add button is now a child of a button somehow! oops. chrome doesn't like this

acao:
@timsuchanek everything is looking good here! found another regex issue with comments

timsuchanek:
Awesome! Thanks for looking into it @acao! I'll look into it tomorrow ✌️

timsuchanek:
This looks good, so I'm merging this now!

PR #2198, opened 2022-02-22
Feature/#1980 examples vue vite monaco

### LINKED ISSUE #1980: Vite example for `monaco-graphql`
Let's add a vite example for `monaco-graphql`? maybe even use vue or svelte instead of react?

https://github.com/vdesjs/vite-plugin-monaco-editor

hinogi:
fixes #1980


acao:
I'm so confused, this example is already here? Why is this PR showing all these as new files when they already exist??

acao (changes requested):
@cshaver and I had no problem building this example last night.

also you are introducing a yarn.lock file to a yarn workspace, which may be why you had issues

I think the issue is the readme steps. I will try a clean checkout later

PR #2202, opened 2022-02-23
fix(examples): :bug: Fixed path to css file in index.html

hinogi:


acao:
We tested the example last night and this worked fine? not sure why it's needed

hinogi:
In firefox the css was not found

PR #2207, opened 2022-02-24
Fix legacy client fetcher example code

dushaniw:
This PR incorrect subscription client object initialization.


acao:
thank you @dushaniw !

PR #2209, opened 2022-02-24
merge interface, parser, types, utils into single lsp service pkg

acao:
## Motivation
This greatly simplifies the monorepo, reduces tooling & cognitive overhead, and simplifies dependency resolution as these core libraries are used everywhere. It will make upgrading and peer dependency resolution issues much easier to manage.

It gives us parity with other LSP ecosystems who use a similar pattern, making it easier for LSP-familiar developers to know where to start.

## Plan

This is the first of several major breaking changes to the LSP! There will be many improvements following this. Enjoy :)

## Todo
- [x] builds
- [x] tests pass
~- [ ] end/clean up the seperate standalone types def file for once and for all! (seperate PR)~
- [x] move around documentation
- [x] decide on legacy path & versioning


acao:
@dotansimha what are the best ways to retire these widely used packages?

I was thinking one last release that tells them to migrate on `postinstall` ?

orta:
npm deprecate with a recommendation of the new library might be enough assuming they are API compatible?

acao:
Oh brilliant! Thank you. I didn't know about this new command https://docs.npmjs.com/cli/v8/commands/npm-deprecate

acao:
Ok @orta this is ready to go FYI, if/when you get a chance to take a look!

acao:
the next step is to "flatten" the interface method parameters using object arguments for a major version release

this will allow us to provide more flexible options, retire parameters, and otherwise more fluid migrations than the current ordered parameters allow

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -24,29 +24,37 @@ import {
...
+  getRange,
+  DIAGNOSTIC_SEVERITY,
+} from 'graphql-language-service/src/interface/getDiagnostics';
 import {
   getDefinitionQueryResultForFragmentSpread,
   getDefinitionQueryResultForDefinitionNode,
   getDefinitionQueryResultForNamedType,
   DefinitionQueryResult,
-} from './getDefinition';
+} from 'graphql-language-service/src/interface/getDefinition';
```
orta, line 53: reaching into the src dir for another module feels quite strange, wouldn't these be dist files?

packages/graphql-language-service/README.md
```diff
@@ -34,3 +34,19 @@ const languageService = new LanguageService({
   schemaConfig: { uri: 'https://my/schema' },
 });
 ```
+
+## Interface
+
+LSP Language Service written in Typescript used by [GraphQL Language Service Server](https://github.com/graphql/graphiql/tree/main/packages/graphql-language-service-server) and [Codemirror GraphQL](https://github.com/graphql/graphiql/tree/main/packages/codemirror-graphql).
```
orta, line 40:
```suggestion
LSP Language Service written in TypeScript used by [GraphQL Language Service Server](https://github.com/graphql/graphiql/tree/main/packages/graphql-language-service-server) and [Codemirror GraphQL](https://github.com/graphql/graphiql/tree/main/packages/codemirror-graphql).
```

packages/graphql-language-service/package.json
```diff
@@ -31,13 +31,22 @@
...
   },
   "dependencies": {
-    "graphql-language-service-interface": "^2.10.2",
-    "graphql-language-service-parser": "^1.10.4",
-    "graphql-language-service-types": "^1.8.7",
-    "graphql-language-service-utils": "^2.7.1"
+    "graphql-config": "^4.1.0",
+    "vscode-languageserver-types": "^3.15.1",
+    "nullthrows": "^1.0.0",
+    "@types/json-schema": "7.0.9"
```
orta, line 37: This dep should probably be in devDeps

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
acao: Ah good catch! Spent so much time cleaning up the autorename bugs

packages/graphql-language-service/package.json
acao: It’s used for user facing interfaces, so users have complained about this in the past with other types. What do you think?

orta: I personally think its not a great call

acao: Or I could just leave this and others as `unknown` and link to the related API doc. I can't imagine it's that useful to have the JSONSchema definition result?

acao:
technically ready, but no rush! that was really fast today, ty so much!!!!!

imolorhe:
This is a nice change!

packages/graphql-language-service/package.json
imolorhe: If the type is exported, wouldn't that automatically make the type available?

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -24,29 +24,34 @@ import {
   Outline,
   OutlineTree,
   GraphQLCache,
-} from 'graphql-language-service-types';
+} from 'graphql-language-service';
```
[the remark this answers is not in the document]
acao: Ty, great catch, this was an adendum

jest.config.base.js
```diff
@@ -20,7 +20,7 @@ module.exports = (dir, env = 'dom') => {
       '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
         'identity-obj-proxy',
       '\\.(css|less)$': 'identity-obj-proxy',
-      '^graphql-language-([^/]+)': `${__dirname}/packages/graphql-language-$1/src`,
+      // '^graphql-language-([^/]+)': `${__dirname}/packages/graphql-language-$1/src`,
```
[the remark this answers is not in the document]
acao: Huh, this should be resolved already today. I re-added the line so tests would pass. Must be cacheing?

packages/graphql-language-service/package.json
acao: the type comes from a third party library's type system, which means that that third party's `@types` library needs to be included in `dependencies` for the type to be available

acao: made the reccomended change and updated the docs, should be all good now!

acao:
@orta @imolorhe I will be opening the follow up PRs for this PR, based on this branch, so let me know what else you think this needs!

please, no one release `graphql-language-service` until all breaking changes are complete

Issue #2211, opened 2022-02-26
Remove IE7 CSS star property hack


hatappi:
The CSS of graphiql contains star property hack for IE7. 
ref: https://stackoverflow.com/questions/14927651/what-does-zoom1-do-in-bootstrap/14927670#14927670

- https://github.com/graphql/graphiql/blob/d3e00052cdfbca4912713e9287aa7d636d947c2f/packages/graphiql/src/css/codemirror.css#L280-L282
- https://github.com/graphql/graphiql/blob/d3e00052cdfbca4912713e9287aa7d636d947c2f/packages/graphiql/src/css/codemirror.css#L417-L420

Can we remove these properties?

## WHY
My project uses [`esbuild`](https://esbuild.github.io/). When I build my codes that includes the CSS of graphiql, the following warning occurs.

```
➜ cat node_modules/graphiql/graphiql.css | npx esbuild --loader=css --minify > /dev/null
▲ [WARNING] Expected identifier but found "*"

    <stdin>:844:2:
      844 │   *zoom:1;
          ╵   ^

▲ [WARNING] Expected identifier but found "*"

    <stdin>:845:2:
      845 │   *display:inline;
          ╵   ^

▲ [WARNING] Expected identifier but found "*"

    <stdin>:972:19:
      972 │ .CodeMirror span { *vertical-align: text-bottom; }
          ╵                    ^

3 warnings
```

I think these warnings make sense because this hack is invalid in current browsers.

And I checked `.browserslistrc` of graphiql.
https://github.com/graphql/graphiql/blob/d3e00052cdfbca4912713e9287aa7d636d947c2f/.browserslistrc
It doesn't include IE5.5 ~ IE7

That's why I opened this issue.

acao:
Yes let’s! We do not support IE anymore. Feel free to open a PR

hatappi:
@acao Thank you for replying. I opened a PR. If you have chance, could you please review that?
https://github.com/graphql/graphiql/pull/2213

acao:
@hatappi this fix should be deployed btw!

PR #2213, opened 2022-02-27
style: remove IE7 CSS star property hack

hatappi:
## WHAT
Remove the [star property hack](https://stackoverflow.com/a/14927670/808699) that was for IE 5.5 to 7 from CSS of graphiql.

## WHY
solve https://github.com/graphql/graphiql/issues/2211


acao:
Honestly embarassing we still had IE7 workarounds 😆

PR #2214, opened 2022-02-27
fix: fix Windows fileUri to resolve type definition

### LINKED ISSUE #2212: Unable to resolve resource on Windows
Original fix #1715
Regression from #2072
Related https://github.com/graphql/vscode-graphql/issues/254

image

After some digging, I've confirmed that the bad path is indeed present in the GraphQLCache for the type.
Supposedly the uri is wrong when caching the schema (I wasn't able to debug at that stage because I'm attaching too late to the process)
https://github.com/graphql/graphiql/blob/484c0523cdd529f9e261d61a38616b6745075c7f/packages/graphql-language-service-server/src/MessageProcessor.ts#L827-L845

I suspect some `vscode-uri.URI.parse(uri).toString()` should use `vscode-uri.URI.parse(uri).fsPath` instead? but my initial attempt failed miserably and I'm a little out of my comfort zone at this point

Cellule:
Fixes #2212
Use `URI.file` instead of `URI.parse` to convert fs path to fileUri 

I found in some places there's the variable `filePath: Uri` which is a bit confusing because type says it's an Uri, but variable name says it's a path.

### DIFF
--- a/.changeset/rare-roses-give.md
+++ b/.changeset/rare-roses-give.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-server': patch
+---
+
+Fixed Windows fileUri when resolving type definition location
--- a/packages/graphql-language-service-server/src/GraphQLCache.ts
+++ b/packages/graphql-language-service-server/src/GraphQLCache.ts
@@ -363,7 +363,7 @@ export class GraphQLCache implements GraphQLCacheInterface {
               // the docs indicate that is what's there :shrug:
               const cacheEntry = globResult.statCache[filePath] as fs.Stats;
               return {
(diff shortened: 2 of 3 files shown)


acao:
Brilliant, thank you! This will be a welcome bugfix for windows users

can you add a changeset for the server as per the comment? The CLI will automatically patch increment

Cellule:
@acao I hope this does the trick :)
Thank for the quick replies, really motivated me to push it to the finish line

acao:
Looking great! i just have one tweak to add, can you give me permissions to write to your branch?

acao:
I will add in another commit

PR #2217, opened 2022-03-01
Move to static imports for all commonJS requires 

kooparse:
WIP...


acao:
@kooparse I appreciate the initiative, but unfortunately this breaks SSR :( dynamic imports will work out much better, but most of our users don't use bundling tools, just the cdn bundle, so we need to make the dynamic imports work in a lot of places

PR #2218, opened 2022-03-03
Cleaning up README based on GraphQL project voice rules

leebyron:
Based on discussion and guidelines that we've had up for discussion for a very long time in https://github.com/graphql/graphql-wg/pull/696, I'm cleaning up this README file.

Please don't take this PR as a value statement on this particular message. The process I'm working through is to first get us to a clean state, after which we can ensure any future messages atop that clean slate adhere to that policy.

Please see our TSC document with a section describing this:

https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#speaking-on-behalf-of-the-graphql-project

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 <!-- @format -->
 
-> # Black Lives Matter 🖤
-
 # GraphQL IDE Monorepo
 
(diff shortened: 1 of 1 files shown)


acao:
Thanks @leebyron, this makes sense! Will submit for Ukraine & Black Lives Matter again!

PR #2220, opened 2022-03-07
fix: Trigger editor update when externalFragments change

dwwoelfel:
Updates codemirror when the `externalFragments` prop changes.

Without this change, GraphiQL must be remounted for any changes to `externalFragments` to take effect.


acao:
thanks for this!

acao:
Oops. No changeset so it won’t be in the changelog :/ you can add a PR to add a changeset and then link to this PR and issue in the changeset description?

PR #2228, opened 2022-03-09
fix: typo in GraphQLCache.ts

eltociear:
becase -> because


acao:
@timsuchanek there is already a PR to add cspell where this and may typos are fixed!

PR #2239, opened 2022-03-14
Initial implementation of vscode-graphql inside the graphiql repo

orta:
This is not a 1 to 1 copy of the current vscode extension (I only ported over stuff I used, which does not include the ability to make calls) nor does the PR handle deployment but it does work


orta:
I got around to updating the status bar to the new vscode API, where we can use the real icon instead of the letters 'graphql' - hovering will give you more information now.

Screenshot from 2022-03-14 11 46 31

orta:
I'll get this green, and I think the tooling should be set up to deploy this to the nightly channel on the vscode extension - then we can either wait for someone who cares about the execute query to re-write those sections of the codebase.

orta:
Sure, folks are welcome to add it 👍🏻 - in the meanwhile, here's an vsix for people who want to use this build.

[vscode-graphql-0.3.52.vsix.zip](https://github.com/graphql/graphiql/files/8244721/vscode-graphql-0.3.52.vsix.zip)

acao:
@orta what do you think of these changes? The best I could manage for a watcher-enabled task, you still need to manually re-load on changes for it to take effect, but it at least builds changes to the entire typescript project references tree!

packages/vscode-graphql/src/extension.ts
```diff
@@ -112,18 +117,19 @@ export function activate(context: ExtensionContext) {
...
+    clientLSPDisposable = client.start();
     context.subscriptions.push(clientLSPDisposable);
 
     outputChannel.appendLine(`GraphQL LSP restarted`);
   });
 }
 
 export function deactivate() {
-  console.log('Extension "vscode-graphql" is now de-active!');
+  console.log('Extension "vscode-graphql" has been de-activated!!');
```
orta, line 132: hah de-active

orta:
Looks good 👍🏻

acao:
one last TODO for me or anyone who has time, just because I know it won't happen in a follow up commit because i always forget to do this, but we should update `CONTRIBUTING.md`, and maybe tweak the readme some along these lines as well.

the `vscode-graphql` readme should ideally not contain duplicate information from what's documented in `graphql-language-service-server`. There is a superset of need-to-know-info in `vscode-graphql` that doesn't apply to the LSP server but it's limited. Often `vscode-graphql` users get stuck on things that are documented in the LSP server readme.

This is something a proper documentation site could handle, but with some README tricks we can improve this part of the DX

acao:
adding the @graphql/tsc, if any members would like input on this decision. the goal here is to complete the reference implementation ecosystem - where we have reference implementations for both web and LSP manifestations of our language services, designed to always support the latest language features

acao:
if anyone wants to borrow the changes to the publishing pipeline from here they can:

https://github.com/graphql/graphiql/pull/2187

acao:
closed the #2187 PR where we tried to merge it with the full git history, in favor of this rewrite with reduced functionality.

PR #2241, opened 2022-03-17
WG cleanup & Add 2022 WG agendas

timsuchanek:
Inspired by the https://github.com/graphql/graphql-wg, I moved the WG documents into folders by year and already created WG agendas for all months.
I suggest adopting the WG protocol from the main GraphQL WG (including time allocation, lead), as that seems to work well there.


acao:
thank you @timsuchanek !

PR #2249, opened 2022-03-25
use dynamic imports, remove usage of `require()`

### LINKED ISSUE #1606: GraphiQL depends on require()

acao:
resolves #1606 by replacing usage of legacy inline `require()` pattern with dynamic import, which is now available in all browsers we target

among other benefits, this allows tools like `esbuild` and `vite` to have a better time with this library


acao:
@timsuchanek merging this one as well, as part of 1.2!

acao:
@timsuchanek lets' merge this one before release as well!

packages/graphiql/src/components/HeaderEditor.tsx
```diff
@@ -59,21 +61,68 @@ export class HeaderEditor extends React.Component<HeaderEditorProps> {
...
+        }}
+        ref={node => {
+          this._node = node as HTMLDivElement;
+        }}
+      />
+    );
+  }
+
+  // @ts-expect-error
+  addonModules = () => [import('codemirror/mode/javascript/javascript')];
```
timsuchanek, line 122: What would the TS error be here?

timsuchanek (approved):
Looks good to me!

packages/graphiql/src/components/HeaderEditor.tsx
acao: that there isn't a type definition for this for `@types/codemirror`, just a legacy codemirror 5 issue

timsuchanek: Ok, thanks for the context! So the afterworld can find this context in this PR

acao: I could just properly override them in a .d.ts file I guess, but this was easier?

acao:
thanks @timsuchanek ! now I'm going to add some esbuild/vite examples for graphql!

PR #2252, opened 2022-03-27
fix operation name regex

acao:


timsuchanek:
The cases without a prefix ` ` space didn't work with the previous regex. This fixes that and adds those as test cases.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -2351,7 +2359,7 @@ function tabsStateEditQueryReducer(value: string, state: TabsState): TabsState {
       }
       return {
         ...tab,
-        title: fuzzyExtractOperationTitle(value),
+        title: operationName ?? fuzzyExtractOperationTitle(value),
```
acao, line 2362: nice find!

Issue #2263, opened 2022-03-30
CodeMirror.on is not a function trying to use tab


eMerzh:
hello, i've just update my project, to the lastest version thus trying to use the new tab feature ✨ ..
i've added the tab props and reloaded but once i type in the editor i have:

> TypeError CodeMirror.on is not a function
> Call Stack
>  onHasCompletion/<
>   node_modules/graphiql/esm/utility/onHasCompletion.js:18:20

arround here:

> 
> export default function onHasCompletion(_cm, data, onHintInformationRender) {
>     import('codemirror').then(function (CodeMirror) {
>         var information;
>         var deprecation;
>         CodeMirror.on(data, 'select', function (ctx, el) {
> 

<img width="570" alt="image" src="https://user-images.githubusercontent.com/177003/160917567-522bbd3e-94bf-43eb-9f29-344956a30742.png">

i've clear up a bunch of customisation to finally have a minimal producing , and it looks like this  : 

```jsx
import ReactDOM from 'react-dom'
import GraphiQL from 'graphiql'
import 'graphiql/graphiql.min.css'
import './styles.css'

function graphQLFetcher(graphQLParams) {
	return fetch('/graphql', {
		method: 'post',
		headers: { 'Content-Type': 'application/json' },
		body: JSON.stringify(graphQLParams),
	}).then(response => response.json())
}

const App = ({ fetcher }) => <GraphiQL style={{ height: '100vh' }} headerEditorEnabled tabs fetcher={fetcher} />
ReactDOM.render(<App fetcher={graphQLFetcher} />, document.getElementById('root'))

```

is there something i can do to fix it? 

note: i'm using webpack

eMerzh:
not sure it's related to tab after all 🤔 
but it triggers on complete if it wasn't clear already

acao:
It’s probably related to a recent change where we switched to dynamic imports instead of inline require for importing codemirror (for SSR support). I will try to re-create the bug soon

eMerzh:
ok thanks @acao  🙌 

it seems to be related to this line https://github.com/graphql/graphiql/blob/main/packages/graphiql/src/utility/onHasCompletion.ts#L29  so yeah import might be the issue 🤔

acao:
Fixed in 1.8.3!

PR #2265, opened 2022-03-31
fix `codemirror` import for `onHasCompletion`

acao:
Not seeing this in the umd build for whatever reason
https://graphiql-test.netlify.app

however #2263 seems to be occuring for some webpack users, and I suspect this might be a fix?

### DIFF
--- a/packages/graphiql/src/utility/onHasCompletion.ts
+++ b/packages/graphiql/src/utility/onHasCompletion.ts
@@ -26,7 +26,7 @@ export default function onHasCompletion(
   data: CM.EditorChange | undefined,
   onHintInformationRender: (el: HTMLDivElement) => void,
 ) {
-  import('codemirror').then(CodeMirror => {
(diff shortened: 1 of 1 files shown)


eMerzh:
as far as i can test this , is seems to work 🤔

acao:
@eMerzh the netlify build is using umd (for cdn bundle users), and that was already working. let me add a changeset here so that it publishes a canary release for you to try with your specific webpack config

eMerzh:
That would be awesome 😎
But I tried using the debugger in my setup and the fix seems to Work... But ofc it would be clearer with a canary

acao:
@eMerzh oh fantastic, that’s great news! Thanks for testing to that extent! I think we can call this fixed then, if both the umd and esm build are working with this change now!

acao:
Thanks @eMerzh, and thanks for reporting this!

PR #2273, opened 2022-04-04
enable `react/jsx-curly-brace-presence` rule and remove `react/jsx-no-literals` rule

B2o5T:
It's anti-practice in React as it's unnecessary JSX expressions and provokes confusing

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,10 +34,10 @@ module.exports = {
   },
 
   extends: [
-    'prettier',
     'plugin:import/typescript',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    'prettier',
   ],
 
   globals: {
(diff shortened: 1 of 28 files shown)


.eslintrc.js
```diff
@@ -34,10 +34,10 @@ module.exports = {
   },
 
   extends: [
-    'prettier',
     'plugin:import/typescript',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    'prettier',
```
B2o5T, line 40: ESLint applies configs cascading, and `eslint-config-prettier` **always** must be last!

acao:
hmm, i think we should leave these for now to make i18n easier

B2o5T:
@acao migration to i18n can be easy with [`@shopify/jsx-no-hardcoded-content`](https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/jsx-no-hardcoded-content.md) eslint rule

thomasheyenbrock (approved):
I like this change! Imho it's the more "idiomatic" way of writing JSX. Regarding i18n, this isn't immediate priority right now so we can deal with this later. Also, the linter rule @B2o5T linked should help us with that 👍

PR #2274, opened 2022-04-04
turn `valid-typeof` as `error`

B2o5T:
`typeof window !== null` **always** returns `false`


acao (approved):
looks good!

acao:
can you add a changeset?

PR #2312, opened 2022-04-12
fix: A query request was sent for subscriptions after tab change

### LINKED ISSUE #2311: GraphiQL sends the wrong request after changing tab
This is a UI bug report. I made the following changes to the CDN sample:
```
    ReactDOM.render(
      React.createElement(GraphiQL, {
        tabs: true,
        fetcher: window.GraphiQL.createFetcher({
          url: "http://localhost:3000/graphql",
          subscriptionUrl: "ws://localhost:3000/graphql"
        }),
        defaultVariableEditorOpen: true,
      }),
      document.getElementById('graphiql'),
    );
```
Then entered this subscription in one tab:
```
subscription S{
  blockAdded(replayFromBlockLevel:2275280) {
    hash
    header {
      level
    }
  }
} 
```
And it works fine. Then I add this query in a new tab:
```
query Q {
  blocks(first: 1) {
    edges {
      node {
        hash
        level
      }
    }
  }
} 
```
And it works fine. Now if I go back to the first tab and run it, I get the following error:
```
{
  "errors": [
    {
      "message": "An internal server error occurred. Please provide error ID '5aff4db5-4303-4475-9f19-99dd288cebd8' if you contact our help desk. (debug-only): requestId {requestId} and {error} Cannot return null for non-nullable field Subscription.blockAdded.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "blockAdded"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot return null for non-nullable field Subscription.blockAdded.",
            "    at completeValue (/home/alireza/work/ECAD/tezgraph/node_modules/graphql/execution/execute.js:559:13)",
            "    at /home/alireza/work/ECAD/tezgraph/node_modules/graphql/execution/execute.js:469:16",
            "    at runMicrotasks (<anonymous>)",
            "    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
            "    at async Promise.all (index 0)",
            "    at async execute (/home/alireza/work/ECAD/tezgraph/node_modules/apollo-server-core/src/requestPipeline.ts:497:14)",
            "    at async processGraphQLRequest (/home/alireza/work/ECAD/tezgraph/node_modules/apollo-server-core/src/requestPipeline.ts:403:22)",
            "    at async processHTTPRequest (/home/alireza/work/ECAD/tezgraph/node_modules/apollo-server-core/src/runHttpQuery.ts:346:24)"
          ]
        }
      }
    }
  ],
  "data": null
}
```
Looking at the network tab of my browser, I realize that GraphiQL is sending a POST request to the Queries url of my service, not the right subscription request. If I make any change in the subscription tab (like adding a single space), this gets fixed. If I go back to the tab with Query in it and come back, it still works fine. If I make an edit in the tab with Query and come back to the tab with Subscription, I get the error again.

So, in short, it appears that some flag is set in GraphiQL that marks the current command as Query or Subscription. It gets updated when changing the text, some times when changing tabs, but not all times.

AlirezaHaghshenas:
fixes #2311

### DIFF
--- a/.changeset/light-rocks-smell.md
+++ b/.changeset/light-rocks-smell.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Fix: After changing to a tab with a subscription, graphiql sends a query request
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -778,7 +778,12 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     this.handleStopQuery();
     this.setState(
       state => stateOnSelectTabReducer(index, state),
-      this.persistTabsState,
+      () => {
+        this.persistTabsState();
+        if (this.state.query) {
+          this.handleEditQuery(this.state.query);
+        }
+      },
     );
   };
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -7439,16 +7439,6 @@ azure-devops-node-api@^11.0.1:
     tunnel "0.0.6"
     typed-rest-client "^1.8.4"
 
-azure-devops-node-api@^7.2.0:
-  version "7.2.0"
-  resolved "https://registry.yarnpkg.com/azure-devops-node-api/-/azure-devops-node-api-7.2.0.tgz#131d4e01cf12ebc6e45569b5e0c5c249e4114d6d"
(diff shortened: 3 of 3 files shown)


acao:
Thank you! Can you add a changeset as per the comment?

AlirezaHaghshenas:
@acao Done

acao:
Thanks! I'll just assume these lockfile changes are necessary? I don't know what's up with that lately

PR #2316, opened 2022-04-14
Fix wrong request type after restoring a subscription

### LINKED ISSUE #2314: With tabs enabled, after graphiql is reloaded and a subscription is restored from storage, a query request is sent
Similar to #2311,
The conditions for this one is: 

- Tabs are enabled
- In the only tab as subscription is stored (probably also if there are multiple tabs and this is the first tab)
- The browser page is refreshed
- Subscription is restored from storage
- Execute query is pressed

Now a query request will be sent.

AlirezaHaghshenas:
closes #2314

### DIFF
--- a/.changeset/old-laws-compare.md
+++ b/.changeset/old-laws-compare.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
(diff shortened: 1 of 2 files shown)


acao:
Thank you @AlirezaHaghshenas !

acao (approved):

PR #2318, opened 2022-04-15
Fix typo in validateWithCustomRules.ts

eltociear:
determing -> determining


timsuchanek:
Merging this without changeset, as it's just a change in a code comment.

Issue #2185, opened 2022-02-20
Move to yarn 3 + pnp, pnpm or latest npm?


acao:
Whatever we decide on is great, just the install time is atrocious for some users, many have complained, and I don't think vscode-graphql is about to help with that 😬

rxliuli:
pnpm is the best known, especially the --filter option currently has no replacement support, I use it in production (about 100+ modules)

acao:
@orta agreed, ideally trimming dependencies is the best way to speed things up

@rxliuli that would be cool, would love to see a PR if someone has time! I imagine we would need to add pnp support libraries for jest, cypress, webpack, vite, etc?

rxliuli:
I can try to create a PR with yarn2 => pnpm if you guys agree. In addition, pnpm does not need to use yarn pnp, it uses soft links to achieve more efficient disk sharing and performance.

ref: https://pnpm.io/zh/motivation

rxliuli:
Some notes when I migrate pnpm inside my team @acao 

## Why use pnpm

- performance: it's the fastest package manager out there, yarn the catfish went in a completely different direction after 2 (platforming, including advanced features like the plugin system)
- monorepo support: filter is by far the best filter in the package manager, probably not one of them (lerna/yarn2/nx don't work as well)
- disk utilization: due to the use of hard links, it is like maven the same package only occupies a space, if you have 100 projects, node_modules will not be copied 100 copies
- Avoiding phantom dependencies: solves the common problem of using dependencies without declaring them, e.g. two modules depend on different versions of lodash, and there is no certainty which version of lodash is used, which can cause very strange errors

![Performance comparison](https://filescdn.proginn.com/8c3365999c38d10430fcd42b9b88e6f7/0acb0915244bea1275009b73563db28a.webp)
![Avoid phantom dependencies](https://pnpm.io/zh/assets/images/node-modules-structure-8ab301ddaed3b7530858b233f5b3be57.jpg)

Actual performance comparison

| comparison | pnpm | yarn3 |
| ---------- | --------- | --------- |
| no cache | 1m49.128s | 7m15.828s |
| with cache | 0m37.468s | 1m20.278s |
| Installing dependencies | 0m2.754s | 0m13.080s |
| not updated | 0m4.034s | 0m12.188s |
| initialize | 0m26.013s | 0m35.381s |

pnpm migration guide

```sh
rm package-lock.json
npm i -g pnpm
cd matrix/
pnpm i/rm
pnpm dev/build
[...]

PR #2323, opened 2022-04-19
yarn => pnpm

rxliuli:
ref: https://github.com/graphql/graphiql/issues/2185


acao:
@rxliuli ah, the email used for CLA has to be the same email that’s used in each commit

acao:
It looks like we need to use a base image that already has pnpm ideally, or we can manually install pnpm for each workflow

rxliuli:
> It looks like we need to use a base image that already has pnpm ideally, or we can use an action to manually install pnpm for each workflow

Just run `npm i -g pnpm`

acao:
@rxliuli don‘t tell me, tell the github actions 😂

acao (changes requested):
please update the github actions to install pnpm, preferrably in the base docker image, so that CI passes. As you can see, all 18 checks are failing

rxliuli:
Do you want to change the command to install dependencies on netlify? I have observed that the dependencies are still installed using npm and then fail

https://app.netlify.com/sites/graphiql-test/deploys/625f4fd9e4fab7000994463a

acao:
@rxliuli weird, not sure why i had to re-approve them!

rxliuli:
Build & Test I know what's wrong, it's fixed. But why does lint fail, it seems strange, do I need to fix these lint and ts errors? Or some configuration was accidentally changed. . .

https://github.com/graphql/graphiql/runs/6087316492?check_suite_focus=true

acao:
@rxliuli something is off with lint staged, so for now you need to run pnpm format.

acao:
Hmm, the typescript version must be resolving differently, either way you can use the ts-expect-error comment if you want since they are mostly just handling errors

rxliuli:
I restored the version of the dependencies from the previous yarn.lock, including typescript, sorry, I didn't notice the change of the dependency version before

rxliuli:
webpack4 is not compatible with pnpm, it's a bit tricky, have you guys done webpack4 => webpack5? Or we can use rollup?
https://github.com/webpack/webpack/issues/5087

timsuchanek:
I think @acao was thinking of Vite, which I'd be totally ok with.
From my side you can go ahead and move to Vite.
Curious what @acao thinks

acao:
Our plan was to migrate to vite instead, and drop babel and webpack altogether. But we may have to upgrade to webpack 5 as a transitional step. It will have to wait as I’m sick with covid and my co maintainers are focused on other things at the moment

acao:
With vite the goal was to get rid of the duplicate typescript project references tree and use vite to bundle each library for it‘s relevant exports, but I’m having trouble finding something that can re-create tsc —build —watch on a monorepo scale. @nx might have something for this, I can’t seem to find any vite and/or rollup plugins that solve this problem

rxliuli:
> This was the blocker we had run into before with pnpm, thus why i raised the issue of webpack support early on! Our plan was to migrate to vite instead, and drop babel and webpack altogether. But we may have to upgrade to webpack 5 as a transitional step. It will have to wait as I’m sick with covid and my co maintainers are focused on other things at the moment

Is there any suggestion, currently only one **Test PR / Cypress E2E Suite (pull_request) ** can not pass, I don't use webpack very much so don't know much about the configuration here
packages/graphiql/resources/webpack.config.js

rxliuli:
> With vite the goal was to get rid of the duplicate typescript project references tree and use vite to bundle each library for it‘s relevant exports, but I’m having trouble finding something that can re-create tsc —build —watch on a monorepo scale. `@nx` might have something for this, I can’t seem to find any vite and/or rollup plugins that solve this problem

In production, we wrote a plugin to handle this, pointing the module to src/index.ts so that we don't have to rebuild dist/index.js every time the lib is modified

ref: https://github.com/rxliuli/liuli-tools/blob/master/libs/rollup-plugin-ts-alias/README.md

acao:
Like I mentioned in #2185, you will need to find a compatibility mode or workaround for webpack 4, jest, cypress, etc, which has been what kept us from switching to pnpm im previous attempts.


Otherwise a webpack 5 upgrade will take some sizeable effort and will definitely need to come in a separate PR, since it heavily impacts the actual bundle we ship to the CDNs, which is used more than even the esm module!

acao:
@rxliuli that plugin looks very promising thanks! Apologies there seems to be a delay with my LTE connection, each of your comments is only visible once I make a comment? So I’m just seeing your reply now

acao:
I think we can do this best in two or more PRs:

1. swap out webpack for vite - we don’t need to worry about bundling every library yet. Leave duplicate project references for now. Merge.
2. then, rebase this PR for pnpm with that upstream change
3. Then, worry about replacing the duplicate references for a modern bundling and even `exports` strategy in an additional PR

timsuchanek:
Thanks for providing this context @acao ! That plan makes total sense to me.
@rxliuli do you have capacity to work on the steps Rikki outlined?

rxliuli:
> Thanks for providing this context @acao ! That plan makes total sense to me. @rxliuli do you have capacity to work on the steps Rikki outlined?

I will try it tomorrow

acao:
closing for #3054

PR #2369, opened 2022-04-21
fix: move @codemirror packages to peer deps, upgrade to 0.20.0

sergeichestakov:
It looks like when [CodeMirror 6 support](https://github.com/graphql/graphiql/issues/2022) was [first added](https://github.com/graphql/graphiql/pull/2035) to the `codemirror-graphql` package, `@codemirror/stream-parser` was mistakenly added as a regular dependency rather than a peer dependency (unlike the CM5 `codemirror` package).

This may lead to conflicts when upgrading upstream CM6 packages in a project that depends on this package (especially when upgrading major/minor versions) since you end up with multiple versions of core packages like `@codemirror/view` which leads to bugs. This is especially relevant as there is now a new release for each of the core CodeMirror packages ([`0.20.0`](https://discuss.codemirror.net/t/release-0-20-0/4302)).

This PR moves `@codemirror/` packages to `peerDependencies` and upgrades the CM6 packages to the latest version in the new release. Note that I remove `@codemirror/stream-parser` in favor of `@codemirror/language` as per the release notes linked above.

### DIFF
--- a/.changeset/sixty-squids-peel.md
+++ b/.changeset/sixty-squids-peel.md
@@ -1,0 +1,5 @@
+---
+'codemirror-graphql': minor
+---
+
+Moved @codemirror/language to peer dependencies and upgraded to 0.20.0
--- a/examples/cm6-graphql-parcel/package.json
+++ b/examples/cm6-graphql-parcel/package.json
@@ -22,8 +22,8 @@
     ]
   },
   "dependencies": {
-    "@codemirror/basic-setup": "^0.19.0",
-    "@codemirror/stream-parser": "^0.19.2",
+    "@codemirror/basic-setup": "^0.20.0",
+    "@codemirror/language": "^0.20.0",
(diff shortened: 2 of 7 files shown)


acao:
@sergeichestakov thanks for this! I was wondering what else we needed for cm6. Perhaps we should upgrade the `codemirror-graphql` peer range for codemirror as well?

perhaps on a follow up PR, we could even upgrade graphiql to use cm6 in legacy mode over a minor version release? This would probably drastically improve accessibility without users needing to wait for monaco which may not come til 2.0

sergeichestakov:
good point! upgraded `codemirror` and `@types/codemirror` to the latest versions too in the latest commit

imolorhe (approved):

acao:
looks great! I think it's ready to merge. the deploy preview looks almost exactly the same!

PR #2370, opened 2022-04-21
feat: add `@graphiql/react` package

thomasheyenbrock:
This PR adds a new package `@graphiql/react` to the monorepo. The idea for this package is to provide all the building blocks (context providers, hooks to work with the state managed by the context, reusable components, etc.) for building a GraphiQL-like UI. The main `graphiql` package will become the reference implementation for how to use the components from `@graphiql/react`.

As a first step towards this goal, we add the following to the new package:
- A React context `ExplorerContext` that shall contain all the state and logic related to the docs/explorer. We wrap with in a component `ExplorerContextProvider` that has to be rendered at the top of the tree that contains other GraphiQL-components that need to access or change this state.
- For easy access in functional components we provide a hook `useExplorerNavStack`.
- We also export the context itself, so that we can use consumer components to access the state in legacy class components.

Notes:
- I needed to update the `eslint-plugin-react` version, otherwise using the `plugin:react/jsx-runtime` did not work.
- I fixed a minor type issue where `GraphQLType` was used instead of the `SchemaReference` type from `codemirror-graphql`.

### DIFF
--- a/packages/graphiql/src/components/DocExplorer.tsx
+++ b/packages/graphiql/src/components/DocExplorer.tsx
@@ -50,7 +50,7 @@ export const DocExplorer = memo(
       push({ name: field.name, def: field });
     }
 
-    let content;
+    let content: ReactNode;
     if (schemaErrors) {
       content = <div className="error-container">Error fetching schema</div>;
     } else if (schema === undefined) {
--- a/packages/graphiql/src/components/DocExplorer.tsx
+++ b/packages/graphiql/src/components/DocExplorer.tsx
@@ -5,9 +5,9 @@
  *  LICENSE file in the root directory of this source tree.
  */
 
-import React, { ReactNode } from 'react';
+import React, { memo, ReactNode } from 'react';
 import { GraphQLSchema, isType, GraphQLNamedType, GraphQLError } from 'graphql';
-import { FieldType } from './DocExplorer/types';
+import { ExplorerFieldDef, useExplorerNavStack } from '@graphiql/react';
 
 import FieldDoc from './DocExplorer/FieldDoc';
 import SchemaDoc from './DocExplorer/SchemaDoc';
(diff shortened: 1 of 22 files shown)


timsuchanek:
Awesome! I'll have a look in a bit!

Regarding eslint, we should get https://github.com/graphql/graphiql/pulls/B2o5T merged, they might even resolve some issues?

packages/graphiql/src/components/DocExplorer.tsx
```diff
@@ -52,32 +37,18 @@ type DocExplorerState = {
...
 
-  render() {
-    const { schema, schemaErrors } = this.props;
-    const navStack = this.state.navStack;
-    const navItem = navStack[navStack.length - 1];
+    function handleClickField(field: ExplorerFieldDef) {
+      push({ name: field.name, def: field });
+    }
 
     let content;
```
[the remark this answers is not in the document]
thomasheyenbrock: Good shout!

acao:
this is a great start for `@graphiql/react`!  what do you think of moving the DocExplorer components to `@graphiql/react` workspace as well?

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -20,20 +20,24 @@ import {
...
   getIntrospectionQuery,
+  GraphQLNamedType,
 } from 'graphql';
 import copyToClipboard from 'copy-to-clipboard';
 import {
   getFragmentDependenciesForAST,
   getOperationFacts,
   VariableToType,
 } from 'graphql-language-service';
+import { SchemaReference } from 'codemirror-graphql/src/utils/SchemaReference';
```
acao, line 37: hmm, perhaps we should make this a top-level public export in `codemirror-graphql` to set a good example.  @imolorhe  does this make sense to you?

thomasheyenbrock:
> what do you think of moving the DocExplorer components to @graphiql/react workspace as well?

Definitely! Will do this in a follow up to not bloat this PR too much 👍

acao (approved):
this is good to go, exactly the direction we need to move in!

packages/graphiql/src/components/GraphiQL.tsx
thomasheyenbrock: As far as I can tell, `codemirror-graphql` currently doesn't have any top-level imports at all. Happy to follow up on this in another PR if it makes sense to you folks 🤗

PR #2371, opened 2022-04-22
draft: merge webpack to vite

rxliuli:
ref: https://github.com/graphql/graphiql/pull/2323#issuecomment-1104859303

### DIFF
--- a/packages/graphiql/index.html
+++ b/packages/graphiql/index.html
@@ -1,0 +1,41 @@
+<!--
+ *  Copyright (c) 2021 GraphQL Contributors
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE file in the root directory of this source tree.
+-->
+<!DOCTYPE html>
+<html>
+  <head>
+    <style>
+      body {
+        height: 100%;
+        margin: 0;
+        width: 100%;
+        overflow: hidden;
+      }
+      #graphiql {
+        height: 100vh;
+      }
+    </style>
+
+    <!--
+      This GraphiQL example depends on Promise and fetch, which are available in
+      modern browsers, but can be "polyfilled" for older browsers.
+      GraphiQL itself depends on React DOM.
+      If you do not want to rely on a CDN, you can host these files locally or
+      include them directly in your favored resource bunder.
+    -->
+
+    <!--
+      These two files can be found in the npm module, however you may wish to
+      copy them directly into your environment, or perhaps include them in your
+      favored resource bundler.
+     -->
+  </head>
+  <body>
+    <div id="graphiql">Loading...</div>
+    <script src="./src/main.tsx"></script>
+  </body>
+</html>
--- a/packages/graphiql/src/main.tsx
+++ b/packages/graphiql/src/main.tsx
@@ -1,0 +1,159 @@
+/**
+ * UMD GraphiQL Example
+ *
+ * This is a simple example that provides a primitive query string parser on top of GraphiQL props
+ * It assumes a global umd GraphiQL, which would be provided by an index.html in the default example
+ *
+ * It is used by:
+ * - the netlify demo
+ * - end to end tests
+ * - webpack dev server
+ */
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import GraphiQL from './cdn';
+import './css/app.css';
+import './css/codemirror.css';
+import './css/foldgutter.css';
+import './css/info.css';
+import './css/jump.css';
+import './css/lint.css';
+import './css/loading.css';
+import './css/show-hint.css';
+
+import './css/doc-explorer.css';
+import './css/history.css';
+
+// Parse the search string to get url parameters.
+var search = window.location.search;
+var parameters = {};
+search
+  .substr(1)
+  .split('&')
+  .forEach(function (entry) {
+    var eq = entry.indexOf('=');
+    if (eq >= 0) {
+      parameters[decodeURIComponent(entry.slice(0, eq))] = decodeURIComponent(
+        entry.slice(eq + 1),
+      );
+    }
+  });
+
+// If variables was provided, try to format it.
+if (parameters.variables) {
+  try {
+    parameters.variables = JSON.stringify(
+      JSON.parse(parameters.variables),
+      null,
+      2,
+    );
+  } catch (e) {
+    // Do nothing, we want to display the invalid JSON as a string, rather
+    // than present an error.
+  }
+}
+
+// If headers was provided, try to format it.
+if (parameters.headers) {
+  try {
+    parameters.headers = JSON.stringify(
+      JSON.parse(parameters.headers),
+      null,
+      2,
+    );
+  } catch (e) {
+    // Do nothing, we want to display the invalid JSON as a string, rather
+    // than present an error.
+  }
+}
+
+// When the query and variables string is edited, update the URL bar so
+// that it can be easily shared.
+function onEditQuery(newQuery) {
+  parameters.query = newQuery;
+  updateURL();
+}
+
+function onEditVariables(newVariables) {
+  parameters.variables = newVariables;
+  updateURL();
+}
+
+function onEditHeaders(newHeaders) {
+  parameters.headers = newHeaders;
+  updateURL();
+}
+
+function onEditOperationName(newOperationName) {
+  parameters.operationName = newOperationName;
+  updateURL();
+}
+
+function onTabChange(tabsState) {
+  const activeTab = tabsState.tabs[tabsState.activeTabIndex];
+  parameters.query = activeTab.query;
+  parameters.variables = activeTab.variables;
+  parameters.headers = activeTab.headers;
+  parameters.operationName = activeTab.operationName;
+  updateURL();
+}
+
+function updateURL() {
+  var newSearch =
+    '?' +
+    Object.keys(parameters)
+      .filter(function (key) {
+        return Boolean(parameters[key]);
+      })
+      .map(function (key) {
+        return (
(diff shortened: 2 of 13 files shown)


packages/graphiql/src/main.tsx
```diff
@@ -26,24 +26,28 @@ import './css/doc-explorer.css';
 import './css/history.css';
```
acao, line 26: I'm not sure how this file ended up in `src` (maybe I moved it?) but it should not be part of the resulting build. It's meant to just be an example implementation. It should be `example.js`, and not be a typescript file. The bundler should not pick up this file

rxliuli: Yes, it seems that I didn't notice that tsc will compile all the files in the directory, and src/main.tsx is only used as the entry script of the web application (the official practice of the vite project). Probably should be excluded from tsc or moved to resources/ directory

acao: ah, yes `src/main.tsx` should be `src/example.js` and should not be used as the bundling entry point, that should be `src/cdn.js` if you see in the webpack config. I originally had this file in the `resources/` directory, it must have been moved during the tabs effort.  `src/index.js` is the ESM main, which is different, and not what you want to treat as the entrypoint with vite.

rxliuli: Fixed, I excluded `src/main.tsx` from `tsconfig.json/tsconfig.esm.json`

acao: this file should not be importing individual css files either, the idea is that it's supposed to show a demo using the CDN bundle and thus the css bundle... I need to clean this file up in another PR!

acao: @rxliuli is it using `src/cdn.tsx` as an entry point now? I don't see the config for it

acao: `build: { lib: { entry: 'src/cdn.js', name: 'GraphiQL', formats: ['es', 'umd'] } }`

rxliuli: I am not using `src/cdn.tsx` as entry file

rxliuli: Understood, I also need to bundle the content of esm/umd, I will add some configuration

rxliuli: An error occurred, code splitting is not supported in umd format

```js
Invalid value "umd" for option "output.format" - UMD and IIFE output formats are not supported for code-splitting builds.
error during build:
Error: Invalid value "umd" for option "output.format" - UMD and IIFE output formats are not supported for code-splitting builds.
    at error (C:\Users\rxliuli\Code\Pkg\graphiql\node_modules\rollup\dist\shared\rollup.js:158:30)
    at validateOptionsForMultiChunkOutput (C:\Users\rxliuli\Code\Pkg\graphiql\node_modules\rollup\dist\shared\rollup.js:16118:16)
    at Bundle.generate (C:\Users\rxliuli\Code\Pkg\graphiql\node_modules\rollup\dist\shared\rollup.js:15957:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async handleGenerateWrite (C:\Users\rxliuli\Code\Pkg\graphiql\node_modules\rollup\dist\shared\rollup.js:23489:23)
    at async doBuild (C:\Users\rxliuli\Code\Pkg\graphiql\packages\graphiql\node_modules\vite\dist\node\chunks\dep-27bc1ab8.js:39180:26)
    at async build (C:\Users\rxliuli\Code\Pkg\graphiql\packages\graphiql\node_modules\vite\dist\node\chunks\dep-27bc1ab8.js:39011:16)
    at async CAC.<anonymous> (C:\Users\rxliuli\Code\Pkg\graphiql\packages\graphiql\node_modules\vite\dist\node\cli.js:738:9)
```

acao: huh weird, webpack supports it for `umd` by just un-splitting the code and inlining the dynamic imports. this is going to be a big problem then.

acao: the esm bundle users want dynamic imports for SSR, and the cdn users just want a single file and don't really need dynamic imports. there has to be a way?

packages/graphiql/src/main.tsx
```diff
@@ -0,0 +1,188 @@
...
+ * This is a simple example that provides a primitive query string parser on top of GraphiQL props
+ * It assumes a global umd GraphiQL, which would be provided by an index.html in the default example
+ *
+ * It is used by:
+ * - the netlify demo
+ * - end to end tests
+ * - webpack dev server
+ */
+
+import React from 'react';
```
acao, line 13: this file should not contain imports. this example was meant to be a CDN usage example, that tests the CDN example using the UMD without esm. this example implementaiton is not a file for users to import, I hope that's clear. the vast amount of our users use the unpkg/jsdelivr bundle, not the esm import. please see the comment at the top of this file

packages/graphiql/vite.config.ts
```diff
@@ -0,0 +1,48 @@
...
+import path from 'path';
+import express from 'express';
+const { graphqlHTTP } = require('express-graphql');
+const schema = require('./test/schema');
+const { schema: badSchema } = require('./test/bad-schema');
+const Router = require('router');
+
+export default defineConfig({
+  plugins: [
+    // reactRefresh(),
```
acao, line 13: Here's an explanation of how library mode works:

https://vitejs.dev/guide/build.html#library-mode

we need to use this, and specify `react` and `react-dom` as externals, so the shipped UMD bundle has parity and doesn't break existing implementations.

we need the demo script for development to use this UMD global, and not use `imports`, because the former is more widely used

acao: I think it will help you a lot to just take some time to study the webpack configuration, and understand how users are consuming this bundle, and try to re-create the same with vite

packages/graphiql/src/main.tsx
acao: @rxliuli it seems some of these solutions might work:

https://github.com/vitejs/vite/issues/2982

thomasheyenbrock: I played around with this too recently. There's a option that inlines all dynamic imports:
```ts
export default defineConfig({
  // ...
  build: {
    lib: { /* libarary mode */ },
    rollupOptions: {
      external: ['react', 'react-dom'],
      output: { inlineDynamicImports: true },
    },
  },
});
```

acao: brilliant!

packages/graphiql/src/main.tsx
```diff
@@ -0,0 +1,188 @@
+/**
```
acao, line 1: yeah, this file just needs to be removed. it should instead import `renderGraphiQL.js` as before, which does not contain any typescript, and uses the UMD global

packages/graphiql/index.html
```diff
@@ -0,0 +1,41 @@
...
+
+    <!--
+      These two files can be found in the npm module, however you may wish to
+      copy them directly into your environment, or perhaps include them in your
+      favored resource bundler.
+     -->
+  </head>
+  <body>
+    <div id="graphiql">Loading...</div>
+    <script src="./src/main.tsx"></script>
```
acao, line 39: remove `main.tsx` and instead point this to `renderGraphiQL.js` as we were already doing. it will expect a UMD bundle, not an esm module.

acao: the HTML file should look like this, to simulate the CDN setup:
https://github.com/graphql/graphiql/blob/eab257ba3e65e47625216fc3545b56426e0c4e61/packages/graphiql/resources/index.html.ejs#L30

rxliuli:
> I plan on coming back to this FYI! I have no wifi at home but now I'm at the office, so I may be able to put more time into helping you on this!

I don't have much time before 05.01, the work content is dead on the 30th, maybe I will have some time to deal with the 5.1-5.5 holiday

acao:
If you like! Would it be ok if I made some commits this week if I have some time?

acao:
Sorry it’s been a long time! But I would love to revive this PR and finish this up to help out the upstream graphiql2 and users who need more/ new formats. @jonathanawesome you may be interested in tinkering with this as well!

rxliuli:
Sorry, I haven't been motivated to do this since using Apollo explorer
https://studio.apollographql.com/sandbox/explorer

shabab477:
Hello @acao any plans to revive this?

acao:
yes I was hoping someone would! just have not had time. I a move to `pnpm` and move all library builds to esbuild & vite, and upgrade webpack example(s) to 5 and update vite examples etc.

acao:
@rxliuli i hope it's ok if we re-open your PR for reference as we make the migration?

rxliuli:
> @rxliuli i hope it's ok if we re-open your PR for reference as we make the migration?

Of course, feel free to use

acao:
@rxliuli thank you and thank you for this helpful draft PR! I'm happy you found a tool that works for you. GraphQL is the only mission!

acao:
@shabab477 let me know if and when you'd like to take this on! I just want to get a few more releases out to the LSP, and some fixes for graphiql@1.x before i dive down this important but inevitably complex "rabbit hole" haha

shabab477:
@acao I went over the repo. I'm open to work this weekend after my "usual" work 😉

shabab477:
@acao i created a PR [here](https://github.com/graphql/graphiql/pull/2638) but seems a lot has changed since this commit has been made. I will have resolve the conflicts.

dimaMachina:
closed by https://github.com/graphql/graphiql/pull/3679

PR #2373, opened 2022-04-22
fix(toolkit): operationName is optional

benjie:
`operationName` can be null, or can be not specified at all; however our TypeScript types mark it as required. This tiny PR fixes that.

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -1335,9 +1335,9 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
 
   private async _fetchQuery(
     query: string,
-    variables: string,
-    headers: string,
-    operationName: string,
+    variables: string | undefined,
+    headers: string | undefined,
+    operationName: string | undefined,
     shouldPersistHeaders: boolean,
     cb: (value: FetcherResult) => any,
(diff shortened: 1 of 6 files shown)


acao:
thanks @benjie ! you'll need to update the implementation in toolkit as well

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -1490,7 +1490,7 @@ export class GraphiQL extends React.Component<GraphiQLProps, GraphiQLState> {
     // Use the edited query after autoCompleteLeafs() runs or,
     // in case autoCompletion fails (the function returns undefined),
     // the current query from the editor.
-    const editedQuery = this.autoCompleteLeafs() || this.state.query;
+    const editedQuery = this.autoCompleteLeafs() || this.state.query || '';
```
benjie, line 1493: `editedQuery` _has_ to be a string because `_fetchQuery` does `query +=` which will only work well if it's already a string. In all the other cases, `undefined` is handled just fine hence the type change to the private `_fetchQuery` parameters.

timsuchanek (approved):
Awesome! Thanks for updating it in the toolkit as well!

packages/graphiql/src/components/GraphiQL.tsx
acao: good find!

PR #2378, opened 2022-04-26
gently trap graphql parse errors in the LS and LSP

### LINKED ISSUE #2376: panic when schemas exists empty type

acao:
I introduced these accidentally and they have caused a lot of issues!

should fix #2376 and many others


acao:
for the longest time I had been blaming `graphql-config` - but this was my mistake when I introduced the `externalFragments` feature!

PR #2387, opened 2022-05-05
fix(types): components that accept children should declare this

benjie:
The React components that accept children should declare as much as a prerequisite for compatibility with React 18, see:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210


timsuchanek (approved):
Looks good! Thanks for the update!

PR #2388, opened 2022-05-05
fix(css): use pointer-events:none on Menu dropdown SVG

benjie:
In React 18 the event handling has changed, I believe this is why in React 18 clicking on the specific arrow on the Menu results in no action (whereas clicking the rest of the button works fine). This is easily solved with the `pointer-events: none` CSS trick for the SVG as used elsewhere.


timsuchanek (approved):
Nice, thanks for updating that!

PR #2391, opened 2022-05-08
feat: provide intellisense for Vue SFC files

AumyF:
This pull request enables intellisense for GraphQL templates in `<script>` and `<script setup>` in `.vue` SFC files.

# Example

```vue
<script setup lang="ts">
// graphql-codegen uses the query to generate some code
`#graphql
query {id}
`;
</script>
```


acao:
so exciting @AumyF !! thanks for doing this!

@stonexer would you like to help review this as you have time?

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -79,26 +111,50 @@ export function findGraphQLTags(
...
 
-  const isTypeScript = ext === '.ts' || ext === '.tsx';
-  if (isTypeScript) {
-    plugins?.push('typescript');
+  const isVue = ext === '.vue';
+
+  let parsedASTs: { [key: string]: any }[] = [];
+
+  if (isVue) {
+    const result = parseVueSFC(text);
```
stonexer, line 119: The line: 110 already has a `result`, another name here would be better.

stonexer (changes requested):
Nice job! 

Maybe we can add some relevant tests?

acao:
@AumyF it still wants you to sign the CLA, are you able to?

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -69,6 +70,37 @@ const BABEL_PLUGINS: ParserPlugin[] = [
   'logicalAssignment',
 ];
 
+function parseVueSFC(
+  source: string,
+):
+  | { type: 'error'; errors: Error[] }
+  | {
```
acao, line 77: perhaps it might be easier and more readable to declare this return type union above rather than inline?

acao:
@AumyF anything I can do to help you finish this PR?

acao:
@AumyF the issue with EasyCLA is easily fixable. You just need to sign the CLA using the email that *all* of your commits are signed with. Sorry for the boilerplate, your next PR will be much easier haha!

acao:
@AumyF also can you add some tests? it might require a rebase, but we have some very specific tests for `findGraphQLTest` that are very easy to write and run

acao:
@AumyF I am going to revive this as a new PR. If you're able to sign the CLA with the same email the commits are signed with this weekend I can give you actual commit credit, but if not, I can give you a git username link attribution in the changelog. Unfortunately our CLA rules are quite strict and I might get in trouble for merging the work of someone who didn't pass the annoying but necessary CLA bot

AumyF:
I'm sorry for the late reply.  I have signed the CLA but I did with a different email address... The signature is invalid, isn't it?

And I finally got my motivation back, so, may I resume the pull request?

acao:
@AumyF it would be an honor! it's a fantastic one and I'd be happy to help you see it through

So, you can either re-sign the CLA using the email address you _did_ use for every commit.

If you prefer not to have to sign the form again, you can perform a `git rebase -i` however you prefer in such a way as to rewrite the email addresses used in the commits to the one you originally signed the CLA with.

AumyF:
I rebased onto latest main and added some commits (I will squash them later)

acao:
@AumyF thank you sooo much! this is so exciting. looks like the very picky CLA bot is  finally happy!! 😆

packages/graphql-language-service-server/src/__tests__/findGraphQLTags-test.ts
```diff
@@ -165,6 +165,32 @@ query Test {
...
+gql\`
+query {id}
+\`;
+</script>
+`;
+    const contents = findGraphQLTags(text, '.vue');
+    expect(contents[0].template).toEqual(`
+query {id}`);
+  });
+
```
acao, line 193: thank you for adding these tests!!

reading more about vue, i wonder if there are cases where .vue files are plain typescript/javascript and whether we should test for those - i `think` i see where you account for this in `findGraphQLTags`? If i understand correctly, render functions may be a case of this? or do these files have standard js/ts files in vue? https://vuejs.org/guide/extras/render-function.html#creating-vnodes

packages/vscode-graphql/esbuild.js
```diff
@@ -11,10 +11,46 @@ build({
   minify: arg === '--minify',
   platform: 'node',
   outdir: 'out/',
-  external: ['vscode'],
   format: 'cjs',
   sourcemap: true,
   watch: isWatchMode,
+  // Avoid bundling @vue/compiler-sfc's dynamic dependencies
```
acao, line 17: ah, is there any way we can more selectively import it by chance?

this only removes it for `vscode-graphql` users, and we have vim, emacs, sublime users as well, who may end up with a bloated bundle over this. are they dynamically loaded into the runtime as needed at least?

packages/graphql-language-service-server/src/__tests__/findGraphQLTags-test.ts
AumyF: As far as I know there is no such cases. Though we can write Vue components as non-SFC files using render functions, they should have a .js, .jsx, .ts or .tsx extension.

However thanks to your comment I realized I forgot to handle JSX in SFC.  `@vue/compiler-sfc` seems to supports JSX out of the box, so I added a test for it.

packages/vscode-graphql/esbuild.js
AumyF: My comment was a bit misleading. I specified these packages as `external` not to reduce bundle size, but to avoid bundling errors. They are template preprocessors and `require()`d at runtime by `@vue/consolidate`. `@vue/consolidate` does not specify the preprocessors as `dependencies` in its package.json, so users need to install them separately.

However, in this case, esbuild tries to load the preprocessors and fails with `package not found` error. 

If the users just install and use `graphql-language-service-cli`, no extra packages installed.

packages/graphql-language-service-server/src/__tests__/findGraphQLTags-test.ts
acao: i think that's what i meant, rad thanks!

acao (approved):
This looks good to go to me!

acao:
@AumyF I tried pushing a rebase but it says it's rejected, I wonder what I'm doing wrong? hmm!

acao:
Screenshot 2022-08-24 at 21 40 20
even without --force! hmm!

acao:
@tobiasdiez it is only missing a rebase! I was hoping to merge this after workspaces support, but I may merge it first because we had to revert workspaces

acao:
@tobiasdiez #2827 will be the one - I'm not able to push the rebase to the PR, so I've re-created it.

PR #2395, opened 2022-05-09
fix vscode publishing with changesets

acao:
This gets `vscode-graphql` publishing properly in concert with `changesets`.

When the workspace is marked `private: true`, we can't expect `postpublish` to fire, but `postversion` should!


acao:

PR #2397, opened 2022-05-11
chore(deps): use React v17 consistently

thomasheyenbrock:
This updates all packages and examples to use React v17. This should help us to avoid any errors related to different versions of React being loaded at the same time.

We also:
- Upgrade to `parcel@2` in the parcel example
- Fix dynamically importing `codemirror` for the parcel example
- Fix the CDN example by using the correct URL for the GraphQL-SWAPI
- Upgrade `@storybook` dependencies to avoid loading React v16 as sub-dependency


packages/graphiql/src/utility/importCodeMirror.ts
```diff
@@ -20,7 +20,11 @@ export async function importCodeMirror(
   addons: Promise<any>[],
   options?: { useCommonAddons?: boolean },
 ) {
-  const { default: CodeMirror } = await import('codemirror');
+  const CodeMirror = await import('codemirror').then(c =>
+    // Depending on bundler and settings the dynamic import either returns a
```
acao, line 24: great find!

acao:
The funky dynamic import logic is all for SSR, but the component is still not compatible with next.js transparently. It worked with earlier versions of next.js but now it doesn’t and you still have to use a funky workaround

PR #2398, opened 2022-05-11
fix bad changeset and avoid warnings

thomasheyenbrock:
I added a bad changeset in #2397 that contained ignored packages. This currently breaks the `release` workflow: https://github.com/graphql/graphiql/runs/6395880044?check_suite_focus=true

I also changed the dependencies in the webpack example to avoid the warnings printed above the error.


timsuchanek (approved):
LGTM!

PR #2404, opened 2022-05-13
add editor context and `useHeaderEditor` hook to `@graphiql/react`

thomasheyenbrock:
Continuing moving over stuff to `@graphiql/react`, next up is a context that stores the editor instances. I also created a `useHeaderEditor` hook that contains all the logic for this editor and just returns the ref that needs to be passed to a `div` element.

Notes (also will add some inline comments):
- I stopped with the header editor to keep the PR size under control, the other editors will follow soon!
- Also the `HeaderEditor` component will be moved in a follow up, I'll try to do this together with design changes.
- I copied over a bunch of utils around editors and completion. Once we moved all the editor logic over to `@graphiql/react` we can remove these utils from the `graphiql` package. (I'll mark these files with inline comments.)
- I removed the `headers` state property from the `GraphiQL` component, we now always derive the current value by looking at the `headers` props and read the editor value if this prop value is `undefined`. The logic for keeping the editor value in sync is all part of the `useHeaderEditor` hook.


packages/graphiql-react/src/editor/common.ts
```diff
@@ -0,0 +1,53 @@
+let isMacOs = false;
```
thomasheyenbrock, line 1: Copied from `packages/graphiql/src/utility/commonKeys.ts`

packages/graphiql-react/src/editor/common.ts
```diff
@@ -0,0 +1,53 @@
...
+  'Ctrl-G': 'findPersistent',
+
+  // Editor improvements
+  'Ctrl-Left': 'goSubwordLeft',
+  'Ctrl-Right': 'goSubwordRight',
+  'Alt-Left': 'goGroupLeft',
+  'Alt-Right': 'goGroupRight',
+};
+
+export const commonCodeMirrorAddons = [
```
thomasheyenbrock, line 20: Copied from `packages/graphiql/src/utility/importCodeMirror.ts`

packages/graphiql-react/src/editor/completion.ts
```diff
@@ -0,0 +1,107 @@
+import type { Editor, EditorChange } from 'codemirror';
```
thomasheyenbrock, line 1: Copied from `packages/graphiql/src/utility/onHasCompletion.ts`

packages/graphiql-react/vite.config.ts
```diff
@@ -6,11 +6,14 @@ export default defineConfig({
...
       entry: 'src/index.ts',
-      fileName: 'index',
+      fileName: 'graphiql-react',
       formats: ['cjs', 'es'],
     },
     rollupOptions: {
       external: ['react', 'react-dom'],
+      output: {
+        chunkFileNames: '[name].[format].js',
+      },
```
thomasheyenbrock, line 16: This makes sure that chunks will be unique when building with Vite in library mode for multiple formats.

packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
```diff
@@ -114,9 +114,11 @@ describe('GraphiQL', () => {
   });
 
   it('defaults to the built-in default query', async () => {
```
thomasheyenbrock, line 116: Fixed up this test which could never fail (`findByText` returns a promise which is always truthy).

timsuchanek (approved):
LGTM!

packages/graphiql-react/src/editor/completion.ts
acao: Is there a reason why the files are treated as new and not index changes? Is it because there are lots of changed lines? Or perhaps because the files aren’t deleted yet? Sorry reviewing from phone. Index change is ideal to retain git history, but if not, let’s have comments at the top of the file to link to to permalinks to the legacy files at least.

acao: My goal was to move each of these `@graphiql/react` iterations would be a set of git mv’d files from `graphiql` workspace, but that’s just what I envisioned

thomasheyenbrock: The files in `graphiql` aren't deleted yet!

I just moved the logic for the header editor in this PR to keep the LOCs changed in check 😅 So we need them in both places atm. Once we moved over the other editors (already working on it) we can remove the files from `graphiql`.

packages/graphiql-react/src/editor/common.ts
```diff
@@ -0,0 +1,53 @@
+let isMacOs = false;
+
+if (typeof window === 'object') {
+  isMacOs = window.navigator.platform === 'MacIntel';
```
JoviDeCroock, line 4: The current M1's seem to return `macOS` here for Brave browser, `MacIntel` for Chrome, `MacIntel` for FireFox and `MacIntel` we could alternatively do something like `window.navigator.platform.toLowerCase().startsWith('mac')`

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -0,0 +1,102 @@
+import { Editor, EditorChange } from 'codemirror';
+import { RefObject, useEffect, useRef } from 'react';
+
+import onHasCompletion from './completion';
+
+export function useSyncValue(editor: Editor | null, value: string | undefined) {
```
JoviDeCroock, line 6: NIT: this could be me but I read this as a way to get a synchronous value rather than to synchronise a value with the editor instance

packages/graphiql-react/src/editor/common.ts
thomasheyenbrock: Interesting, when I run this in Brave on my M1 Max I get `MacIntel` 🤔 However, I don't see anything that speaks against making this more loose 👍

JoviDeCroock: Heh, could've been me using `navigator.userAgentData.platform` as well 😅 apologies

PR #2408, opened 2022-05-17
add `useQueryEditor` hook to `@graphiql/react`

thomasheyenbrock:
This PR moves over the logic for the query editor into a hook exported from `@graphiql/react`.

For now we keep the query also in the state of the `GraphiQL` component. It's a bit more work to move this over to the editor context and hooks as we need to derive the operation facts when the query changes. Will do this in a follow-up PR.


packages/graphiql-react/src/editor/query-editor.tsx
```diff
@@ -0,0 +1,280 @@
...
+  useEffect(() => {
+    importCodeMirror([
+      import('codemirror/addon/comment/comment'),
+      import('codemirror/addon/search/search'),
+      import('codemirror-graphql/esm/hint'),
+      import('codemirror-graphql/esm/lint'),
+      import('codemirror-graphql/esm/info'),
+      import('codemirror-graphql/esm/jump'),
+      import('codemirror-graphql/esm/mode'),
+    ]).then(CodeMirror => {
```
JoviDeCroock, line 84: Since this effect can be cancelled (deps) do we want to do something like

```
let active = true;
// current stuff
return () => { active = false }
```

This way you can check in your `.then` whether you are in an aborted invocation which could avoid race conditions

JoviDeCroock (approved):
Minor comment about the slim chance of running into a race condition, looks great!

packages/graphiql-react/src/editor/context.tsx
```diff
@@ -1,24 +1,32 @@
...
+  queryEditor: CodeMirrorEditor | null;
+  setHeaderEditor(newEditor: CodeMirrorEditor): void;
+  setQueryEditor(newEditor: CodeMirrorEditor): void;
 };
 
 export const EditorContext = createContext<EditorContextType>({
   headerEditor: null,
+  queryEditor: null,
   setHeaderEditor() {},
+  setQueryEditor() {},
```
timsuchanek, line 16: Just out of curiosity - why are these empty functions?

packages/graphiql/__mocks__/@graphiql/react.ts
```diff
@@ -0,0 +1,98 @@
...
+      setCode(newValue);
+      onEdit?.(newValue);
+    });
+
+    const mockWrapper = document.createElement('div');
+    mockWrapper.appendChild(mockGutter);
+    mockWrapper.appendChild(mockTextArea);
+
+    ref.current.appendChild(mockWrapper);
+  });
```
timsuchanek, line 64: Are you sure you don't want to add a `[]` dependency here? With this, `useEffect` will be called on every render.

timsuchanek (approved):
Looks good to me! Just 2 clarification questions from my side.

packages/graphiql-react/src/editor/context.tsx
thomasheyenbrock: TypeScript wants you to provide a default value when creating a context. When actually rendering the context provider we'll pass in the actual value that contains the functions to set the editors in state.

packages/graphiql/__mocks__/@graphiql/react.ts
thomasheyenbrock: Good call 👍

package.json
```diff
@@ -30,13 +30,16 @@
     "npm": "please_use_yarn_instead"
   },
   "scripts": {
-    "build": "yarn workspace @graphiql/react run build && yarn tsc --clean && yarn tsc",
+    "build": "yarn build:clean && yarn build:graphiql-react && yarn build:packages && yarn build:graphiql",
```
thomasheyenbrock, line 33: I had to make sure the builds are executed in correct order since we use Vite for `@graphiql/react`:
- First build all packages
- Then build `@graphiql/react`
- Then build `graphiql`

packages/graphiql/__mocks__/@graphiql/react.ts
thomasheyenbrock: This effect is now still executing on each render, but we return early if the ref is set and already contains children. That way we also make sure that we insert the mock text area if the ref is not yet set on the initial render.

package.json
```diff
@@ -30,13 +30,16 @@
     "npm": "please_use_yarn_instead"
   },
   "scripts": {
-    "build": "yarn workspace @graphiql/react run build && yarn tsc --clean && yarn tsc",
+    "build": "yarn build:clean && yarn build:packages && yarn build:graphiql-react && yarn build:graphiql",
```
thomasheyenbrock, line 33: I had to make sure the builds are executed in correct order since we use Vite for `@graphiql/react`:
- First build all packages
- Then build `@graphiql/react`
- Then build `graphiql`

PR #2409, opened 2022-05-18
add `useVariableEditor` hook to `@graphiql/react`

thomasheyenbrock:
Next editor: Variables! Similar to #2404 and #2408 we create a hook `useVariableEditor` that contains all the logic for setting up Codemirror and returns a ref that we just need to attach to a DOM element. We also remove the `variable` property from `GraphiQLState` and use the editor state as source of truth.

Notes:
- We can now remove the `onHasCompletion` function from the `graphiql` package, we already copied this over to `@graphiql/react` in #2404.
- To avoid breaking changes we still re-export `onHasCompletion` from `graphiql`. Invoking this function prints a deprecation warning, we'll remove this export for the next major version.
- I fixed a bug introduced in #2404 around the header editor: Changes in the editor after the initial render were not seen for requests. (See [this commit](https://github.com/graphql/graphiql/commit/267ae5a96083a062f4f085520a1320f97c0c3273).)


acao:

timsuchanek (approved):
LGTM! I like the cleanup here!

PR #2411, opened 2022-05-19
add `useResponseEditor` hook to `@graphiql/react`

thomasheyenbrock:
This moves over the logic for the last (read-only) editor: The response editor. We also move the `ImagePreview` component to `@graphiql/react` as it's only used in the `useResponseEditor` hook. This allows us to remove the utilities from `graphiql` that have been temporarily copied over in #2404 (`commonKeys`, `importCodeMirror`) as well as the `CodeMirrorSizer` utility (each editor takes care of resizing itself now using the `useResizeEditor` hook).


packages/graphiql-react/src/editor/components/image-preview.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    if (width !== state.width || height !== state.height) {
+      setState(current => ({ ...current, height, width }));
+    }
+  }
+
+  useEffect(() => {
+    updateMetadata();
+  });
```
JoviDeCroock, line 49: should this run on every render?

thomasheyenbrock: Tbh it was like that before and I blindly migrated the logic 😄 But this is a good chance to break that up a bit, we can be more resourceful here 👍

JoviDeCroock: Yes splitting it up between the fetch and a resizeObserver might be a good shout, WDYT?

JoviDeCroock: no other benefits, wasn't sure whether you needed to respond to the user resizing the window

timsuchanek (approved):
LGTM

acao:

PR #2412, opened 2022-05-19
move storage classes to toolkit

thomasheyenbrock:
The `StorageAPI` and the things that build on top of it (`QueryStore` and `HistoryStore`) are quite generic and reusable. This PR moves all three of them over into the toolkit package. I sneaked in minor code improvements but overall there didn't change anything about how these classes function.


acao:

PR #2420, opened 2022-05-21
add `SchemaContext` to `@graphiql/react`

thomasheyenbrock:
Another rather big PR: Moving over the logic for fetching and validating the schema into a new context in `@graphiql/react`.

We still keep the `schema` inside `GraphiQL` state for now so that deriving operation facts can continue to work like before. Moving this logic over is next!


timsuchanek:
Wouldn't this even potentially fit into `@graphiql/toolkit` as it's mostly UI agnostic?
Not saying that it's a high prio, happy to optimize for speed here.

timsuchanek (approved):
LGTM!

acao:
@timsuchanek it would seem that way, but react context is still react specific.

I was thinking that if we move the state to something like zustand or xstate eventually, then we would have a js framework agnostic state implementation for toolkit. But this should be fine for now. 

In vue they might use vuex, and sveltekit for svelte, so state manager implementations have become quite framework specific anyways

JoviDeCroock (approved):

PR #2425, opened 2022-05-22
remove erroneously re-added package.json files from #2208

acao:
Accidentally added these files from an old PR #2208 - bad rebase. these were the only erroneous changes from this PR


acao:

PR #2427, opened 2022-05-22
fix linting in query and variable editor

thomasheyenbrock:
Linting in the query and variable editor broke with the latest upgrade of the codemirror version: https://github.com/graphql/graphiql/commit/eab257ba3e65e47625216fc3545b56426e0c4e61

The reason are internal changes to the lint addon implementation, in particular the options object is not split up between `options` and `linterOptions` where our custom properties (e.g. `schema`) are added to the latter: https://github.com/codemirror/CodeMirror/commit/70c615c5ff7d25e91dd50190945ef295b9ce7f09

To fix this we also modify the `linterOptions` object in the editor state when we want to update one of our custom options.

Props @benjie who noticed the broken linting 👏

### DIFF
--- a/.changeset/olive-cooks-juggle.md
+++ b/.changeset/olive-cooks-juggle.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Mark `graphql` as external dependency to avoid importing multiple instances
--- a/packages/graphiql-react/src/editor/query-editor.tsx
+++ b/packages/graphiql-react/src/editor/query-editor.tsx
@@ -235,6 +235,7 @@ function useSynchronizeSchema(
 
     const didChange = editor.options.lint.schema !== schema;
 
+    editor.state.lint.linterOptions.schema = schema;
     editor.options.lint.schema = schema;
     editor.options.hintOptions.schema = schema;
     editor.options.info.schema = schema;
@@ -258,6 +259,7 @@ function useSynchronizeValidationRules(
 
     const didChange = editor.options.lint.validationRules !== validationRules;
 
+    editor.state.lint.linterOptions.validationRules = validationRules;
     editor.options.lint.validationRules = validationRules;
 
     if (didChange && codeMirrorRef.current) {
@@ -279,6 +281,7 @@ function useSynchronizeExternalFragments(
     const didChange =
       editor.options.lint.externalFragments !== externalFragments;
 
+    editor.state.lint.linterOptions.externalFragments = externalFragments;
     editor.options.lint.externalFragments = externalFragments;
     editor.options.hintOptions.externalFragments = externalFragments;
 
--- a/.changeset/orange-clocks-march.md
+++ b/.changeset/orange-clocks-march.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Fix linting by also updating the options object in the internal codemirror state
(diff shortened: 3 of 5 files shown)


acao:
Good catch! This warrants an end to end test perhaps?

thomasheyenbrock:
Good call @acao! Added a test suite that asserts that the linter is working for query and variables 👍

acao:
innnnnteresting...
Screenshot 2022-05-23 at 09 42 54

acao:
@thomasheyenbrock perhaps it's the specificity and a change in error formatting between versions, if i recall. perhaps try just two words that are consistent like `Unexpected character`?

.changeset/olive-cooks-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Mark `graphql` as external dependency to avoid importing multiple instances
```
acao, line 5: i like to see that you are trying to provide separate changelog entries like this to provide a clean changelog, but in the future, for multiple changes in the same commit you can just use a markdown list in the same changeset file!

acao (approved):
This is all looking great, and thank you for the thorough e2e suite! Looks good to merge.

Down the road, when monaco comes along we can build on the pattern you've introduced and provide a different set of language feature assertions for each of monaco or codemirror, and/or be used for a suite just for `codemirror-graphql` if after introducing monaco, the `graphiql` cypress suite no longer covers `codemirror-graphql`

packages/graphiql/cypress/integration/lint.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+          test {
+            id
+          }
+          +++
+        }
+      `,
+    }).assertLinterMarkWithMessage(
+      '+++',
+      'error',
+      'Syntax Error: Unexpected character: "+".',
```
acao, line 16: here you could use a version flag using `graphql/version`, and find the difference between these strings. annoying but it allows us to provide some backwards compatible support for 15

https://github.com/thomasheyenbrock/graphiql/blob/6efac7e5a565a13f9c3b2069098f51dbc3adfb4d/packages/graphiql/cypress/integration/incrementalDelivery.spec.ts#L6

first,

`import { version as graphqlVersion } from 'graphql/version';`

This is the only difference between major versions if i recall - the new syntax error ends with a `.`. there is probably a more terse way of doing this of course
```ts

// 
const syntaxError = graphqlVersion.includes('15') ? 'Syntax Error: Unexpected character: "+"' : 'Syntax Error: Unexpected character: "+".'

cy.visitWithOp({
      query: /* GraphQL */ `
        {
          doesNotExist
          test {
            id
          }
          +++
        }
      `,
    }).assertLinterMarkWithMessage(
      '+++',
      'error',
      syntaxError,
      )
 ```

File: packages/graphiql-react/src/editor/query-editor.tsx [hunk 3]
```diff
@@ -235,6 +235,7 @@ function useSynchronizeSchema(
 
     const didChange = editor.options.lint.schema !== schema;
 
+    editor.state.lint.linterOptions.schema = schema;
```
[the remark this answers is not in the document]
benjie: :man_shrugging:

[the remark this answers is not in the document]
thomasheyenbrock: This is specific to the codemirror version we have added as a dependency. We do in a sense rely on implementation details of codemirror here, but I don't see another way of solving this atm 😞 That said, I'd prefer treating this as it's always gonna be there and letting it break if it isn't (which should only happen if we upgrade codemirror).

packages/graphiql/cypress/integration/lint.spec.ts
thomasheyenbrock: Thanks for the hint! Used a ternary based on the GraphQL version to resolve this as you proposed 👍

PR #2435, opened 2022-05-25
move logic for deriving operation facts to `@graphiql/react`

thomasheyenbrock:
The operation facts are currently derived in on-change handlers as part of the huge `GraphiQL` component and are stored in state. We move this logic to `@graphiql/react` and make it a bit more light-weight. Instead of storing it as React state, we store the operation facts as properties on the query editor instance. (Updating the query already triggers a re-render that one could react upon, so it seemed excessive to trigger another one for the operation facts.)

We also fix setting the initial editor values by also looking at the things that might be stored in `localStorage`.


timsuchanek (approved):
Nice cleanup! LGTM

PR #2436, opened 2022-05-25
move state for doc explorer visibility to explorer context in `@graphiql/react`

thomasheyenbrock:
In addition we can now also move the logic for clicking references to get to the docs into `@graphiql/react`, that way we don't need to pass the `onClickReference` and `onHintInformationRender` props to the editors anymore.


timsuchanek (approved):
LGTM!

PR #2438, opened 2022-05-25
chore: use the same TypeScript version throughout the monorepository

n1ru4l:
GraphQL.js 17 requires TypeScript >= 4.4; This is a base PR for a stacked GraphQL v17 update PR.

### DIFF
--- a/examples/graphiql-parcel/package.json
+++ b/examples/graphiql-parcel/package.json
@@ -26,7 +26,7 @@
     "graphql": "^16.4.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "typescript": "^3.4.4"
+    "typescript": "^4.6.3"
   },
   "devDependencies": {
     "parcel": "^2.5.0",
--- a/packages/graphiql/src/components/ToolbarButton.tsx
+++ b/packages/graphiql/src/components/ToolbarButton.tsx
@@ -49,7 +49,11 @@ export class ToolbarButton extends React.Component<
       this.props.onClick();
       this.setState({ error: null });
     } catch (error) {
-      this.setState({ error });
+      if (error instanceof Error) {
+        this.setState({ error });
+        return;
+      }
+      throw error;
     }
   };
 }
--- a/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
@@ -15,6 +15,7 @@ import {
   NamedTypeNode,
   ValidationRule,
   FieldNode,
+  GraphQLError,
 } from 'graphql';
 
 import {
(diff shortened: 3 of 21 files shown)


examples/graphiql-parcel/package.json
```diff
@@ -26,7 +26,7 @@
     "graphql": "^16.4.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "typescript": "^3.4.4"
```
acao, line 29: see above

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -153,15 +154,22 @@ export class GraphQLLanguageService {
...
+          {
+            severity: DIAGNOSTIC_SEVERITY.Error,
+            message: error.message,
+            source: 'GraphQL: Syntax',
+            range,
+          },
+        ];
+      }
+
+      throw error;
```
acao, line 172: do not `throw` error here or it will crash the language server. please remove!

acao (changes requested):

packages/graphql-language-service/src/interface/getDiagnostics.ts
```diff
@@ -81,15 +81,22 @@ export function getDiagnostics(
...
+      return [
+        {
+          severity: DIAGNOSTIC_SEVERITY.Error as DiagnosticSeverity,
+          message: error.message,
+          source: 'GraphQL: Syntax',
+          range,
+        },
+      ];
+    }
+    throw error;
```
acao, line 99: Please do not throw this error. we trap  and report them as needed in language services, we don't throw. This would crash graphiql or the LSP server on syntax errors

acao (changes requested):
There are a lot changes in this PR that would lead to bugs throughout the LSP ecosystem.

As a rule of thumb, you should not be `throw`ing exceptions on `catch` when they are being trapped already. If you don't see `throw`, then that means that exception should not bubble up. 

This is a golden rule of language tooling, and how you prevent crashing the server or the browser client on simple validation or network errors.

packages/graphql-language-service/src/interface/getDiagnostics.ts
acao: ohh right ok. in that case, I would prefer it log the message but still trap the exception

packages/graphiql/src/components/ToolbarButton.tsx
```diff
@@ -49,7 +49,11 @@ export class ToolbarButton extends React.Component<
       this.props.onClick();
       this.setState({ error: null });
     } catch (error) {
-      this.setState({ error });
+      if (error instanceof Error) {
+        this.setState({ error });
+        return;
+      }
+      throw error;
```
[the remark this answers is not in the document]
acao: @n1ru4l nevermind, this code has all moved a bunch!

acao:
We just want the exceptions in the language service and thus the language service server to not bubble up unless that is desired.

By `throw`ing in some of these cases that effect the language server, the exceptions don't bubble like you'd hope. 

Even though we trap the exceptions  in `startServer`, if you throw an exception from the `MessageProcessor` methods, it seems to cause the server to restart 5x times.  

However, it may be that this bug was caused by an uncaught exception? Seems like we need some more unit tests for this at least.

acao:
Perhaps this is why some exceptions in the MessageProcessor are uncaught?

Screenshot 2022-06-08 at 15 44 08

acao:
rebasing and testing locally now FYI!

acao (approved):

acao:
This is probably fine, let's just do some manual testing with the language server before merging? You can just use the vscode launcher. As a rule, I don't merge changes to the language server without manually checking locally, because the language server nor the vscode client have functional test coverage.

acao:
A local smoketest with the LSP server reveals no regressions that I can detect, all looks good, thanks @n1ru4l ! merging this soon

PR #2439, opened 2022-05-25
update to GraphQL 17

n1ru4l:


acao:
@n1ru4l can you update the regression matrix in github actions then?

PR #2454, opened 2022-05-30
remove private methods, deprecate public methods, and fix ref forwarding

thomasheyenbrock:
- We remove and replace some private methods:
  - All `handleEditX` were just wrapping the callback props
  - We replace the `handleSelectHistoryQuery` method with a custom hook
- All remaining public methods are now marked as deprecated
- I noticed that we broke these methods with `1.9.0` anyways because we forgot to forward the ref to the class component. Nobody complained, still I'd propose to avoid breaking changes until we release a new major version, so I added ref forwarding.

### DIFF
--- a/.changeset/clean-jokes-build.md
+++ b/.changeset/clean-jokes-build.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Deprecate the public methods `getQueryEditor`, `getVariableEditor`, `getHeaderEditor`, and `refresh` on the `GraphiQL` class.
--- a/.changeset/kind-carrots-lick.md
+++ b/.changeset/kind-carrots-lick.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Continue forwarding the ref to the class component to not break public methods
--- a/packages/graphiql-react/src/editor/query-editor.tsx
+++ b/packages/graphiql-react/src/editor/query-editor.tsx
@@ -1,6 +1,7 @@
-import { getSelectedOperationName, StorageAPI } from '@graphiql/toolkit';
+import { getSelectedOperationName } from '@graphiql/toolkit';
 import type { SchemaReference } from 'codemirror-graphql/utils/SchemaReference';
 import type {
+  DocumentNode,
   FragmentDefinitionNode,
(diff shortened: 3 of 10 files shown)


timsuchanek (approved):

acao:
@thomasheyenbrock that's a good sign that there haven't been complaints. I had pinned an issue about deprecating these interfaces for a few years that is now buried, but between that and a comment I had on the readme for a while, I think with 1.x we scared people into using the documented APIs.

in 2.x if not sooner we should drop assigning GraphiQL component to `window.g` as well, this is internal only but can present a big problem. Our usage of `window.g` in the e2e suites is an antipattern I introduced when I created them, as we should be using cypress to keyboard input these values instead of programatically setting them

The only react API anti-pattern we should worry about keeping legacy support for is GraphiQL.Logo, GraphiQL.Toolbar etc. Many custom implementations override these components, and I see this in use in the wild all the time. It's currently the only good way to add custom buttons or a custom logo. We will need feature parity for this in 2.x at some point.

thomasheyenbrock:
Thanks for the context @acao, that all makes sense to me! 👍

acao:
@thomasheyenbrock it looks like the component instance is still available in the latest version? Maybe you meant something else changed in 1.9.0?

Screenshot 2022-05-31 at 10 36 52

thomasheyenbrock:
`window.g` was never broken, passing a ref to the exported `GraphiQL` had a breaking change:
- In `1.8.10` the exported `GraphiQL` component was still the main class component, so you could call its method via the ref.
- Starting with `1.9.0` we wrapped the class component with a function component that wraps it with context providers, so passing a ref would not give you the class component anymore.

acao:
@thomasheyenbrock I understand better, thank you! I don't imagine people are calling it by the ref. `window.g` and `GraphiQL.Logo` were documented interfaces from early on up until last year, so I would only expect issues with those. Did the ref forwarding negatively impact `GraphiQL.Logo`? Could you still do this during the regression?

`GraphiQL.Logo = () => <span>My Company</span>`

thomasheyenbrock:
This should not have impacted `GraphiQL.Logo` or any of the other exposed components 👍

PR #2456, opened 2022-05-30
update eslint to latest, eslint plugins

B2o5T:


.eslintrc.js
```diff
@@ -53,6 +53,7 @@ module.exports = {
     // Possible Errors (http://eslint.org/docs/rules/#possible-errors)
     'no-cond-assign': 1,
     'no-console': 1,
+    'no-constant-binary-expression': 2,
```
B2o5T, line 56: new rule that should be always enabled to catch possible errors

package.json
```diff
@@ -53,7 +53,7 @@
     "jest": "jest",
     "license-check": "jsgl --local ./",
     "lint": "yarn eslint && yarn lint-check && yarn pretty-check",
-    "lint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
```
B2o5T, line 56: https://github.com/prettier/eslint-config-prettier#legacy

package.json
```diff
@@ -103,26 +103,25 @@
     "@types/set-value": "^4.0.1",
     "@types/theme-ui": "^0.3.1",
     "@types/ws": "^7.4.0",
-    "@typescript-eslint/eslint-plugin": "^4.14.0",
-    "@typescript-eslint/parser": "^4.14.0",
+    "@typescript-eslint/eslint-plugin": "^5.27.0",
+    "@typescript-eslint/parser": "^5.27.0",
     "aws-serverless-express": "^3.4.0",
-    "babel-eslint": "^10.1.0",
```
B2o5T, line 109: never used, just removed

package.json
```diff
@@ -53,7 +53,7 @@
     "jest": "jest",
     "license-check": "jsgl --local ./",
     "lint": "yarn eslint && yarn lint-check && yarn pretty-check",
-    "lint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    "lint-check": "eslint --print-config .eslintrc.js",
```
thomasheyenbrock, line 56: Looking at the link you posted, shouldn't we now do this to check for conflicting rules? The current script only prints the config.
```suggestion
    "lint-check": "eslint-config-prettier .eslintrc.js",
```

thomasheyenbrock (commented):
Thanks for the housekeeping 🧹 🙏 just one question about the `eslint-config-prettier` update

acao (approved):
Looks fine to me, thanks for this! We can consider disabling the linter for `examples/` btw. the RFC code is outdated and will be deleted soon (just used for reference, from the 1st redesign effort in 2020) and is creating a lot of linter noise

thomasheyenbrock (approved):

PR #2463, opened 2022-06-01
show proper error when headers or variables are invalid json

thomasheyenbrock:
I messed this up in #2453 (which is not published yet), there is no global try-catch anymore, so we need to avoid throwing early.

### DIFF
--- a/packages/graphiql-react/src/execution.tsx
+++ b/packages/graphiql-react/src/execution.tsx
@@ -98,18 +98,29 @@ export function ExecutionContextProvider(props: ExecutionContextProviderProps) {
       let query = autoCompleteLeafs() || queryEditor.getValue();
 
       const variablesString = variableEditor?.getValue();
-      const variables = tryParseJson(
-        variablesString,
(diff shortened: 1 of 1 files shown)


packages/graphiql-react/src/execution.tsx
```diff
@@ -98,18 +98,29 @@ export function ExecutionContextProvider(props: ExecutionContextProviderProps) {
...
-        throw new Error('Headers are not a JSON object.');
+      let headers: Record<string, unknown> | undefined;
+      try {
+        headers = tryParseJsonObject({
+          json: headersString,
+          errorMessageParse: 'Headers are invalid JSON',
+          errorMessageType: 'Headers are not a JSON object.',
+        });
+      } catch (error) {
+        setResponse(error instanceof Error ? error.message : `${error}`);
```
timsuchanek, line 122: What else might error be besides an instance of Error?

timsuchanek: Sounds good!

PR #2469, opened 2022-06-03
migrate rfc to examples workspace, out of yarn workspace

acao:
this reduces the yarn lockfile by almost 4k lines 👀 

it also frees us up from weird resolutions caused by this awesome and useful but essentially now an artefect/point of reference


acao:
One of these days we won't even have babel in this repo! such lean deps we will have! 😍

Issue #2356, opened 2021-05-29
Tagged template literals not working with TypeScript type parameters


emmiep:

acao:
<img width="356" alt="Screenshot 2022-06-04 at 14 14 52" src="https://user-images.githubusercontent.com/1368727/171998522-bedd3777-11d3-44ec-b043-84a7fb58ecbb.png">

Finally got it working!

acao:
I cut a release, but the extension did not publish from the github action yet, so I will publish 0.4.10 manually tonight

acao:
ok. 0.4.11 is the release, let me know if there's any other issues! I haven't checked but this *should* fix #2444 which is the same bug just embedded in markdown

PR #2474, opened 2022-06-04
Ensure highlighting works with typescript type parameters

acao:
At least a partial fix for #2356, enables highlighting for template strings with typescript type parameters

Borrowed from https://github.com/apollographql/vscode-graphql/blob/main/syntaxes/graphql.js.json

Also casually bump the version for vscode-graphql once since 0.4.9 is released already from my local, so 0.4.10 becomes the next resolution we want.

This is not the most ideal way of accomplishing this, but I couldn't get it working with a single pattern

type param example

### DIFF
--- a/.changeset/sharp-bags-appear.md
+++ b/.changeset/sharp-bags-appear.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql': patch
+---
+
+Fix bug with typed parameters on the gql/graphql/etc tagged templates!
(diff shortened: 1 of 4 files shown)


acao:
@captbaritone @mjmahone et al - here's a great example of where apollo's grammars are still some of the best out there for vscode. would love to see if they are willing to open source them as an npm package or a highlighting-only-extension like we were discussing?

PR #2477, opened 2022-06-04
simplify doc explorer components

thomasheyenbrock:
Instead of passing down the `schema` and callbacks for clicking types and fields via props, we can
- consume the schema context directly in the component where we need the schema
- push to the nav stack directly in the `TypeLink` and `FieldLink` components

ℹ️ much more easy to review with hidden whitespace ℹ️ 

No changeset needed as non of the components where the props changed are public.


packages/graphiql/src/components/DocExplorer/SearchResults.tsx
```diff
@@ -6,152 +6,123 @@
...
-                isMatch(arg.name, searchValue),
-              );
-              if (matchingArgs.length === 0) {
-                return;
-              }
-            } else {
+    if (type && 'getFields' in type) {
+      const fields = type.getFields();
+      Object.keys(fields).forEach(fieldName => {
+        const field = fields[fieldName];
```
timsuchanek, line 61: Object.entries can make this a bit more elegant.
And according to my tests so far `for of` is always faster than `forEach`

packages/graphiql/src/components/DocExplorer/TypeDoc.tsx
```diff
@@ -5,207 +5,176 @@
...
-          {types.map(subtype => (
-            <div key={subtype.name} className="doc-category-item">
-              <TypeLink type={subtype} onClick={onClickType} />
-            </div>
+  // InputObject and Object
+  let fieldsDef;
+  let deprecatedFieldsDef;
+  if (type && 'getFields' in type) {
+    const fieldMap = type.getFields();
+    const fields = Object.keys(fieldMap).map(name => fieldMap[name]);
```
timsuchanek, line 74: Object.values()

packages/graphiql/src/components/DocExplorer/TypeDoc.tsx
```diff
@@ -5,207 +5,176 @@
...
-  handleShowDeprecated = () => this.setState({ showDeprecated: true });
+  return (
+    <div>
+      <MarkdownContent
+        className="doc-type-description"
+        markdown={
+          ('description' in type && type.description) || 'No Description'
+        }
+      />
+      {type instanceof GraphQLObjectType && typesDef}
```
timsuchanek, line 159: I'd probably use the helpers from graphql-js here, because those will also work across multiple GraphQL versions https://github.com/graphql/graphql-js/pull/3617

thomasheyenbrock:
Very nice comments @timsuchanek! 👏 Didn't look too closely at the implementation of the component yet. But while we're at it we can as well do this 👍

packages/graphiql/src/components/DocExplorer/SearchResults.tsx
thomasheyenbrock: Sadly `type.getFields()` does not return an iterator for input object types according to typescript, so we can't use for-of here

packages/graphiql/src/components/DocExplorer/TypeDoc.tsx
thomasheyenbrock: Same as above. It would actually work, but typescript would infer `any` as the return type of `Object.values(fieldMap)`

thomasheyenbrock:
After looking into it I could only do the last one for typescript reasons

timsuchanek (approved):
Nice, LGTM now!

acao:
@thomasheyenbrock since there are so many projects who use the DocExplorer export directly (which is documented usage), can we add something to the graphiql/react readme to show how to use the schema context and the DocExplorer component? Or make the schema prop optional?

packages/graphiql/src/components/DocExplorer/TypeLink.tsx
```diff
@@ -5,43 +5,45 @@
...
+      </>
+    );
   }
-  if (type instanceof GraphQLList) {
-    return <span>[{renderType(type.ofType, onClick)}]</span>;
+  if (isListType(type)) {
+    return (
+      <>
+        [<TypeLink type={type.ofType} />]
+      </>
```
B2o5T, line 35: no need for fragment wrapper
```suggestion
      <TypeLink type={type.ofType} />
```

thomasheyenbrock:
@acao good shout! I added back the `schema` prop to the `DocExplorer` for now and marked it as deprecated to be removed in v2.

thomasheyenbrock:
@benjie I can imagine that we would offer multiple versions of the doc explorer as plugins where one will be built-in (either the current one or the more advanced new one, tbd 😅 )

Issue #2121, opened 2021-12-15
`monaco-graphql` tabs through intellisense too aggressively


jeffsee55:
When you enter `tab` so select an intellisense option, it immediately drops you into the nested properties. This is different from codemirror graphql where it just puts you at the end of the word. This is mainly an issue when you have arugments you'd like to provide (especially when they're required)

https://www.loom.com/share/e20881c74d2f49f4858d508539c79a7a

acao:
first feedback about this new feature 🎂 ! recently introduced this to `graphql-language-service-server` and `vscode-graphql` as well

this is very helpful feedback! I hear this and could see how this would get annoying.

it's just so temptingly easy to add this on `completion` because of `command` and `insertText` in `monaco.languages.CompletionItem`. but also this seems like bad DX in retrospect, and makes field arguments *more* difficult to use

a few easier options for now:
1. make this an opt-in feature
2. only expand selection set for leaf fields without *any* arguments? 
3. only expand selection set for leaf fields with *required* arguments? (more expensive, is that even useful if you want to provide arguments though?)

more difficult,  may take longer:
4. if the field type has required arguments, expand the selection set automatically below as an `additionalInsertText`, but move the cursor to complete for the first argument. like this:
```graphql
query {
	chosenField(|) {
	  #selection set expanded, but cursor is in arguments
	}
}
```
5. choose to use an editor command/keybinding for this sort of behaviour similarly to `graphiql`?  (I think this may be the best option to avoid unexpected behavior for now). another or perhaps complementary option along this route would be a code action lightbulb that expands a selection set if you've left one un-expanded

definitely will take longer:
6. i have a WIP from this effort for required field and argument completion as well, but there are still some open DX questions and issues

also while we're here, can I ask - would it be helpful for it to be clearer in suggestions when fields or arguments are required?  I've also thought of modifying the suggestion sort to boost required fields or arguments to the top of the suggestion list.

jeffsee55:
These are really interesting options! It's really hard to say because this is such a _feel_ thing. I'd rather step through a bit more slowly because so much of GraphQL is about discovery 🤔 . In general, 5 sounds like the best option to me.

> would it be helpful for it to be clearer in suggestions when fields or arguments are required?

How would it be more clear? I see the `String!` in the hint box but maybe there's more to it?

[...]

PR #2488, opened 2022-06-08
Disable`fillLeafsOnComplete` by default

acao:
Users found this generally annoying by default, especially when there are required arguments

Fixes https://github.com/graphql/graphiql/issues/2121

Without automatically prompting autocompletion of required arguments as well as lead expansion, it makes the extension harder to use

You can now supply this in your graphql config:

`config.extensions.languageService.fillLeafsOnComplete`

Setting it to to `true` will enable this feature.
Will soon add the ability to manually enable this in `monaco-graphql` as well.

For both, this kind of behavior would be better as a keyboard command, context menu item &/or codelens prompt. We could even re-create the graphiql behavior for this, where after choosing an autocomplete item, `CMD` + `return` will automatically fill leaf fields

### DIFF
--- a/.changeset/khaki-otters-brush.md
+++ b/.changeset/khaki-otters-brush.md
@@ -1,0 +1,20 @@
+---
+'graphql-language-service-server': patch
+'monaco-graphql': patch
+'vscode-graphql': patch
+---
+
+Disable`fillLeafsOnComplete` by default
+    
+Users found this generally annoying by default, especially when there are required arguments
+    
+Without automatically prompting autocompletion of required arguments as well as lead expansion, it makes the extension harder to use
+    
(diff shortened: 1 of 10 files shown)


acao:
Bah, forgot I need to change a bunch of tests

acao:
@stonexer what do you think - updating these tests with minor changes to the language service are cumbersome to the point that I wonder if snapshot tests would be more ideal sometimes? 

the snapshot diffs would tell us everything we need to know about the changes. and then i could remove the boilerplate for asserting the expected output

stonexer:
I also think the snapshot test might be more efficient on the language service. And we can be a little more detailed when we review the snapshot diff.

PR #2491, opened 2022-06-08
add `.idea/` to `.gitignore` to make JetBrains users happy

B2o5T:


.gitignore
```diff
@@ -17,15 +17,14 @@ dist
 esm
 bundle
 cypress/screenshots
-node_modules/
```
B2o5T, line 20: already defined in line 9

.gitignore
```diff
@@ -17,15 +17,14 @@ dist
...
-npm-debug.log
```
B2o5T, line 21: already defined in line 6

PR #2501, opened 2022-06-14
add keymap option (allow vim, emacs keymap in codemirror 5)

acao:
This allows users to specify an alternative keymap, any of the default keymaps shipped with codemirror 5.


packages/graphiql-react/src/editor/types.ts
```diff
@@ -3,3 +3,5 @@ import type { Editor } from 'codemirror';
 export type CodeMirrorType = typeof import('codemirror');
 
 export type CodeMirrorEditor = Editor & { options?: any };
+
+export type KeyMapsEnum = 'sublime' | 'emacs' | 'vim';
```
thomasheyenbrock, line 7: Should we just name it `KeyMap`?
- The `Enum` suffix raises the expectation that this is actually an enum and not just a string union type.
- Imho using singular makes more sense as plural because this type represents one possible keymap and not multiple ones at the same time.

thomasheyenbrock (changes requested):
Awesome 🙌  thanks also for cleaning up the exports ❤️ Just one naming nit

thomasheyenbrock (approved):

PR #2502, opened 2022-06-14
prettier!!!

acao:
Inspired by the prettier support we introduced in `monaco-graphql` in 2020! And endless demands to add prettier support to graphiql

### DIFF
--- a/.changeset/fair-snails-drum.md
+++ b/.changeset/fair-snails-drum.md
@@ -1,0 +1,6 @@
+---
+'graphiql': patch
+'@graphiql/react': patch
+---
+
+Introduce prettier formatting support to graphiql
--- a/packages/graphiql-react/package.json
+++ b/packages/graphiql-react/package.json
@@ -37,7 +37,8 @@
     "escape-html": "^1.0.3",
     "graphql-language-service": "^5.0.5",
     "markdown-it": "^12.2.0",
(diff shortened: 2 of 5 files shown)


thomasheyenbrock (commented):
What's the motivation for moving to prettier instead of "vanilla" printing/parsing? The latter feels more idiomatic to me, as far as I see prettier implements its own parser and printer instead of using the ones from `graphql-js`.

thomasheyenbrock:
Just checked again, prettier does use `parse` from `graphql-js`, but implements its own printer. I'm basically wondering how fast `prettier` would be in terms of supporting changes to the GraphQL language syntax. Is this something that's done by GraphQL maintainers or by prettier maintainers?

acao:
@benjie and I were just discussing, the official graphql spec now suggests prettier for graphql formatting! The graphql formatter plugin is kept up to date, and this would help them even more!

acao:
This would lead to a very slow format on the first use of prettier, b/c of the initial dynamic import. Webpack allows you to specify magic comments to make dynamic import lazy but it wouldn’t work across bundlers I don’t think. Options could be passed from props as well

acao:
@thomasheyenbrock terribly sorry, I had meant to merge the `keyMaps` one and not this one, which requires more decision making and possibly community input

PR #2507, opened 2022-06-15
chore: extend `eslint:recommended` config

B2o5T:
Extending `eslint:recommended` config is a best practice while using ESLint, also I made cleanup of rules already defined in [`eslint:recommended` config](https://github.com/eslint/eslint/blob/main/conf/eslint-recommended.js) except:
- `no-case-declarations` as he is turned off
- `no-empty` as he has `allowEmptyCatch: true` option


examples/graphiql-2-rfc-context/src/components/GraphiQL.tsx
```diff
@@ -361,7 +361,7 @@ class GraphiQLInternals extends React.Component<
             console: variables,
           }}
           navPanels={[
-            // TODO: rewrite this for plugin API
+            // eslint-disable-next-line no-constant-condition -- TODO: rewrite this for plugin API
```
B2o5T, line 364: it's better to use `eslint-disable` directive that turns rule as warning for this particular case

examples/graphiql-2-rfc-context/src/components/DocExplorer/TypeDoc.tsx
```diff
@@ -235,11 +235,7 @@ function Field({ type, field, onClickType, onClickField }: FieldProps) {
   );
 }
 
-type EnumValue = {
-  value: GraphQLEnumValue;
-};
-
-function EnumValue({ value }: EnumValue) {
+function EnumValue({ value }: { value: GraphQLEnumValue }) {
```
B2o5T, line 238: Component name matches type name, just removed type to omit error report

.eslintrc.js
```diff
@@ -355,5 +312,13 @@ module.exports = {
         'no-await-in-loop': 0,
       },
     },
+    {
+      // TODO: Remove this overrides after extending `plugin:@typescript-eslint/recommended` config
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'no-redeclare': 'off', // ts(2451)
+        'no-undef': 'off', // ts(2304)
```
B2o5T, line 320: the following rules are turned off in `plugin:@typescript-eslint/recommended` config as they already catched by `tsc`

acao (approved):
looks great! i started removing `no-undef` in another PR but this is much more comprehensive

acao:
the more complexity we can remove, and the more we can lean on presets, the better

PR #2508, opened 2022-06-16
feat: add prettier!

acao:
This introduces prettier formatting by dynamically importing prettier's core library and the graphql plugin.  This is how we've imported prettier in `monaco-graphql`'s webworker compatible language service since 2020!

It's not ideal for UX, because the dependency is fetched only when using `format` for the first time. I would prefer to use pre-fetching, but I'm not sure how to implement `import()` prefetching in a cross-bundler compliant way. We could also manually prefetch on a triggered event, such as blur/focus of the format button.

### DIFF
--- a/packages/graphiql-react/src/editor/hooks.ts
+++ b/packages/graphiql-react/src/editor/hooks.ts
@@ -1,7 +1,7 @@
 import { fillLeafs, GetDefaultFieldNamesFn, mergeAst } from '@graphiql/toolkit';
 import { EditorChange } from 'codemirror';
 import copyToClipboard from 'copy-to-clipboard';
-import { parse, print } from 'graphql';
+import { print } from 'graphql';
 import { useCallback, useEffect } from 'react';
 
 import { useExplorerContext } from '../explorer';
@@ -208,12 +208,22 @@ export function usePrettifyEditors({
     }
 
     if (queryEditor) {
-      const editorContent = queryEditor.getValue();
-      const prettifiedEditorContent = print(parse(editorContent));
-
-      if (prettifiedEditorContent !== editorContent) {
-        queryEditor.setValue(prettifiedEditorContent);
-      }
+      import('prettier/standalone').then(({ default: prettier }) => {
+        import('prettier/parser-graphql').then(({ default: graphqlPlugin }) => {
+          try {
+            const editorContent = queryEditor.getValue();
+            const prettifiedEditorContent = prettier.format(editorContent, {
+              parser: 'graphql',
+              plugins: [graphqlPlugin],
+            });
+            if (prettifiedEditorContent !== editorContent) {
+              queryEditor.setValue(prettifiedEditorContent);
+            }
+          } catch {
+            // as always on format, no-op on parser failures
+          }
+        });
+      });
     }
   }, [queryEditor, variableEditor, headerEditor]);
 }
--- a/.changeset/fair-snails-drum.md
+++ b/.changeset/fair-snails-drum.md
@@ -1,0 +1,6 @@
+---
+'graphiql': patch
+'@graphiql/react': patch
+---
+
+Introduce prettier formatting support to graphiql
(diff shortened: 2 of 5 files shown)


acao:
@benjie can you speak to this from the language spec perspective? is using prettier's formatting for graphql safe for us from a spec compliance perspective? graphql support is shipped with prettier core, and is a frequently maintained and updated capability as the language spec changes it would seem

benjie:
I cannot speak on behalf of prettier, and I certainly don't know what it means for experimental features of GraphQL. I would imagine that there wouldn't be much pushback from prettier for supporting newer syntax so long as it is officially supported by graphql-js; however I do not know who is currently doing that work.

thomasheyenbrock:
Total sidenote: I started hacking on my own GraphQL implementation in JS (more modular compared to the all-in-one `graphql-js`) which is storing comments on the AST nodes: https://github.com/thomasheyenbrock/graphql-modular/blob/main/packages/language/src/index.ts

This is far from being useable (implemented traversal but no printing yet), just something I wanted to drop here.

saihaj:
we do have an issue to support `prettyPrint` so it is prettier like formatting https://github.com/graphql/graphql-js/issues/3230

saihaj (approved):
yes it for sure will be harder to work with new language features to get prettier support. But in reality I don't think stable `graphiql`  should be running an experiment`graphql-js` release so I think it is fine to merge this. So all those experimental features should be on some experimental `graphiql` release.

This comes back to the question me and @n1ru4l raised on how we can get `graphql-js` to easily support multiple RFC stage features

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -208,12 +208,22 @@ export function usePrettifyEditors({
...
 
     if (queryEditor) {
-      const editorContent = queryEditor.getValue();
-      const prettifiedEditorContent = print(parse(editorContent));
-
-      if (prettifiedEditorContent !== editorContent) {
-        queryEditor.setValue(prettifiedEditorContent);
-      }
+      import('prettier/standalone').then(({ default: prettier }) => {
+        import('prettier/parser-graphql').then(({ default: graphqlPlugin }) => {
```
B2o5T, line 212: I guess you can use `Promise.all` in this case, and async/await.
```suggestion
const [{ default: prettier }, { default: graphqlPlugin }] = await Promise.all([
  import('prettier/standalone'),
  import('prettier/parser-graphql'),
]);
```

acao: I would need to make the parent function async as well, I can try this

acao:
@saihaj @thomasheyenbrock what if we „incubate“ this formatting utility in graphiql as an internal detail, and then propose it as a PR to graphql-js?

saihaj:
> @saihaj @thomasheyenbrock what if we „incubate“ this formatting utility in graphiql as an internal detail, and then propose it as a PR to graphql-js?

`graphql-js` wouldn’t take this since it would require prettier as a dependency

acao:
@saihaj apologies for the confusion- we are proposing writing a custom formatter function that we maintain internally and/or contribute to graphql-js. Prettier would not be a dependency. See @thomasheyenbrock ‘s example above!

saihaj:
> @saihaj apologies for the confusion- we are proposing writing a custom formatter function that we maintain internally and/or contribute to graphql-js. Prettier would not be a dependency. See @thomasheyenbrock ‘s example above!

Ohhhh yes that would work!

thomasheyenbrock:
That sounds like a good proposal to me @acao! Just a custom printer that preserves comments would already be a huge improvement for GraphiQL, and we could iteratively see how to maybe incorporate that into `graphql-js` 👍 

Does that mean we abandon this PR in favor of a custom printer, or would we still want to add prettier as a first step and do the custom printer as next iteration?

PR #2509, opened 2022-06-16
feat: add gql call expressions support

Chnapy:
This PR enables highlight & intellisense on common `gql()` calls without need to add a `#graphql` tag or equivalent.

Before:
image

After:
image

### DIFF
--- a/packages/graphql-language-service-server/src/findGraphQLTags.ts
+++ b/packages/graphql-language-service-server/src/findGraphQLTags.ts
@@ -100,10 +100,48 @@ export function findGraphQLTags(
   }
   const ast = parsedAST!;
 
+  const parseTemplateLiteral = (node: TemplateLiteral) => {
+    const loc = node.quasis[0].loc;
+    if (loc) {
+      if (node.quasis.length > 1) {
+        const last = node.quasis.pop();
+        if (last?.loc?.end) {
+          loc.end = last.loc.end;
+        }
+      }
+      const template =
+        node.quasis.length > 1
+          ? node.quasis.map(quasi => quasi.value.raw).join('')
+          : node.quasis[0].value.raw;
+      const range = new Range(
+        new Position(loc.start.line - 1, loc.start.column),
+        new Position(loc.end.line - 1, loc.end.column),
+      );
+      result.push({
+        tag: '',
+        template,
+        range,
+      });
+    }
+  };
+
   const visitors = {
     CallExpression: (node: Expression) => {
       if ('callee' in node) {
         const callee = node.callee;
+
+        if (
+          callee.type === 'Identifier' &&
+          getGraphQLTagName(callee) &&
+          'arguments' in node
+        ) {
+          const templateLiteral = node.arguments[0];
+          if (templateLiteral && templateLiteral.type === 'TemplateLiteral') {
+            parseTemplateLiteral(templateLiteral);
+            return;
+          }
+        }
+
         if (
           !(
             (callee.type === 'Identifier' &&
@@ -208,28 +246,7 @@ export function findGraphQLTags(
         node.leadingComments?.[0]?.value.match(/^\s*GraphQL\s*$/),
       );
       if (hasGraphQLPrefix || hasGraphQLComment) {
-        const loc = node.quasis[0].loc;
-        if (loc) {
-          if (node.quasis.length > 1) {
-            const last = node.quasis.pop();
-            if (last?.loc?.end) {
-              loc.end = last.loc.end;
-            }
-          }
-          const template =
-            node.quasis.length > 1
-              ? node.quasis.map(quasi => quasi.value.raw).join('')
-              : node.quasis[0].value.raw;
-          const range = new Range(
-            new Position(loc.start.line - 1, loc.start.column),
-            new Position(loc.end.line - 1, loc.end.column),
-          );
-          result.push({
-            tag: '',
-            template,
-            range,
-          });
-        }
+        parseTemplateLiteral(node);
       }
     },
   };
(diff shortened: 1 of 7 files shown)


saihaj (approved):

acao:
@Chnapy this adds highlighting support only fyi!  It would seem language support already exists in the LSP server. Can you add a changeset?

acao:
@Chnapy I haven't seen this pattern used. Usually with `graphql-tag` and relay and such, it's

```graphql
const myQuery = gql`
  query {
    field
  }
`

```
as you can see, even github's ace editor supports this pattern!

whereas, this pattern is not supported in most of the third party ecosystem as you can see:

```graphql
const myQuery = gql(`
  query {
    field
  }
`)

```

do you have examples of libraries that use this pattern? I'm worried about creating confusion, because if someone used `graphql-tag` or similar libraries this way, it would look like it's working when it would break at runtime

acao:
@Chnapy my apologies, we can add support for this syntax no problem, I don't need examples, there is no sense holding back innovation.  

To clarify, the `vscode-graphql` grammars only add syntax highlighting support, and only to `vscode-graphql` client and not the LSP server itself. On the language support side, I'm actually not sure if our implementation of `@babel/core `findGraphQLTags` in `graphql-language-service-server`, I think it may actually already cover that, and appeared to not work because the grammar did not support syntax highlighting yet.

Chnapy:
Thanks for your feedbacks !

I worked on this add because of a small typescript plugin I'm working on ([ts-gql-plugin](https://github.com/Chnapy/ts-gql-plugin)).
Since I change typing returned by `gql` calls (`TypedDocumentNode` instead of `DocumentNode`) I cannot use template literal calls because of a [TS type-safety issue](https://github.com/microsoft/TypeScript/issues/33304). So I should use ``gql(`...`)`` instead of  ``gql`...` ``.

There is no difference between these 2 ways of doing with `graphql-tag`. To check that I made a basic test on https://npm.runkit.com/graphql-tag :
```js
require("graphql/package.json"); // graphql is a peer dependency. 
const { gql } = require("graphql-tag")

const a = gql`
    query Foo {
      bar {
        id
      }
    }
`;
  
const b = gql`
    query Foo {
      bar {
        ${'id'}
      }
    }
`;

const c = gql(`
    query Foo {
      bar {
        id
      }
    }
`);
  
const d = gql(`
    query Foo {
      bar {
        ${'id'}
      }
    }
`);

const strA = JSON.stringify(a);
const strB = JSON.stringify(b);
const strC = JSON.stringify(c);
const strD = JSON.stringify(d);

console.log(
a, b, c, d,
strA.length, strB.length, strC.length, strD.length,   // same length
strA === strB, strC === strD, strA === strC  // true
);
```
--

I admit this is a very specific use-case, based on one random librairy. I thought this add would cost quite nothing in maintainability (?). Also since it is simply possible to call `gql()` I find nice to handle this case, for projects with specific linter rules for example.

> To clarify, the `vscode-graphql` grammars only add syntax highlighting support, and only to `vscode-graphql` client and not the LSP server itself. On the language support side, I'm actually not sure if our implementation of `@babel/core` `findGraphQLTags` in `graphql-language-service-server`, I think it may actually already cover that, and appeared to not work because the grammar did not support syntax highlighting yet.

Hmm I see, I think I found [the part to work on](https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service-server/src/findGraphQLTags.ts#L203-L210). I'll pass some time and update this PR if I succeed something 👍

acao:
@Chnapy yes my bad, `graphql-tag` does work both ways now! a long time ago I think it didn't

the visitor you linked to is used for comment tags, which is a different visitor than what we need.

what we need to do is repeat this for non-template tag expressions:

https://github.com/graphql/graphiql/blob/a390b40de18f7d85d9f51ebecaf944746186f722/packages/graphql-language-service-server/src/findGraphQLTags.ts#L174

Chnapy:
I made few changes on `findGraphQLTags`, language support works well now following my local testing 👍 

I put the logic into `CallExpression` function to ensure tag name check, and reuse `TemplateLiteral` logic.

Some tests added in `MessageProcessor-test.ts`, and I created `findGraphQLTags-test.ts` as you suggested. In this file I put just few tests, but major part of MessageProcessor tests can be copied to it.

acao:
@Chnapy brilliant work, thanks for this! Looks great so far and I will give it a spin later today

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -100,10 +100,48 @@ export function findGraphQLTags(
   }
   const ast = parsedAST!;
 
+  const parseTemplateLiteral = (node: TemplateLiteral) => {
```
acao, line 103: this is exactly the abstraction I was hoping for!

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -100,10 +100,48 @@ export function findGraphQLTags(
...
+  };
+
   const visitors = {
     CallExpression: (node: Expression) => {
       if ('callee' in node) {
         const callee = node.callee;
+
+        if (
+          callee.type === 'Identifier' &&
+          getGraphQLTagName(callee) &&
```
acao, line 135: 💯

acao (approved):
@Chnapy reviewed it locally and everything works great. thanks for all your work on this!

acao:
@Chnapy there is just one upstream issue when rebasing with cspell, for `randomthing` in the tests, I added the ignore and tried to push to your branch but it seems I am not able to

acao:
@Chnapy I think I've found a substantial section of code that is never used within `CallExpression` visitor.  Tests confirm that everything appears to be ok - I just wonder what test case I'm missing that this was supposed to cover?

acao:
@stonexer if you'd like to take a look, and make sure you can't find a test case that I'm overlooking by deleting this seemingly unused functionality??

acao:
merged! thanks @Chnapy ! 🥳

Issue #2325, opened 2022-04-19
Implement new design with CodeMirror 5


timsuchanek:
To achieve the [new design](https://github.com/graphql/graphiql/discussions/2216) in an incremental fashion, we want to first implement as much as possible of it with the current GraphiQL technology stack before moving to Monaco / modernize any components.
This is the tracking issue for that work.

thomasheyenbrock:
The new design has been implemented and shipped with `graphiql@2`

Issue #2328, opened 2022-04-19
Path to GraphiQL v2


timsuchanek:
We have [ambitious plans](https://github.com/graphql/graphiql/discussions/2216) for the future of GraphiQL and want to go there step by step.

This is the path I suggest to move forward:

## 1. Implement [new design](https://www.figma.com/file/p8f8MaG7ZRnuCc4lVv7w6t/GraphiQL-2.0?node-id=0%3A1) with existing components
- [x] https://github.com/graphql/graphiql/issues/2325
## 2. Modernize components step by step
As the new design has a slightly different layout, in Step 1, the components can already be modernized as needed.
If necessary, we'll create a separate issue for tracking work that's left.

## 3. Introduce monaco
- [ ] https://github.com/graphql/graphiql/issues/2326

## 4. Implement the new sidebar.
The new sidebar is quite ambitious, that's why we want to work on it last.
- [ ] https://github.com/graphql/graphiql/issues/2327
<img width="502" alt="image" src="https://user-images.githubusercontent.com/1094804/164035768-1fbc91f3-5115-4764-9f08-9cc662aba818.png">

## Additional tasks
While working on this, we want to revisit and consolidate the boards @acao created:
- [ ] https://github.com/orgs/graphql/projects/1/views/1?query=is%3Aopen+sort%3Aupdated-desc
- [ ] https://github.com/orgs/graphql/projects/1/views/4?query=is%3Aopen+sort%3Aupdated-desc
- [ ] https://github.com/orgs/graphql/projects/1/views/5?query=is%3Aopen+sort%3Aupdated-desc
- [ ] https://github.com/orgs/graphql/projects/1/views/7?query=is%3Aopen+sort%3Aupdated-desc

Some of the issues will be fixed by our tasks above, some are additional improvements, bug fixes, and features.

We'll have @thomasheyenbrock from GraphCDN focus on steps 1-4 in the coming weeks.
Anyone who's interested in collaborating, please let us know and we can organize code pairing sessions!

We also commit to 1h of issue triaging per week, as some of them grew a bit old. We start May 2nd at 4 pm UTC. Let me know if you want to join. Calendar link:

https://calendar.google.com/calendar/u/0/embed?src=linuxfoundation.org_ik79t9uuj2p32i3r203dgv5mo8@group.calendar.google.com

You can either answer here or [join the graphiql Discord channel.](https://discord.gg/cXKXpdMyTt)

acao:
[...]

PR #2513, opened 2022-06-21
[redesign]: editor in new design

thomasheyenbrock:
Part of #2325 

This implements the new design for all the editors and moves the components to `@graphiql/react`. I left some more notes as inline comments.

Preview at https://deploy-preview-2513--graphiql-test.netlify.app/


.changeset/pre.json
```diff
@@ -0,0 +1,19 @@
+{
+  "mode": "pre",
```
thomasheyenbrock, line 2: We're in pre-release mode now! After merging this PR, all new versions will be published as alpha.

packages/graphiql-react/src/editor/completion.ts
```diff
@@ -18,28 +18,122 @@ export function onHasCompletion(
...
+          deprecationLabel.innerText = 'Deprecated';
+          deprecation.appendChild(deprecationLabel);
+
+          deprecationReason = document.createElement('div');
+          deprecationReason.className =
+            'CodeMirror-hint-information-deprecation-reason';
+          deprecation.appendChild(deprecationReason);
+
+          /**
+           * This is a bit hacky: By default, codemirror renders all hints
```
thomasheyenbrock, line 89: That was probably the trickiest part: Getting the two-column layout working for the typeahead hints. I hope this comment explains how I made this work with the restrictions imposed by codemirror.

Also, if you see a better way, happy to change the implementation!

thomasheyenbrock:
@acao do you want to get in #2502 before we start with the breaking changes and the pre-release? Or are you fine with waiting a couple of weeks for v2? (Shipping prettier with v2 could also be an option.)

JoviDeCroock (approved):

.changeset/pre.json
acao: So excited!

acao:
@thomasheyenbrock let’s see how it plays out! Let’s not let one block the other, but if we can make them part of the same breaking release that would be cool!

packages/graphiql-react/src/style/root.css
```diff
@@ -0,0 +1,61 @@
+.graphiql-container,
+.CodeMirror-info,
+.CodeMirror-lint-tooltip {
+  /* Colors */
```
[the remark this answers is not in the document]
thomasheyenbrock: Makes sense! I added an issue for this: https://github.com/graphql/graphiql/issues/2516

acao:
@thomasheyenbrock should I assume this one comes first, then #2515 ?

packages/codemirror-graphql/src/info.ts
```diff
@@ -74,36 +74,51 @@ CodeMirror.registerHelper(
       (kind === 'Field' && step === 0 && typeInfo.fieldDef) ||
       (kind === 'AliasedField' && step === 2 && typeInfo.fieldDef)
     ) {
+      const header = document.createElement('div');
+      header.className = 'CodeMirror-info-header';
+      renderField(header, typeInfo, options);
       const into = document.createElement('div');
-      renderField(into, typeInfo, options);
+      into.appendChild(header);
```
acao, line 81: I would like for maintainers such as @yoshiakis and @imolorhe to be able to review this change, to make sure that it doesn't clash with other popular `codemirror-graphql` implementations outside of `graphiql`

.changeset/pre.json
```diff
@@ -0,0 +1,19 @@
...
+  "tag": "alpha",
+  "initialVersions": {
```
acao, line 4: `initialVersions` here should be `graphiql`, `@graphiql/react`, `codemirror-graphql` and `@graphiql/toolkit` only. 

I imagine the changests CLI added all of these just because they had active diffs or changes, but there is no reason the language service, server, cli,  monaco-graphql, vscode-graphql, or any of these examples need to go into pre-release for these changes

packages/graphiql-react/src/editor/completion.ts
```diff
@@ -56,67 +151,87 @@ export function onHasCompletion(
...
+            const renderType = (type: GraphQLType) => {
+              if (isNonNullType(type)) {
+                typeNameSuffix!.innerText = '!' + typeNameSuffix!.innerText;
+                renderType(type.ofType);
+              } else if (isListType(type)) {
+                typeNamePrefix!.innerText += '[';
+                typeNameSuffix!.innerText = ']' + typeNameSuffix!.innerText;
+                renderType(type.ofType);
+              } else {
+                typeName!.innerText = type.name;
```
acao, line 185: yikes! we need to escape this input as before, or else it risks a security regression for a known past vulnerability that we have already issued a CVE for

all of this unsanitized html-string-building we can do away with via monaco-graphql! in fact we will be able to delete this entire file, and lots of other complex logic in the app, but for now, we need to keep this secure and to not re-introduce a security flaw.

@graphql/tsc if you can, help me keep an eye on these PRs so we don't accidentally re-introduce security flaws elsewhere. I don't have time to review every one, and considering someone has already approved this, it shows we need to keep a careful eye on things. Thomas is balancing a lot, and especially because the file history here is gone, he may not have known the historical reason for escaping user input

acao (changes requested):
Let's make sure we are escaping schema as user input here, so we don't re-introduce the XSS exploit we patched & published last year

https://www.cvedetails.com/cve/CVE-2021-41248/

I will come back for a more thorough review later

packages/graphiql-react/src/editor/completion.ts
acao: Since the patch, in order to exploit this vulnerability, a user has to enable `dangerouslyAssumeSchemaIsValid`. 

However it seems many are using this setting, and on the server bypassing type name validation using flags like [`GRAPHQL_NO_NAME_WARNING`](https://github.com/graphql/graphql-js/blob/831598ba76f015078ecb6c5c1fbaf133302f3f8e/src/utilities/assertValidName.js#L14-L16), leaving any unsuspecting user who uses such an implementation vulnerable to the attack.

thomasheyenbrock: We now use `innerText` instead of `innerHTML`, so this does not suffer from any injection vulnerabilities anymore. (That's why I also removed `escape-html`.)

packages/codemirror-graphql/src/info.ts
thomasheyenbrock: Makes sense! This would be a new major version for `codemirror-graphql`, so it would not break any dependents. Still good to know if this would be blocking others to update to this new version in any way.

@yoshiakis @imolorhe to sum up the changes for easy digestion:
- For fields the type prefix was removed, i.e. `MyType.myField` -> `myField`
- For args, the type and field was removed, i.e. `MyType.myField(myArg: MyArgType)` -> `myArg: MyArgType`
- The DOM structure of the info tooltip changed (for styling reasons):
  - The first section (i.e. the clickable parts like type and field name) are wrapped in an additional `div`
  - The markdown content for deprecation reasons is wrapped in an additional `div`

.changeset/pre.json
acao: the dedicated branch is how things went south before, because then the main and dedicated branch (2.x at the time) got terribly out of sync!

that said, as long as we don't make any changes to the graphiql packages from main while this branch exists, and the plan is to merge it back into main, I think we will be fine? just rebase often!

packages/graphiql-react/src/editor/completion.ts
acao: that's an important difference! it's usually safe, [but there are cases where it isn't](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#usually-safe-methods)

acao (approved):

packages/codemirror-graphql/src/info.ts
acao: also CC @stonexer @benjie @sgrove  and others who might be interested in this conversation!

acao: @imolorhe yes, essentially just by making it a major version change and adding the equivalent conventional-commits style breaking change note in bold in the changeset markdown! Then it will appear in the gh release and in the changelog

thomasheyenbrock:
@imolorhe For context, this PR essentially is part of #2325, which doesn't contain any information. The "parent issue" #2328 actually contains more information around how we want to more forward. And at last, the design itself was discussed in #2216. Hope that helps 😊

.changeset/pre.json
thomasheyenbrock: Yeah I also don't really favor this approach, but I don't see a better option 😕 (I also hope that we can get all of this done within the next couple of weeks, so this branch won't be around for a longer time 🤞 )

Then let's do it like this! I'll create a branch `graphiql-v2` where we'll merge all the breaking changes and the new design into. Once we reviewed and collected everthing we'll merge this branch into `main`. I'll also rebase the branch regularly so that it won't get out of sync and doesn't block other PRs being merged into `main` in the meantime.

packages/codemirror-graphql/src/info.ts
thomasheyenbrock: Yes, this will show up in the changelog! I also made the changeset message more elaborate and added the bullets from my comment above.

PR #2515, opened 2022-06-24
[redesign] layout for editor components

thomasheyenbrock:
Part of #2325.

This implements the new design for the layout for the editor components:
- Writable editors with white background (query, variables, headers)
- Read-only response editor with gray background
- Toggle button for the editor tools (variable & header editors)

PR #2523, opened 2022-06-29
graphiql@2

thomasheyenbrock:
This branch collects all the breaking changes for the next major version of `graphiql` and will be merged once we're ready to release it.

For context, we decided on this approach because using changeset pre-releases would also block other packages from this repo (`vscode-graphql` etc.) from publishing non-prerelease versions (see [here](https://github.com/graphql/graphiql/pull/2513#discussion_r909434787))

All the changes here have already been reviewed and approved individually:
- #2513
- #2515
- #2540
- #2524
- #2539
- #2543
- #2571
- #2581
- #2582
- #2586
- #2588
- #2592
- #2595
- #2596


acao:
@thomasheyenbrock i had an idea for automating @next releases! This I think this will solve the problem completely.

in the canary script, we can just add conditional behavior for `graphiql-v2` and non-forked branches with `next/` branch, which skips the canary logic.

instead, it attempts a pre-release if a `.changeset/pre.json` is present. However in most cases when pushing commits, if the versions haven't been incremented since the last next release, nothing will happen.

so, when you're ready to publish another series of `@next` tags, just increment the versions where you need them and you're good to go!  You may also want to increment in `dependencies`, though this can feel redundant with ^ resolution, it's more important when it's ready for a stable release to set a working baseline ofc.

for example, you won't need to make another `codemirror-graphql` pre-release again unless it changes, so you will be able to just push a commit that increments the next tag versions for `graphiql` and `@graphiql/react` and others as needed.

it's actually much simpler than the canary logic itself so it should be easier to implement.

thomasheyenbrock:
@acao sounds like it should work. I'm just wondering how many `next` releases we actually need before having v2 ready, so not sure if it's worth the effort now. (At least for me who hasn't touched any of this canary logic it would take me some time to dig into it.)

For future major versions this would definitely be very nice to have!

acao:
@thomasheyenbrock or here rather :)

PR #2531, opened 2022-07-01
clean up repository references, more readmes and links

acao:
the absolute links are for npm and other repo mirrors of course
can't wait to improve our typedoc output and use it in a proper docs website soon!

the gatsby docs site experiment I did about 2 years ago is a good starting point, the branch is somewhere, but there are many approaches we can look at. graphql dot org was being rebuilt with gatsby at the time so I figured so should we. whatever folks are comfortable using is fine


thomasheyenbrock (approved):
Nice cleanup! 🧹

PR #2533, opened 2022-07-03
add more tests for findGraphQLTags, fix cache/schema tests

acao:
another boring incremental improvement PR 😆 

- re-enable some tests for GraphQLCache that were disabled after an update to `graphql-config` after I found a proper mocking workaround
- add more tests for cases in `findGraphQLTags.ts` as a follow up to #2509


acao:
Nice. Almost 4% coverage increase!

Screenshot 2022-07-03 at 16 53 43

PR #2535, opened 2022-07-04
fix error display for websocket fetchers

### LINKED ISSUE #2534: Getting an Error for Invalid Subscription Requests
I'm currently working on integrating GraphQL subscriptions in Ballerina language to a GraphiQL web based client. It uses websockets for subscriptions and works fine with valid subscription requests. 
But for invalid subscription requests, an error occurs like this.
image

The GraphiQL client receives this websocket payload.
```
{"type":"error", "id":"b2dd2712-1ac5-48fa-9f25-8b88d3d80bf0", "payload":{"errors":[{"message":"Cannot query field \"messag\" on type \"MsgType\".", "locations":[{"line":4, "column":5}]}]}}
```

thomasheyenbrock:
Fixes #2534

The issue is caused by the default ws-fetcher which wraps joins multiple error messages and wraps them in a new error.

Open TODOs:
- [x] add changeset
- [x] write some tests for displaying correctly formatted errors in general

### DIFF
--- a/.changeset/thirty-grapes-pay.md
+++ b/.changeset/thirty-grapes-pay.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/toolkit': patch
+---
+
+Fix error formatting for websocket requests and make error formatting more generic in general
--- a/packages/graphiql-react/src/execution.tsx
+++ b/packages/graphiql-react/src/execution.tsx
@@ -302,23 +302,17 @@ export function ExecutionContextProvider(props: ExecutionContextProviderProps) {
             }),
           );
         } else if (isAsyncIterable(value)) {
-          (async () => {
-            try {
-              for await (const result of value) {
-                handleResponse(result);
-              }
-              setIsFetching(false);
-              setSubscription(null);
-            } catch (error) {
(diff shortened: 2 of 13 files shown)


acao:
yeah i think this is a longtime issue, maybe not a regression, so feel free to fix it in 2.0.0. will take a look at this later today!

thomasheyenbrock:
We could also fix that with a new patch in v1 as we decided not to do the prereleases and this does not contain breaking changes 😉

acao:
ah yes of course! i was hoping to make it easier on you but that's exactly what we agreed to isn't it, i forgot we have a path for this now! so nice to have, looking forward to using pre-release more often

it looks like there is a new cypress failure here, are you familar with this one?

https://github.com/graphql/graphiql/runs/7176376103?check_suite_focus=true#step:8:412

acao:
Looks good! I approve just need to fix the e2e suite


Screenshot 2022-07-13 at 16 01 47

acao (approved):

acao:
oh my bad i see the write tests task

PR #2540, opened 2022-07-04
[redesign] footer

thomasheyenbrock:
Quick one: This fixes the layout for the custom footer that is displayed beneath the response editor. As there is none in the preview, here's the visual diff:

| Before | After |
| --- | --- |
| CleanShot 2022-07-04 at 17 55 33@2x | CleanShot 2022-07-04 at 17 55 50@2x |

Note: No changeset here, we already have one for the redesign in #2523


timsuchanek (approved):
LGTM!

PR #2543, opened 2022-07-06
[redesign] implement sidebar

thomasheyenbrock:
Part of #2325 

This adds the new sidebar on the left side that contains the toggle buttons for docs and history, as well as (currently non-functional) buttons for showing the short-keys and the settings modal. It also moves over the doc explorer to the right side.

💡 Reviewing with hidden whitespaces reduces the changed LOCs significantly!

Preview is here: https://deploy-preview-2543--graphiql-test.netlify.app/

Note: No changeset here as we already have one for the redesign in #2523.


acao (approved):
Well this looks fantastic!

Is there a plan to move some more of these components into small component files in `graphiql-react` perhaps, as well as some implementation specific components in `graphiql` as needed? Just seeing `graphiql.tsx` grow and grow makes me nervous for contributors haha! 

There is still a plan for adding the component library and storybook/styleguidest/ladle/or what have you for some baseline, css-var and class selector? theme-able components, yes?

update: just re-checked and wow, this codebase changes quickly! i see there are many components in `@graphiql/react`, not just context, hooks and icons.  So you'll probably be eventually moving these to separate component files in `@graphiql/react` then great!

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -826,98 +843,115 @@ class GraphiQLWithContext extends React.Component<
...
+                              'second'
+                            ) {
+                              this.props.editorToolsResize.setHiddenElement(
+                                null,
+                              );
+                            } else {
+                              this.props.editorToolsResize.setHiddenElement(
+                                'second',
+                              );
+                            }
```
B2o5T, line 888:
```suggestion
                            this.props.editorToolsResize.setHiddenElement(
                              this.props.editorToolsResize.hiddenElement ===
                              'second' ? null : 'second',
                            );
```

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -826,98 +843,115 @@ class GraphiQLWithContext extends React.Component<
...
-                  />
-                  {footer}
+                <div ref={this.props.editorResize.dragBarRef}>
+                  <div className="graphiql-horizontal-drag-bar" />
+                </div>
+                <div ref={this.props.editorResize.secondRef}>
+                  <div className="graphiql-response">
+                    {this.props.executionContext.isFetching ? (
+                      <div className="graphiql-spinner" />
+                    ) : null}
```
B2o5T, line 941:
```suggestion
                    {this.props.executionContext.isFetching && (
                      <div className="graphiql-spinner" />
                    )}
```

B2o5T:
Looks nice!
Not related to your PR but does there is some plan to look better GraphiQL on mobile? 
image

I guess it could be awesome to change `flex-direction` on some breakpoints for `graphiql-query-editor`

B2o5T (approved):

hebertcisco (approved):

acao (approved):

thomasheyenbrock:
@acao definitely worth exploring a storybook-like component library in the future. I'm focussing on getting the redesign out first (basically `graphiql@2`), after that we can also do more iterations on `@graphiql/react` that are non-breaking anymore for the `graphiql` package. I'd still keep `@graphiql/react` in `0.x` to indicate that this is not yet 100% stable and there might be some more changes down the road.

@B2o5T good point! But also here I'd defer this to a later point in time, just to have better focus on the redesign.

packages/graphiql/src/components/GraphiQL.tsx
thomasheyenbrock: This seems to me like a clash of personal preferences, because I like ternaries better 😄 (for the simple reason that it avoids the pitfall of including `0` in the markup when the variable you base the assertion on is a number)

I'm open to being more opinionated here (no matter which one we choose), but if we think that this brings us value we should imho look into validating this using `eslint` and uniformly adapt the whole codebase. So I'm just leaving this as is for now. ❤️

PR #2545, opened 2022-07-06
avoid top-level dynamic codemirror imports

### LINKED ISSUE #2368: ReferenceError: navigator is not defined (remix react)
Hi There,
I am trying to use GraphiQL with [Hansanghyeon/error-remix-graphiql](https://github.com/Hansanghyeon/error-remix-graphiql) but I am getting following error:

There was a similar issue. #118

```
ReferenceError: navigator is not defined
    at /Users/imweb/@lab/blog-tutorial/node_modules/codemirror/lib/codemirror.js:18:19
    at /Users/imweb/@lab/blog-tutorial/node_modules/codemirror/lib/codemirror.js:11:83
    at Object.<anonymous> (/Users/imweb/@lab/blog-tutorial/node_modules/codemirror/lib/codemirror.js:14:2)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at /Users/imweb/@lab/blog-tutorial/node_modules/codemirror/addon/hint/show-hint.js:8:9
    at Object.<anonymous> (/Users/imweb/@lab/blog-tutorial/node_modules/codemirror/addon/hint/show-hint.js:13:3)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
```

thomasheyenbrock:
Fixes #2368 

This should fix breaking imports of `@graphiql/react` (or its dependent `graphiql`) in non-browser environments like Node.js. Remix for example breaks during SSR because these imports also get loaded and executed. `codemirror` assumes to always run in a browser env though.

### DIFF
--- a/packages/graphiql-react/src/editor/common.ts
+++ b/packages/graphiql-react/src/editor/common.ts
@@ -17,20 +17,6 @@ export const commonKeys = {
   'Alt-Right': 'goGroupRight',
 };
 
-export const commonCodeMirrorAddons = [
-  import('codemirror/addon/hint/show-hint'),
-  import('codemirror/addon/edit/matchbrackets'),
-  import('codemirror/addon/edit/closebrackets'),
-  import('codemirror/addon/fold/brace-fold'),
-  import('codemirror/addon/fold/foldgutter'),
-  import('codemirror/addon/lint/lint'),
-  import('codemirror/addon/search/searchcursor'),
-  import('codemirror/addon/search/jump-to-line'),
-  import('codemirror/addon/dialog/dialog'),
-  // @ts-expect-error
-  import('codemirror/keymap/sublime'),
-];
-
 /**
  * Dynamically import codemirror and dependencies
  * This works for codemirror 5, not sure if the same imports work for 6
@@ -47,7 +33,20 @@ export async function importCodeMirror(
   const allAddons =
     options?.useCommonAddons === false
       ? addons
-      : commonCodeMirrorAddons.concat(addons);
+      : [
+          import('codemirror/addon/hint/show-hint'),
+          import('codemirror/addon/edit/matchbrackets'),
+          import('codemirror/addon/edit/closebrackets'),
+          import('codemirror/addon/fold/brace-fold'),
+          import('codemirror/addon/fold/foldgutter'),
+          import('codemirror/addon/lint/lint'),
+          import('codemirror/addon/search/searchcursor'),
+          import('codemirror/addon/search/jump-to-line'),
+          import('codemirror/addon/dialog/dialog'),
+          // @ts-expect-error
+          import('codemirror/keymap/sublime'),
+          ...addons,
+        ];
   await Promise.all(allAddons.map(addon => addon));
   return CodeMirror;
 }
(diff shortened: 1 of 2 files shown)


acao:
Why are they resolving unresolved promises though? The import should only happen on `await import()` right?

packages/graphiql-react/src/editor/common.ts
```diff
@@ -47,7 +33,20 @@ export async function importCodeMirror(
...
+          import('codemirror/addon/fold/foldgutter'),
+          import('codemirror/addon/lint/lint'),
+          import('codemirror/addon/search/searchcursor'),
+          import('codemirror/addon/search/jump-to-line'),
+          import('codemirror/addon/dialog/dialog'),
+          // @ts-expect-error
+          import('codemirror/keymap/sublime'),
+          ...addons,
+        ];
   await Promise.all(allAddons.map(addon => addon));
```
B2o5T, line 50: Not related to your PR, but why we need it? I guess it should be
```suggestion
  await Promise.all(allAddons);
```

acao (approved):
Hopefully this resolves the issue!

packages/graphiql-react/src/editor/common.ts
acao: Yeah this was my original implementation! I wonder what happened?

B2o5T: In this place it makes no sens to do it 🙃

thomasheyenbrock:
This has in fact nothing to do with bundlers after thinking about it again.

In the moment you call `import()` Node.js will dispatch the promise and push it on the call stack. On the next available tick `import` will load the module, thus executing its content.

As an example, you could do the following in Node.js:
- Have a file `dynamic.mjs` like this
  ```js
  console.log("i'm running");
  ```
- Have a file `index.mjs` like this:
  ```js
  function wait(ms) {
    return new Promise((r) => setTimeout(r, ms));
  }

  import("./dynamic.mjs");

  wait(2000).then(() => console.log("done"));
  ```

What you'll see printed is first `i'm running` and then `done`, no matter if you await the dynamic import or not.

packages/graphiql-react/src/editor/common.ts
thomasheyenbrock: Oops, now I forgot to include this 🙈

acao:
Indeed! I had a similar realization in the original issue comments. Thus why `React.lazy(() => import())` works for top level declarations, because of the function closure!

packages/graphiql-react/src/editor/common.ts
thomasheyenbrock: See #2559

PR #2553, opened 2022-07-08
fix the lsp cli for vim users

acao:
Re-creating the PR created by user @xuanduc987


acao:
@xuanduc987  try `npm install -g graphql-language-service-cli@3.2.30-canary-e992d7a6.0` and changing the install path for `graphql-lsp` or whatever works? i don't know how `nvim.coc` works beyond the default config

PR #2561, opened 2022-07-11
add missing effect deps for `keyMap` prop

thomasheyenbrock:
Follow-up on #2501, we forgot to add the `keyMap` prop as effect dep.

### DIFF
--- a/.changeset/wild-wasps-retire.md
+++ b/.changeset/wild-wasps-retire.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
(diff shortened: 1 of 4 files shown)


acao:
Moving forward, on config changes we should probably just use this `setOption` method instead of re-creating the editor instance!
https://codemirror.net/5/doc/manual.html#api_configuration

thomasheyenbrock:
@acao good shout! Refactored this in https://github.com/graphql/graphiql/pull/2563

PR #2567, opened 2022-07-13
[redesign] implement new design for history plugin

thomasheyenbrock:
Part of #2325 

Notes:
- I added two more changesets I forgot in previous PRs
- The functionality changed a bit, namely the input doesn't disappear on blur anymore (since we now have a `Save` button next to it). Instead you have now two ways of exiting input: Clicking either button to the right (save or close without saving) or pressing `escape` while focussing the input.
- I took the creative freedom to add a separate close button (which was not in the [design](https://www.figma.com/file/p8f8MaG7ZRnuCc4lVv7w6t/GraphiQL-2.0?node-id=2004%3A19365)), otherwise there'd be no visible way how to exit edit-mode without saving the changes.
- The one thing from the design that I didn't implement is the highlighted history item that says `now` next to it. There are multiple reasons for why this needs more consideration (we can iterate over this feature later if we want to):
  - We don't store timestamps with history items, so `now` might not be accurate
  - In general, the current query might be totally different compared to the last history item
  - Highlighting an item based on if the current query matches the historic one could lead to multiple highlighted items as one query might appear more than once in the history.

Preview is here: https://deploy-preview-2567--graphiql-test.netlify.app/


packages/graphiql/src/cdn.ts
```diff
@@ -16,7 +16,6 @@ import './style.css';
 // Legacy styles
 import './css/app.css';
 import './css/doc-explorer.css';
-import './css/history.css';
```
acao, line 19: nice

acao (approved):
beautiful! I love where all of this is going. your consistency and naming conventions are refreshing

PR #2569, opened 2022-07-13
docs: removes redundant copy re: EasyCLA in readme

jonathanawesome:
This PR removes a bit of redundant copy in the readme around EasyCLA in the `Contributing` section. Also, should get me an EasyCLA prompt.

Two birds, one pull request.


acao:
great! yes, looks like we're all set up to merge your first contribution then! :)

PR #2573, opened 2022-07-16
move vscode grammars, basic language support to standalone syntax extension

acao:
`vscode-graphql` now depends on this extension, and provides language server support in addition. this allows alternative LSP servers and non-LSP graphql extensions to use our highlighting only, or allows users to just install the syntax highlighting extension if they don't need LSP server features.


acao:
CC @captbaritone @orta this is the PR for creating the syntax only extension by the way!

I think I can finish setting this up today. i think we will use the namespace `graphql.vscode-graphql-syntax` because the naming convention then reflects that it is part of this project.

acao:
Just thought that the @graphql/tsc might like to know this really exciting change happened today, as well as this related one for relay:

https://github.com/facebook/relay/pull/4032

PR #2574, opened 2022-07-18
feat(graphiql): pass introspection data to `schema` prop

thomasheyenbrock:
This is a small QoL improvement: The `schema` prop now also accepts introspection data as an alternative to a `GraphQLSchema` instance. Also, this fixes a bug where the schema was not updated after rerunning introspection.

We had a use-case for this feature at Stellate: We store introspection data for different endpoints, and we want to use this to render GraphiQL on the client without running introspection (because this is likely to be disabled for lots of production APIs). We'd now have to add `graphql` as first-class dependency just to call `buildClientSchema` once. With this tiny change the `SchemaContextProvider` does that for us, which is a lot nicer DX.

💡 Reviewing with whitespace turned off halves the LOCs!

### DIFF
--- a/.changeset/nine-elephants-retire.md
+++ b/.changeset/nine-elephants-retire.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': minor
+---
+
+Allow passing introspection data to the `schema` prop of the `SchemaContextProvider` component
--- a/.changeset/nine-eyes-relax.md
+++ b/.changeset/nine-eyes-relax.md
@@ -1,0 +1,5 @@
+---
+'graphiql': minor
+---
+
+Allow passing introspection data to the `schema` prop of the `GraphiQL` component
--- a/.changeset/serious-ties-compete.md
+++ b/.changeset/serious-ties-compete.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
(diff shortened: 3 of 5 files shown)


acao:
I’m all for it! What do you think of `introdpectionSchema` as a seperate prop to provide it, would that be more complicated than using the same `schema` prop and just detecting it each time?

thomasheyenbrock:
I thought about this but decided against it for two reasons:
- When having two props you could pass two different schemas and we'd have to either explicitly throw or implicitly take one over the other.
- I actually like the naming also for this new case. The introspection data is still describing the schema, it is essentially just a different representation of the same thing.

connorshea:
Do you have any example code of how this is actually used? I can't quite figure that part out 😅 

I _think_ this'd allow me to use a static JSON representation of my GraphQL schema to have GraphiQL load automatically, but I'm not 100% sure how to provide the GraphiQL component with the right object.

thomasheyenbrock:
Hey @connorshea 👋 the object you can pass here is the data returned from a "standard" introspection query (you can check the `getIntrospectionQuery` method from `graphql-js` for how that query looks). The object looks something like `{ "__schema": { /* schema introspection data */ } }`. Does that help with your question?

connorshea:
yup, that helps! thanks

PR #2582, opened 2022-07-22
[redesign] move docs components to `@graphiql/react`

thomasheyenbrock:
Part of #2325

This PR moves over the "primitive" components that make up the doc explorer:
- type and field links
- default value
- arguments
- directives
- markdown content

Some more notes:
- I didn't work out which parts of the CSS for the docs are now obsolete. Instead we'll remove all the legacy styles in the next PR.
- I combined a lot of changesets related to new components added to `@graphqil/react` into one to have a better overview.


acao:
this is looking lovely! i think the changeset automation/canary release behavior expects a *new* changest file unfortauntely, so I think you can just delete all of these and consolidate them into a _new_ file

acao:
some slight spacing nits -

is there supposed to be this gap on the border? i need to look at the figma

Screenshot 2022-07-24 at 15 48 14

Screenshot 2022-07-24 at 15 49 25

packages/graphiql-react/src/markdown.ts
```diff
@@ -1,3 +1,6 @@
 import MarkdownIt from 'markdown-it';
```
acao, line 1: I hate to be annoying about tiny things but this would be a tiny but great addition to `@graphiql/toolkit`. perhaps also consider exporting our default config so that folks can create their own instance with custom additional config without adding to the singleton (I've had users talk about wanting to use custom markdown-it plugins for internal docs links/etc!)

An alternative to the singleton could even be customization props for the `MarkdownContent` component and even `GraphiQL` component for the `markdown-it` instantiation.

packages/graphiql-react/src/ui/markdown.css
```diff
@@ -6,6 +6,8 @@
  *   (field description any optionally deprecation reason).
```
acao, line 6: ok just leaving an idea here, could be done in a different PR - what do you think of CSS, design primitives and other non-react-specific elements going in the `@graphiql/toolkit` directory?

thomasheyenbrock:
@acao the new docs design is not fully implemented yet 😅 there will be likely two more PRs for this to be done, I just splitted the changes because it would have been too much otherwise.

The comments about `@graphiql/toolkit` are great but out of scope here. I'll open issues to track these and so that we don't forget 👍

packages/graphiql-react/src/markdown.ts
thomasheyenbrock: Opened https://github.com/graphql/graphiql/issues/2584 to track this

packages/graphiql-react/src/ui/markdown.css
thomasheyenbrock: Opened a discussion here to align on this (I don't think this is actionable yet): https://github.com/graphql/graphiql/discussions/2585

thomasheyenbrock:
@acao regarding changesets, we're not merging this PR into `main` but into `graphiql-v2`, so all changesets will accumulate there before we're then actually merging this branch into `main`.

acao:
Ok it all looks good! Sorry i forgot about that part haha

acao (approved):

PR #2586, opened 2022-07-25
[redesign] implement new search combobox

thomasheyenbrock:
Part of #2325 

This PR adds a `Search` component to `@graphiql/react` that allows searching through schema items. There is still some final work left to do for the docs, so the UI you'll see in the preview deployment is not the final one! Expect another PR that finalizes the docs design.

To help implement the combobox and assure that it aligns with all a11y requirements I used `@reach/combobox`. It seemed pretty light-weight while also being very flexible and customizable. I'm basically proposing this now for the first iteration, but I'm not too opinionated on this choice. It seems good enough at this point and unless there are major blockers I'd recommend taking it. We can always follow up on minor points and maybe switch out the implementation behind the scenes after releasing v2.


acao:
beautiful!

packages/graphiql-react/src/explorer/components/search.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+type TypeMatch = { type: GraphQLNamedType };
+
+type FieldMatch = {
+  type: GraphQLNamedType;
+  field: GraphQLField<unknown, unknown> | GraphQLInputField;
+  argument?: GraphQLArgument;
+};
+
+export function useSearchResults(caller?: Function) {
```
acao, line 176: oh wow, this is really clever! excellent for both the SDK and graphiql

thomasheyenbrock:
Extending the search logic should be easy enough to do later on:
- The `useSearchResults` hook contains the logic of what items to return from a given schema instance
- The `Search` component takes the results and renders it

acao:
@thomasheyenbrock this is amazing! exactly what I hoped for, and better than I could've done!

acao (approved):

PR #2587, opened 2022-07-26
move remaining doc explorer component to `@graphiql/react`

thomasheyenbrock:
An intermediate PR that moves over the remaining doc explorer components to `@graphiql/react` *without* adjusting them. This way we can hopefully preserve the git history. (When moving *and* editing files at the same time, git seems to not always pick up that the file has been moved.)


acao:
@thomasheyenbrock indeed it's trucky! if more than 90% of lines changed it usually can't read an index change. this is a great solution

you can also do this in one PR - make one commit that moves them and then another that changes them, but then for that to work you must use rebase & merge instead of squash

acao (approved):

PR #2590, opened 2022-07-26
[housekeeping] remove unused shared dependencies

thomasheyenbrock:
Did a quick run through, and this is all stuff that we don't use anywhere for sure.

@acao not sure why we had the jest environment packages installed explicitly, they seem to already come via transitive dependencies of `jest` itself.


acao:
Looks good to me!

The context rfc example is nor (or shouldn't be!) included in the workspace so it won't resolve those packages on yarn install anyways

acao (approved):

PR #2592, opened 2022-07-27
[redesign] implement new design for tabs

thomasheyenbrock:
Part of #2325

This PR implements the new design for tabs. With this, we can remove the last of the legacy stylesheets, i.e. from this point on all elements are in the new design 💅


packages/graphiql-react/src/ui/tabs.tsx
```diff
@@ -1,67 +1,62 @@
...
-
-/**
- * Generic tab component that implements wai-aria tab spec
- */
-export function Tab(props: TabProps): React.ReactElement {
+function TabButton(props: JSX.IntrinsicElements['button']) {
   return (
-    <button
-      {...props.tabProps}
-      role="tab"
```
acao, line 33: oops! let's keep this. I don't know if this was on purpose, but anything that is a tab should have this role for the other aria props to work!
Screenshot 2022-07-27 at 19 44 38

otherwise it appears everything is in order with the aria props, just this one is missing. it seems the tab components have been re-structured, but as you probably know, it should be used for the tab elements anywhere that tabpanel and tablist are used

thomasheyenbrock: Very good catch! That should have moved to the `Tab` component, but seems like I messed that up.

PR #2593, opened 2022-07-27
[housekeeping] fix install issue by cleaning up jest dependencies

thomasheyenbrock:
This should fix the issue that where you get an error trying to add a dependency to any individual project. Also, it cleans up a lot of stuff in our lockfile 💆


acao (approved):
You are fantastic @thomasheyenbrock !!

PR #2595, opened 2022-07-27
[redesign] add settings dialog

thomasheyenbrock:
Part of #2325

This adds the settings dialog. The only option for now is to clear the storage, theme selection will be added in the next PR.

### DIFF
--- a/.changeset/five-pillows-fail.md
+++ b/.changeset/five-pillows-fail.md
@@ -3,7 +3,7 @@
 ---
 
 Add new components:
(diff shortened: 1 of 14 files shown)


acao:
this is so exciting!

acao:
this looks great so far, and the clear data toggle works, and is a fantastic start.


found this though, is this a known issue?
Screenshot 2022-07-27 at 19 27 55

thomasheyenbrock:
Good catch! CodeMirror has some `z-index` set on the scrollbar containers which caused them to be above the dialog, should be fixed now.

acao (approved):
💯 ⚙️ 🤓 🚀

PR #2601, opened 2022-07-29
add a dialog that shows all available shortkeys

thomasheyenbrock:
This adds a dialog that shows all the current short keys.

CleanShot 2022-07-29 at 16 20 55@2x

### DIFF
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -1023,46 +1023,70 @@ class GraphiQLWithContext extends React.Component<
             />
           </div>
           <div className="graphiql-dialog-section">
-            <ul className="graphiql-short-keys">
-              <li>
-                {modifier}
-                {' + '}
-                <code className="graphiql-key">F</code> : Search in editor
-              </li>
-              <li>
-                {modifier}
-                {' + '}
-                <code className="graphiql-key">K</code> : Search in
-                documentation
-              </li>
-              <li>
-                {modifier}
-                {' + '}
-                <code className="graphiql-key">Enter</code> : Execute query
-              </li>
-              <li>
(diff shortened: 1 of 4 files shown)


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -997,6 +999,62 @@ class GraphiQLWithContext extends React.Component<
...
+          <div className="graphiql-dialog-header">
+            <div className="graphiql-dialog-title">Short Keys</div>
+            <Dialog.Close
+              onClick={() => {
+                this.setState({ showShortKeys: false });
+              }}
+            />
+          </div>
+          <div className="graphiql-dialog-section">
+            <ul className="graphiql-short-keys">
```
acao, line 1016: Do you plan to move this to JSON in later iterations?

acao, line 1016: I'm guessing you plan on moving this to JSON eventually?

Also, you could add a little note on this dialog if you want that contains a little link to these docs, keyed by `prop.keyMap`:
https://codemirror.net/5/demo/sublime.html
https://codemirror.net/5/demo/vim.html

also, just noticed i was completely wrong about that vim PR! this does have insert mode already, i forgot, this should be good to say we already have vim support. i will correct myself there.

acao (approved):
Looks great to me! And the design system appears to be working out wonderfully here! Need another next release? I'm going to suggest an improvement as a new issue to not slow you down!

acao:
@thomasheyenbrock another thought for later perhaps, what if this were an html table or used table cell layout or similar to achieve alignment of the right-hand entries?

thomasheyenbrock:
Nice suggestions @acao! Here's the second iteration:
CleanShot 2022-08-01 at 19 08 52@2x

PR #2608, opened 2022-07-31
Initialize the project cache more selectively

acao:
Massive optimization for multi-project users!

Warning: this changes the cache lifecycle a lot, especially after config change. Might need some fine tuning 😆
I should add lots of tests actually.

### Changes

`WorkspaceMessageProcessor._updateGraphQLConfig()` becomes `_loadAndCacheProjectConfig(projectName: string)`, which has grown up and become more discerning and instead of re-building all project config caches, we just load the project config it needs to handle that event, of course! What a poor design of mine before.

Because of this, the cache invalidation strategy on configuration and settings changes must change, and also allow project caches to lazily load as needed.

### TODO:
- [x] as part of this, we must get rid of `this._isInitialized()`, should be detected per-project. otherwise one project cache that had an error initializing (we perform parsing and load files, etc) will prevent all other projects from having a cache or language features.
- [x] figure out what to do on `MessageProcessor.handleDidChangeConfiguration` - perhaps re-create the cache reset behaviour using a public method for `WorkspaceMessageProcessor`?
- [x] will building the project cache lazily on editor open or save events (not workspace) make working with the language? perhaps we need to add per-project cache initialization to onChange as well. again, this fires only once per project config per server init or config file or settings change events. 
- [ ] make `package.json` change events ignored more quickly via a quick parse-and-check for the top-level `graphql` namespace.
- [ ] on workspace file changes, gently update just the cache for that file (maybe seperate PR)

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -216,12 +216,19 @@ export class MessageProcessor {
   }
 
   async handleDidChangeConfiguration(
-    _params: DidChangeConfigurationParams,
+    params: DidChangeConfigurationParams,
   ): Promise<DidChangeConfigurationRegistrationOptions> {
+    if (!params?.settings || params?.settings.length === 0) {
+      return {};
+    }
+
+    // reset all the workspace caches
+    // and prepare for them to lazily re-build based
+    // on where the user opens and saves files
     await Promise.all(
-      Array.from(this._processors.values()).map(processor =>
-        processor._updateGraphQLConfig(),
-      ),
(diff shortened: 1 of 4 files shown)


acao:
actually, I think i have to get rid of `this.isInitialized()` entirely!

Foo-x:
This looks great and the TODO is very clear!

acao:
@Foo-x getting back to it today!

acao:
ok! now it's ready for review 🥳 

now that we (re) initialize the project caches more selectively in this PR, i much feel better about multi-workspace support
we even ensure the project is cached on workspace changes, and on change, not just on did open or save

PR #2620, opened 2022-08-01
feat: added full codemirror6 graphql lang support

imolorhe:
This adds full graphql support for codemirror 6 with custom lezer grammar + parser for graphql based on the [October 2021 graphql spec](https://spec.graphql.org/October2021/).

Features in codemirror 5 reimplemented for codemirror 6:

- autocomplete (hint)
- lint
- jump (cmd-click)
- custom info render

There are still some kinks that would need to be worked on to make the experience as smooth as codemirror 5, e.g. in the auto indentation.


acao:
@imolorhe looking forward to reviewing this! the ci failure is just a prettier fix, I can get it

thomasheyenbrock:
@imolorhe this is awesome! 🚀 

I just want to come back to the question that we started discussing in Discord: Do we want to create a new package for this or replace the current `codemirror-graphql` with this new codebase and publish it under a new major version?

My thoughts on this:
- **Maintainance overhead**: This comes down to if we want to keep publishing changes for CM5. If yes, then a new package makes our lives significantly easier for us. If no, then having a new package feels like just adding additional "load" on our monorepo (in a sense that `codemirror-graphql` would basically become unmaintained).
- **Relating this to GraphiQL v2**: I'd prefer publishing GraphiQL v2 with CM5, just to keep the breaking changes to a minimum. I think there's no shame in publishing another major in quick succession if we wanna switch to CM6 afterwards. This doesn't really work when we would publish these changes with a new major of `codemirror-graphql` already now. An alternative could be to keep this PR open until after we released GraphiQL v2.

Would be very interested to hear your thoughts on this as well!

imolorhe:
I think we should keep publishing changes for CM5, at least in the near future. CM5 and CM6 have vastly different APIs, and requiring folks to have to migrate from CM5 to CM6 will be a challenging requirement. It will be better to consider them as two different editors the same way codemirror is different from monaco in this case.

Regarding GraphiQL v2, depending on how the interop is built will determine how difficult it will be to switch out CM5 for CM6. Considering @acao's comment to be able to potentially swap out the editor in the context. i.e. `<EditorContext editor="monaco" />` or `<EditorContext editor="cm5" />` or `<EditorContext editor="cm6" />`.

I agree though that providing support for a new editor in general _is_ an additional maintenance overhead, so it is a call that we have to make if we want to provide support for.

We could also decide not to handle the CM6 support in this repo, and leave it for a separate repo (e.g. `@codemirror/lang-graphql` which would follow the pattern for other CM6 language mode support), but the advantage of having it here is being able to manage it more closely in this repo along with the other graphql JS related packages.

acao:
I agree with @imolorhe here, it makes sense as as a separate package. we rarely have issues with `codemirror-graphql` itself as it were, and we can say that the codemirror 5 implementation will be maintained for bugfixes, but may not be kept up to date for spec compliance if we end up using cm6 for graphiql 2. 

People will end up contributing backwards compliant support eventually, but we won't take the initiative to keep both a codemirror 5 and 6 implementation up to graphql spec compliance, and focus on the latter.

Having both projects in the repo doesn't mean `graphiql@react` has to provide support for both simultaneously however, it is definitely not necessary to have reference implementation hooks APIs in react for both versions simultaneously! That is `graphiql@react'`s call to make, but I would assume that the cm5 hooks implementation would just be replaced, and whatever breaking props changes as a result for 2.x

acao:
a few other things for this PR:

- we probably should not revert to mocha for this PR if we can avoid it
- rollup is fine but may be replaced with esbuild or vite shortly when we replace all the build tooling
- ideally you would add this to the typescript project references trees in `resources/tsconfig.build.<module format>.json`, so that it can be developed alongside the other typescript packages using the monorepo-wide typescript project references config. it looks like this config was broke for `codemirror-graphql` recently as well, and so I guess ignore this until I can fix the build tooling which seems to circumvent project references throughout the repo

imolorhe:
@acao @thomasheyenbrock Updated the PR according to the suggestions

examples/cm6-graphql-parcel/yarn.lock
```diff
@@ -0,0 +1,2329 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
acao, line 1: These should be git ignored, yes?

resources/tsconfig.build.cjs.json
```diff
@@ -14,6 +14,9 @@
     {
       "path": "../packages/codemirror-graphql"
     },
+    {
```
acao, line 17: Make sure to add to build.esm.json as well, and then we can remove rollup yeah?

acao (commented):
Just a few more thingd!

examples/cm6-graphql-parcel/yarn.lock
webberwang: `yarn.lock` should be included in git for version control

acao: We already have yarn.lock at the root for all package workspaces. This is an example. We don’t want the example lockfiles because they will trigger things like dependabot or security alerts that are not applicable

imolorhe: Yeah I'll remove that.

resources/tsconfig.build.cjs.json
imolorhe: I've added an entry. However I'm not sure we can remove rollup just yet. It depends on a custom [codemirror build helper](https://github.com/codemirror/buildhelper) which uses rollup

jest.config.base.js
```diff
@@ -20,6 +20,7 @@ module.exports = (dir, env = 'jsdom') => {
       '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
         'identity-obj-proxy',
       '\\.(css|less)$': 'identity-obj-proxy',
+      '^graphql-language-service-([^/]+)': `${__dirname}/packages/graphql-language-service/src/$1`,
```
imolorhe, line 23: This is required for proper resolution of the `graphql-language-service-*` packages since the other one is only valid for `graphql-language-service` and `graphql-language-service-cli` cases

packages/graphiql-react/src/custom.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '*.svg' {
```
imolorhe, line 1: Not sure how the build was passing previously without this, but this provides typings for imported SCG files so typescript doesn't complain (even though typescript doesn't compile this, we use `tsc --emitDeclarationOnly` so which fails since typescript doesn't understand how to type svg files)

resources/tsconfig.build.cjs.json
acao: Well if you want the cross repo tsc —build and tsc —build —watch to work, you just need these and the matching tsconfig files which just extend the base files. Is the rollup required to bundle the plugin or all of codemirror? The bundle treats codemirror as an external yes

examples/graphiql-explorer/src/vite-env.d.ts
```diff
@@ -1 +1 @@
-/// <reference types="vite/client" />
+// / <reference types="vite/client" />
```
thomasheyenbrock, line 1: Somehow prettier or our setup of it is flawed and always makes this change, please revert that

thomasheyenbrock (changes requested):

packages/graphiql-react/src/custom.d.ts
thomasheyenbrock: This worked before because we load the `vite/client` types. After reverting https://github.com/graphql/graphiql/pull/2620#discussion_r957142262 this should no longer be necessary

packages/graphiql-react/src/vite-env.d.ts
```diff
@@ -1 +1 @@
-/// <reference types="vite/client" />
+// / <reference types="vite/client" />
```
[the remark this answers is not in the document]
imolorhe: I can't seem to make that change in the CLI since the prettier runs before the commit, so making it here instead

examples/graphiql-explorer/src/vite-env.d.ts
imolorhe:
```suggestion
/// <reference types="vite/client" />
```

resources/tsconfig.build.cjs.json
imolorhe: Looking at the source code of the build helper, it depends on rollup. We could potentially have a custom build script without depending on the codemirror build helper and remove the need for rollup, but this is the path of least resistance.

imolorhe:
@acao @thomasheyenbrock not sure why the license check is failing. It passes locally for me though

acao:
@imolorhe it's just that a new package of ours is missing the `license` field, I will take care of this in a seperate PR and you can rebase

Screenshot 2022-09-03 at 12 41 33

acao:
@imolorhe i took care of this the other day! everything looks good to go! thanks again for all of ths

packages/cm6-graphql/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@codemirror/lint": "6.0.0",
+    "@codemirror/state": "6.1.0",
+    "@codemirror/view": "6.1.2",
+    "@lezer/common": "^1.0.0",
+    "@lezer/generator": "^1.1.0",
+    "@lezer/highlight": "^1.0.0",
+    "@lezer/lr": "^1.1.0",
+    "esbuild": "^0.14.47",
+    "graphql": "^16.5.0",
+    "graphql-language-service-interface": "2.8.4",
```
acao, line 29: @imolorhe these are retired and outdated, we now export all of these from `graphql-language-service`.

packages/cm6-graphql/src/completions.ts
```diff
@@ -0,0 +1,58 @@
+import { Completion, CompletionContext } from '@codemirror/autocomplete';
+import { getAutocompleteSuggestions } from 'graphql-language-service-interface';
```
acao, line 2:
```suggestion
import { getAutocompleteSuggestions } from 'graphql-language-service';
```

packages/cm6-graphql/src/interfaces.ts
```diff
@@ -0,0 +1,21 @@
+import { Completion, CompletionContext } from '@codemirror/autocomplete';
+import { EditorView } from '@codemirror/view';
+import { GraphQLSchema } from 'graphql';
+import { ContextToken } from 'graphql-language-service-parser';
```
acao, line 4:
```suggestion
import { ContextToken } from 'graphql-language-service';
```

packages/cm6-graphql/src/interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+import { CompletionItem } from 'graphql-language-service-types';
```
acao, line 5:
```suggestion
import { CompletionItem } from 'graphql-language-service';
```

packages/cm6-graphql/src/jump.ts
```diff
@@ -0,0 +1,31 @@
+import { EditorView } from '@codemirror/view';
+import {
+  getTokenAtPosition,
+  getTypeInfo,
+} from 'graphql-language-service-interface';
```
acao, line 5:
```suggestion
} from 'graphql-language-service';
```

packages/cm6-graphql/src/lint.ts
```diff
@@ -0,0 +1,37 @@
+import { Diagnostic, linter } from '@codemirror/lint';
+import { getDiagnostics } from 'graphql-language-service-interface';
```
acao, line 2:
```suggestion
import { getDiagnostics } from 'graphql-language-service';
```

acao (changes requested):
Just need to move to the new `graphql-language-service` module to fix typescript & yarn resolution errors in CI. this is bringing in an outdated version of dependencies from last year in which is causing a few issues with the workflows

imolorhe:
@acao I've updated the graphql-language-service dependencies

acao (approved):

acao:
@imolorhe it's looking fantastic! thanks for all your hard work on this much needed language mode

acao:
awesome work @imolorhe! further improvements can happen iteratively, this seems good to release

thomasheyenbrock:
> @thomasheyenbrock are you ok with this one?

Sorry, being super late to the party here 🙈 I'm totally good with this, that's a solid start and we can go from here!

PR #2621, opened 2022-08-01
adds styles to contain the result window above any footer content

jonathanawesome:
A simple, initial PR to ensure that I'm branching/PRing correctly for v2 work.

I branched off of `graphiql-v2` for this fix and would like to merge this commit into `graphiql-v2`. Is this the correct process?

### DIFF
--- a/packages/graphiql/src/style.css
+++ b/packages/graphiql/src/style.css
@@ -182,7 +182,7 @@ button.graphiql-tab-add > svg {
 /* The results editor wrapping container */
 .graphiql-container .graphiql-response .result-window {
   position: relative;
-  flex: 1 1 100%;
+  flex: 1;
 }
 
 /* The footer below the response view */
--- a/packages/graphiql/src/style.css
+++ b/packages/graphiql/src/style.css
@@ -179,6 +179,12 @@ button.graphiql-tab-add > svg {
   position: relative;
 }
 
+/* The results editor wrapping container */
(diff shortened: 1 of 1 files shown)


packages/graphiql/src/style.css
```diff
@@ -179,6 +179,12 @@ button.graphiql-tab-add > svg {
   position: relative;
 }
 
+/* The results editor wrapping container */
+.graphiql-container .graphiql-response .result-window {
+  position: relative;
+  flex: 1 1 100%;
```
thomasheyenbrock, line 185: Could we just get away with `flex: 1` here? If there's no footer this is the only child so it will take up the full height anyways.

thomasheyenbrock (commented):
Nice catch and thanks for fixing this! This worked at some point but seems like I messed that up 😅 

I'd like to continue opening PRs against `graphiql-v2` so that we can review incoming changes in isolation. That would be way harder when doing it in a branch that already has thousands of changed LOCs compared to `main`.

thomasheyenbrock:
Oops, closing was a misclick 🙈 

And FYI, I rebased your branch after I rebased `graphiql-v2` on `main` just now.

packages/graphiql/src/style.css
jonathanawesome: great catch! fixed.

thomasheyenbrock (approved):
🙌

PR #2626, opened 2022-08-02
reduce CO2 usage by running less unneccesary GH actions workflows

acao:
because it's good for the earth 🌍 and the patience of contributors 😆 

i wish you could use `paths-ignore` further down, to skip individual tasks

note:

i decided to upgrade `actions/setup-node` usage as well for similar reasons, it appears to be faster but brings some bugs, so i will finish this up tonight

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - '**.md'
+      - 'examples'
+      - '!examples/monaco-graphql-webpack'
 
 jobs:
   publish-canary:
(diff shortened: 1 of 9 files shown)


acao:
a lower priority fix - how to configure workflows to only re-run when _only_ that workflow file changed? probably by using these same configs

acao:
this makes me want to decouple cypress tests into seperate workflows, as they and vscode-graphql have seperate tests that should only need to be run together if `graphql-language-service` changes which is rare but usually a good thing haha

acao:
another possibility, I wonder if some of these extremely edge-case node and graphql-js version matrix tests could just be run on merge to `main`? that way, a release breaks if it should, but it would be rare. maybe a few times a year at most we have an issue with either throughout the entire monorepo

acao:
so you know what's fun? mocking various fetch polyfill libraries across node versions, where TextDecoder may or may not be present and more excitement

PR #2627, opened 2022-08-02
make dropdown menu accessible

thomasheyenbrock:
Some more polish for v2: This replaces the simply `Dropdown` component with a `Menu` component which ensures that the UI is accessible. Currently the execute button doesn't at all work with keyboard navigation. This even allows us to kick out the custom on-key-down handler as all of this is built-in now.


acao:
@thomasheyenbrock this is awesome and this custom key event handler will open the floodgates for accessibility and keboard navigation in general!

acao (approved):

PR #2628, opened 2022-08-02
forward refs and props for UI components

thomasheyenbrock:
I think this is a best-practice for simple UI component: All "native" props for the element that is being rendered (often times it's the only one) are now passed to that component. We also wrap the components with `forwardRef` so that users can access the DOM elements returned from these components.

### DIFF
--- a/packages/graphiql-react/src/toolbar/button.tsx
+++ b/packages/graphiql-react/src/toolbar/button.tsx
@@ -1,14 +1,18 @@
-import { useState } from 'react';
+import { forwardRef, useState } from 'react';
 
 import { UnStyledButton } from '../ui';
 
 import './button.css';
 
-export function ToolbarButton(props: JSX.IntrinsicElements['button']) {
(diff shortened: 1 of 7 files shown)


acao:
this looks great! i had a question but it was answered here:
https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

so let's make a todo somewhere for 2.x.x docs to note a migration path here (if it is necessary to do so)? the common and usually more advanced usage after `GraphiQL.Logo` is the custom `Toolbar` and/or `ToolbarButton`, so that may be a good case to show where you might need to factor in for `forwardRef`? I'm trying to imagine what else this would enable that wasn't possible before

thomasheyenbrock:
@acao we only add ref forwarding to new components that we added to `@graphiql/react`, so in that sense it's not a breaking change for `graphiql`. It's certainly worth mentioning this though, I'll see to include that in the migration guide that I already started collecting notes for 👍

PR #2635, opened 2022-08-06
DO-NOT-MERGE: Theme colors prop exploration

jonathanawesome:
My experiment towards providing an easy-to-use theme prop for graphiql-v2 has come to a dead end. While it's something I feel we must provide in future revisions of GraphiQL, it's a mountain to climb if we're to get v2 out in a reasonable time frame. 

TLDR; light mode actually works pretty well to accept a base color and automatically generate alpha accent colors (there's no way to restrict poor color choices, but, generally, it works well), but dark mode is a different situation. There's just no way to map the same, single, input color value across dark and light modes. We all know this already, but I took a swing.

Before we consider how to provide theme-ability for GraphiQL we need to fix the existing WCAG contrast issues with the current light and dark modes. If we can launch v2 with _compliant_ light and dark modes, that's huge.

So, how do we get to theme-ability for future versions of GraphiQL?

I think we first have to decide how _much_ theme-ability we're willing to provide. If an implementor wants to change or supplement only the light or dark theme, we might want to allow for that. If they want to change one of the modes and not expose the other as an option, we might want to support that. If they would like to change or supplement both, we need to figure out _how much_ to guide them within the exposed interface for modifying themes (really, this is true for all of aforementioned options).

Second, I think it's critical to have a design file that we can rely on as the source of truth. Both @thomasheyenbrock and I have built interfaces using the same base Figma file and they're wildly different. Neither is better or worse, but we've both interpreted and extended the design file in different ways. If we're to allow the level of extensibility that we'd like in future versions of GraphiQL, the foundation of that documentation should live in the design file.

So, don't merge this PR. Have a look if you're interested (I merged in a load of unrelated commits, so only those that are mine are critical for understanding), then I'm happy for it and the branch to be deleted.

Onward!


thomasheyenbrock:
That's a great start! Some thoughts:
- Renaming the color variables makes soooo much sense! 🤯 Especially because users might use a totally different color for what we internally call `--color-pink`! We should definitely get this into v2.
- I opened #2637 inspired by this. We reduce the set of CSS variables to basically the color tokens that you propose here. In fact, everything still looks the same in the UI after switching to transparent colors! I hope this will pave the way for adding a theme prop without breaking changes, so we can take our time here and don't need to prolong v2. (Didn't rename the color variables in my PR, but again we should definitely do that 😅 )

jonathanawesome:
Glad it kicked off some new work! I'm going to close this and move the conversation to #2637.

PR #2636, opened 2022-08-06
[@graphiql/react] deduplicate props

thomasheyenbrock:
In the midst of some final cleanup for v2 I noticed that after the initial refactoring there are some props which are now "duplicated" in some places. In particular:
- `externalFragments` was passed separately to `ExecutionContextProvider` and `QueryEditor`
- `shouldPersistHeaders` was passed separately to `EditorContextProvider`, `ExecutionContextProvider` and `HeaderEditor`

This makes it possible to have somewhat inconsistent state. This PR attempts to clean this up by only passing these two props to `EditorContextProvider` (semantically it makes the most sense to have it there). The provider component includes the props in the context value, so they can be consumed by other parts of the UI.

I also did the same for `validationRules` even though it was only passed to one component (`QueryEditor`). It felt inconsistent for me to have `externalFragment` on the editor context value and `validationRules` not. (Both are used for the same purpose, i.e. to customize how operations are validated.)

Sidenote: Technically this is a breaking change for `@graphiql/react`, but we only do a minor version bump as we're still in the `0.x` range and the package is not stable yet.

### DIFF
--- a/.changeset/afraid-kiwis-notice.md
+++ b/.changeset/afraid-kiwis-notice.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': minor
+---
+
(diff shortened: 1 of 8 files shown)


jonathanawesome:
I don't have anything technical to add in this comment, but thanks for looking out for these types of problems after a _massive_ refactor.

jonathanawesome (approved):

PR #2637, opened 2022-08-06
use hsl values and reduce number of CSS variables

thomasheyenbrock:
Inspired by #2635, this is a proposal for how we can prepare v2 so that adding a theme prop down the road won't be a breaking change.

The main idea is to move the color conversion (non-transparent to transparent) into CSS instead of doing it in JS. We achieve this by storing the HSL values as list in the CSS variables. To applying a color we can then just do `hsl(var(--color-pink)`. To apply a transparent shade of a color we can do `hsla(var(--color-pink), 0.15)`.

This reduces the number of CSS variables that users need to care about when creating their custom theme, and it should align nicely with the idea of @jonathanawesome to have a theme prop like this in future versions:
```jsx
<GraphiQL theme={{ colors: { primary: "<some-color>" } }} />
```

### DIFF
--- a/packages/graphiql-react/src/style/root.css
+++ b/packages/graphiql-react/src/style/root.css
@@ -3,29 +3,15 @@
 .CodeMirror-lint-tooltip,
 reach-portal {
   /* Colors */
-  --color-pink: #d60690;
-  --color-purple: #6e6acf;
-  --color-blue: #007eea;
-  --color-orche: #d37f00;
-  --color-red: #f85b30;
-  --color-green: #2bab7c;
-  --color-teal: #009fb8;
-  --color-neutral-100: #3b4b68;
-  --color-neutral-60: #8993a4;
-  --color-neutral-40: #b1b7c3;
-  --color-neutral-15: #e2e4e8;
-  --color-neutral-10: #ebedf0;
-  --color-neutral-7: #f1f2f4;
-  --color-neutral-0: #ffffff;
-
-  --color-pink-background: rgb(214, 6, 144, 0.1);
-  --color-pink-background-dark: rgb(214, 6, 144, 0.15);
-  --color-pink-dark: #ab0573;
-  --color-orche-background: rgba(211, 127, 0, 0.07);
-  --color-orche-background-dark: rgba(211, 127, 0, 0.12);
-  --color-red-background: rgba(248, 91, 48, 0.12);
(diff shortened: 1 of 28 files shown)


jonathanawesome:
The switch to alpha values is _especially_ apparent in dark mode...it's a drastic difference and much more accurate to the design in the Figma file.

Before/after screenshots for those that aren't interested in running the app.

Before:
Screen Shot 2022-08-06 at 9 10 58 AM

After:
Screen Shot 2022-08-06 at 9 11 18 AM

packages/graphiql-react/src/style/root.css
```diff
@@ -73,39 +59,25 @@ reach-portal {
...
-  --color-neutral-10: #26303f;
-  --color-neutral-7: #293242;
-  --color-neutral-0: #222b39;
-
-  --color-pink-dark: #ff357f;
+  --color-pink: 338, 100%, 67%;
+  --color-purple: 243, 100%, 77%;
+  --color-blue: 208, 100%, 72%;
+  --color-orche: 30, 100%, 80%;
+  --color-neutral: 219, 29%, 78%;
```
jonathanawesome, line 79: This is where we may run into extensibility issues in the future. I did some testing auto-generating dark mode values was less reliable than it was in light mode. It's a well-understood dark theme problem, and we can decide how to deal with it down the road.

packages/graphiql-react/src/style/root.css
```diff
@@ -3,29 +3,15 @@
...
-
-  --color-pink-background: rgb(214, 6, 144, 0.1);
-  --color-pink-background-dark: rgb(214, 6, 144, 0.15);
-  --color-pink-dark: #ab0573;
-  --color-orche-background: rgba(211, 127, 0, 0.07);
-  --color-orche-background-dark: rgba(211, 127, 0, 0.12);
-  --color-red-background: rgba(248, 91, 48, 0.12);
-  --color-green-background: rgba(43, 171, 124, 0.12);
-  --color-neutral-background: rgba(59, 75, 104, 0.1);
+  --color-pink: 320, 95%, 43%;
```
jonathanawesome, line 6: This is _much_ more clean, well done! Once the theme API is opened up, it may be a stretch to require implementors to provide hsl values, but we can do some conversion if we have to, and once we get there.

jonathanawesome (approved):

PR #2642, opened 2022-08-08
fix operation name props

thomasheyenbrock:
Currently the `operationName` and `onEditOperationName` props do not really work. I can't trace back exactly when this was introduced, but it's likely that it's again a bug introduced when we started pulling stuff out into `@graphiql/react`.

Now it should work as expected:
- The `operationName` prop controls the operation name sent with the request when executing. It does not check if there actually exists an operation with this name, this has to be done in userland. When this prop is set, the `ExecutionButton` skips showing the dropdown if there are multiple operations.
- The `onEditOperationName` prop is now triggered whenever the operation name is changed. That includes
  - when changing the operation name in the query editor
  - when executing a different operation from the same document
  - when selecting the operation that should be executed from the dropdown menu triggered by the execution button

### DIFF
--- a/.changeset/eight-poems-call.md
+++ b/.changeset/eight-poems-call.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Fix controlling the operation name sent with the request using the `operationName` prop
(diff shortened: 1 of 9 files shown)


acao:
good find @thomasheyenbrock ! lots of expected logic added here. looks great to me!

PR #2653, opened 2022-08-09
fix: fix fetchError persisting across different fetchers

dylanowen:
If you construct a `GraphiQL` instance with a bad fetcher `fetchError` gets persisted in `SchemaContextType` forever. You can still fetch a new schema but the `DocExplorer` will always show "Error fetching schema"

### DIFF
--- a/packages/graphiql-react/src/schema.tsx
+++ b/packages/graphiql-react/src/schema.tsx
@@ -168,7 +168,6 @@ export function SchemaContextProvider(props: SchemaContextProviderProps) {
       }
 
       setIsFetching(true);
-      // clear any previous fetch errors
       setFetchError(null);
 
       let result = await fetch;
--- a/packages/graphiql-react/src/schema.tsx
+++ b/packages/graphiql-react/src/schema.tsx
@@ -168,6 +168,8 @@ export function SchemaContextProvider(props: SchemaContextProviderProps) {
       }
 
       setIsFetching(true);
+      // clear any previous fetch errors
+      setFetchError(null);
 
       let result = await fetch;
 
--- a/packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
+++ b/packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
@@ -98,7 +98,7 @@ describe('GraphiQL', () => {
     expect(secondCalled).toEqual(true);
   });
 
-  it('should refetch schema after a fetchError', async () => {
+  it('should refresh schema with new fetcher after a fetchError', async () => {
     function firstFetcher() {
       return Promise.reject('Schema Error');
     }
--- a/packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
+++ b/packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
@@ -98,6 +98,31 @@ describe('GraphiQL', () => {
     expect(secondCalled).toEqual(true);
   });
 
+  it('should refetch schema after a fetchError', async () => {
+    function firstFetcher() {
+      return Promise.reject('Schema Error');
+    }
+    function secondFetcher() {
+      return Promise.resolve(simpleIntrospection);
+    }
+
+    // Use a bad fetcher for our initial render
+    const { rerender, container } = render(<GraphiQL fetcher={firstFetcher} />);
+    await wait();
+
+    expect(
+      container.querySelector('.doc-explorer-contents .error-container'),
+    ).toBeTruthy();
+
+    // Re-render with valid fetcher
+    rerender(<GraphiQL fetcher={secondFetcher} />);
+    await wait();
+
+    expect(
+      container.querySelector('.doc-explorer-contents .error-container'),
+    ).not.toBeTruthy();
+  });
+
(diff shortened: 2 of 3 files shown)


packages/graphiql-react/src/schema.tsx
```diff
@@ -168,6 +168,8 @@ export function SchemaContextProvider(props: SchemaContextProviderProps) {
       }
 
       setIsFetching(true);
+      // clear any previous fetch errors
```
thomasheyenbrock, line 171: I don't think this comment is necessary, it's very clear what the following line does.
```suggestion
```

packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
```diff
@@ -98,6 +98,31 @@ describe('GraphiQL', () => {
     expect(secondCalled).toEqual(true);
   });
 
+  it('should refetch schema after a fetchError', async () => {
```
thomasheyenbrock, line 101: The test description sounds like the refetch would be automatically triggered after a fetch error. Can we make it more clear that the refetch has to be triggered "manually" (i.e. providing a different fetcher)

thomasheyenbrock (changes requested):
Thanks so much @dylanowen for catching this! That's indeed a bug that got in with one of the refactorings we did in the past months (most likely my bad 😓 ). The fix itself is perfect, just have some tiny nits!

Can you also add a changeset here that describes the fix? (I just noticed that this is still missing in our `CONTRIBUTING.md` 🙈 ). Just run `yarn changeset` in the repository root and the CLI will walk you through how to create a changeset which is just a markdown file that you need to commit and push to the branch.

thomasheyenbrock:
@dylanowen and you need to sign the CLA in order to become a contributor. Just click on the details link next to the failing GitHub check, that should walk you through everything.

lwilli (approved):

packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
dylanowen: 👍  How does

> should refresh schema with new fetcher after a fetchError

 sound?

dylanowen:
@thomasheyenbrock thanks for the quick review! I'm still working on the CLA but should have that ready soon.

packages/graphiql/src/components/__tests__/GraphiQL.spec.tsx
thomasheyenbrock: Sounds good!

acao:
thanks for this! @dylanowen make sure to sign the CLA bot with the email you used to sign the commits, it's very specific about this but won't tell you this is why 😆

dylanowen:
@acao Sorry for the CLA delay but it should be all set now

acao (approved):
Looks good to merge!

pitgrap:
looks related to https://github.com/graphql/graphiql/pull/2653

jonathanawesome:
@dylanowen I'm working on a small PR (#2778 ) to update some styling that's unrelated to this PR, but getting a failing test that _is_ part of this PR.

```
```

dylanowen:
@jonathanawesome I'm pretty sure this is caused by the long lag time between my PR being tested and me getting my CLA signed. When I opened this PR those selectors existed: https://github.com/graphql/graphiql/blob/48e7710a80739931565e1d472ed38136cc73865f/packages/graphiql/src/components/DocExplorer.tsx#L123 but I didn't realize they were changed before my merge.

jonathanawesome:
ahh...there it is. Thanks for the explanation, makes sense now.

I've got the fix in with #2778.

dylanowen:
Thanks for the quick fix! I was wondering if it'd be better to not depend on the `DocExplorer` for the test so I was working on 
```javascript
    let foundError = null;
    function ErrorIcon() {
      foundError = useSchemaContext({nonNull: true}).fetchError;
      return <div/>;
    }
    const ERROR_PLUGIN = {
      title: 'test',
      icon: ErrorIcon,
      content: () => <div />,
    };
    function firstFetcher() {
      return Promise.reject('Schema Error');
    }
    function secondFetcher() {
      return Promise.resolve(simpleIntrospection);
    }

    // Use a bad fetcher for our initial render
    const { rerender } = render(<GraphiQL fetcher={firstFetcher} plugins={[ERROR_PLUGIN]} />);
    await wait();

    expect(foundError).toBeTruthy()

    // Re-render with valid fetcher
    rerender(<GraphiQL fetcher={secondFetcher} plugins={[ERROR_PLUGIN]} />);
    await wait();

    expect(foundError).not.toBeTruthy();
```
But I'll defer to whatever solution you think is best.

jonathanawesome:
I think you've a good point...the Doc Explorer isn't in the DOM by default anymore (it must have been when you added your test), so I added a fireEvent to the test to ensure that it's there. 

I also added an additional test [closer to the component](https://github.com/graphql/graphiql/pull/2778/commits/ec90db510209d743301cd9daae4e3ad2a3f606ee). 

These are good, the-way-the-user-sees-it tests, so I think they're sufficient.

dylanowen:
Ahhh nice, even better 👍

PR #2654, opened 2022-08-09
[language-services] Improve autocomplete suggestions when editing type system definitions

cshaver:
When editing type system definitions, the autocomplete suggestions aren't super helpful (and kinda hard not to accidentally trigger at the top-level of the document):
image

In this PR I've updated one of the examples (so I could play with this behavior) to have an editor for the SDL that gets loaded in, which is applied to the query editor upon editing (I've added `fakeField` in the screenshot):
<details><summary>🖼 Screenshot of modified example</summary></details>

And I've added some autocomplete suggestions for the right keywords at the top-level, plus keywords that go after `extend`:
<details>
<summary>🖼 Gif </summary></details>

This is still rough, but I wanted to get some feedback on distinguishing between documents that contain documents executable and type system.

@acao (not sure who else to tag!)

### DIFF
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -1310,12 +1310,14 @@ function unwrapType(state: State): State {
   if (
     state.prevState &&
     state.kind &&
-    ([
-      RuleKinds.NAMED_TYPE,
-      RuleKinds.LIST_TYPE,
-      RuleKinds.TYPE,
-      RuleKinds.NON_NULL_TYPE,
-    ] as RuleKind[]).includes(state.kind)
+    (
+      [
+        RuleKinds.NAMED_TYPE,
+        RuleKinds.LIST_TYPE,
+        RuleKinds.TYPE,
+        RuleKinds.NON_NULL_TYPE,
+      ] as RuleKind[]
+    ).includes(state.kind)
   ) {
     return unwrapType(state.prevState);
   }
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -1271,7 +1271,7 @@ export function getTypeInfo(
   };
 }
 
-enum GraphQLDocumentMode {
+export enum GraphQLDocumentMode {
   TYPE_SYSTEM = 'TYPE_SYSTEM',
   EXECUTABLE = 'EXECUTABLE',
 }
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -25,6 +25,11 @@ import {
   GraphQLArgument,
   isListType,
   isNonNullType,
+  isScalarType,
+  isObjectType,
+  isUnionType,
+  isEnumType,
+  isInputObjectType,
 } from 'graphql';
 
 import {
@@ -51,6 +56,7 @@ import {
   isCompositeType,
   isInputType,
   visit,
+  BREAK,
   parse,
 } from 'graphql';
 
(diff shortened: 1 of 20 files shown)


acao:
Thanks for this! This is right off the LSP roadmap

acao:
I have a whole laundry list of improvements planned on this area! I want to make the language services work better without schema as well. Hope to make a new major version where they use a single object param and we can pass more options and make schema nullable!

acao:
CC @jonathanawesome - just as a preview in case you're interested in the more obscure seeming parts of the monorepo - this is what I was referring to about monaco's closesness to LSP and lower overhead for changes compared to codemirror which is not LSP-esque, and UX prescribed like monaco. Here with not many changes, and good test coverage, we can unlock some _really_ useful UX improvements for `monaco-graphql`, and this works in `vscode-graphql` and every other IDE extension too!

packages/monaco-graphql/src/languageFeatures.ts
```diff
@@ -238,13 +241,13 @@ export class CompletionAdapter
...
       const worker = await this._worker(model.uri);
-      const completionItems = await worker.doComplete(
+      const completionItems = await worker?.doComplete(
         resource.toString(),
         position,
       );
       return {
         incomplete: true,
-        suggestions: completionItems.map(toCompletion),
+        suggestions: completionItems?.map(toCompletion) || [],
```
acao, line 250: hm... completion LS should _always_ return an array! empty or not

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -308,6 +454,32 @@ const getInsertText = (field: GraphQLField<null, null>) => {
   return null;
 };
 
+function getSuggestionsForTypeSystemDefinitions(
```
acao, line 457: 💯

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -554,7 +726,7 @@ function getSuggestionsForFragmentTypeConditions(
     }
   } else {
     const typeMap = schema.getTypeMap();
-    possibleTypes = objectValues(typeMap).filter(isCompositeType);
+    possibleTypes = objectValues(typeMap).filter(type => isCompositeType(type) && !type.name.startsWith('__'));
```
acao, line 729: you even accounted for introspection types?! 🤩

acao:
This is _soooo close_ if the only thing blocking you are the CI issues:

1. `yarn format` because something is funky with the precommit hook again
1. `codemirror-graphql` tests are failing because you accidentally enhanced `codemirror-graphql` as a side effect 😆 
1. CLA bot is mad but not telling you why specifically (because it wants the commits to be signed with the same email you signed the CLA with 😩 )

acao:
tagging in @imolorhe as well, as he will probably appreciate this improvement that will land in both cm6 and cm6-graphql!

also @stonexer for LSP matters

packages/monaco-graphql/src/languageFeatures.ts
cshaver: Ah, I should take out the changes in this file — I was seeing errors in the console for trying to access `doValidation` on null from the worker, but this isn't really related  to the main changes anyway

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -120,6 +172,8 @@ export function getAutocompleteSuggestions(
   const state =
     token.state.kind === 'Invalid' ? token.state.prevState : token.state;
 
+  const mode = options?.mode || (options?.uri?.endsWith('.graphqls') || hasTypeSystemDefinitions(queryText)) ? 'TypeSystem' : 'Executable';
```
imolorhe, line 175: Maybe we should make mode an enum instead? Or string constants

imolorhe (commented):
Neat!

packages/monaco-graphql/src/languageFeatures.ts
cshaver: found the actual fix for this in https://github.com/graphql/graphiql/pull/2654/commits/3659928b5f30e6ca7c695ebe6722185569c98693!

cshaver:
cc: @acao this should be ready now, but I'm not sure why the prettier check is failing; `yarn lint` seems to pass when running locally

cshaver:
> cc: @acao this should be ready now, but I'm not sure why the prettier check is failing; `yarn lint` seems to pass when running locally

ah, just had to rebase with `main`! 🎉

stonexer (approved):
This is amazing!

acao:
@cshaver thank you for all of this, including cleaning up the examples! I think this looks ready to go, tests are all looking good.

I think for `codemorror-graphql` in 5 at least, we are actually overriding this I think? if we want to unlock this capability there we should look into this.

Wait - I just noticed something - `extend input` - I don't see that in the specs - only `extend type`.. hmm!!! Is there a draft for `extend input type` somewhere? if so, if it's advanced enough then that's fine. Either way, I'm really excited someone has this feature somewhere 😃

cshaver:
> Wait - I just noticed something - `extend input` - I don't see that in the specs - only `extend type`.. hmm!!! Is there an RFC for `extend input type` somewhere? if so, if it's advanced enough then that's fine. Either way, I'm really excited someone has this feature somewhere 😃

@acao [It is in the spec here!](https://spec.graphql.org/June2018/#InputObjectTypeExtension) I was surprised to learn scalars can be extended as well to add directives!

acao:
@cshaver i see it now! My method of searching for code examples caused me to miss it. Thanks for teaching me about the spec 😂

acao (approved):

cshaver:
@acao I don't have write access, so feel free to merge this whenever!

acao:
Ok sounds good! The one thing I wanted to add to the changleog isn’t necessary

acao:
@cshaver unfortunately we need to revert this

PR #2657, opened 2022-08-10
add documention for `@graphiql/react`

thomasheyenbrock:
This adds some more documentation in the simplest form possible: JSDoc comments :D Those should also be picked up by typedoc and enhance the docs we generate and push to Netlify already.

TODOs:
- [x] write README for `@graphiql/react`

### DIFF
--- a/.changeset/chilly-radios-tap.md
+++ b/.changeset/chilly-radios-tap.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': minor
+---
+
+BREAKING: The `onHasCompletion` export has been removed as it is only meant to be used internally.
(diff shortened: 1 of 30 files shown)


acao:
Love it!

I need to fix the typedoc config, but this is looking really nice! Of course it will be great in vscode/codesandbox/etc but also:

https://deploy-preview-2657--graphiql-test.netlify.app/typedoc/modules/graphiql_react.html#editorcontextproviderprops-5

i love these kinds of docs improvements because they show up _everywhere_ and we can make them show up in even more places!

would be cool to add `@examples` using backtick codeblocks, and then perhaps we can do some fun things with with a doc site tool like docusaurus or what have you - mdx, plugins, etc to link to show inline examples that link to templated codesandboxes for examples/etc

perhaps you'll want to keep the examples to the readme for now, whatever you prefer!

acao (approved):

PR #2658, opened 2022-08-10
Fix TypeScript GraphQL config file loading.

notjosh:
Hiya 👋  This is likely somewhat incomplete, as I'm really only focusing on fixing VSCode. If there's more to do for other packages, let me know!

This PR bumps `ts-node` to 10.x, so that TypeScript based configs (i.e. `.graphqlrc.ts`) will continue to work. It also bumps to the latest patch releases of `graphql-config` fixed several issues with TypeScript loading ([v4.3.2](https://github.com/kamilkisiela/graphql-config/releases/tag/v4.3.2), [v4.3.3](https://github.com/kamilkisiela/graphql-config/releases/tag/v4.3.3)).

## Background

Before this upgrade, the following appeared in my logs:

```8/10/2022, 9:23:24 PM [1] (pid: 8154) graphql-language-service-usage-logs: WARNING: graphql-config error, only highlighting is enabled:
Failed to compile TypeScript: Debug Failure. False expression: Non-string value passed to `ts.resolveTypeReferenceDirective`, likely by a wrapping package working with an outdated `resolveTypeReferenceDirectives` signature. This is probably not a problem in TS itself.
for more information on using 'graphql-config' with 'graphql-language-service-server', 
see the documentation at https://www.npmjs.com/package/graphql-language-service-server#user-content-graphql-configuration-file
```

Other folks online (i.e. [here](https://stackoverflow.com/questions/72488958/false-expression-non-string-value-passed-to-ts-resolvetypereferencedirective)) mentioned that `ts-node` needed updating to resolve this error.

### DIFF
--- a/packages/graphql-language-service/package.json
+++ b/packages/graphql-language-service/package.json
@@ -44,10 +44,10 @@
     "@types/json-schema": "7.0.9",
     "benchmark": "^2.1.4",
     "graphql": "^16.4.0",
-    "graphql-config": "4.3.0",
+    "graphql-config": "4.3.3",
     "lodash": "^4.17.15",
     "platform": "^1.3.5",
-    "ts-node": "^8.10.2",
+    "ts-node": "^10.8.1",
     "typescript": "^4.6.3"
   },
   "scripts": {
--- a/packages/vscode-graphql/esbuild.js
+++ b/packages/vscode-graphql/esbuild.js
@@ -11,7 +11,11 @@ build({
   minify: arg === '--minify',
   platform: 'node',
   outdir: 'out/',
-  external: ['vscode'],
+  external: [
+    'vscode',
+    '@cspotcode/source-map-support', // workaround for https://github.com/cspotcode/node-source-map-support/issues/43
+    './transpilers/swc.js', // suppresses warning via ts-node, but we won't be using it, so we can exclude it
+  ],
   format: 'cjs',
   sourcemap: true,
   watch: isWatchMode,
--- a/custom-words.txt
+++ b/custom-words.txt
@@ -68,6 +68,7 @@ codemirror
 codesandbox
 commitlint
 cosmicconfig
+cspotcode
 dompurify
 dset
 esbuild
--- a/package.json
+++ b/package.json
@@ -138,6 +138,6 @@
     "wsrun": "^5.2.4"
   },
   "resolutions": {
-    "graphql-config": "4.3.0"
+    "graphql-config": "4.3.3"
   }
 }
--- a/packages/graphql-language-service-server/package.json
+++ b/packages/graphql-language-service-server/package.json
@@ -37,7 +37,7 @@
     "@babel/types": "^7.17.0",
     "@babel/parser": "^7.17.3",
     "dotenv": "8.2.0",
-    "graphql-config": "4.3.0",
+    "graphql-config": "4.3.3",
     "graphql-language-service": "^5.0.6",
     "mkdirp": "^1.0.4",
     "node-fetch": "^2.6.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3691,6 +3691,13 @@
   resolved "https://registry.yarnpkg.com/@cspell/dict-vue/-/dict-vue-2.0.1.tgz#7514875f760ae755d2a6ef1fd00917d107682fe1"
   integrity sha512-n9So2C2Zw+uSDRzb2h9wq3PjZBqoHx+vBvu6a34H2qpumNjZ6HaEronrzX5tXJJXzOtocIQYrLxdd128TAU3+g==
 
+"@cspotcode/source-map-support@^0.8.0":
+  version "0.8.1"
+  resolved "https://registry.yarnpkg.com/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz#00629c35a688e05a88b1cda684fb9d5e73f000a1"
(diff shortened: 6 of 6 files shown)


packages/vscode-graphql/esbuild.js
```diff
@@ -11,7 +11,11 @@ build({
   minify: arg === '--minify',
   platform: 'node',
   outdir: 'out/',
-  external: ['vscode'],
+  external: [
+    'vscode',
+    '@cspotcode/source-map-support', // workaround for https://github.com/cspotcode/node-source-map-support/issues/43
+    './transpilers/swc.js', // suppresses warning via ts-node, but we won't be using it, so we can exclude it
```
notjosh, line 17: The warning was:

```
$ node esbuild
▲ [WARNING] "./transpilers/swc.js" should be marked as external for use with "require.resolve" [require-resolve-not-external]

    ../../node_modules/ts-node/dist/index.js:192:37:
      192 │         transpiler = require.resolve('./transpilers/swc.js');
          ╵                                      ~~~~~~~~~~~~~~~~~~~~~~

{
  detail: undefined,
  id: 'require-resolve-not-external',
  location: {
    column: 37,
    file: '../../node_modules/ts-node/dist/index.js',
    length: 22,
    line: 192,
    lineText: "        transpiler = require.resolve('./transpilers/swc.js');",
    namespace: '',
    suggestion: ''
  },
  notes: [],
```

acao:
Thank you! This appears to be an issue with the cosmic config plugin for typescript  I added to graphql-config. 

The resolution pin is important to fix an even larger bug. 

Is this working with remote introspection? We had to pin the version just last week because graphql-config‘s undici usage was breaking vscode graphql for even more users!

i generally prefer to not add dependencies that we don’t import to resolve an issue, because these can lead to issues later on

IMHO, we should be fixing the issue in graphql-config‘s resolution of the cosmicconfig plugin or the plugin itself

acao:
When i tried this fix last week, i got the jest mocks working with the latest whatwg fetch replacement they used in graphql tools url loader, but schema introspection still failed, so I rolled it back & pinned. I would link to the PR but I am on my phone & headed to sleep, so I will check in tomorrow or possibly later if this dependency drift keeps me up another night again 😂😭

acao:
@notjosh did you confirm this works for url schemas yet? I won’t have time to test the schema changes locally until after work in 10 hours

acao:
It appears to fix the issue for typescript files, but schema-as-url context appears to be failing silently, I can't get it to work at all with rick & morty API, whereas I was able to before off `main` before checking out this PR. I will look into it some more and see if we can at least get the error to surface. Last week when I tried upgrading `graphql-config` to the latest, to this same version in fact, the schema fetch failure was very visible, now this what appears to be a silent schema fetch failure is even more worrisome! It must be there was another patch version of `@graphql-tools/url-loader`

acao (changes requested):
Let's confirm that we can get this working in schema-as-url mode first. I can not get a schema introspection fetch to succeed from the LSP server in `vscode-graphql` after this change 😟 !

packages/graphql-language-service/package.json
```diff
@@ -44,10 +44,10 @@
     "@types/json-schema": "7.0.9",
     "benchmark": "^2.1.4",
     "graphql": "^16.4.0",
-    "graphql-config": "4.3.0",
+    "graphql-config": "4.3.3",
     "lodash": "^4.17.15",
     "platform": "^1.3.5",
-    "ts-node": "^8.10.2",
```
acao, line 50: wait.... how did we end up with this dependency in the first place? I think it was added by the GSOC intern who added the benchmarks. It's unfortunate that dev dependency resolutions like this impact the whole build just for a benchmark suite!

notjosh:
> @notjosh did you confirm this works for url schemas yet? I won’t have time to test the changes locally until after work in 10 hours

Ah, sorry, it was end of day for me yesterday. I'll noodle around a bit with it through the day and see what I can figure out. Seems like we'll need to find the ✨Goldilocks✨ combination of dependencies to cover all the use cases.

notjosh:
Hrm, it seems to be working okay for me, as far as I can tell.

I created a small sample project with a schema via URL, and a single document (a basic query). When I force the GraphQL LSP to restart, it takes a moment to resolve, but seems to handle the introspection (and therefore autocompletion) just fine:

image

But it's possible I'm missing something entirely, and there's a better way to tell if the introspection is actually working as expected (vs somehow using a cached version after failing). Any thoughts?

acao:
@notjosh indeed sounds like it's working fine then! false negative on my side. thanks for checking! we're progressively quieting the logs by default to only useful errors and warnings (still a few to go as you can see), so the usual fountain of output for normal language features isn't there by default unless you enable debug mode

acao (approved):

acao:
@notjosh published earlier, forgot to mention

I'm noticing @divyenduz 's original `vscode-graphql` package.json had pinned `dependencies` and I think I understand why. ^ versions when you're shipping shared libraries work pretty great. But for end consuming clients that have to ship a fixed bundle? And can't benefit from patch updates? Pin the `dependencies` to reduce the unnecessary surprises 😬 😆 !

acao:
Ok looks like we need to revert this. Last week repeating itself haha

acao:
@notjosh can you open a ticket for this bug now that it's been re-introduced?

PR #2659, opened 2022-08-11
disable codecov for now

acao:
disabling codecov but not removing.

we need to redo the test suite for graphiql anyways, and the LSP server coverage info is useful, however I tend to use it more for adding coverage with coverage codecov gutter extensions. 

also i sometimes feel like the codecov diff data is innacurate  anyways, probably because of misconfifguration

also clutter. this shrouds the CLA bot and changeset bot comments which are more important

we have many practical ways to display coverage as needed when adding coverage, and perhaps we will want to show coverage diffs in comments again


acao:
tried using codespaces for this and it was a lil funky 👀

PR #2665, opened 2022-08-11
re-introduce query exec as separate extension

acao:
a starting point, based on
https://github.com/graphql/vscode-graphql

TODO:
- [x]  as with `vscode-graphql`, port to the latest `vscode-anguageserver-client` using `vscode-anguageserver-client/node` import
- [x] fix embarassing package.json typos haha
- [ ] perhaps port the source provider codelens uses to locate tags to use `parseDocument` from `graphql-language-service-server`, and share parity with the LSP server's tag selection (expands on what is currently supported here
- [ ] add test to make sure codelens is working
- [ ] separate graphql config into a separate plugin? and how?
- [ ] add e2e suite because this could be flaky in the past, and making it a seperate extension from `vscode-graphql` makes manual testing more cumbersome


acao:
oof... i accidentally clobbered a ton of progress on this with a force push, and the only surviving  local copy of this branch with enhancements is on another computer

PR #2670, opened 2022-08-14
Pre release next

acao:
I _think_ this is what we need to do to be able to pre-release branches pre-fixed with `/next`

This would work from the changeset files alone!

How it works
1. if the branch has `next/` prefix
1. If the packages changed in the branch are already in pre-release state
1. new pre-releases will be re-released when new changesets are added!

The branch just needs to be prefixed with `next/`

CC @thomasheyenbrock


acao:
@dotansimha what do you think of this approach? Do you foresee any issues?

acao:
labelling do not merge until we are able to test this PR. I guess it should only impact PRs prefixed with `next/`

PR #2674, opened 2022-08-15
[v2] add plugin api for docs and history

thomasheyenbrock:
After playing around with v2 and the OneGraph Explorer I quickly realized that upgrading won't be feasible without being able to integrate the explorer into the plugin pane that we also use for docs and history. The only real solution that's not specific to the explorer is a plugin API. In the spirit of trying to keep the changes for v2 as small as possible, this was the easiest way I could define a plugin API.

In a nutshell, a plugin is an object that consists of:
- A title that has to be unique across all plugins
- A component that renders the contents in the plugin pane
- A component that renders an icon that we can display in the sidebar

All the visibility and toggle callbacks is handled by a new context called `PluginContext`. This actually allows us to remove quite a bit of code from the `ExplorerContext`, the `HistoryContext` and the `GraphiQL` component! Functionality-wise, nothing really changes for GraphiQL, the docs and the history are the two "built-in" plugins.


acao:
@thomasheyenbrock this is so exciting!

acao:
@sgrove it's finally happening 😃 !!

acao:
This is so exciting all around! Letting react & context API power the plugin architecture for you! Just as we envisioned

One idea for you - what about the dual service layer idea ala playground?

`<GraphiQLApp plugins={plugins} settings="" url="" headers="" /> `

which then looks like this:
```ts
const GraphiQLApp = ({plugins, settings}) => {
  const pluginManager = 
  const fetcher = pluginManager.buildFetcher()
  const router = pluginManager.buildRouter() // just something to parse query params with perhaps?
  // more logic to build <GraphiQL /> that is shaped by evolving plugin logic
  return (
  <PluginContext router={router} pluginManager={pluginManager}> <GraphiQL fetcher={fetcher}  /> </PluginContext>
  )
}
```

this way, the complexities of plugins aren't baked into `GraphiQL`, and plugins and settings can be an abstraction that feeds props to `GraphiQL`. This is a very lo-fi psuedocode example but just a service layer design modification I thought might be worth considering!

acao (approved):
I think this is a fantastic and simple start to plugins, and leverages the power of react to keep the plugin API simple as @leebyron, @benjie , @orta and everyone had envisioned. Thanks for this!

the service layer design comment I will break out into a discussion

PR #2683, opened 2022-08-20
add example for `graphiql-explorer`

thomasheyenbrock:
This adds an example for how to add the popular [`graphiql-explorer`](https://github.com/OneGraph/graphiql-explorer) as plugin in GraphiQL v2. There's some additional styling thrown in to make it blend in with the new design, I'm pretty happy with the looks for the minimal amount of style overrides one needs to pass in. Eventually it would be awesome if `graphiql-explorer` follows suit and adapts its default design to match the new GraphiQL design. But for now this should give folks something that is useable.

I deployed a preview at https://graphiql-explorer-thomasheyenbrock.vercel.app/

### DIFF
--- a/custom-words.txt
+++ b/custom-words.txt
@@ -63,6 +63,7 @@ yoshiakis
 // packages and tools
 argparse
 browserslist
+bundlephobia
 changesets
 codemirror
 codesandbox
@@ -82,6 +83,7 @@ manypkg
 markdownit
 meros
 nullthrows
+onegraph
 openvsx
 ovsx
 picomatch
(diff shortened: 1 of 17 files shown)


acao:
@thomasheyenbrock as we've discussed before, onegraph wanted us to adopt the graphiql explorer, so we were going to pull it into the repo. It's up to us to style it for the new graphiql!

thomasheyenbrock:
Thanks for the info @acao, I didn't know! I think this example is a good starter for now and we can iterate over that after we have v2 out of the door 👍

PR #2693, opened 2022-08-24
Deprecate graphql 15.5

acao:
actually wait we may not need to do this


acao:
instead we will fix the code that breaks graphql 15.5

PR #2700, opened 2022-08-25
Instantiate `isHeadersEditorEnabled` early

### LINKED ISSUE #2698: [v2] initialHeaders not loading
Setting headers in the Headers tab and then reloading the window crashes the page
image

image

image

This was previously working on the last version. The only way around this for us at the moment is to clear the Local Storage of the browser for graphiql.

patrick91:
This should fix #2698, it looks like that when there are headers in localstorage we try to access `isHeadersEditorEnabled` before it is instantiated 😊

### DIFF
--- a/.changeset/wise-pans-hear.md
+++ b/.changeset/wise-pans-hear.md
@@ -1,0 +1,5 @@
+---
+'graphiql': patch
+---
+
+Fix cannot access `initialHeaders` before initialization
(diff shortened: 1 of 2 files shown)


patrick91:
@acao is it fine to bump just GraphiQL?

acao:
@patrick91 that explains it, thank you so much!

acao:
@patrick91 in this case, yes, because the bug is located in `GraphiQL` itself!

PR #2703, opened 2022-08-25
fix: Broken image in README

ind1go:


thomasheyenbrock (approved):
Thanks for catching that and opening a PR! 😍

PR #2687, opened 2022-08-23
overhaul READMEs and add migratio guide for `graphiql@2`

thomasheyenbrock:
That's it, last PR for v2! (Unless someone finds a bug 😄 )

This refines the repo README, the README for the `graphiql` package and adds a migration guide that highlights all the breaking changes and how to upgrade a `graphiql@1.x` implementation to `graphiql@2`.

I removed the section with the architecture diagram from the repo README as it's no longer accurately reflecting the current state. Might still be valuable to have this, open for adding it back in an upgraded version later on!

PR #2704, opened 2022-08-25
Fix image path for graphiql UI screenshot

andrew-demb:
With #2687 image extension was changed from .jpg to .png, but the readme link was not updated according to the new file extension


thomasheyenbrock:
Thanks a lot for catching that and opening a PR 🙌 in fact I just merged that same fix that came in an hour earlier in https://github.com/graphql/graphiql/pull/2703, so I'm closing this.

PR #2708, opened 2022-08-26
fix initial state computation to avoid duplicate tabs

### LINKED ISSUE #2380: Using headers prop creates new tab on every reload
I've created a repro here: https://stackblitz.com/edit/react-ts-9dagjc?file=GraphiQL.tsx

Try hitting the refresh button on the preview on the right. It seems when i remove the `headers` prop the problem disappears.

thomasheyenbrock:
Fixes #2380 

I refactored the code a bit so that the computation of the initial editor values and the initial tabs happen at the same place. To avoid any inconsistencies we now store the tab state directly after deriving it in the initial render.

### DIFF
--- a/packages/graphiql-react/src/editor/tabs.ts
+++ b/packages/graphiql-react/src/editor/tabs.ts
@@ -62,11 +62,13 @@ export type TabsState = {
 };
 
 export function getDefaultTabState({
+  defaultQuery,
   headers,
   query,
   variables,
   storage,
 }: {
+  defaultQuery: string;
   headers: string | null;
   query: string | null;
   variables: string | null;
@@ -118,7 +120,7 @@ export function getDefaultTabState({
   } catch (err) {
     return {
       activeTabIndex: 0,
-      tabs: [createTab({ query: query ?? DEFAULT_QUERY, variables, headers })],
+      tabs: [createTab({ query: query ?? defaultQuery, variables, headers })],
     };
   }
 }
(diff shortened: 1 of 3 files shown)


packages/graphiql-react/src/editor/tabs.ts
```diff
@@ -108,15 +110,18 @@ export function getDefaultTabState({
...
 
       return parsed;
     } else {
       throw new Error('Storage for tabs is invalid');
     }
   } catch (err) {
-    storage?.set(STORAGE_KEY, '');
-    return { activeTabIndex: 0, tabs: [emptyTab()] };
+    return {
+      activeTabIndex: 0,
```
acao, line 122: ah, makes sense!

packages/graphiql-react/src/editor/tabs.ts
```diff
@@ -252,14 +257,18 @@ export function useSetEditorValues({
...
-export function emptyTab(): TabState {
+export function createTab({
+  query = null,
+  variables = null,
+  headers = null,
+}: Partial<Pick<TabState, 'query' | 'variables' | 'headers'>> = {}): TabState {
   return {
     id: guid(),
-    hash: hashFromTabContents({ query: null, variables: null, headers: null }),
+    hash: hashFromTabContents({ query, variables, headers }),
```
acao, line 267: ✅

acao (approved):
looks good to me!

PR #2710, opened 2022-08-26
remove rfc example code

thomasheyenbrock:
After v2 is released I propose to remove the RFC code from the repo. If we ever need it again, we can always restore and checkout this branch of revert the PR.


acao:
yep! it will be in the history. less code to lint/etc, good for me!

PR #2716, opened 2022-08-28
fix: make `@types/codemirror` a dependency of `@graphiql/react`

SimenB:
Without this, I get type errors in my project

```
../../node_modules/@graphiql/react/types/editor/components/image-preview.d.ts:1:28 - error TS7016: Could not find a declaration file for module 'codemirror'. '/Users/simen/repos/monorepo/node_modules/codemirror/lib/codemirror.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/codemirror` if it exists or add a new declaration (.d.ts) file containing `declare module 'codemirror';`

1 import type { Token } from 'codemirror';
                             ~~~~~~~~~~~~

../../node_modules/@graphiql/react/types/editor/hooks.d.ts:2:42 - error TS7016: Could not find a declaration file for module 'codemirror'. '/Users/simen/repos/monorepo/node_modules/codemirror/lib/codemirror.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/codemirror` if it exists or add a new declaration (.d.ts) file containing `declare module 'codemirror';`

2 import type { EditorConfiguration } from 'codemirror';
                                           ~~~~~~~~~~~~

../../node_modules/@graphiql/react/types/editor/response-editor.d.ts:1:38 - error TS7016: Could not find a declaration file for module 'codemirror'. '/Users/simen/repos/monorepo/node_modules/codemirror/lib/codemirror.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/codemirror` if it exists or add a new declaration (.d.ts) file containing `declare module 'codemirror';`

1 import type { Position, Token } from 'codemirror';
                                       ~~~~~~~~~~~~

../../node_modules/@graphiql/react/types/editor/types.d.ts:1:29 - error TS7016: Could not find a declaration file for module 'codemirror'. '/Users/simen/repos/monorepo/node_modules/codemirror/lib/codemirror.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/codemirror` if it exists or add a new declaration (.d.ts) file containing `declare module 'codemirror';`

1 import type { Editor } from 'codemirror';
                              ~~~~~~~~~~~~

../../node_modules/@graphiql/react/types/editor/types.d.ts:2:52 - error TS7016: Could not find a declaration file for module 'codemirror'. '/Users/simen/repos/monorepo/node_modules/codemirror/lib/codemirror.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/codemirror` if it exists or add a new declaration (.d.ts) file containing `declare module 'codemirror';`

2 export declare type CodeMirrorType = typeof import('codemirror');
```


thomasheyenbrock (dismissed):
Hey @SimenB and thanks for opening this PR 🙏 

I ran into this myself before and had the same thought. My concern here is that this would enforce a TypeScript-specific dependency to all users of this package, no matter if they use  TS of JS. Instead I would prefer to guide users to install `@types/codemirror` themselves if they use TS (basically what the error message you posted says). This is something that we should definitely call out in the README.

I'm requesting changes because of this concern for now. I'd be very interested to see how other libraries solve that problem before moving forward with this approach.

andreialecu:
> My concern here is that this would enforce a TypeScript-specific dependency to all users of this package, no matter if they use TS of JS. Instead I would prefer to guide users to install `@types/codemirror` themselves if they use TS (basically what the error message you posted says). This is something that we should definitely call out in the README.

If this is done, it should be set as an optional peer dependency. Otherwise users of yarn2/3 in PnP mode will encounter errors due to the undeclared dependency.

See: https://yarnpkg.com/advanced/rulebook/#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

thomasheyenbrock:
Thanks for the suggestion @andreialecu! Making it an optional peer dependency sounds like the best solution for me.

andreialecu:
> I personally prefer adding the dependency just to not force users who don't care one bit about what your type dependencies are to add it. It would also fall apart if `@graphiql/react` needed another version of `codemirror` than one a dependent needed.

+1 for this

acao:
On top of all the good reasons for this given by @SimenB , it's also harmless to make it a dependency from a bundling standpoint. So I generally support this approach for shipping libraries, if it matters.

The flip side of this is that npm@7 (and matching pnpm releases) automatically installs peer dependencies by default again

When we move to monaco editor and eventually codemirror 6, we wont need to import a `@types` package

B2o5T:
I am fine to include this in v3, as @acao said, anyway, we'll take away this `@types` dependency in v4 with `monaco-editor` integration

B2o5T (approved):
I am fine to include this in v3, as @acao said, anyway, we'll take away this `@types` dependency in v4 with `monaco-editor` integration

PR #2718, opened 2022-08-28
chore: use `createFetcher` in CDN example

SimenB:
The current `graphQLFetcher` in the example doesn't support headers. But it seems easier to just use `createFetcher`?


thomasheyenbrock (approved):
Good catch, the recommended approach is definitely using `createFetcher` 👍

PR #2720, opened 2022-08-28
Fixing completions

### LINKED ISSUE #2671: No language features for .graphql files since v0.5.0 (only colorization)
I've been trying to make the latest versions of this extension to work but it seems like completion (and most other language features) broke since `v0.5.0`. Note that everything works fine in `v0.4.15`.

The project I'm currently working on is [here](https://github.com/MariaSolOs/paradigm), in case you want to check out my setup. But as a summary, my `graphql.config.json` file looks like this:
```
{
    "schema": "./graphql-server/introspection.json",
    "documents": "./graphql-server/**/*.graphql"
}
```
And I'm using [codegen](https://www.graphql-code-generator.com/) with the following configuration:
```
schema: './graphql-server/schema.ts'
documents: './graphql-server/**/*.graphql'
generates:
  ./graphql-server/introspection.json:
    plugins:
      - introspection
    config:
      minify: true
  ./graphql-server/sdk.ts:
    plugins:
      - add:
          content: '// @ts-nocheck Ignore this file, it is just codegen output'
      - typescript
      - typescript-operations
      - typescript-graphql-request
      - plugin-typescript-swr
    config:
      skipTypename: true
  ./graphql-server/resolvers-types.ts:
    plugins:
      - add:
          content: '// @ts-nocheck Ignore this file, it is just codegen output'
      - typescript
      - typescript-resolvers
```

The logs only show the following:
```
8/14/2022, 2:55:11 PM [3] (pid: 7005) graphql-language-service-usage-logs: {"type":"usage","messageType":"initialize"}
```

MariaSolOs:
There seems to be an issue with completions caused by the [online parser](https://github.com/graphql/graphiql/blob/82c7aac12106c00dc0b3d97e50f2d773ca309e8c/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts#L1010) reading a version of the document that doesn't include the latest character typed. 

When debugging this query:
```
query getSnippets($query: String) {
    snippets(query: $query) {
        i <- This character is the completion trigger
    }
}
```
The online parser receives the following query text:
```
query getSnippets($query: String) {
    snippets(query: $query) {
        
    }
}
```
And then the wrong token is identified because this [check](https://github.com/graphql/graphiql/blob/82c7aac12106c00dc0b3d97e50f2d773ca309e8c/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts#L976) doesn't take into account that the `cursor` parameter corresponds to a newer text than the one that the `CharacterStream` was built with.

Although this feels like a caching issue, this PR handles the edge case separately.

Should fix #2671

### DIFF
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -973,10 +973,7 @@ export function getTokenAtPosition(
   let stringAtCursor = null;
   const token = runOnlineParser(queryText, (stream, state, style, index) => {
     if (index === cursor.line) {
-      if (
-        stream.getCurrentPosition() >= cursor.character ||
-        stream.sourceLength() < cursor.character
-      ) {
+      if (stream.getCurrentPosition() >= cursor.character) {
         styleAtCursor = style;
         stateAtCursor = { ...state };
         stringAtCursor = stream.current();
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -973,7 +973,10 @@ export function getTokenAtPosition(
   let stringAtCursor = null;
   const token = runOnlineParser(queryText, (stream, state, style, index) => {
     if (index === cursor.line) {
-      if (stream.getCurrentPosition() + 1 >= cursor.character) {
+      if (
+        stream.getCurrentPosition() >= cursor.character ||
+        stream.sourceLength() < cursor.character
+      ) {
         styleAtCursor = style;
         stateAtCursor = { ...state };
         stringAtCursor = stream.current();
@@ -1310,14 +1313,12 @@ function unwrapType(state: State): State {
   if (
     state.prevState &&
     state.kind &&
-    (
-      [
-        RuleKinds.NAMED_TYPE,
-        RuleKinds.LIST_TYPE,
-        RuleKinds.TYPE,
-        RuleKinds.NON_NULL_TYPE,
-      ] as RuleKind[]
-    ).includes(state.kind)
+    ([
+      RuleKinds.NAMED_TYPE,
+      RuleKinds.LIST_TYPE,
+      RuleKinds.TYPE,
+      RuleKinds.NON_NULL_TYPE,
+    ] as RuleKind[]).includes(state.kind)
   ) {
     return unwrapType(state.prevState);
   }
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -973,7 +973,7 @@ export function getTokenAtPosition(
   let stringAtCursor = null;
   const token = runOnlineParser(queryText, (stream, state, style, index) => {
     if (index === cursor.line) {
-      if (stream.getCurrentPosition() >= cursor.character) {
+      if (stream.getCurrentPosition() + 1 >= cursor.character) {
         styleAtCursor = style;
         stateAtCursor = { ...state };
         stringAtCursor = stream.current();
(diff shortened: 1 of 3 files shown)


acao:
@MariaSolOs thank you!!! This seems like the likely cause - I feel like I remember a user talking about this exact issue with the language service and/or server not long ago! I am off to bed but I will review soon. I am on break but this is high priority so I will try to review tomorrow or the next day.

CC @stonexer, @orta

MariaSolOs:
Thanks to YOU @acao for the hints! I think that the root cause of the problem comes from this change in [`vscode-languageserver-node`](https://github.com/microsoft/vscode-languageserver-node#3170-protocol-800-json-rpc-800-client-and-800-server): 
> all sendNotification methods now return a promise. Returning a promise was necessary since the actual writing of the message to the underlying transport is async and a client for example could not determine if a notification was handed off to the transport. This is a breaking change in the sense that it might result in floating promise and might be flagged by a linter.

This makes me think that [this comment](https://github.com/graphql/graphiql/blob/621253d5e855c0147f6eff50dbe9a827ef44c464/packages/graphql-language-service-server/src/MessageProcessor.ts#L562-L565) is no longer valid, and that the asynchronous `textDocument/didChange` notification might result in the document cache being updated post/during the completion request.

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -1310,14 +1313,12 @@ function unwrapType(state: State): State {
...
     state.kind &&
-    (
-      [
-        RuleKinds.NAMED_TYPE,
-        RuleKinds.LIST_TYPE,
-        RuleKinds.TYPE,
-        RuleKinds.NON_NULL_TYPE,
-      ] as RuleKind[]
-    ).includes(state.kind)
+    ([
```
acao, line 1316: It seems the prettier --check wants you to revert this formatting change

acao:
@MariaSolOs you are absolutely correct

> the asynchronous textDocument/didChange notification might result in the document cache being updated post/during the completion request.

I'm noticing that we no longer actually set the document cache (this._invalidateCache) on either `textDocument/didChange` or `textDocument/completion`, we only call `this._getCachedDocument()`... I think this is the cacheing issue you are talking about!

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
MariaSolOs: Interesting, I didn't make this change lol

MariaSolOs:
> I'm noticing that we no longer actually set the document cache (`this._invalidateCache`) on `textDocument/didChange`, we only call `this._getCachedDocument()`... I think this is related to the cacheing issue you are talking about!
@acao so here are my 2¢ on that document cache: I think that with LSP becoming more and more asynchronous (and trust me, *I know this is a fact*), we should sacrifice the performance gain of caching to regain correctness. Once the features are working properly again, we can start thinking about possible optimizations.

MariaSolOs:
Okay so it seems like we *should* still receive the `textDocument/didChange` notification before the completion request: https://github.com/microsoft/vscode-languageserver-node/issues/1025

So maybe a package update will solve the issue?

MariaSolOs:
Great news! I just tried updating [this package](https://github.com/graphql/graphiql/blob/92a17490c3842b4f83ed1065b73a803f73d02a17/packages/vscode-graphql/package.json#L177) to `^8.0.2` and completions are working properly again :)

I can then just close this PR and maybe this bump can be included in #2736 instead?

acao:
@MariaSolOs i think you could just pin the resolution to the working one in this PR, how does that sound? Even though it’s just upgrading the one dependency, we still need to make a change to vscode graphql to make a release happen (i think)?

acao:
@MariaSolOs actually I can make the PR real quick later no problem!

MariaSolOs:
@acao oops you created the new PR seconds before I updated this one hehe. Let's keep yours and thanks again!

acao:
@MariaSolOs just seeing that haha! Just adding a co-author credit too!

MariaSolOs:
> @MariaSolOs 0.7.5 fix should be released now! Thank you for spending so much time on this and painstaikingly discovering the fix. You highlighted a few important issues in this initiative!
> 
> Many more improvements to come of course! Please feel free to open any other PRs or contribute in other ways! Looking forward to hopefully working together more

Thanks to you for helping me fix it! I really like this extension and I was happy to contribute a bit to it :)
And for sure! You will see me around hehe

PR #2726, opened 2022-08-29
Use GraphiQLPluginExplorer.useExplorerPlugin instead of useExplorerPlugin

patrick91:
Just a small fix, otherwise this works pretty well!


thomasheyenbrock (approved):
Good catch!

PR #2742, opened 2022-09-02
better color contrasts

thomasheyenbrock:
This tries to improve the color contrasts for GraphiQL:
- Increasing the alpha for secondary and tertiary neutral colors
- Adding the box shadow from the design around the main editor card
- Using a yellow color for matching brackets

| Before | After |
| --- | --- |
| CleanShot 2022-09-02 at 17 39 29@2x | CleanShot 2022-09-02 at 17 37 37@2x |
| CleanShot 2022-09-02 at 17 39 18@2x | CleanShot 2022-09-02 at 17 38 35@2x |

### DIFF
--- a/packages/graphiql-react/src/editor/style/codemirror.css
+++ b/packages/graphiql-react/src/editor/style/codemirror.css
@@ -35,7 +35,7 @@
 
 .cm-s-graphiql {
   /* Default to punctuation */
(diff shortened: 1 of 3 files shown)


thomasheyenbrock:
I consulted with @julianbauer (who created the design for v2) and we are good to move ahead with this. Some future ideas around the new explorer might require some adjustments in terms of visual hierarchy, but for the current UI this is fine.

PR #2769, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/pr-graphql-compat-check.yml
+++ b/.github/workflows/pr-graphql-compat-check.yml
@@ -14,6 +14,9 @@ on:
 
 # TODO: test matrix?
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   build:
     name: Build & Test
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,6 @@ jobs:
   release:
     permissions:
       contents: read # to checkout the code
-      pull-requests: write
 
     environment: deploy
     name: Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,13 @@ on:
   push:
     branches:
       - main
+permissions: {}
 jobs:
   release:
+    permissions:
+      contents: write # to create release 
+      pull-requests: write # to create pull request 
+
     environment: deploy
     name: Release
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
           publish: yarn release
         env:
           # only use GH token here, because GITHUB_TOKEN is no longer allowed to create PRs
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           # for vscode marketplace, see https://github.com/microsoft/vscode-vsce/blob/194d59b975523696362ead891dc0f3ddd277b3bd/README.md#linux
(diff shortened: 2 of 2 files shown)


acao:
thank you @sashashura ! I think this means we can revert to `github-actions` user instead of using the custom token for `release` workflow on PR creation.

sashashura:
You must be having the check-box in the settings to prevent actions from creating pull requests. This is good. In that case the pull-request permission is not needed, because you use a dedicated token for that. Please, take a look if the last commit is OK.

acao:
@sashashura ah ok, so even though the github actions are creating Pull Requests, the read permission is enough? I thought we would need to request write permission as well? we won't know if it works until a PR is merged with a changeset

sashashura:
The `permissions:` instructs what privileges are granted to the temporary token `GITHUB_TOKEN`. The custom `GH_TOKEN` is not affected by this as it got permissions at the time of creation. Additionally to this there is a repo/org setting to disallow creation of pull requests for GITHUB_TOKEN`, this is why I guess you have the custom token.

acao:
@sashashura but I have removed the custom token, because now we have a solution to the permissions issue on PR creation and pushing commits. so i think we need to set a write permission here then yeah?

acao:
maybe you missed this commit?

https://github.com/graphql/graphiql/pull/2769/commits/ac98fc11a1089a1b848e8cacae9d0e292352ac0e

sashashura:
I didn't... There are two ways to implement what you need:
1.
- a. No permissions in the yaml file - token is generated according to the repo/org settings. By default it receives all write permissions (issues, pull requests, content, discussions, packages, etc.).
- b. Permissions in the yaml grant `pull-request: write` and `contents: read` to the generated token.
However, even though it had all permissions you needed a dedicated token to create a pull request. It makes me think you have the checkbox unchecked in your repo/org - "Allow GitHub actions to create and approve pull requests" 
image
If you want to implement it this way you need to re-enable it. Though, it is a good to have it disabled from security point of view...
2. 
- a. No permissions in the yaml file (all write by default) and a dedicated token to create the pull request.
- b. Permissions `contents: read` to checkout the code with `actions/checkout` and the dedicated token to create the pull request (which is not affected by `permissions:` but gets its permissions at creation time).

Hope that explains.

acao:
>  "Allow GitHub actions to create and approve pull requests"

ah yes, we won't be able to enable this for legal reasons, so I think we will need to go with option 2 with a dedicated token. I am hoping to create an org user though so not every automated PR is created by me but a bot user we all have access to

CC @leebyron

sashashura:
Coul you please then revert your last commit? Then I think it is good to be merged.

acao:
@sashashura I need to revert both commits I think! we should restore the custom token, and then remove the permission i added. I will come back around before the weekend hopefully

acao:
oh dear! this broke release. https://github.com/graphql/graphiql/actions/runs/3217885196/jobs/5261386138

 @sashashura I will revert

PR #2773, opened 2022-09-22
docs: fix usePrettifyQuery -> usePrettifyEditors

benjie:
There doesn't seem to be a `usePrettifyQuery` export; I assume `usePrettifyEditors` was what was intended?


thomasheyenbrock (approved):
Thanks! The naming was intended as it also prettifies the JSON for variables and headers.

Issue #2774, opened 2022-09-23
First response of a subscriptions wrongfully clears `isFetching` state


simhnna:

thomasheyenbrock:
Hey @simhnna 👋  This behavior is intended. The full view should be available once there is a first response. For subscriptions the play button as you noted is currently the indicator if the request is still running or not.

I'm with you that the current state is not ideal:
- Previous payloads are overridden once new ones come in
- The UI is lacking in terms of a11y

thomasheyenbrock:
I don't think we have an issue yet that tracks the bigger picture, i.e. how we can improve the UI for subscriptions. Would you be up for broadening the scope of this issue? I think I would be better thinking that through first before jumping on smaller items like the loading spinner and the `isFetching` state. What's your take on this?

simhnna:
Yeah we can broaden this issue I guess.

I'm no a11y expert, I'd rather leave that to someone else. But yeah I see 3 main pain points with the current implementation:
1. The execute button doesn't show the correct state for subscriptions (which is what I was mainly interested in fixing)
2. The loading indicator shown above the results for subscriptions. I could live with that for the time being
3. No history

## Execute button
For our internal deployment of graphiql I'd rather have that fixed quickly, and it's a small change (but it surfaces issue 2...)
We usually have long running subscriptions and currently there is no indicator that the subscription is still going. And IMO the fix improves the situation

## The loading indicator
I guess it's well placed for queries and mutations. Subscriptions could be running for a long time, and I don't think reserving that much space is a good idea. Maybe switch over to a loading bar? Not sure where to place it though. 
If subscriptions use a loading bar, maybe the queries/mutations should also switch to that...

## History
That's a big point. I currently mainly use [altair](https://altairgraphql.dev/) for subscriptions but I can't say I'm happy with how the subscriptions are displayed there. I actually very much like the current implementation in that only shows the most recent value. I do understand that showing the history is something people will want to have.  
If a history is added for subscriptions, I'd prefer it to be a separate view that you can switch to. I'd find a history view that can highlight diffs or only show specific keys of the response quite useful. Just having tens of large objects that you need to open/close that jump around when new entries are added isn't that useful to me...

acao:
@thomasheyenbrock the original issue for improved subscriptions UI is #436. it was in the github projects roadmap I shared!

simhnna:
Could we maybe discuss a MVP so I can get this bugfix in?

thomasheyenbrock:
[...]

PR #2775, opened 2022-09-23
Do not clear isFetching state for subscriptions until they complete

simhnna:
Fixes: #2774

### DIFF
--- a/.changeset/pink-mails-pump.md
+++ b/.changeset/pink-mails-pump.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Fix subscriptions clearing `isFetching` state after first emit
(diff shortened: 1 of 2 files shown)


simhnna:
I've added a changeset. Should I also bump graphiql?  
Regarding the CLA, I'm checking how in my company can sign the agreement.  

Previously the spinner only showed until the first result came in. Now it shows until the subscription is completed

Here's how it looks now 
now

It's better that the play/stop button now shows the correct state. But one could argue that the loading spinner shouldn't be that prominent after the first result is shown. I'm not sure how to best address that, but I'd track that in a separate issue I guess.

thomasheyenbrock (changes requested):
As described in https://github.com/graphql/graphiql/issues/2774#issuecomment-1256201649, I think we should align on where we wanna go with the UX for subscriptions first before jumping on smaller items and improvements. Blocking this PR for now until we had a discussion about this in the mentioned issue or a new one that tracks this (the code changes would be perfectly fine in itself).

simhnna:
Not needed anymore since the same functionality was now implemented in #2907

Issue #2777, opened 2022-09-25
GraphiQL Sidebar Styles


noahsb:
I'm using GraphiQL with the [Mantine](https://mantine.dev/) component library. When I render GraphiQL, the sidebar buttons aren't visible and the other buttons are significantly smaller:

<img width="1792" alt="Screen Shot 2022-09-25 at 6 48 25 PM" src="https://user-images.githubusercontent.com/49932437/192171090-0b551236-12dd-41ee-ae9f-2c80665ad715.png">

I believe this is an issue with how/when the styling is applied. This works fine when I create a new create react app without a component library.

Example code:

```
<MantineProvider withNormalizeCss withGlobalStyles>
  <GraphiQL fetcher={fetcher} />
</MantineProvider>
```

This only applies with graphiql@v2.0^

jonathanawesome:
If you could stand up a reproduction that'd be super helpful.

noahsb:
Sure thing! Here's a simple repo with the reproduction:

https://github.com/noahsb/graphiql-mantine-v2

jonathanawesome:
Thanks for the repro. Looks like Mantine is using the very common `box-sizing` reset via the `withGlobalStyles` prop. The quick solution is to _not_ use that prop. Here's the related [Mantine docs page](https://mantine.dev/theming/mantine-provider/#css-reset-and-global-styles), though it's actually kind of confusing because 1) it doesn't mention the box-sizing reset in the list of styles added ([here's the source](https://github.com/mantinedev/mantine/blob/14a3ff014b0f339fe1f1d329c95193c7a358b414/src/mantine-styles/src/theme/GlobalStyles.tsx#L9)) and 2) you'd think this reset would be part of the `withNormalizeCSS` prop rather than the `withGlobalStyles`. 

@thomasheyenbrock, how do you feel about adding a light reset to the `@graphiql/react` root css file? I think it would improve the overall compatibility of the package.

noahsb:
Thanks for the great explanation. Would definitely help my team to have a light reset of the css file, but will plan on removing the `withGlobalStyles` or using a version below 2.0 in the meantime. Thank you!

thomasheyenbrock:
@jonathanawesome having a light reset that is scoped to the container element where we render GraphiQL sounds good to me 👍

PR #2778, opened 2022-09-28
Simple box model reset

jonathanawesome:
This PR address the issues described in #2777 by adding the well-known box model reset to all children of the container into which we render GraphiQL.

A new class and associated styles are added to contain the sidebar controls, and adjustments are made to the toolbar button css to fit the new box model.

### DIFF
--- a/.changeset/tricky-glasses-try.md
+++ b/.changeset/tricky-glasses-try.md
@@ -3,4 +3,4 @@
 "graphiql": patch
 ---
 
-Adds a box-model reset for all children of the .graphiql-container class. This change facilitated another change to the --sidebar-width variable. 
+Adds a box-model reset for all children of the `.graphiql-container` class. This change facilitated another change to the `--sidebar-width` variable. 
--- a/.changeset/tricky-glasses-try.md
+++ b/.changeset/tricky-glasses-try.md
@@ -1,0 +1,6 @@
+---
+"@graphiql/react": patch
+"graphiql": patch
+---
+
+Adds a box-model reset for all children of the .graphiql-container class. This change facilitated another change to the --sidebar-width variable. 
--- a/packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
+++ b/packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
@@ -14,13 +14,12 @@ const defaultSchemaContext: SchemaContextType = {
 };
 
 const withErrorSchemaContext: SchemaContextType = {
-  fetchError: "Error fetching schema",
+  fetchError: 'Error fetching schema',
   introspect() {},
   isFetching: false,
   schema: new GraphQLSchema({ description: 'GraphQL Schema for testing' }),
   validationErrors: [],
 };
-
 
 function DocExplorerWithContext() {
   return (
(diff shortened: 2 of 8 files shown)


packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
```diff
@@ -59,4 +67,24 @@ describe('DocExplorer', () => {
       container.querySelector('.graphiql-markdown-description'),
     ).toHaveTextContent('GraphQL Schema for testing');
   });
+  it('renders correctly with schema error', () => {
+    const { rerender, container } = render(
+      <SchemaContext.Provider value={withErrorSchemaContext}>
```
acao, line 72: the providers make testing so much easier, don't they?!

acao (approved):
Looks great to me!

thomasheyenbrock (dismissed):
We're missing a changeset here. I think we should do a minor version bump here for GraphiQL as this potentially might affect style overrides that customize the `--sidebar-width` variable. (I doubt that anyone is using that though 😅 )

thomasheyenbrock (commented):
Other than the changeset, this looks great! 👏

acao:
> (I doubt that anyone is using that though 😅 )

you'd be surprised, with this many users, things move fast haha! there's no harm in a minor bump, the changelog tells all as to how minor it is. shows up in their automated tooling/etc even! and `^2.0.0` will resolve new minor versions as you know. agreed on the minor version bump now that you've explained it!

.changeset/tricky-glasses-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphiql/react": patch
+"graphiql": patch
+---
```
TallTed, line 4: This appears to be meant for humans to read and act on, with lines 2 and 3 being code, and lines 1 and 4 being code fences.  If that's correct, lines 1 and 4 should become triple-backticks ```` ``` ```` or triple-tildes `~~~` (whichever style you prefer), instead of triple-hyphens `---`
````suggestion
```
"@graphiql/react": patch
"graphiql": patch
```
````

File: .changeset/tricky-glasses-try.md [hunk 3]
```diff
@@ -0,0 +1,6 @@
...
+
+Adds a box-model reset for all children of the .graphiql-container class. This change facilitated another change to the --sidebar-width variable.
```
TallTed, line 6:
```suggestion
Adds a box-model reset for all children of the `.graphiql-container` class. This change facilitated another change to the `--sidebar-width` variable. 
```

TallTed (changes requested):
Minor touchup for human consumers of the MD file

.changeset/tricky-glasses-try.md
jonathanawesome: This is a changeset file, I'll leave it as is.

.changeset/tricky-glasses-try.md
jonathanawesome: This bit _is_ meant to be human readable...good catch!

.changeset/tricky-glasses-try.md
acao: @TallTed this file is machine read, the resulting changelog.md. entry is meant for humans to read

acao: @TallTed this file is machine read, the resulting changelog.md. entry is meant for humans to read

PR #2784, opened 2022-10-02
finally add issue forms & templates

acao:
Just this to start? And then keep improving as we go?


ADTC:
Some templates look incomplete. When I was creating issue #2843 I couldn't type in anything. There was only a checkbox. I had to submit it, then edit the first comment.

PR #2796, opened 2022-10-09
fix: update bug tracker links

joyceerhl:
The existing bug tracker links reference labels which no longer exist in this repository.


thomasheyenbrock (approved):
Thanks! 👏

PR #2808, opened 2022-10-10
Revert "Initialize the project cache more selectively"

acao:
Reverts graphql/graphiql#2608


acao:
this definately seems to fix the bug locally

PR #2814, opened 2022-10-12
Support Default values in lint variables

eMerzh:
At least variable section do a big red warning for missing values of field with a default value, 

this intend to silence that.

### DIFF
--- a/packages/codemirror-graphql/src/variables/lint.ts
+++ b/packages/codemirror-graphql/src/variables/lint.ts
@@ -170,9 +170,9 @@ function validateValue(
 
     // Look for missing non-nullable fields.
     Object.keys(type.getFields()).forEach(fieldName => {
+      const field = type.getFields()[fieldName];
       if (!providedFields[fieldName]) {
-        const fieldType = type.getFields()[fieldName].type;
-        if (fieldType instanceof GraphQLNonNull) {
+        if (field.type instanceof GraphQLNonNull && !field.defaultValue) {
           fieldErrors.push([
             valueAST,
             `Object of type "${type}" is missing required field "${fieldName}".`,


packages/codemirror-graphql/src/variables/lint.ts
```diff
@@ -172,7 +172,7 @@ function validateValue(
     Object.keys(type.getFields()).forEach(fieldName => {
       if (!providedFields[fieldName]) {
         const fieldType = type.getFields()[fieldName].type;
-        if (fieldType instanceof GraphQLNonNull) {
+        if (fieldType instanceof GraphQLNonNull && !type.getFields()[fieldName].defaultValue) {
```
thomasheyenbrock, line 175: Now that I see this, we have a lot of calls to `type.getFields()` in the branch of the code that handles input object. Can we store that in a variable somewhere so that we don't need to call it multiple times?

thomasheyenbrock (commented):
Thanks for fixing this 👏 I only have one small suggestion

packages/codemirror-graphql/src/variables/lint.ts
eMerzh: of course...
I  did it this way because there was already a bunch of getFileds so I chose the less invasive patch :)

acao:
It was only recently that we copied this logic [over from the language service](https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service/src/interface/getDiagnostics.ts). It bothers me that they are so similar yet so diverged. I will reflect this over there so this change takes for monaco and the language server

acao:
@jonathanawesome I don't think this is a problem in monaco yes? Json schema considers a required field with a default value valid?

jonathanawesome:
> @jonathanawesome I don't think this is a problem in monaco yes? Json schema considers a required field with a default value valid?

This doesn't seem to be an issue in Monaco, but I also can't replicate in GraphiQL v2 so maybe I don't understand the core issue. Is there a screenshot or a repro?

eMerzh:
@jonathanawesome i can't really give access to my api , but what i experience is :
schema looks like this : 
```graphql 
input StoryEditInput {
  id: ID!
  publish: Boolean! = false
  removeBody: Boolean! = false
  removeCoverAsset: Boolean! = false
  removeMainAsset: Boolean! = false
  releaseLock: Boolean! = true
  lockToken: String
}
type Mutation {
  storyEdit(input: StoryEditInput!): StoryEditPayload
}
```

and in graphiql i have : 
Screenshot 2022-10-17 at 17 03 01


but if i inline the variables, it's fine 

Screenshot 2022-10-17 at 17 07 28

jonathanawesome:
@eMerzh thank you for the screenshots 👍

I now understand the issue (thanks for the fix!) and can confirm for @acao that this is not applicable to Monaco.

thomasheyenbrock (approved):
Nice!

thomasheyenbrock:
This has been fixed in `codemirror-graphql@2.0.1` and `graphiql@2.0.12`!

PR #2821, opened 2022-10-16
Initial tabs

### LINKED ISSUE #2705: Support defining initial tabs
[graphql-playground](https://github.com/graphql/graphql-playground#properties) supports loading pre-defined tabs.

It would be useful to get the same feature in GraphiQL v2.

avaly:
Closes #2705

### DIFF
--- a/packages/graphiql-react/src/editor/context.tsx
+++ b/packages/graphiql-react/src/editor/context.tsx
@@ -172,6 +172,16 @@ export type EditorContextProviderProps = {
   /**
    * This prop can be used to defined the initial set of tabs with their queries,
    * variables and headers.
+   *
+   * @example
+   * ```tsx
+   * <GraphiQL
+   *   initialTabs={[
+   *     { query: 'query myExampleQuery {}' },
+   *     { query: '{ id }' }
+   *   ]}
+   * />
+   *```
    */
   initialTabs?: TabDefinition[];
   /**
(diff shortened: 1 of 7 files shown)


packages/graphiql-react/src/editor/context.tsx
```diff
@@ -168,6 +169,11 @@ export type EditorContextProviderProps = {
    * typing in the editor.
    */
   headers?: string;
+  /**
+   * This prop can be used to defined the initial set of tabs with their queries,
+   * variables and headers.
+   */
```
acao, line 175: can you add an `@example` like this by chance? would like to start doing this to make the props docs more helpful

```suggestion
   * @example
   * ```tsx
   * <GraphiQL initialTabs={[
   *    { query: 'query myExampleQuery {}' }, 
   *    { query: '{ id }' }
   *   ]}
   * />
   *```
   */
```

packages/graphiql-react/src/editor/tabs.ts
```diff
@@ -5,10 +5,25 @@ import debounce from '../utility/debounce';
...
+   */
+  query: string | null;
+  /**
+   * The contents of the variable editor of this tab.
+   */
+  variables: string | null;
+  /**
+   * The contents of the headers editor of this tab.
+   */
+  headers: string | null;
```
acao, line 20: is there a reason `headers?: string` isn't used instead? 

so, when specifying this, users will need to provide `{ query: '{ id }', variables: null, headers: null }` if they only want to specify a query. I think allowing it to be undefined is preferrable over requiring null here, what do you think @thomasheyenbrock  or @jonathanawesome ?

acao (changes requested):

acao:
thanks @avaly !

PR #2824, opened 2022-10-17
fix: prevent keydown event when pressing escape to close autocomplete dialogs

TheMightyPenguin:
# Description

Prevents keydown event when pressing escape to close autocomplete dialogs. 

We are using GraphiQL in our app ([stellate.co](https://stellate.co/)), and we show the GraphiQL editor as a modal. In this scenario, whenever we pressed the escape key inside of the editor, specifically when closing the autocomplete dialog, our whole modal was closing down as we have keydown listeners in that component, that will close it when 'Escape' is pressed.

These changes stop a `keydown` event from propagating out of the editor if the escape key was pressed to close an autocomplete dialog.

This also added e2e tests for this behavior.

**Small demo of the changes**: https://www.loom.com/share/478449efefc94be8be944cd304910038

**More context on the approach used**: I was reading Codemirror docs and checking some of their source, and I didn't find anything in their state indicating if the popup was open or not. I did found that they fire the `startCompletion` and `endCompletion` events whenever the popup was open or closed (closed by any means, including pressing the escape key), so I used these events to control state on our side to prevent propagating the keydown event.

### Questions

- Should this be made a customizable option/prop or if it's ok to be the default behavior?


packages/graphiql-react/src/editor/query-editor.tsx
```diff
@@ -148,6 +148,8 @@ export function useQueryEditor(
         return;
       }
 
+      let showingHints = false;
```
thomasheyenbrock, line 151: Nit: Can we move that variable down where we define the handlers to have better coloaction?

thomasheyenbrock (commented):
Awesome!

> Should this be made a customizable option/prop or if it's ok to be the default behavior?

I'm fine with this being the default, we're already doing the same thing for the search menu in the docs.

.changeset/wicked-jars-push.md
```diff
@@ -0,0 +1,6 @@
+---
+'graphiql': patch
```
thomasheyenbrock, line 2: This is actually just changing `@graphiql/react`
```suggestion
```

thomasheyenbrock (changes requested):

thomasheyenbrock (approved):
Perfect! 👏 Congrats to the first GraphiQL contribution 🥳

PR #2827, opened 2022-10-22
Vue sfc intellisense

acao:
re-creating #2391 so that I can rebase it before merging!

This introduces vue support for intellisense.

Thanks @AumyF for making this PR!


flotothemoon:
(Adding here from Discord for searchability & awareness.) 

I can't get VSCode Intellisense in `.vue` files to work, even though GraphQL Intellisense works in `.graphql` and with the same `graphql` tag in `.ts` files. My `.graphqlrc` includes `.vue` files and I'm running the latest version of the extension, which should include Vue support as I understand.

What might I be doing wrong?

flotothemoon:
After some debugging it Vue VSCode Intellisense works if the setup scripts comes first (before template & script), but fails in any other order. If the setup script comes second, GraphQL errors are overlaid on whatever is actually there as if the script came first. So it looks like the setup script offset isn't added somewhere.   

I'm not familiar with this codebase, so maybe this is an upstream issue with the vue sfc compiler.

PR #2839, opened 2022-10-25
expose PluginContextProvider

ClemensSahs:
Any other ContextProvider are exposed, with exception PluginContextProvider.

So in my opinion https://github.com/graphql/graphiql/pull/2674 only miss this expose

### DIFF
--- a/packages/graphiql/__mocks__/@graphiql/react.tsx
+++ b/packages/graphiql/__mocks__/@graphiql/react.tsx
@@ -62,7 +62,7 @@ export {
   PenIcon,
   PlayIcon,
   PluginContext,
-  PluginContextProfider,
+  PluginContextProvider,
   PlusIcon,
   PrettifyIcon,
   ReloadIcon,
--- a/packages/graphiql/__mocks__/@graphiql/react.tsx
+++ b/packages/graphiql/__mocks__/@graphiql/react.tsx
@@ -62,6 +62,7 @@ export {
   PenIcon,
   PlayIcon,
   PluginContext,
+  PluginContextProfider,
   PlusIcon,
   PrettifyIcon,
   ReloadIcon,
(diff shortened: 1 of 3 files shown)


thomasheyenbrock (commented):
That is definitely an oversight, good catch! Can you please also re-export the `PluginContextProvider` in the mock file of the `graphiql` package here: https://github.com/graphql/graphiql/blob/aaa43e401aeb4a1b9cef8adc0bf5702dac92745c/packages/graphiql/__mocks__/%40graphiql/react.tsx#L14

(For context, we're mocking part of the package for testing and just re-export everything else . It's quite annoying to keep this in sync, but nobody found the time to improve that yet.)

ClemensSahs:
> That is definitely an oversight, good catch! Can you please also re-export the `PluginContextProvider` in the mock file of the `graphiql` package here:
> 
> https://github.com/graphql/graphiql/blob/aaa43e401aeb4a1b9cef8adc0bf5702dac92745c/packages/graphiql/__mocks__/%40graphiql/react.tsx#L14
> 
> (For context, we're mocking part of the package for testing and just re-export everything else . It's quite annoying to keep this in sync, but nobody found the time to improve that yet.)

yes, sure. DONE ;)

side context:

only following things that are mocked right anything else will be re-exported, right?
```
useHeaderEditor
useQueryEditor
useResponseEditor
useVariableEditor
HeaderEditor
QueryEditor
ResponseEditor
VariableEditor
```

thomasheyenbrock:
@ClemensSahs CI has some failures, can you take a look?

packages/graphiql/__mocks__/@graphiql/react.tsx
```diff
@@ -62,6 +62,7 @@ export {
   PenIcon,
   PlayIcon,
   PluginContext,
+  PluginContextProfider,
```
thomasheyenbrock, line 65: That's what CI is complaining about
```suggestion
  PluginContextProvider,
```

thomasheyenbrock (approved):
I also added a changeset, thanks again!

Issue #2843, opened 2022-10-27
[vscode-graphql-syntax] Text on the same line after `#graphql` inside a Javascript string is not formatted as a comment


ADTC:
### Is there an existing issue for this?

- [X] I have searched the existing issues

ADTC:
It also seems that using `/* GraphiQL */` as a comment before the variable doesn't seem to work.

acao:
@ADTC thanks for reporting this bug!

regarding the second comment - you need to use `/* GraphQL  (other comment errata here) */` - let me check the new readme hopefully I didn't mess that up!

PR #2849, opened 2022-10-28
docs: typo - `/* GraphQL */` not `/* GraphiQL */`!

acao:
huge typo. caused this user bug https://github.com/graphql/graphiql/issues/2843#issuecomment-1294685543

### DIFF
--- a/.changeset/spicy-ducks-fail.md
+++ b/.changeset/spicy-ducks-fail.md
@@ -1,0 +1,6 @@
+---
+"vscode-graphql-syntax": patch
+"vscode-graphql": patch
+---
+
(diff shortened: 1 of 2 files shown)


ADTC:
@acao I should have mentioned that I had actually tried the different spelling and it still didn't work:

image

But this version works:

image

So does it have to be positioned after the equals (`=`) sign? This is a `.jsx` file btw.

acao:
@ADTC very good question - thanks for asking! both should work. looks like the regex for the syntax highlighting for case #1 is broken, then! Are you able to see validation/autocompletion even withought highlighting in case #1?

ADTC:
@acao I'm sorry, but I'm only using the syntax highlighting extension right now. Hope you can check that on your end. Thank you!

ADTC:
Hope you managed to check. But it's okay. I kinda like the second version TBH. It doesn't take up one extra line.

PR #2867, opened 2022-11-05
fix: fixed "Mark decorations may not be empty"  error

imolorhe:
fix: fixed "Mark decorations may not be empty"  error in codemirror 6 lint extension.


acao:
thank you @imolorhe !

PR #2872, opened 2022-11-08
enable `react/jsx-filename-extension`

B2o5T:
Back to contribution 🫶. This PR allows `jsx/tsx` extension only if there is JSX syntax inside file


acao (approved):
thank you for these painstaking but important eslint improvements

PR #2873, opened 2022-11-08
[graphiql] remove useless ternary

B2o5T:


acao:
thanks @B2o5T ! good find

PR #2885, opened 2022-11-10
Do not show dropdown for execute button if an operation is currently running

simhnna:
If you have multiple operations defined in the editor, clicking the execute button will show a dropdown to choose which operation to run.

If an operation is running, the stop button still shows the dropdown. Since only one operation can be running at any point in time, showing the dropdown is useless. This merge request makes sure that the dropdown is only shown, if no operation is currently running

### DIFF
--- a/.changeset/poor-vans-kiss.md
+++ b/.changeset/poor-vans-kiss.md
@@ -1,0 +1,6 @@
+---
+'graphiql': patch
+'@graphiql/react': patch
+---
+
+Fix stop execution button showing a dropdown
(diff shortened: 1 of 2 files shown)


.changeset/poor-vans-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+'graphiql': patch
+'@graphiql/react': patch
+---
+
+Fix stop execution button showing a dropdown
```
acao, line 6:
```suggestion
Bugfix: Stop execution button from showing a dropdown during operation fetch
```

acao (approved):
looks good to me!

Issue #2294, opened 2021-11-11
Decorators break completion


sdykae:
## Actual Behavior
![image](https://user-images.githubusercontent.com/50224521/141252579-a2f9177d-8729-4a4c-bd25-aa2f0a15e8fe.png)

## Expected Behavior
![image](https://user-images.githubusercontent.com/50224521/141252692-6f9c20eb-04fc-4c2f-b3b0-108a39dd2b35.png)

## Steps to Reproduce the Problem Or Description
![image](https://user-images.githubusercontent.com/50224521/141252771-35c15b4a-f170-4048-b960-541e9ec1999e.png)

## Specifications

  - GraphQL for VSCode Extension Version:  0.3.18

  - VSCode Version: latest
  - OS Name: linux
  - OS Version: ubuntu 20.04

#fix              
https://github.com/babel/babel-eslint/issues/774
https://github.com/babel/babel-eslint/issues/662
https://stackoverflow.com/questions/53483066/babel-7-using-the-export-keyword-between-a-decorator-and-a-class-is-not-allowed/54516632

Why babel-eslint is triggered? Wtf?
My project don't use babel, or eslint.

fajarclodeo:
Affect my nestjs project as well.

acivier-serial:
Any update on this issue ?

acao:
I’m in the midst of a big project at work, but if someone wants to open a PR with those changes to the babel parser options and a test for findGraphQLTags that proves the fix, it would be very welcome!

acao:
At this service level we have `decorators-legacy` available for this type, however I get this:

<img width="564" alt="Screenshot 2022-07-16 at 01 21 56" src="https://user-images.githubusercontent.com/1368727/179323638-7368c4a5-2f13-4855-b7a5-c6ac410c44dd.png">

we can add an option to allow someone to optionally enable support for `decorators-legacy`, which would disable support for `decorators`? one could use projects to filter which files have this setting enabled?

joepjoosten:
https://github.com/graphql/graphiql/blob/48e7710a80739931565e1d472ed38136cc73865f/packages/graphql-language-service-server/src/findGraphQLTags.ts#L43

jrwpatterson:
any updates on this. is there a way just to block it looking at the typescript with decorators, we are currently running a monorepo with the graph queries defined in a react app and a nest server. The nest server doesn't need any graph queries but because it has decorators its breaking the rest of the thing..

acao:
[...]

PR #2889, opened 2022-11-10
Feat/add opt in for legacy decorators

acao:
Fix for #2294 thanks to @mvnhrd

hey there @mvnhrd thanks so much for your work on this!

I'm going to create a PR and rebase this branch now, and maybe make a few more commits, is that ok?

### DIFF
--- a/packages/graphql-language-service-server/README.md
+++ b/packages/graphql-language-service-server/README.md
@@ -147,6 +147,8 @@ module.exports = {
       cacheSchemaFileForLookup: true,
       // undefined by default which has the same effect as `true`, set to `false` if you are already using // `graphql-eslint` or some other tool for validating graphql in your IDE. Must be explicitly `false` to disable this feature, not just "falsy"
       enableValidation: true,
+      // opt-in for allowing legacy decorators, which will set the `decorators-legacy` option in the Babel parser configuration. Must be explicitly `true` to enable, not just "truthy".
+      enableLegacyDecorators: true,
     },
   },
 };
@@ -188,6 +190,7 @@ export default {
     languageService: {
       cacheSchemaFileForLookup: false,
       enableValidation: true,
+      enableLegacyDecorators: false,
     },
   },
 };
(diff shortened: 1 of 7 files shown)


acao:
you can get commit credit if you like, if you just sign the CLA using the same email the commits are signed with. if you'd rather not I can just give you a thank you in the git commit message!

acao:
oh wait i see now you are also in CET so perhaps a bit too late to ask, haha!

mvnrhd:
just signed it, but feel free to go ahead if it's too late now! 🙂

acao:
@mvnrhd i finally reviewed it in more depth last night, this is a great first PR. For future reference, the change to `vscode-graphql` or `startServer` is not needed - the `MessageProcessor` accesses workspace config using `this._connection.workspace.getConfiguration()`  already!


This is a quite difficult case as it turns out looking more closely. Instantiating the parser properly based on config change events is proving to be tricky but I think I have a solution. for example, on MessageProcessor constructor() _or_ on the LSP server initialise event, we do not even have access to workspace settings yet!  This impacts vim users as well, and I'm finding bugs with how all the settings are loaded. I might have to do this PR myself on closer inspection, though I can re-use the logic from `findGraphQLTags`! 

It requires writing a a set of tests as well. I may create a separate PR that gives you attribution though! Please don't take it the wrong way, this PR helps me think about the problem, but the resulting PR will look quite different, is that ok?

mvnrhd:
Hey @acao no worries! As said, I was not able to test my changes locally since I didn't find clear instructions how to actually run & debug the extension, the same applies for running tests. That's why I was already expecting that there will be further code changes needed. I'm just happy that we're tackling this now, so thanks a lot for your help. 😉

acao:
@mvnrhd sorry we need to improve the docs for vscode - the vscode debugger tasks provide the language (LSP) and execution extension runners, of which you want the former. and the build tooling it uses changed because of graphiql so it doesn't always work right away. first a good `yarn build` across the repo and a `yarn compile` in `packages/vscode-graphql/` will ensure you can run the task for the language extension

acao:
@mvnrhd it ended up being a really simple fix because of my misunderstanding - no settings needed, just `decorators-legacy`!

acao:
@mvnrhd thank you for everything on this! The fix is released now. Sorry I made it so overcomplicated!

Issue #2793, opened 2022-10-07
[graphiql] user-facing setting for persisting headers


thomasheyenbrock:
The `GraphiQL` currently provides a prop `shouldPersistHeaders` that implementors can use to decide if header values should be persisted in `localStorage`. By default, no headers are persisted as a security measure (header values often times contain sensitive tokens). This defaults leads to many GraphiQL implementations that are not persisting headers.

As a GraphiQL user I would like to override this setting. I want to be able to choose if I want my headers to be persisted or not.

Proposal: Add a section to the settings dialog where users can override the default behavior of the GraphiQL implementation around persisting headers. This setting should only have effect if the `shouldPersistHeaders` prop is not set, or in other words, the GraphiQL implementation should still be able to explicitly enforce a choice about persisting headers and GraphiQL users can only choose their preference if the implementation is not explicit about this setting.

TheMightyPenguin:
Gonna give this one a shot 👀

TheMightyPenguin:
Found an accessibility issue when working on this one https://github.com/graphql/graphiql/issues/2891

Issue #2892, opened 2022-11-11
[graphiql] Seeing a lot of `act` related warnings in tests


TheMightyPenguin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A few of the react unit tests are reporting warnings as they're not using `act` to wrap events that update the state. Also this is kind of tech debt, not sure what labels shout it have 
<img width="869" alt="image" src="https://user-images.githubusercontent.com/11748696/201421701-c4bf7370-3364-4000-8356-1635ef9a408f.png">

### Expected Behavior

We could wrap with `act` where necessary to get rid of these and remove some noise when running tests.

### Steps To Reproduce

```Markdown
1. Run `yarn test Graphiql.spec` and you will see a lot of these (printing the minified source as well which adds even more noise)
```

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version: latest
- OS: Mac OS 12.5
- Browser: Chrome 107
- Bundler:
- `react` Version: 17
- `graphql` Version: 16
```

TheMightyPenguin:
@jonathanawesome you working on this one? If not I can help with it one next week

jonathanawesome:
@TheMightyPenguin I'm on it!

Issue #2891, opened 2022-11-11
[graphiql] Settings dialog button group does not comply with accessibility behaviors


TheMightyPenguin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The ButtonGroup in the Settings dialog, specifically when changing the current theme:
<img width="484" alt="image" src="https://user-images.githubusercontent.com/11748696/201409160-9c2c0b29-3e95-426a-b0f3-e1250e74bd02.png">

Behaves more like a RadioGroup control, and it doesn't have the appropriate keyboard event handling or aria attributes to indicate this: https://www.w3.org/WAI/ARIA/apg/patterns/radiobutton/

### Expected Behavior

These buttons should comply with the standard for accessible radio groups.

### Steps To Reproduce

```Markdown
1. Open the GraphiQL demo http://graphql.org/swapi-graphql
2. Open the settings dialog
3. Navigate it using `tab` and once you land on the "System" button, try navigating using the arrow keys
4. Also inspect the elements and you'll notice they don't have the appropriate attributes
```

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version: latest
- OS: Mac OS 12.5
- Browser: Chrome 107
- Bundler:
- `react` Version: 17
- `graphql` Version: 16
```

TheMightyPenguin:
I'd like to give this one a go after I finish #2973

PR #2895, opened 2022-11-11
Add user facing setting for persisting headers

TheMightyPenguin:
Resolves https://github.com/graphql/graphiql/issues/2793

# Description

Adds a user control in the settings dialog for persisting headers. This is only visible if the `shouldPersistHeaders` is explicitly set as `false`.

Demo and brief explanation: https://www.loom.com/share/d72cf65025b04b1f9d6d5e518445bc25

image

I also found 2 potential issues when working on this:
- https://github.com/graphql/graphiql/issues/2891 (Accessibility related, I think the ButtonGroup we use in the settings dialog doesn't have the correct accessibility behavior and attributes. I decided to keep it as a separate issue to keep this PR smaller, I'll work on that one once this one is completed)
- https://github.com/graphql/graphiql/issues/2892

**Note for the future:** I think it would be nice to somehow refactor this out into a user settings module, not the UI but the logic to store and access user settings. At the moment we have Theme and now Persist Headers, but there could be more user controlled settings going forward and having a central place for this logic (including defaults and such) could be nice.

### Questions
1. As headers are not persisted as a security measure, should we include a warning for the user in the dialog?
2. I added logic to cleanup the headers storage if this option is turned off, do we also need to do the same for tabs state?

### DIFF
--- a/.changeset/sweet-foxes-drive.md
+++ b/.changeset/sweet-foxes-drive.md
@@ -1,6 +1,6 @@
 ---
-'graphiql': patch
-'@graphiql/react': patch
+'graphiql': minor
+'@graphiql/react': minor
 ---
 
 Add user facing setting for persisting headers
--- a/.changeset/sweet-foxes-drive.md
+++ b/.changeset/sweet-foxes-drive.md
@@ -1,0 +1,6 @@
+---
+'graphiql': patch
+'@graphiql/react': patch
+---
+
+Add user facing setting for persisting headers
--- a/packages/graphiql-react/src/editor/__tests__/tabs.spec.ts
+++ b/packages/graphiql-react/src/editor/__tests__/tabs.spec.ts
@@ -1,7 +1,10 @@
+import { StorageAPI } from '@graphiql/toolkit';
 import {
   createTab,
   fuzzyExtractOperationName,
   getDefaultTabState,
+  clearHeadersFromTabs,
+  STORAGE_KEY,
 } from '../tabs';
 
 describe('createTab', () => {
@@ -139,5 +142,31 @@ describe('getDefaultTabState', () => {
         }),
       ],
     });
+  });
+});
+
+describe('clearHeadersFromTabs', () => {
+  const createMockStorage = () => {
+    const mockStorage = new Map();
+    return mockStorage as unknown as StorageAPI;
+  };
+
+  it('preserves tab state except for headers', () => {
+    const storage = createMockStorage();
+    const stateWithoutHeaders = {
+      hash: 123,
+      response: {
+        a: 'test',
+      },
+    };
+    const stateWithHeaders = {
+      ...stateWithoutHeaders,
+      headers: `{ "authorization": "secret" }`,
+    };
+    storage.set(STORAGE_KEY, JSON.stringify(stateWithHeaders));
+
+    clearHeadersFromTabs(storage);
+
+    expect(JSON.parse(storage.get(STORAGE_KEY)!)).toEqual(stateWithoutHeaders);
   });
 });
--- a/packages/graphiql-react/src/editor/__tests__/tabs.spec.ts
+++ b/packages/graphiql-react/src/editor/__tests__/tabs.spec.ts
@@ -167,6 +167,9 @@ describe('clearHeadersFromTabs', () => {
 
     clearHeadersFromTabs(storage);
 
-    expect(JSON.parse(storage.get(STORAGE_KEY)!)).toEqual(stateWithoutHeaders);
+    expect(JSON.parse(storage.get(STORAGE_KEY)!)).toEqual({
+      ...stateWithoutHeaders,
+      headers: null,
+    });
   });
 });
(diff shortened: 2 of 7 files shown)


acao:
thank you so much for this @TheMightyPenguin !

acao:
> should we include a warning for the user in the dialog?
I think so, perhaps just a warning to make sure they trust the device they are persisting these secrets to.

> do we also need to do the same for tabs state?
it would be lovely if you could cover that case as well!

acao:
@TheMightyPenguin merged an upstream PR where @jonathanawesome fixed the `act()` issue, sorry for the merge conflicts!

acao:
> This is only visible if the `shouldPersistHeaders` is not passed to the GraphQL component.

This seems like a sensible choice, however - what if this prop just determines the default value, and the user can choose to override it?

TheMightyPenguin:
> > This is only visible if the `shouldPersistHeaders` is not passed to the GraphQL component.
> 
> this seems sensible! what do you think of this?
> 
> * if `shouldPersistHeaders` is`false` explicitly, we should not show the setting
> * otherwise, if it is `true` or not specified, the user defined setting is available, defaulting to enabled/disabled respectively

I like it, seems considerate with the end user being in control of their privacy, I'll include this logic in the next commit!

TheMightyPenguin:
@acao updated PR, now we only show the setting if the property is not set (`undefined`) or `true`, and when it's set to false we also clean up the tabs storage, removing any headers key in there, I added tests for this logic.

I also added a warning message (using the warning color, let me know is the styling is not right 😅):
image

thomasheyenbrock (changes requested):
I'm not sure if we should add the state for user-controlling this setting to the `EditorContextProvider`. I feel that this is more of a thing that should happen in user-land (from the perspective of `@graphiql/react`).

To give an example: If I'm building my own custom GraphQL IDE with the components from `@graphiql/react`, I might have a clear opinion on how I want to persist headers or not. I already have the `shouldPersistHeaders` prop that I can use to control this behavior. Changing this state with the `setShouldPersistHeaders` method added in this PR gives me too much control that I don't need.

My proposal here is to move the state for the new toggle button in the settings dialog to the `GraphiQL` component. This component can combine this state with the value it receives as `shouldPersistHeaders` prop and pass the "final" value for that prop to the `GraphiQLProvider`. It also passes the value and the state-setter down to the `GraphiQLInterface` component which will use it to render the toggle button in the settings dialog. That way we don't need to expose any more state or methods from the `EditorContext` which keep it more easy to use.

Finally: Having a chunk of logic in the `EditorContextProvider` that clears the persisted header state when the `shouldPersistHeaders` prop changes is a fantastic addition, so we should definitely leave that in! (In fact, I would even consider it a bug that we're not doing it yet.)

TheMightyPenguin:
> I'm not sure if we should add the state for user-controlling this setting to the `EditorContextProvider`. I feel that this is more of a thing that should happen in user-land (from the perspective of `@graphiql/react`).
> 
> To give an example: If I'm building my own custom GraphQL IDE with the components from `@graphiql/react`, I might have a clear opinion on how I want to persist headers or not. I already have the `shouldPersistHeaders` prop that I can use to control this behavior. Changing this state with the `setShouldPersistHeaders` method added in this PR gives me too much control that I don't need.
> 
> My proposal here is to move the state for the new toggle button in the settings dialog to the `GraphiQL` component. This component can combine this state with the value it receives as `shouldPersistHeaders` prop and pass the "final" value for that prop to the `GraphiQLProvider`. It also passes the value and the state-setter down to the `GraphiQLInterface` component which will use it to render the toggle button in the settings dialog. That way we don't need to expose any more state or methods from the `EditorContext` which keep it more easy to use.
> 
> Finally: Having a chunk of logic in the `EditorContextProvider` that clears the persisted header state when the `shouldPersistHeaders` prop changes is a fantastic addition, so we should definitely leave that in! (In fact, I would even consider it a bug that we're not doing it yet.)

@thomasheyenbrock I tried implementing this, but there's an issue as the `GraphiQL` would not have access to the storage context. We could make this work somehow but not sure how clean the solution would be 😅. Gonna think this one a bit more to see what workaround we could use.

Additionally, I'd say we could always show the setting in the GraphiQL component, removing the logic we have for showing this in only certain scenarios (e.g. when the prop is `true` or `undefined`) would simplify the code.

TheMightyPenguin:
@acao @thomasheyenbrock pushed a few commits addressing the comments!

acao (approved):
Everything looks in order to me, and I concur with your choice to make the option always available!

jonathanawesome:
Looks good to me, thanks @TheMightyPenguin for the thorough PR!

jonathanawesome (approved):

TheMightyPenguin:
@acao @thomasheyenbrock @jonathanawesome anything missing from my side in this PR? (just making sure I am not blocking anything)

thomasheyenbrock (approved):
Awesome, this looks fine for me 💯

PR #2903, opened 2022-11-12
Fix react testling library context errors

### LINKED ISSUE #2892: [graphiql] Seeing a lot of `act` related warnings in tests

jonathanawesome:
This PR removes the act() warnings we've been experiencing with tests. Apologies for the lack of more precise commits, turned out there were additional issues that I discovered while updating the tests that just need to be fixed in place. These included updating tests that tested _no longer used_ props and removing the ts-nocheck comment and fixing up the TS bits.

I've also updated the testing-library packages to the nearest allowed version with our React 17 requirement.

There are still two `no-op` errors in the output that, based on [this](https://github.com/facebook/react/pull/22114) Facebook PR, are likely false positives. These errors have been removed in React 18 but won't be back-ported to React 17.

Resolves #2892

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "@cypress/webpack-preprocessor": "^5.5.0",
     "@manypkg/get-packages": "^1.1.3",
     "@strictsoftware/typedoc-plugin-monorepo": "^0.3.1",
-    "@testing-library/jest-dom": "^5.4.0",
+    "@testing-library/jest-dom": "5.16.5",
     "@types/aws-serverless-express": "^3.3.3",
     "@types/codemirror": "^0.0.90",
     "@types/express": "^4.17.11",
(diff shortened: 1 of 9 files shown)


acao (approved):
Thank you for all of this!!! The tests run so much more happily now

PR #2905, opened 2022-11-13
fix: prevent default event for graphiql-doc-explorer-back link

ccbrown:
This PR prevents the default event when clicking the doc explorer's back button. Without this change, clicking the back button will cause the page to navigate away in some use-cases.

### DIFF
--- a/.changeset/loud-guests-drive.md
+++ b/.changeset/loud-guests-drive.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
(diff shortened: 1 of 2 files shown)


acao:
thank you @ccbrown ! can you add a changeset for the bugfix?

ccbrown:
> thank you @ccbrown ! can you add a changeset for the bugfix? and i think you need to run `yarn format` if the commit hook doesn't run.

Done!

thomasheyenbrock (approved):
Thanks 🙌

Issue #2904, opened 2022-11-12
alternative state management RFC


jonathanawesome:
As per the discussion at this week's working group meeting, I'll be putting together a proposal for replacing the use of React Context with an alternative state manager. Goals for this RFC are to reduce component complexity in the `graphiql-react` package and to unlock the possibility to more easily provide GraphQL IDE building blocks for non-React frameworks.

Here are the tasks that I'd like to accomplish to make this a successful RFC:

- [x] Add a Ladle installation as a component development environment
- [x] Pull `explorer` plugin out of `graphiql-react` and into a separate plugin package
  - [x] rewrite context -> zustand 
  - [x] refactor tests
- [x] Pull `history` plugin out of `graphiql-react` and into a separate plugin package
  - [x] rewrite context -> zustand 
  - [x] refactor tests
 - [ ] Add a new `store` folder to `graphiql-toolkit` that contains state slices that match up with existing contexts
  - [x] storage
  - [x] editor
  - [x] schema
  - [x] execution
  - [x] plugin

dimaMachina:
closed by https://github.com/graphql/graphiql/pull/3946

PR #2906, opened 2022-11-13
DO NOT MERGE: alternative state management RFC

jonathanawesome:
Tracking issue #2904


acao:
Decided not to invite y’all as proper reviewers yet because it’s still WIP but @B2o5T and @thomasheyenbrock we feel that we can rework state management in a 2.x refactor to make the rest of the planned features and plugin API easier to build, and for more intelligent & efficient component rendering - what do you think?

ladle/src/utils/testSchema.ts
```diff
@@ -0,0 +1,398 @@
...
+      },
+    },
+  }),
+});
+
+const sleep = async (timeout: number) =>
+  new Promise(res => setTimeout(res, timeout));
+
+const longDescription = `
+The \`longDescriptionType\` field on the \`Test\` type has a long, verbose, description to test inline field docs.
```
TallTed, line 205:
```suggestion
The \`longDescriptionType\` field on the \`Test\` type has a long, verbose description to test inline field docs.
```

thomasheyenbrock (commented):
Love the idea of moving the doc explorer into its own plugin package! Aligns nicely with all the other plugins, both existing ones and future ones.

My main question here is if we can make this backwards compatible with the existing hooks from v2. In other words, can we keep exporting the `ExplorerContext` and the `useExplorerContext` hook while moving all the actual state management from the `ExplorerContextProvider` component into `zustand`?

I guess it would not require a major version bump for `graphiql` either way, but we should decide if we're fine with breaking stuff for folks that use `@graphiql/react` directly.

acao:
@thomasheyenbrock yes this was part of the original plugin proposal, the idea was that the plugin API should be able to support doc explorer, query history, etc.

ladle/yarn.lock
```diff
@@ -0,0 +1,4354 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
B2o5T, line 1: this lock file should be removed since you added laddle to workspace

packages/graphiql-plugin-react-doc-explorer/.eslintrc
```diff
@@ -0,0 +1,22 @@
+{
+  "extends": [
```
B2o5T, line 2:
```suggestion
  "root": true,
  "extends": [
```
this config do nothing, without `root: true` always root eslint config is taken in priority

packages/graphiql-plugin-react-doc-explorer/src/components/DocExplorer/DocExplorer.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <section
+      className="graphiql-doc-explorer"
+      aria-label="Documentation Explorer"
+    >
+      <div className="graphiql-doc-explorer-header">
+        <div className="graphiql-doc-explorer-header-content">
+          {prevName && (
+            <button
+              // href="#"
```
B2o5T, line 78:
```suggestion
```

ladle/.eslintrc
```diff
@@ -0,0 +1,19 @@
+{
```
B2o5T, line 1:
```suggestion
{
  "root": true,
```

same here

ladle/.eslintrc
```diff
@@ -0,0 +1,19 @@
...
+  "extends": [
+    "plugin:prettier/recommended",
```
B2o5T, line 3:
```suggestion
```
I don't like `eslint-plugin-pretter` since he make a lot of noise, anyway prettier check action is enabled in ci

dimaMachina:
repalced by https://github.com/graphql/graphiql/pull/3946, tracking progress in https://github.com/graphql/graphiql/issues/3874

PR #2907, opened 2022-11-14
show multipart requests as running until they are finished or cancelled

thomasheyenbrock:
I noticed this after looking at https://github.com/graphql/graphiql/pull/2885: We're currently not showing the stop button while a multipart request (e.g. a subscription) is running. The only state provided by the context is the boolean `isFetching` which is `true` only until the first (partial) response is returned.

This PR solves the issue by exposing another boolean `isSubscribed` which is `true` if there is an active subscription running (i.e. the fetcher function returned an unsubscribable).

Sidenote: We can't just change the semantics of the `isFetching` state. It also controls the loading spinner which should indeed only be shown while the first response is loading.

### DIFF
--- a/.changeset/rich-humans-rescue.md
+++ b/.changeset/rich-humans-rescue.md
@@ -1,0 +1,7 @@
+---
+'@graphiql/react': minor
+---
+
+Clearly separate between the fetching and subscription state for multipart
+requests (like subscriptions) and show the stop-button as long as the
+subscription is running
--- a/.changeset/rich-humans-rescue.md
+++ b/.changeset/rich-humans-rescue.md
@@ -2,6 +2,6 @@
 '@graphiql/react': minor
 ---
 
-Clearly separate between the fetching and subscription state for multipart
+Clearly separate the fetching and subscription states for multipart
 requests (like subscriptions) and show the stop-button as long as the
 subscription is running
--- a/packages/graphiql-react/src/execution.tsx
+++ b/packages/graphiql-react/src/execution.tsx
@@ -19,11 +19,17 @@ import { createContextHook, createNullableContext } from './utility/context';
 
 export type ExecutionContextType = {
   /**
-   * If there is currently a GraphQL request in-flight. For long-running
-   * requests like subscriptions this will be `true` until the request is
-   * stopped manually.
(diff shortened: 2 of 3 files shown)


thomasheyenbrock:
Also related to https://github.com/graphql/graphiql/issues/2774

.changeset/rich-humans-rescue.md
```diff
@@ -0,0 +1,7 @@
+---
+'@graphiql/react': minor
+---
+
+Clearly separate between the fetching and subscription state for multipart
```
TallTed, line 5:
```suggestion
Clearly separate the fetching and subscription states for multipart
```

packages/graphiql-react/src/execution.tsx
```diff
@@ -19,11 +19,17 @@ import { createContextHook, createNullableContext } from './utility/context';
 
 export type ExecutionContextType = {
   /**
-   * If there is currently a GraphQL request in-flight. For long-running
-   * requests like subscriptions this will be `true` until the request is
-   * stopped manually.
+   * If there is currently a GraphQL request in-flight. For multi-part
+   * requests like subscriptions this will be `true` while fetching the
```
TallTed, line 23:
```suggestion
   * requests like subscriptions, this will be `true` while fetching the
```

packages/graphiql-react/src/execution.tsx
```diff
@@ -19,11 +19,17 @@ import { createContextHook, createNullableContext } from './utility/context';
...
+   * first partial response and `false` while fetching subsequent batches.
    */
   isFetching: boolean;
+  /**
+   * If there is currently a GraphQL request in-flight. For multi-part
+   * requests like subscriptions this will be `true` until the last batch
```
TallTed, line 29:
```suggestion
   * requests like subscriptions, this will be `true` until the last batch
```

.changeset/rich-humans-rescue.md
thomasheyenbrock: Improve changeset wording

thomasheyenbrock:
Awesome, thanks for the review @TallTed 🙌

PR #2908, opened 2022-11-14
rename the `initialTabs` prop to `defaultTabs`

thomasheyenbrock:
Following up on https://github.com/graphql/graphiql/pull/2821 (I didn't manage to review even though it was open for a while, sorry for that 🙈 ).

This PR suggests a different naming for the `initialTabs` prop. The word "initial" suggests that GraphiQL will always spawn with a certain set of tabs if I pass them to this prop. But it's actually just the default set of tabs that will be shown if there is no tab state persisted in `localStorage`. Imho, `defaultTabs` would be a better wording here and also aligns with other props like `defaultQuery`.

As renaming would be a breaking change now (the new prop is already released), we deprecate the existing prop and add a new one. The deprecated prop can then be removed in a future major version.

### DIFF
--- a/packages/graphiql-react/src/editor/context.tsx
+++ b/packages/graphiql-react/src/editor/context.tsx
@@ -174,8 +174,9 @@ export type EditorContextProviderProps = {
    */
   initialTabs?: TabDefinition[];
   /**
-   * This prop can be used to defined the initial set of tabs with their queries,
-   * variables and headers.
+   * This prop can be used to define the default set of tabs, with their
+   * queries, variables, and headers. It will be used as default only if
+   * there is no tab state persisted in storage.
    *
    * @example
    * ```tsx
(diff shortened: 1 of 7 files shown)


packages/graphiql-react/src/editor/context.tsx
```diff
@@ -169,21 +169,25 @@ export type EditorContextProviderProps = {
    * typing in the editor.
    */
   headers?: string;
+  /**
+   * @deprecated Use `defaultTabs` instead.
+   */
+  initialTabs?: TabDefinition[];
   /**
    * This prop can be used to defined the initial set of tabs with their queries,
```
jonathanawesome, line 177: Agreed that `defaultTabs` is more appropriate than `initialTabs`. Can we make the comment more clear? From your PR comment: "This prop is used to define the default set of tabs, with their queries, variables, and headers, that will be shown if there is no tab state persisted in localStorage".

thomasheyenbrock: Definitely! Adjusted the text 👍

PR #2913, opened 2022-11-16
Reproduction of a bug in validation of undefined variables

11bit:
Hello! 

This PR adds a unit test with the reproduction of a bug in undefined variables validation: `getDiagnostics` function returns an "undefined variable" error twice and has the wrong end position for the second entry.

For instance for a following query on star wars schema:

`{ hero(episode: $ep) { name } }`

1. it returns two "Variable \\"$ep\\" is not defined." erros instead of one 
2. The second error has end position = 48 which is greater than the query itself

### DIFF
--- a/packages/graphql-language-service/src/interface/__tests__/getDiagnostics-test.ts
+++ b/packages/graphql-language-service/src/interface/__tests__/getDiagnostics-test.ts
@@ -46,33 +46,37 @@ describe('getDiagnostics', () => {
     expect(error.source).toEqual('GraphQL: Validation');
   });
 
-  it('validates undefined variables', () => {
+  it('catches with multiple highlighted nodes', () => {
     const errors = validateQuery(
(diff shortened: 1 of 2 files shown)


acao:
I think I've found the issue!

It seems that `graphql.validate()` now returns two nodes in this case, and the language service is currently set up to create the same error, but one is for operation node, and the other is for the variable callsite.

I'm sure there is a better way to do this, I think multi-range error nodes are possible for LSP, need to look into this?

image

acao:
Also, fixed a bug! This fixes a bug with range and other error details. The duplicate error will make more sense with this change, because it will apply to different ranges

image

11bit:
Thanks for debugging! This seems to be the right way to fix it. It will handle all the errors with multiple positions like duplicate names

11bit:
@acao I've copied your fix to PR. Now tests work, thanks!

Issue #2685, opened 2022-08-21
monaco-graphql recursive variables not working


dvh91:
schema def:
```
schema {
  query: query_root
}

input string_options {
  _eq: String
  _ilike: String
}

type issues {
  name: String
}

input issues_where_input {
  _and: [issues_where_input!]
  name: string_options
}

type query_root {
  issues(where: issues_where_input): [issues!]!
}
```

query:
```
query Example(
  $where: issues_where_input! = {}
) {
  issues(where: $where) {
    name
  }
}
```

note the `_and` in the `issues_where_input`. when it's there, the variables json failed to work properly and the whole monaco intellisense based on the query variables is broken. 

Not sure if this meant to be working or it's a known limitation of nested json schema generation behind the scenes

dvh91:
@acao maybe you can have a look on this one?
Saw this which I guess is related piece of code and thought you might have some insights on this 🙏 
https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts#L135

acao:
Can you open a PR to add a failing test case for this? I’m glad you found one!

dvh91:
> Can you open a PR to add a failing test case for this? Thank you for finding this bug!

Sure thing, thanks!
https://github.com/graphql/graphiql/pull/2686

acao:
it looks like this bugfix will take a while

PR #2686, opened 2022-08-22
chore: add recursive schema test

dvh91:

PR #2917, opened 2022-11-23
fix: avoid infinity recursiveness in getVariablesJSONSchema

woodensail:
add a marker to mark analyzed InputObject to avoid infinity recursiveness while generating JsonSchema.

see https://github.com/graphql/graphiql/issues/2685

### DIFF
--- a/.changeset/fifty-points-share.md
+++ b/.changeset/fifty-points-share.md
@@ -2,4 +2,4 @@
 "graphql-language-service": patch
 ---
 
-Fix infinity recursiveness in getVariablesJSONSchema when the schema contains types that reference themselves
+Fix infinite recursiveness in getVariablesJSONSchema when the schema contains types that reference themselves
--- a/.changeset/fifty-points-share.md
+++ b/.changeset/fifty-points-share.md
@@ -1,0 +1,5 @@
+---
+"graphql-language-service": patch
+---
+
+Fix infinity recursiveness in getVariablesJSONSchema when the schema contains types that reference themselves
--- a/packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
+++ b/packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
@@ -106,14 +106,29 @@ const scalarTypesMap: { [key: string]: JSONSchema6TypeName } = {
   DateTime: 'string',
 };
 
+class Marker {
+  private set = new Set<string>()
+  mark(name: string): boolean {
+    if (this.set.has(name)) {
+      return false
+    } else {
+      this.set.add(name)
+      return true
+    }
+  }
+}
+
 /**
  *
  * @param type {GraphQLInputType}
+ * @param options
+ * @param definitionMarker
  * @returns {DefinitionResult}
  */
 function getJSONSchemaFromGraphQLType(
   type: GraphQLInputType | GraphQLInputField,
   options?: JSONSchemaOptions,
+  definitionMarker: Marker = new Marker(),
 ): DefinitionResult {
   let required = false;
   let definition: CombinedSchema = Object.create(null);
@@ -137,6 +152,7 @@ function getJSONSchemaFromGraphQLType(
     const { definition: def, definitions: defs } = getJSONSchemaFromGraphQLType(
       type.ofType,
       options,
+      definitionMarker,
     );
     if (def.$ref) {
       definition.items = { $ref: def.$ref };
@@ -154,6 +170,7 @@ function getJSONSchemaFromGraphQLType(
     const { definition: def, definitions: defs } = getJSONSchemaFromGraphQLType(
       type.ofType,
       options,
+      definitionMarker,
     );
     definition = def;
     if (defs) {
@@ -162,7 +179,7 @@ function getJSONSchemaFromGraphQLType(
       });
     }
   }
-  if (isInputObjectType(type)) {
+  if (isInputObjectType(type) && definitionMarker.mark(type.name)) {
     definition.$ref = `#/definitions/${type.name}`;
     const fields = type.getFields();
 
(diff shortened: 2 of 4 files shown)


acao:
Thank you @woodensail !!

acao (changes requested):
Thank you for this fix! It was much needed, and you chose a path that it seems will improve performance greatly. I see you've signed the CLA already, awesome thank you!


One thing I'm wondering about:

the `definitionMarker` argument - I don't think it's necessary to provide, because we do not export `getJSONFromGraphQLType()` method itself. Do you have a use case? If so, we just need to add it to the `getVariablesJSONSchema`, perhaps it can just be added to the `options` object? The options object pattern I'm trying to use more (across the language service in general) because it behaves more like named parameters, and thus we can more easily extend the public interface,  deprecate options, etc without worrying about ordered arguments.


Beyond that, just a few clean-up items:

1. it appears one of the unit tests was actually false positive and is thus now failing - it was missing `markdownDescription` when it should have been expected.
1. quick `yarn format`
1. add a test case for this like we have here https://github.com/graphql/graphiql/pull/2686/files#diff-e6ad170730835d9d63d3161ad437190a1549198ee61a0da3fd6b19b0aa8947a4R261
1. the comment above mentions a changeset, if you could, to document the change.

woodensail:
thank you for your reply. This is my first commit to an open source project and I'm thrilled to hear from you.

It is indeed an inelegant way to use definitionMarker as an optional parameter, because I couldn't find a good place to put it at the time. 
Putting it into options is indeed a good way, but it seems that this will expose the marker to the upper layer. I plan to encapsulate JSONSchemaOptions as follows. 
```
export type JSONSchemaOptions = {
  useMarkdownDescription?: boolean;
};
// the wrapped type
type JSONSchemaRunningOptions = JSONSchemaOptions & {
  definitionMarker?: Marker;
}

const runtimeOptions:JSONSchemaRunningOptions = {
  ...options,
  definitionMarker: new Marker(),
}
getJSONSchemaFromGraphQLType(type, runtimeOptions);
```

Do you think this modification is appropriate? Looking forward to your reply. In order to avoid submitting too much useless code, I will submit it later until I receive your reply.

Finally, since I am not good at English, I am using Google Translate to communicate, I hope it will not affect your reading.

packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
```diff
@@ -106,14 +106,29 @@ const scalarTypesMap: { [key: string]: JSONSchema6TypeName } = {
...
+ * @param options
+ * @param definitionMarker
  * @returns {DefinitionResult}
  */
 function getJSONSchemaFromGraphQLType(
   type: GraphQLInputType | GraphQLInputField,
   options?: JSONSchemaOptions,
+  definitionMarker: Marker = new Marker(),
 ): DefinitionResult {
   let required = false;
```
acao, line 133: Just one change I'd like to suggest here:

```ts
function getJSONSchemaFromGraphQLType(
  type: GraphQLInputType | GraphQLInputField,
  options?: JSONSchemaOptions,
  definitionMarker: Marker = new Marker(),
): DefinitionResult {
  let required = false;
  ````
  
I think should become:
  
  ```ts
function getJSONSchemaFromGraphQLType(
  type: GraphQLInputType | GraphQLInputField,
  options?: JSONSchemaOptions,
): DefinitionResult {
  const definitionMarker = new Marker();
  let required = false;
  ````
  
  Unless there is a reason we need to expose this argument on this private, internal method?

**Comment by acao** [comment]
> thank you for your reply. This is my first commit to an open source project and I'm thrilled to hear from you.

that's so exciting! this is an incredibly useful contribution for your first open source commit :)

File: packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts (hunk 1 above)
Comment (acao) [reply packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts:133]: > Putting it into options is indeed a good way, but it seems that this will expose the marker to the upper layer. I plan to encapsulate JSONSchemaOptions as follows.

Ohhh I understand better now, that makes a lot of sense!

**Comment by acao** [comment]
Everything looks good to me! Now just these items left @woodensail

>     1. test fix - it appears one of the unit test was actually false positive and is thus now failing - it was missing `markdownDescription` when it should have been expected.
>     2. runk `yarn format` to fix the linting
>     3. add a test case for this like we have here [#2686 (files)](https://github.com/graphql/graphiql/pull/2686/files#diff-e6ad170730835d9d63d3161ad437190a1549198ee61a0da3fd6b19b0aa8947a4R261) if you could?
>     4. the `changeset-bot` comment above mentions adding a changeset, if you could, to document the change.

**Review by acao — APPROVED** [review]

**Comment by acao** [comment]
ah, i see, from what I'm getting in tests,  the bugfix itself is not quite finished yet. I will continue with this this week, and push these changes to this branch

**Comment by acao** [comment]
This change makes sure that the references are correct, and fixes a few bugs
image

 but the descriptions are not correct
 
image

**Comment by woodensail** [comment]
Sorry it's been a long time since the last submission, I've been a little busy recently.

In the new commit, I put the marker into the running options, and by the way fixed a previous error.
In the previous submission, I skipped the follow-up logic directly after detecting duplicate objects, resulting in the lack of `$ref` field in the parsed schema. This time I added the `$ref` field, although the two layers of if look a bit ugly.
This error will cause some test cases to report errors before. I am not sure that this is not the [false positive test case] you mentioned before, because I did not see the error message about markdownDescription. But anyway, the test cases can pass now.

about other issues
1：I ran `yarn format` and it modified many lines, even parts of the code that I didn't touch
2:  I just copied the test case you mentioned. Is it inappropriate to do so? I don't know how to do it correctly, should I roll back the commit of the test case and let you directly merge the pull request mentioned above?
3：The change-set has been submitted, although there is only one line of content

File: .changeset/fifty-points-share.md [hunk 2]
```diff
@@ -0,0 +1,5 @@
+---
+"graphql-language-service": patch
+---
+
+Fix infinity recursiveness in getVariablesJSONSchema when the schema contains types that reference themselves
```
TallTed, line 5:
```suggestion
Fix infinite recursiveness in getVariablesJSONSchema when the schema contains types that reference themselves
```

acao:
@woodensail looks good to me! the test case should have merged. the formatting would have happened like that if there were upstream formatting issues still in your branch. thanks for this!

acao (approved):

PR #2923, opened 2022-11-29
Remove side-effect in StorageAPI that overrides localStorage.clear

TheMightyPenguin:
# Description

I noticed we were overriding `localStorage.clear` method when the `StorageAPI` constructor ran, and this could cause unexpected behavior when someone is building a GraphQL Editor and using the `StorageContext`.

This makes sure we are not affecting the local storage prototype.

### DIFF
--- a/packages/graphiql-toolkit/src/storage/__tests__/base.spec.ts
+++ b/packages/graphiql-toolkit/src/storage/__tests__/base.spec.ts
@@ -1,7 +1,11 @@
 import { StorageAPI } from '../base';
 
 describe('StorageAPI', () => {
-  const storage = new StorageAPI();
+  let storage = new StorageAPI();
+
+  beforeEach(() => {
+    storage = new StorageAPI();
+  });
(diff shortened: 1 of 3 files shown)


packages/graphiql-toolkit/src/storage/__tests__/base.spec.ts
```diff
@@ -1,7 +1,11 @@
 import { StorageAPI } from '../base';
 
 describe('StorageAPI', () => {
-  const storage = new StorageAPI();
+  let storage = new StorageAPI();
+
+  beforeEach(() => {
+    storage = new StorageAPI();
+  });
```
TheMightyPenguin, line 8: these changes are not required but I thought it could be nice to have a clean slate in every test.

acao: thank you!

acao (approved):

PR #2407, opened 2022-05-17
Moving `graphql-ws` to a dynamic import and display a nice message when not found

benjdlambert:
Fixes #2405

PR #2925, opened 2022-11-30
DRAFT: Migrate graphiql package to vite

jonathanawesome:
This PR is the latest attempt to move away from Webpack and into Vite. 

The Vite config turned out to be much more non-standard that I'd expected. Because we need to prep a minified bundle and a non-minified bundle (for the Netlify deployments), and because Vite's library build mode doesn't allow processing of the required `index.html` file, there's some hacky bits regarding the Vite configs that I'm not particularly fond of. That said, it mostly works.

I've left all of the webpack-related code in for now, except the build scripts, for reference.

Here are the important notes for reviewers:

#### Broken subscriptions/e2e
Most critical is an error originating from `createFetcher` and the usage of `require` for providing a user-friendly error when `graphql-ws` is not installed as a dependency (introduced in #2407). This will die in the browser when attempting to execute a subscription (screenshot below) and, most importantly, causes the e2e test run to fail. I haven't thought of a fix nor was I interested in cramming a fix into this PR without first reviewing as a group.

Screen Shot 2022-11-29 at 5 47 20 PM

#### Removal of `graphiql.min.css.map`
I struggled for a short while to get Vite to cough up a minified css source map without a plugin, then decided to abandon. I don't think removing the css source map on the _development_ Netlify build is that big of a deal, but I'm happy to be wrong upon review and find a solution.

#### Bundle file target directory
I've changed the target directory for the bundled files, moving them from the root of the package to `/build`. If this interferes with packaging it can be changed, but we'll need to perform some additional Vite magic because Vite expects an `index.html` to live in the root and we can't have the bundler overwriting it. The new `index.html` is a template that is used in development mode _and_ to produce the `dev.html` and `index.html` files that are used for the Netlify deployments...the bulk of this PR deals with hacking around this file and Vite's requirements for it. Happy to be wrong about my implementation and adjust upon review, but I _think_ I've got it as clean as could be.

#### Non ideal dev mode
The script for dev mode starts the e2e server, starts a Vite build mode to watch for changes to `/components/GraphiQL.tsx` and rebuild the bundle, and starts the Vite dev server. It's complicated and not as lightning fast as you might expect with Vite, but necessary unless we want to jump through hoops to adhere to the "standard" Vite library mode setup. Early on in this work I had Ladle running and pointed directly at the GraphiQL component file, and that worked great, but it seemed even more awkward than the current state without a more well thought out approach to UI development throughout the repository. Ideally, we'll bring back Ladle or Storybook at some point in the near future to cover _all_ of the UI development needs for UI-related packages and leave those packages to just handle bundling. @thomasheyenbrock has the most recent experience here and, I assume, was dev'ing `/graphiql-react` via `graphiql`...is that assumption correct?

---

Here's what I've got for a punch list:
- [ ]  Find a solution for the `createFetcher`/`require` situation
- [ ]  Come to a decision about the removal of `graphiql.min.css.map`
- [ ]  Come to a decision about the modified bundle target directory
- [ ]  Remove unused Webpack-related code
- [ ]  Optional: use this opportunity to remove _other_ old and unused build/tooling related code


packages/graphiql/resources/renderExample.js
```diff
@@ -67,7 +67,7 @@ function updateURL() {
 }
 
 function getSchemaUrl() {
-  const isDev = window.location.hostname.match(/localhost$/);
+  const isDev = window.location.hostname.match(/127.0.0.1$/);
```
B2o5T, line 70:
```suggestion
  const isDev = window.location.hostname.match(/127\.0\.0\.1$/);
```
in regex dot means any character

packages/graphiql/vite/GraphiQLHTMLPlugin.ts
```diff
@@ -0,0 +1,63 @@
...
+      } else {
+        // we're in vite's "production" mode, so this is replacing/building/writing build/index.html
+        index = doReplace({min: true});
+
+        // we have to handle on our own the writing of our "dev.html" file
+        // this "dev.html" is used for the uncompressed netlify deployment
+        // i wonder if this might not be relevant any longer...?
+        const dev = doReplace({min: false});
+        
+        fs.writeFile(`${BUILD_DIR}/dev.html`, dev, err => {
```
B2o5T, line 51: async version can be used instead of callbacks

packages/graphiql/vite/sharedBuildConfig.ts
```diff
@@ -0,0 +1,32 @@
+import type { BuildOptions } from 'vite'
+
+// constants
+import { BUILD_DIR } from './constants'
+
+export const sharedBuildConfig: BuildOptions =  {
+  sourcemap: process.env.MIN ? true : false,
+  minify: process.env.MIN ? true : false,
```
B2o5T, line 8:
```suggestion
  sourcemap: !!process.env.MIN,
  minify: !!process.env.MIN,
```

packages/graphiql/vite/sharedCSSConfig.ts
```diff
@@ -0,0 +1,10 @@
+import type { CSSOptions } from 'vite';
+
+// utils
+import path from 'path';
+
+export const sharedCSSConfig: CSSOptions =  {
+    postcss: {
+      file: path.resolve(__dirname, 'src/css'),
```
B2o5T, line 8:
```suggestion
    postcss: {
      file: path.resolve(__dirname, 'src', 'css'),
```

acao:
this gets us so much closer to fixing the `build` vs `build-bundles` issue!
now, i'm going to just rename some of these plugins to use `build-bundles` with `wsrun -s` so it executes in order of dependencies.

then `build` can be back to pure `tsc --build` project references, for all the type checks I see for underlying `graphiql` plugins etc

acao:
@jonathanawesome in fact... I think i've figured out how to consolidate all of the graphiql packages & plugins into one consolidated vite config, by automatically resolving `rollupOptions.entry` and `rollupOptions.output` and/or with `lib.entry` https://github.com/vitejs/vite/discussions/1736#discussioncomment-3812904. We can generate all of the config from the package names at build time.

acao:
Oof. So it may be that this pattern for dev/prod builds will need to continue. Worth noting this _is_ possible in webpack 5, technically it's also possible in webpack 4 by returning two configs as an array, but I have the env flag split approach there as well.

image

However everything else - harmony between the build and build-bundles and monorepo wide dev server will still be doable!

acao:
i think this could be our happy place... 😻  this would be a deus ex machina if it can work for us... it would solve the `build` vs `build-bundles` riddle transparently.  turborepo runs up against some of my preferences (per-workspace scripts for example) but we can add some plop generators for adding new workspaces (graphiql plugin or generic for starters)

https://github.com/vercel/turbo/blob/b7bcfc312e41367e2ec7f5fb37a6ee88a6b6545a/examples/kitchen-sink/package.json#L6

cometkim:
Heya! :wave: @acao  turbocache is a very simple demo project showing how to host a turborepo remote cache on Cloudflare Workers, so depending on your use case you may need to customize it, but it's very easy!

If you're looking for a bundler for libraries/binary that doesn't use css or html entries, you can keep watching at [nanobundle](https://github.com/cometkim/nanobundle). Very similar to microbundle, but uses esbuild. I'm are working on adding multi-entry support.

acao:
`turborepo` uses `tsup` in many examples which uses both `swc` and `esbuild`?

really, anywhere that `tsup` is used we could `nanobundle`, and the options are almost the same? for example `--dts` flag. haha interesting

we can also still use vite library mode with `turborepo` as well!

acao:
@cometkim I don't think we need multi-entrypoint support actually if we can use turborepo, i think `nanobundle` will suit us just fine as is for many cases (except for the few with css)! It has dts support, and support for multiple output formats, and many configuration options for the cases where we might need them, which is all we really need.

acao:
i hope to return to this effort next week!

acao:
I've got it working again on a new branch by plucking these changes onto the current main with `git checkout`. vite has since upgraded to 3.3.x 

> Vite expects an index.html to live in the root and we can't have the bundler overwriting it. 

agreed! this is how `graphiql` has always worked as well, at least in terms of where the index.html is always _written_ to. The source being there is funky. we can publish the demo from any directory, netlify isn't required to publish `packages/graphiql` as the root, but we _do_ need to make sure `graphiql.min.js` and .css and such are still at least _aliased_ to the root import paths for unpkg/jsdelivr etc

acao:
I've managed to address the require() issue in toolkit, bundling the library itself in a compatible way.
Now something is causing vite to not encode this unicode character properly in reach ui, have you seen this before?

https://github.com/reach/reach-ui/blob/43f450db7bcb25a743121fe31355f2294065a049/packages/popover/src/reach-popover.tsx#L338

malte-behrendt-gr:
> I've managed to address the require() issue in toolkit, bundling the library itself in a compatible way. Now something is causing vite to not encode this unicode character properly in reach ui, have you seen this before?
> 
> https://github.com/reach/reach-ui/blob/43f450db7bcb25a743121fe31355f2294065a049/packages/popover/src/reach-popover.tsx#L338

@acao Just to clarify: is the issue that something does not get rendered or that the variable name, restoreTabIndexTupl**é**s, contains a unicode character? 

If the first: do you have a link to a error message? 
If the latter, can't we just a normal 'e'?

Sytten:
What there ever a fix for the require, I am trying to use the package and its PITA

dimaMachina:
closed by https://github.com/graphql/graphiql/pull/3679

PR #2926, opened 2022-11-30
fix: support cts/mts extensions

### LINKED ISSUE #2855: [vscode-graphql-syntax] Configure extra file extensions for VSCode extension

The list of default extensions covered by the extensions does not include the new `cts`/`mts` TypeScript extensions, and as far as I can tell there is no way on client side to configure this.

Is it possible to add these extensions, or add a configuration option to allow for extra extensions to be added?

elijaholmos:
Resolves #2855 

Introduces support for the `.cts` and `.mts` in the VSCode extension(s) and the LSP server

note: no tests were created for this patch

### DIFF
--- a/.changeset/heavy-dancers-deliver.md
+++ b/.changeset/heavy-dancers-deliver.md
@@ -1,0 +1,7 @@
+---
+'graphql-language-service-server': patch
+'vscode-graphql': patch
+'vscode-graphql-syntax': patch
+---
+
+support cts and mts file extensions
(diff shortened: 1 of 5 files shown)


acao:
@elijaholmos looks good to me, thank you! shouldn't need additional tests since we are already testing the same logic with `.ts` files

I still get a spooky feeling from the idea of CJS + TS but haha I'm sure there are use cases

elijaholmos:
@acao no problem!

> I still get a spooky feeling from the idea of CJS + TS but haha I'm sure there are use cases

same... I can't wait until ESM is standardized

Issue #1620, opened 2020-07-21
Support more customization options? 


blacksider:
Currently ***graphiql*** only support several view customization options(children: Logo, ToolBar, Menu,...etc),  but I can't find a direct way to hide or show some elements, such as the whole toolbar, the resultviewer, or the docs panel, it would be great if there are options to control the ***visibility*** of some elements, and change the ***style*** of those elements.
Thanks!

lasmil:
+1

thomasheyenbrock:
With the release of `graphiql@2` last year, we also published a new package called `@graphiql/react`. This new package solves this issue, but not quite in the expected way. Let me explain.

Building in more customization options into GraphiQL would ever increase the API surface and maintenance complexity, which isn't something we wanted to do. Instead, GraphiQL is what we call the "**reference implementation for a GraphQL IDE**". It's merely using all the building blocks from `@graphiql/react` and sticks them together in a particular way.

If you want to customize GraphiQL, you have several options. Instead of reciting them here, here's a [link](https://github.com/graphql/graphiql/blob/main/packages/graphiql/README.md#customize) to the section of the README that goes into more detail. If these customization options do not solve your use-case, we encourage you to take the `GraphiQL` component as a template, create your own component and customize it to your needs.

As a final note, there might be things (state, UI components, ...) that are not exposed yet through `@graphiql/react`. If you find something that you'd need, feel free to open a new issue specifically for that thing.

PR #2929, opened 2022-12-01
1620: added customize options for some panels

lasmil:
- closes https://github.com/graphql/graphiql/issues/1620

- added the ability to hide/show the 'second panel'
- added the ability to hide/show just the 'editor panel' (inside second panel)
- added the ability to hide/show just the 'results panel' (inside second panel)
- added the ability to apply custom styles to the 'second panel' container
- added the ability to apply custom styles to the 'editor panel' container
- added the ability to apply custom styles to the 'results panel' container

I checked how the hiding of these panels would work and I noticed that the best option to implement the feature is using the 'display' CSS property when would like to hide the specific panel. The reason behind this is that the panel 'ref' does some calculations when the page is loaded. These are lost if we are "react" removing the panel, ie using the
`{ !hideSecondPanel && <div> ... </div> `}
syntax. Toggling the "displayness (display: none to flex)" fixes this issue and if we conditionally toggle these panels visibility, everything would work as it should.

I added separate optional properties related to styling. In this way, the user won't be confused when trying to use them.

Backward compatibility is provided, hence these new props are optional.

### DIFF
--- a/.changeset/rich-lamps-beg.md
+++ b/.changeset/rich-lamps-beg.md
@@ -1,0 +1,20 @@
+---
+'graphiql': minor
+---
+
+1) Added optional props which can be used to toggle the visibility of
+   - second panel — with the `hideSecondPanel` prop
+   - editor panel — with the `hideEditorPanel` prop
+   - results panel — with the `hideResultsPanel` prop
+
+These work independently of each other, so you can have the second panel visible, but the editor panel hidden.
+NOTE: If the second panel is hidden, the other two props will be ignored.
+
+2) Added optional props which can be used to add custom styles through class names:
+   - second panel — with the `secondPanelClassName` prop
+   - editor panel — with the `editorPanelClassName` prop
+   - results panel — with the `resultsPanelClassName` prop
+
+NOTE: These props are used only if the corresponding panel is not hidden.
+
+
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -198,10 +198,53 @@ export type GraphiQLInterfaceProps = WriteableEditorProps &
      * editor.
      */
     toolbar?: GraphiQLToolbarConfig;
+    /**
+     * Toggle if the second panel should be shown.
+     * @default false
+     */
+    hideSecondPanel?: boolean;
+    /**
+     * Toggle if the editor panel should be shown.
+     * @default false
+     */
+    hideEditorPanel?: boolean;
+    /**
+     * Toggle if the results panel should be shown.
+     * @default false
+     */
+    hideResultsPanel?: boolean;
+    /**
+     * Styles to be applied to the second panel.
+     * @default ''
+     */
+    secondPanelClassName?: String;
+    /**
+     * Styles to be applied to the editor panel.
+     * @default ''
+     */
+    editorPanelClassName?: String;
+    /**
+     * Styles to be applied to the results panel.
+     * @default ''
+     */
+    resultsPanelClassName?: String;
   };
 
 export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
   const isHeadersEditorEnabled = props.isHeadersEditorEnabled ?? true;
+  const hideSecondPanel = props.hideSecondPanel ?? false;
+  const hideEditorPanel = props.hideEditorPanel ?? false;
+  const hideResultsPanel = props.hideResultsPanel ?? false;
+
+  const secondPanelClassName = hideSecondPanel
+    ? 'graphiql-panel-hidden'
+    : `${props.secondPanelClassName} graphiql-panel-visible`;
+  const editorPanelClassName = hideEditorPanel
+    ? 'graphiql-panel-hidden'
+    : `${props.editorPanelClassName} graphiql-panel-visible`;
+  const resultsPanelClassName = hideResultsPanel
+    ? 'graphiql-panel-hidden'
+    : `${props.resultsPanelClassName} graphiql-panel-visible`;
 
   const editorContext = useEditorContext({ nonNull: true });
   const executionContext = useExecutionContext({ nonNull: true });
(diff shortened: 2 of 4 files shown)


.changeset/rich-lamps-beg.md
```diff
@@ -0,0 +1,20 @@
+---
+'graphiql': minor
+---
+
+1) Added optional props which can be used to toggle the visibility of
+- second panel - with the `hideSecondPanel` prop
+- editor panel - with the `hideEditorPanel` prop
+- results panel - with the `hideResultsPanel` prop
```
TallTed, line 8:
```suggestion
1) Added optional props which can be used to toggle the visibility of
   - second panel — with the `hideSecondPanel` prop
   - editor panel — with the `hideEditorPanel` prop
   - results panel — with the `hideResultsPanel` prop
```

.changeset/rich-lamps-beg.md
```diff
@@ -0,0 +1,20 @@
...
+
+These work independently of each other, so you can have the second panel visible, but the editor panel hidden.
+NOTE: If the second panel is hidden, the other two props will be ignored.
+
+2) Added optional props which can be used to add custom styles to:
+- second panel - with the `secondPanelStyle` prop
+- editor panel - with the `editorPanelStyle` prop
+- results panel - with the `resultsPanelStyle` prop
```
TallTed, line 16:
```suggestion
   - second panel — with the `secondPanelStyle` prop
   - editor panel — with the `editorPanelStyle` prop
   - results panel — with the `resultsPanelStyle` prop
```

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -198,10 +199,53 @@ export type GraphiQLInterfaceProps = WriteableEditorProps &
      * editor.
      */
     toolbar?: GraphiQLToolbarConfig;
+    /**
+     * Toggle if the second panel should be shown.
+     * @default false
+     */
+    hideSecondPanel?: boolean;
+    /**
+     * Toggle if editor panel should be shown.
```
TallTed, line 208:
```suggestion
     * Toggle if the editor panel should be shown.
```

acao:
@thomasheyenbrock and @jonathanawesome would you agree that this change should come along with:

1. overriding plugins with `plugins` prop
2. deprecating other props for hiding/disabling panes

another thought here @lasmil  - we are trying to use css classes for overriding styles more at this point  - I'll defer to my colleauges whether this new pattern of passing raw css-in-js props is what they had in mind

simoneb:
This contribution is being made by NearForm. We're working on the CLA.

lasmil:
> @thomasheyenbrock and @jonathanawesome would you agree that this change should come along with:
> 
> 1. overriding plugins with `plugins` prop
> 2. deprecating other props for hiding/disabling panes
> 
> another thought here @lasmil - we are trying to use css classes for overriding styles more at this point - I'll defer to my colleauges whether this new pattern of passing raw css-in-js props is what they had in mind

Sure thing. I can work with both options. I went with this option though because it was one of the easiest way to keep backward compatibility and also keep certain already present style values.
If you choose to go with css classes then I can either try to update this PR accordingly or we can create another issue for future reference.

We need this functionality in one of our repos, thus I ended up going with the easiest solution.

jonathanawesome:
Hey there @lasmil & @simoneb, thanks for the PR!

I'm not against expanding the API surface of `GraphiQLInterfaceProps`, especially if they're optional and tested, but I do think that an effort should be made to work within the convention of using class names for styling overrides. If adding class names to the target divs can get you folks the functionality you're after, that should be the way forward.

There was consideration early on in the new design that had an alternative view for plugins that needed more space and didn't need to interface directly with the editor(s):
Screen Shot 2022-12-01 at 12 45 41 PM

Outside of more general discussions around enhancing the plugin API, we haven't discussed specifically how this might work, though I wonder if this view might be a better fit for the plugin you're building. If so, we should take the opportunity to discuss!

lasmil:
> Hey there @lasmil & @simoneb, thanks for the PR!
> 
> I'm not against expanding the API surface of `GraphiQLInterfaceProps`, especially if they're optional and tested, but I do think that an effort should be made to work within the convention of using class names for styling overrides. If adding class names to the target divs can get you folks the functionality you're after, that should be the way forward.
> 
> There was consideration early on in the new design that had an alternative view for plugins that needed more space and didn't need to interface directly with the editor(s): Screen Shot 2022-12-01 at 12 45 41 PM
> 
> Outside of more general discussions around enhancing the plugin API, we haven't discussed specifically how this might work, though I wonder if this view might be a better fit for the plugin you're building. If so, we should take the opportunity to discuss!

Alright, I will see what can we do with the class names. 
The plugin we are building requires the entire second panel to be hidden, so I have taken this task to make the inner panels configurable as well.

acao:
@lasmil though these props will definitely be useful to users, especially dynamically - for your project in particular, you may find yourself limited in overriding layout, styling approach, etc as a whole - if so, consider just using `@graphiql/react` and `@graphiql/toolkit` directly?

`graphiql` itself is now one big layout component. It's meant to be a starting point for more complex use cases, then you can cut out/replace half the UI as needed, and still be able to use plugins if you import their CSS
https://github.com/graphql/graphiql/blob/main/packages/graphiql/src/components/GraphiQL.tsx

`@graphiql/react` components forward their refs so you can apply whatever css-in-js approach you like.

lasmil:
> @lasmil [Here's a sandbox](https://codesandbox.io/s/graphiql-full-view-plugin-eigzsu) that may be helpful!
> 
> Also, I'm on the [discord](https://discord.com/channels/625400653321076807/862957469155721236) if you'd like to chat.

Ohh I see what you did there. Alright, for our plugin, this can work and then we don't really need any of these changes.
Nevertheless we may keep these changes as well which can be useful in some scenarios as @acao mentioned.

lasmil:
PR updated with using class names instead of style objects.

thomasheyenbrock (changes requested):
Hey @lasmil 👋 thanks for the PR! There hasn't been a lot of activity here for a while, is the PR still relevant / up-to-date?

I have to admit, I'm not fond of adding a bunch of new props. It makes the API surface even more complex (it's already huge) and just harder to maintain in the long run.

GraphiQL is meant to be the reference implementation for a GraphQL IDE, and I see all panels as an essential part of that. Of course there is a good use-case for hiding one or the other, but there's good use-cases for a lot of things. That's also why we published `@graphiql/react` to enable everybody to build their own GraphQL IDEs with their own custom behavior.

Instead of having props for this, I could also imagine adding a simple `display: none` to the DOM elements that wrap the respective panels in user land. Would that also be a viable solution?

PR #2931, opened 2022-12-02
enable `no-negated-condition` and `no-else-return` rules

B2o5T:
they help improve code readability


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "graphiql-monorepo",
-  "version": "independent",
+  "version": "0.0.0",
```
B2o5T, line 3: this was incorrect 
image

acao: @B2o5T changing this used to impact other tooling, but we haven't used `lerna publish` for a while, so it should be fine now

acao:
> they help improve code readability

I agree so much! I've had two jobs and several reviews in this repo where if/else is considered more readable and it truly baffles me

acao (approved):

PR #2933, opened 2022-12-02
enable `@typescript-eslint/no-unused-expressions` rule

B2o5T:
found code that does nothing https://github.com/graphql/graphiql/blob/695100bd317940ff3ffd8f56b54248c1dba1ac04/packages/graphiql-react/src/utility/resize.ts#L158

a good case to enable this rule

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -142,6 +142,7 @@ module.exports = {
       'error',
       { argsIgnorePattern: '^_', ignoreRestSiblings: true },
     ],
+    '@typescript-eslint/no-unused-expressions': 'error',
 
     'no-use-before-define': 0,
 
--- a/examples/monaco-graphql-react-vite/src/App.tsx
+++ b/examples/monaco-graphql-react-vite/src/App.tsx
@@ -108,7 +108,7 @@ export default function App() {
     const variablesModel = getOrCreateModel('variables.json', defaultVariables);
     const resultsModel = getOrCreateModel('results.json', '{}');
 
-    queryEditor ??
+    if (!queryEditor) {
       setQueryEditor(
         createEditor(opsRef, {
           theme: 'vs-dark',
@@ -116,14 +116,16 @@ export default function App() {
           language: 'graphql',
         }),
       );
-    variablesEditor ??
+    }
+    if (!variablesEditor) {
       setVariablesEditor(
         createEditor(varsRef, {
           theme: 'vs-dark',
           model: variablesModel,
         }),
       );
-    resultsViewer ??
+    }
+    if (!resultsViewer) {
       setResultsViewer(
         createEditor(resultsRef, {
           theme: 'vs-dark',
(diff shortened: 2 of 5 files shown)


examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -108,22 +108,24 @@ export default function App() {
     const variablesModel = getOrCreateModel('variables.json', defaultVariables);
     const resultsModel = getOrCreateModel('results.json', '{}');
 
-    queryEditor ??
```
acao, line 111: it seems the eslint rule doesn't recognize these valid expressions though? I mean I'm not worried about making the code more clear, but this code works as intended?

packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -665,7 +665,9 @@ export class GraphQLCache implements GraphQLCacheInterface {
     const schemaKey = this._getSchemaCacheKeyForProject(
       projectConfig,
     ) as string;
-    schemaKey && this._schemaMap.delete(schemaKey);
```
acao, line 668: this is a valid expression as well? confused

acao (commented):
I'd like to see about enabling a few of these opt-in features for the rule.

My main nit with this rule is that the label `no-unused-expressions` makes it seem like some of these expressions are never executed. It's one thing to disable short circuit patterns for stylistic reasons, but the label of this rule makes it seem like it's invalid altogether? I don't understand why some of these rules are disabled by default, though I'm sure there is some good reason

packages/graphql-language-service-server/src/GraphQLCache.ts
B2o5T: It’s valid but with `if` statement it’s more readable and clear

examples/monaco-graphql-react-vite/src/App.tsx
B2o5T: it’s a valid, code works as intended, but refactoring to use `if` statement is more readable,

.eslintrc.js
```diff
@@ -142,6 +142,7 @@ module.exports = {
       'error',
       { argsIgnorePattern: '^_', ignoreRestSiblings: true },
     ],
+    '@typescript-eslint/no-unused-expressions': 'error',
```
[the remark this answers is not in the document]
B2o5T: I would still like to have `allowShortCircuit: false` since it will forbid code style in favour of `if` statements, but if you dislike just click on the `Commit suggestion` 😜

B2o5T:
> I don't understand why some of these rules are disabled by default, though I'm sure there is some good reason

@acao a lot of useful eslint rules are turned off by default and doesn't include in `recommended` config because they are stylistic rules, the `recommended` config contains only `problem` rules

thomasheyenbrock (approved):
+1 for enforcing `if` statements over short circuts for improved readability

B2o5T:
let's merge it as-is and do rule config adjustment in another pr in case of false negative cases

PR #2939, opened 2022-12-02
removes regenerator-runtime from cdn.ts

### LINKED ISSUE #2868: [graphiql] generated bundles of graphiql is incompatible with strict CSP

### Current Behavior

The top level readme in the repo states:
> codemirror-graphql and the graphiql browser bundle use the [.browserslistrc](https://github.com/graphql/graphiql/blob/main/.browserslistrc), which targets modern browsers to keep bundle size small and keep the language services performant where async/await is used, and especially to avoid the requirement of regenerator-runtime or special babel configuration.

But in fact regenerator-runtime is used [here](https://github.com/graphql/graphiql/blob/main/packages/graphiql/src/cdn.ts#L10)

If I run `build-bundles` in graphiql the generated graphiql.min.js file includes regenerator-runtime and actually differs from the version served here https://graphiql-test.netlify.app/ (the hosted version uses a newer build of regenerator-runtime which doesn't have this issue)

The easiest solution would be to remove regenerator-runtime by just removing the one line.

regenerator-runtime was updated recently to fix the issue https://github.com/facebook/regenerator/commit/cc0cde9d90f975e5876df16c4b852c97f35da436

If the lib should be kept, we could update regenerator-runtime to 0.14.8 or make sure that strict mode is _not_ used

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

_No response_

### Anything else?

_No response_

jonathanawesome:
resolves #2868

### DIFF
--- a/.changeset/honest-windows-scream.md
+++ b/.changeset/honest-windows-scream.md
@@ -1,0 +1,5 @@
+---
+"graphiql": patch
+---
+
+removes regenerator-runtime from cdn.ts, resolves #2868
--- a/packages/graphiql/src/cdn.ts
+++ b/packages/graphiql/src/cdn.ts
@@ -7,7 +7,6 @@
 import * as GraphiQLReact from '@graphiql/react';
 import { createGraphiQLFetcher } from '@graphiql/toolkit';
 import * as GraphQL from 'graphql';
-import 'regenerator-runtime/runtime';
 import { GraphiQL } from './components/GraphiQL';
 
 import '@graphiql/react/font/roboto.css';


acao:
thanks! don't forget the changeset so that we can publish a bug fix
I'm surprised it's still there! I'm pretty sure the browserslist we have now (is our browserslist up to date?)  means this is no longer required.

acao (approved):

PR #2940, opened 2022-12-02
🛠️ ESLint: enable `unicorn/prefer-node-protocol` rule

B2o5T:


packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -8,8 +8,9 @@
  */
 
 import mkdirp from 'mkdirp';
+// can't use node:fs since we mock in jest `fs` module
```
thomasheyenbrock, line 11: Would it work if we replace `jest.mock('fs')` with `jest.mock('node:fs')` in the test file? As an alternative, we could remove mock for the whole module and just spy on `readFileSync` in the individual tests.

B2o5T: I tried `jest.mock('node:fs')` but for some reason it not worked 🥲

B2o5T: @thomasheyenbrock you were right, I forget to modify `jest.requireMock('fs')` as well

thomasheyenbrock (approved):

PR #2952, opened 2022-12-06
feat: add panels hiding hooks to context for plugin use

matijagaspar:
Add the resizable panel state and setters to the context, so plugins can hide/show resizable panels and react on the changes in their visibility.

The context exposes the following object: 
```
export type InterfaceContextType = {
  pluginResize: {
    setHiddenElement: ReturnType<typeof useDragResize>['setHiddenElement'];
    hiddenElement: string | null;
  };
  editorResize: {
    setHiddenElement: ReturnType<typeof useDragResize>['setHiddenElement'];
    hiddenElement: string | null;
  };
  editorToolsResize: {
    setHiddenElement: ReturnType<typeof useDragResize>['setHiddenElement'];
    hiddenElement: string | null;
  };
};
```

A plugin can than use the said context to set hidden element or react on the change of it.
For example (inside a plugin):
```
const { pluginResize } = useInterfaceContext()
  useEffect(() => {
    pluginResize.setHiddenElement('second')
  }, [pluginResize])
```
This would allow the plugin to show as "full screen" hiding the editor by default.

The names and function of the variables mirror the ones in the GraphiQLInterface, which is a bit confusing, but this is how it works:

For each of the 3 resize objects, the hidden element can either be `null` `frist` `second`. Null means none are hidden and both are shown, for `first` and `second` tells which part of the panel is hidden (see screenshots)

## pluginResize
chrome_VOWqsPoVt5

## editorResize
chrome_hyWUEccVzm

## editorToolsResize
chrome_D0ZkWzYQEb


jonathanawesome:
Hey there @matijagaspar 👋, thanks for the comprehensive PR!

Is this related to #2929? If so, can you help us understand the need for more than one solution and why the [workaround](https://github.com/graphql/graphiql/pull/2929#issuecomment-1334482831) isn't working for y'all?

matijagaspar:
Hi @jonathanawesome 

Regarding #2929 we found out is not really suitable to our needs. The customization there is so the `<GraphiQL>` component can be customized via its own properties, however we are developing a GraphiQL plugin (to be passed into the `plugins` prop). 

One reason for the use of context is to keep the plugin contained and not interfere with any other plugin that might be passed. The other reason is that this specific plugin has a serverside counterpart, that is a plugin to mercurious, and the default setup does not allow us to specify what props mercurious can pass to the `<GraphiQL>` component.

You can take a look at the example on the repo: https://github.com/nearform/mercurius-federation-info-graphiql-plugin/tree/master/example

However  #2929 might still be of use to others and that is why we are keeping it open. This PR does not close the #1620 and does not provide any additional customization to the users of the `<GraphiQL>` component. 

So in essence this PR adds an additional API to the plugin API, while the #2929 adds customization options to the `<GraphiQL>` component.

thomasheyenbrock (changes requested):
Hey @matijagaspar 👋 I found this PR which has been open for quite a while (sorry for not coming back to this earlier 🙏 ). First question: Is the topic even still relevant for you?

Looking at the proposed code changes, I'm not sure I feel comfortable exposing this level of "implementation details". I agree that there is a use-case for programmatically controlling plugin sizes. However, exposing the specific plugin resize elements creates additional API surface that we need to maintain and that we can't simply break without publishing a new major version. I don't have a better idea at this point though 😕

PR #2961, opened 2022-12-08
Update changeset's `config.json`

B2o5T:


thomasheyenbrock:
Why did we make this change? I'm not sure we really need changelogs for examples 🤔

B2o5T:
@thomasheyenbrock because I removed examples from workspace and forget update changeset's `config.json` that ignore examples

Issue #2894, opened 2022-11-11
ESLint improvements


dimaMachina:
Below are my vision about improving ESLint linting in `graphiql` repository, it will be done with separated PRs to make review easy!

- [x] fix all warnings to have clear ESLint output https://github.com/graphql/graphiql/pull/2962
- [x] add `--max-warnings=0` and `--cache` CLI flags https://github.com/graphql/graphiql/pull/2962
- [x] <s>extend `plugin:jest/recommended` config for test files</s> **was already extended**
- [x] extend `plugin:cypress/recommended` config for e2e tests https://github.com/graphql/graphiql/pull/3035
- [x] extend `plugin:import/recommended` config #2922
- [ ] extend `plugin:jsx-a11y/recommended` config
- [x] extend `plugin:@typescript-eslint/recommended` config https://github.com/graphql/graphiql/pull/2967
- [ ] extend `plugin:@typescript-eslint/recommended-requiring-type-checking` config
- [x] ensure all code-blocks in markdown files doesn't contain syntax errors with `eslint-plugin-mdx`

jonathanawesome:
@B2o5T very much looking forward to these changes!

dimaMachina:
Most of theme were improved

PR #2962, opened 2022-12-08
🛠️ ESLint: clean all ESLint warnings, add `--max-warnings=0` and `--cache` flags

B2o5T:
the day of the pure ESLint output is coming

related https://github.com/graphql/graphiql/issues/2894


.eslintrc.js
```diff
@@ -14,9 +14,6 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 6,
     sourceType: 'module',
-    ecmaFeatures: {
-      jsx: true,
-    },
```
B2o5T, line 19: will be set while extending react recommended config

.eslintrc.js
```diff
@@ -29,10 +26,7 @@ module.exports = {
     es6: true,
     node: true,
     browser: true,
-    jest: true,
-    'jest/globals': true,
```
B2o5T, line 33: no need to set this for all files

.eslintrc.js
```diff
@@ -291,11 +283,12 @@ module.exports = {
         '**/*.spec.{ts,js.jsx.tsx}',
       ],
       extends: ['plugin:jest/recommended'],
-      env: {
-        'jest/globals': true,
-      },
```
B2o5T, line 296: will be set while extending jest recommend config

packages/cm6-graphql/__tests__/test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/expect-expect */
```
B2o5T, line 1: I guess this file is never run, because it's missing `{test,spec}` suffix, what do we need to do with it?

packages/graphiql/.eslintrc.js.old
```diff
@@ -1,298 +0,0 @@
-module.exports = {
```
B2o5T, line 1: some outdated config, that can be send to trash :)

packages/cm6-graphql/__tests__/test.ts
thomasheyenbrock: I think this has since been renamed to `test.spec.ts` on `main`

PR #2963, opened 2022-12-08
enable `prefer-destructuring` rule

B2o5T:
with friendly config (only objects) and only declaration


packages/codemirror-graphql/src/info.ts
```diff
@@ -62,9 +62,8 @@ CodeMirror.registerHelper(
       return;
     }
 
-    const state = token.state;
-    const kind = state.kind;
-    const step = state.step;
+    const { state } = token;
+    const { kind, step } = state;
```
thomasheyenbrock, line 66: Could we even enforce this?
```suggestion
    const { state: { kind, step } } = token;
```

B2o5T: I don't think :)
But if we force I would prefer to
`const { kind, step } = token.state;`

also for better destructuring there is https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-destructuring.md

acao:
looks good to me! much of this code predates object destructuring, haha! this rule makes it more readable, thank you!

PR #2971, opened 2022-12-12
feat(graphiql-react): add useTheme to context

### LINKED ISSUE #2956: [graphiql/react] useTheme hook does not update when user changes the theme

### Current Behavior

Please have a look at this [Codesandbox example](https://codesandbox.io/s/graphiql-usetheme-does-not-update-n7b433?file=/src/App.tsx)

I created a `TestPlugin` that uses the `useTheme` hook to get the current theme. 
When the plugin loads, it gets the correct theme, but if the user changes it, the value returned by the hook does not update.
I need to close and reopen the plugin tab to get the updated value.

Please, have a look at the following video:

https://user-images.githubusercontent.com/1233582/206126554-715784c9-2f07-469f-9f7d-f13a5f0dce0c.mov

### Expected Behavior

The value returned by `useTheme` hook should update when the user changes the theme in the settings.

### Steps To Reproduce

_No response_

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version: 2.2.0
- OS:
- Browser: Chrome
- Bundler:
- `react` Version: 18.0.0
- `graphql` Version: 16.6.0
```

### Anything else?

_No response_

matijagaspar:
Fixes #2956

Makes `useTheme` a context instead of local state so it can be used in plugins (or any children of `<GraphiQL/>`) when using:

`const {theme, setTheme} = useTheme()`


jonathanawesome:
It would be great to get the return values updated as well, see [here](https://github.com/graphql/graphiql/issues/2956#issuecomment-1357574162).

packages/graphiql-react/src/theme.tsx
```diff
@@ -53,4 +63,17 @@ export function useTheme() {
...
+export function ThemeContextProvider(props: ThemeContextProviderProps) {
+  const theme = useThemeInternal();
+  return (
+    <ThemeContext.Provider value={theme}>
+      {props.children}
+    </ThemeContext.Provider>
+  );
+}
+
+export const useTheme = createContextHook(ThemeContext);
```
acao, line 77: can you add a documentation block above, for the api docs?

matijagaspar: Added

acao:
@thomasheyenbrock and @jonathanawesome  - should we ask for tests for this?

packages/graphiql-react/src/theme.tsx
```diff
@@ -53,4 +63,23 @@ export function useTheme() {
   return useMemo(() => ({ theme, setTheme }), [theme, setTheme]);
 }
 
+type ThemeContextProviderProps = { children: ReactNode };
+
+export function ThemeContextProvider(props: ThemeContextProviderProps) {
+  const theme = useThemeInternal();
```
thomasheyenbrock, line 69: Nit: We could just inline the contents from that hook here and avoid the extra function.

thomasheyenbrock (commented):
Thanks a lot @matijagaspar, that's an awesome addition!

Starting to add some tests to this would be awesome, but I wouldn't put that on @matijagaspar - unless you're up to it!

PR #2835, opened 2022-10-23
fix(sonar-lint): :bug: fixed doctype and missing html tags

hinogi:

PR #2972, opened 2022-12-13
fix for graphiql-plugin-explorer example

jonathanawesome:
The example `index.html` was overwritten in #2835 with an incorrect unpkg URL.


thomasheyenbrock (approved):
...and I didn't notice in the review 🙈 thanks for fixing! 💜

jonathanawesome:
> ...and I didn't notice in the review 🙈 thanks for fixing! 💜

An easy line to miss. Thanks for the quick review!

PR #2986, opened 2023-01-05
fix: VariableJSONSchema for custom scalars

bboure:
`any` is not an accepted value for [type](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-type), which gives an error in validation.

In order to accept any value, we should just not add the `type` property at all.

before:
image

after:
image


acao:
@bboure thank you, this makes a lot more sense!

Issue #1635, opened 2020-08-07
Allow to run multiple operations (for query batching)


leoloso:
Executing multiple queries in a single query is not part of the spec yet, but [it has been requested](https://github.com/graphql/graphql-spec/issues/375), and several servers already implement it (such as [Apollo](https://www.apollographql.com/blog/batching-client-graphql-queries-a685f5bcd41b/)).

Then, GraphiQL should allow to execute multiple operations against the server. However, when there are multiple operations in the document, the Run button only allows to select either one, but not many of them, or all of them.

Please notice in this screenshot, the 2 operations are expected to be executed together, but that's not possible:

<img width="479" alt="Screenshot 2020-08-07 at 10 46 55 PM" src="https://user-images.githubusercontent.com/1981996/89657665-fe202100-d8ff-11ea-81f5-c5855cfb57c0.png">

## Proposal

Add a configuration option `"runMultipleQueries"` which, when `true`, if there is more than 1 operation in the document, also adds entry `"Execute all"` in the Run button dropdown:

<img width="479" alt="Screenshot 2020-08-07 at 10 50 54 PM" src="https://user-images.githubusercontent.com/1981996/89657979-8999b200-d900-11ea-88ec-cad51bd8de2c.png">

acao:
@leoloso since we are the official spec reference implementation, then someone who is advancing this spec is free to open a PR to graphiql, however it won't ship with the official GraphiQL preset until it's official spec!

acao:
actually, i will re-open this to make sure we enable this at a relatively early iteration of plugin scope. this would be a cool feature to help advance, but we are currently focused on proposals at more advanced stages right now.

more good news @leoloso  - though we can't support this "officially" yet, and there is no plugin API yet, there _are_ props that would allow you to do this, kinda. you'd still get the dropdown, but clicking any operation name would execute all of them if you want, as  you can modify how the outbound http request is handled by `GraphiQL`

the `FetcherParams` you get when implementing the `fetcher` async function prop would allow you to implement this, and whatever request pattern you want. you can even use apollo client. when serializing the body for `fetch` as we normally do for graphiql, you could just specify `operationNames` instead of `operationName`, or no `operationName` to execute all operations, or however folks are implementing this spec

i wonder though, without implenting this in `graphql-js`, it may be that a lot of cases would fail validation

leoloso:
That's great, thanks @acao.

[...]

PR #2994, opened 2023-01-13
feat: Add @graphiql/plugin-batch-request

samuelAndalon:
## Description

Hi all, this is Sam from Expedia Group, we recently [added graphiql to graphql-kotlin](https://github.com/ExpediaGroup/graphql-kotlin/pull/1623) with explorer and exporter plugin, [at EG we heavily rely on batched requests](https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/#5853) from clients to GraphQL servers.

The `graphiql@2` plugin API gave us an idea of a plugin that we really need, a way to create batched requests from an IDE, as currently playground, graphiql, insomina, postman offer that capabilty, forcing us to use raw http clients to make some requests for testing.

This PR that adds a batch-request plugin, relies on the `editorContext` to access to tabs and operations on each tab, then it presents the operations in a file tree format to have user to select which operations are going to be sent in a batch.

Please let us know your feedback and if this is indeed a candidate for a plugin.

Couple of notes:

1. The plugin scope is for sending multiple graphQL operations with two approaches `aliasing` and `arrays` (server requires to be configured for array inputs).  
2. we could consider this PR as a continuation of [this opened issue](https://github.com/graphql/graphiql/issues/1635)
3. we use [react-checkbox-tree](https://github.com/jakezatecky/react-checkbox-tree) to present tabs and operations.
4. still need to work on the code-mirror `graphql-results` mode, as setting the batch response array in response editor causes some parsing issues.

Demo video:

https://user-images.githubusercontent.com/6611331/216736177-2d8d6153-b246-48ef-8e97-687beea6f9fc.mov


samuelAndalon:
Hi @acao yes, sorry! forgot to switch the npm registry, will apply the changes, along with some fixes regarding building the payload for the HTTP request. 
Also if needed we could add an option to batch using aliases as an alternative as described here
https://the-guild.dev/graphql/tools/docs/batch-execution

will push registry changes along with other fixes.

samuelAndalon:
@acao just make the changes, also update the README.md

packages/graphiql-plugin-batch-request/README.md
```diff
@@ -1 +1,58 @@
-# GraphiQL Batch Request Plugin
\ No newline at end of file
+# GraphiQL Batch Request Plugin
+
+This package provides a plugin that allows sending a batch request to a GraphQL Server into the GraphiQI UI.
```
TallTed, line 3:
```suggestion
This package provides a plugin that allows sending a batch request to a GraphQL Server and thence into the GraphiQI UI.
```

packages/graphiql-plugin-batch-request/README.md
```diff
@@ -1 +1,58 @@
...
+The following packages are peer dependencies, so make sure you have them installed as well:
+
+```sh
+npm i -S react react-dom graphql
+```
+
+## Usage
+
+The plugin scope is for sending multiple GraphQL operations as an array, so GraphQL server requires to be configured to allow
+arrays.
```
TallTed, line 22:
```suggestion
The plugin scope is for sending multiple GraphQL operations as an array, so the GraphQL server must be configured to allow arrays.
```

samuelAndalon:
Hi all, an update here!, just added an option to also batch using aliases, took the batching algorithm from 

https://github.com/ardatan/graphql-tools/blob/master/packages/batch-execute/src/mergeRequests.ts

samuelAndalon:
Hi everyone, just circling back here, would it still make sense to have this plugin ? we are already using using but publishing it internally. We believe that this is something that could benefit the GraphQL community

thomasheyenbrock:
Hey @samuelAndalon 👋 this plugin looks nothing short of awesome 😍 props for putting that together 👏 

I do believe that this plugin is meaningful and probably delivers value also for other folks. I'm just not sure if this repo is the right place for it. A main motivator for introducing plugins was to enable folks to build and maintain their own things without us needing to build all feature requests first-class into GraphiQL. We do have some plugins (the explorer and code exporter) which are fairly general purpose and should also give an example of what you can build with the plugin API.

However, I do not think that all plugins should live in this repository. In particular, plugins that are opinionated or that focus on supporting GraphQL features that are not (yet) part of the latest spec version should _not_ be maintained in this repo. This plugin doesn't match with that as batched requests are an extension of GraphQL that is not part of the spec. This is my personal opinion and we haven't aligned on this yet as maintainers, so I encourage other opinions here (cc @acao and @jonathanawesome, we could also discuss that in future working group sessions if desired).

I can definitely encourage you to publish this publicly on your end and share it with the community for now ⭐ that way you can share this with a broader audience until we figured out our strategy on which plugins should be maintained by us and which shouldn't.

acao:
I think this is awesome!

But I would rather see `@graphiql/toolkit` `createGraphiQLFetcher()`  introduce a plugin API to support this behavior than to create an alternate query execution mechanism.

trevor-scheer:
Thanks for your work on this @samuelAndalon! +1 @thomasheyenbrock - I'm going to close this for these reasons and also since it's quite old at this point. @samuelAndalon if you did end up publishing this publicly I think it'd be interesting to create or add to a list of community plugins here (I'm not sure if we have one!)

PR #3012, opened 2023-02-01
feat: maintain DocExplorer stack on schema update

benjie:
Previously:

- when you click the schema reload button, the doc explorer stack was reset even if the schema is unchanged 😞 
- when you passed a new schema in, the doc explorer stack is _unaffected_, meaning it can render invalid content or even crash in some circumstances 🐞

I'm not sure what the need for resetting the explorer stack is, so I've raised this PR to begin discussions. I believe that the doc explorer should only be reset when an incompatible schema is loaded, and even then we should aim to keep the compatible parts of the stack where possible. I've attempted to achieve this (and included tests for the same) in this PR.

### DIFF
--- a/packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
+++ b/packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
@@ -1,15 +1,35 @@
 import { render } from '@testing-library/react';
-import { GraphQLSchema } from 'graphql';
+import { GraphQLInt, GraphQLObjectType, GraphQLSchema } from 'graphql';
+import { useContext, useEffect } from 'react';
 
 import { SchemaContext, SchemaContextType } from '../../../schema';
-import { ExplorerContextProvider } from '../../context';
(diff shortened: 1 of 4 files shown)


benjie:
> @graphiql/react@npm:^0.17.0-canary-208963ff.0: No candidates found

acao:
@benjie yep, just need my `acao` npm user to be added to `graphiql` organisation again

acao:
@benjie just tested manually and it works! thanks for this :)

packages/graphiql-react/src/explorer/components/__tests__/doc-explorer.spec.tsx
```diff
@@ -87,4 +107,134 @@ describe('DocExplorer', () => {
     const errors = container.querySelectorAll('.graphiql-doc-explorer-error');
     expect(errors).toHaveLength(0);
   });
+  it('maintains nav stack when possible', () => {
```
acao, line 110: <3

acao (approved):

PR #3017, opened 2023-02-07
avoid bundling code from `react/jsx-runtime`

thomasheyenbrock:
We currently include code from `react/jsx-runtime` in the bundled output files. That prevents anybody from using these packages with Preact, as using React-packages with Preact relies on import aliasing. We solve that for all affected packages in two different ways:
- For `@graphiql/react` we're simply marking `react/jsx-runtime` as external.
- For the plugin packages (`@graphiql/plugin-explorer` and `@graphiql/plugin-code-exporter`) the above doesn't work because these packages also provide a UMD build. There's currently no file provided by `react` that would allow to load the JSX runtime transform as UMD bundle. So instead we fall back to the "classic" JSX transform that requires us to `import React from 'react'` in all `tsx` files. However, these plugins are quite lean and only consist of one such file, so the overhead is minimal.

### DIFF
--- a/.changeset/twenty-flies-unite.md
+++ b/.changeset/twenty-flies-unite.md
@@ -1,0 +1,7 @@
+---
+'@graphiql/plugin-code-exporter': patch
+'@graphiql/plugin-explorer': patch
+'@graphiql/react': patch
+---
+
+Avoid bundling code from `react/jsx-runtime` so that the package can be used with Preact
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -132,7 +132,11 @@ module.exports = {
 
     '@typescript-eslint/no-unused-vars': [
       'error',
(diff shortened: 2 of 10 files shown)


acao (approved):
thanks for this @thomasheyenbrock !

acao:
this reminds me that perhaps we should bring back some of the example builds as validation in CI steps

we used to confirm webpack/etc consuming builds still worked by making some of the `examples/` build in CI and/or netlify

B2o5T (commented):
Can we somehow test it? So in the future, if anybody removes import React we’ll know

acao:
@B2o5T ideally our `examples/` should be used to test for potential bundler issues

acao:
@thomasheyenbrock rebased and ready to go, thanks for this!

PR #3019, opened 2023-02-09
feature: add fragment arguments syntax highlights

mjmahone:
Adding fragment argument syntax highlighting to VSCode. This is needed as per discussion in https://github.com/graphql/graphql-wg/discussions/1239#discussioncomment-4908699

Before:
vscode-graphql-missing-fragment-args

After:
vscode-graphql-with-fragment-args

To test this:
- Download VSCode from https://code.visualstudio.com/docs
- Download the *GraphQL Foundation* **GraphQL Syntax Highlighting** extension (in my case, v1.0.6)
- Write a `test.graphql` file and ensure syntax highlighting is working as expected.
- Copy `graphiql/packages/vscode-graphql-syntax/grammars/graphql.json` to `~/.vscode/extensions/graphql.vscode-graphql-syntax-1.0.6/grammars/graphql.json`
- Reload VSCode

*Probably* the fast-devex way to make changes here is to have some `ln -s` symlink from `~/.vscode/extensions/` to `graphiql/packages/vscode-graphql-syntax`, but this change was simple enough that I didn't need that quick an iteration cycle.

### DIFF
--- a/.changeset/eighty-rabbits-ring.md
+++ b/.changeset/eighty-rabbits-ring.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql-syntax': minor
+---
+
+Adds syntax highlighting for arguments on fragment spreads as well as variable definitions on fragments.
--- a/packages/vscode-graphql-syntax/grammars/graphql.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.json
@@ -31,15 +31,21 @@
     },
     "graphql-fragment-definition": {
       "name": "meta.fragment.graphql",
-      "begin": "\\s*(?:(\\bfragment\\b)\\s*([_A-Za-z][_0-9A-Za-z]*)?\\s*(?:(\\bon\\b)\\s*([_A-Za-z][_0-9A-Za-z]*)))",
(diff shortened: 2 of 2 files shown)


orta:
👍🏻  looks right overall to me, didn't run it locally but the regexes seem right

acao:
looks great, thank you @mjmahone !

we hope to be able to publish again by next week

mjmahone:
> we hope to be able to publish again by next week... though I guess there is a way I can publish just the vscode syntax extension locally...

Next week would certainly be soon enough: I think right now I'm just making sure all the work items for fragment args are making progress and figuring out what, if any, of the blockers are. Syntax highlighting was the biggest blocker to trying it in Relay, but there are still a few other (namely, graphql-js syntax additions and Relay parsing support) that have not yet been merged. Thanks @acao !

acao (approved):

acao:
going to go ahead and merge for now, thanks again!

just noting for the future - this feels like a regression that may have been introduced with another fix in the recent past. we don't have a great way to test this yet, though there is now a way to test the static highlighting it would seem

PR #3032, opened 2023-02-23
fix: limit all code mirror classes to .graphiql-container scope

Yahkob:
This PR is to fix an issue I saw using GraphiQL in a repo that already uses code mirror. I was confused at first the code mirror components I added were blowing up to 100% width and height but saw the style rules came from graphiql's style sheet

Example
image

I'm not sure if fixing this problem is as simple as this PR but figured I'd open a PR instead of an issue if it is a quick fix 😀

The work around I've found is to just scope each instance of CodeMirror you use and override the graphiql css rules through specificity

### DIFF
--- a/packages/graphiql-react/src/editor/style/codemirror.css
+++ b/packages/graphiql-react/src/editor/style/codemirror.css
@@ -1,31 +1,31 @@
 @import url('codemirror/lib/codemirror.css');
 
 /* Make the editors fill up their container and make them scrollable */
-.CodeMirror {
+.graphiql-container .CodeMirror {
   height: 100%;
   position: absolute;
   width: 100%;
 }
 
 /* Override font settings */
-.CodeMirror {
+.graphiql-container .CodeMirror {
   font-family: var(--font-family-mono);
 }
 
 /* Set default background color */
-.CodeMirror,
-.CodeMirror-gutters {
+.graphiql-container .CodeMirror,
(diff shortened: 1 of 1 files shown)


acao:
thank you @Yahkob ! this makes perfect sense and follows with our original approach, we definately shouldn't be overriding vendor styles globally.

just a housekeeping note, in order for the EasyCLA bot to be happy and the one check to pass, you'll need to make sure you sign the CLA with the same email you signed the commits with

Yahkob:
@acao sorry for the delay! Made sure my commit was signed/verified and somehow I broke all of the CIs though 🤦

acao:
@Yahkob no worries, the breakage is on our end. Your next commit should cause the netlify deploy preview to pass! And the github actions should be fine, I have just approved them

Yahkob:
@acao I feel bad adding nonsensical commits but I just slightly changed a css comment to see if it would fix other CI steps, I guess the alternative was opening a new PR

Yahkob:
@acao just a ping on this to check if I have to do anything else 😄

thomasheyenbrock:
@Yahkob you need to sign the CLA agreement before we can consider merging this PR. You should be able to do that by clicking on the "Details" next to the failed check on this PR.

Yahkob:
@thomasheyenbrock I did but for some reason it's not registering? Let me try to reopen this PR.

PR #3035, opened 2023-02-24
extend `plugin:cypress/recommended` config

B2o5T:


acao (approved):
nice to see we don't need .wait() anymore!

PR #3041, opened 2023-02-25
[ESLint] enable `no-var` rule

B2o5T:


packages/graphiql/resources/renderExample.js
```diff
@@ -13,12 +13,12 @@
  */
 
 // Parse the search string to get url parameters.
-var parameters = {};
+const parameters = {};
```
acao, line 16: the reason I use `var` here is for browser compatibility. this file is not compiled.

B2o5T: This file is used for cypress, also every browser now supports let/const

TallTed: It may not matter to this specific, but "every browser now supports [whatever]" does not account for users who are stuck on older versions, for whatever reason, plenty of which are valid. It is important to keep stock of what the actual user environment requirements are, and make them known, rather than silently relying on what "every browser now supports."

B2o5T: I don't need to teach lessons of usage `var` keyword in the script file, graphiql is bundled to esm and cjs with es5 target where `var` is used instead of `let`/`const`.

You should avoid using `var` in your code and give this work for bundlers.

acao: this example script has always been understood as file that is loaded without a bundler. the html may load a cdn bundled version of graphiql, but this code should run in any browser without any transpiler. many of graphiql's users use an approach such as this, just on jsdelivr about 500,000/month are served using the CDN bundle directly like this, and thus no transpiler is used. this would probably be much higher combined with unpkg stats if we had them, and which we now use in our documentation and is commonly referred to in examples when users have issues

thomasheyenbrock: `const` and `let` have been fully supported by all modern browsers since the death of Internet Explorer, I think it should be fine to make this change: https://caniuse.com/?search=const

thomasheyenbrock (approved):

PR #3043, opened 2023-02-25
[ESLint] enable `unicorn/prefer-string-replace-all` rule

B2o5T:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md


acao (approved):
this will probably fix some bugs!

PR #3053, opened 2023-02-25
finish upgrading webpack & babel dependencies, fix monaco example

acao:
- [x] reintroduce the `monaco-graphql-webpack` workspace for local resolution, to honor `DEVELOPMENT.md`
- [x] tweak webpack configs for upgrade, to work on clean build
- [x] upgrade babel versions and webpack loaders/plugins and implementations
- [x] consolidate versions across the monorepo so that, despite adding a workspace, our yarn.lock file only grows by ~200 lines


acao:
I have tested this exhaustively across browsers, I hope it's ok, I had to tweak the postcss configuration, but after reviewing the changelogs I don't think we should expect changes that would impact our users (theming css variables, etc). There are a couple tweaks and improvements that are still dev related - for example, react fast refresh vs HMR plugin, etc, but those can wait or just be replaced by another bundler soon if we still go that route

Issue #986, opened 2019-10-22
Feature request: add a copy as cURL command button


judepereira:
I'm well aware that this is feasible via the dev tools for Chrome/Safari, however, that also copies all the cookies for my domain.

Additionally, Safari, for example, does not allow one to exclude cookies :(

benjie:
I don't think we want this in core as most users won't need it, but I think it would be a good candidate for a plugin in our new plugin system.

PR #3055, opened 2023-02-26
[wip] introduce `@graphiql/plugin-copy-as`

B2o5T:
fixes https://github.com/graphql/graphiql/issues/986

- [ ] remove import from `graphiql` package and add cypress tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ yarn-1.18.0.js
 
 examples/*/yarn.lock
 .eslintcache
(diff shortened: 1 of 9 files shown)


acao:
@B2o5T I think this work could be combined with the code exporter plugin - it already lets you provide custom snippets for copying your graphql query in different forms. should we not just add this button to that plugin?

B2o5T:
@acao sure! I’ll take a look at this PR after Monaco-editor will be integrated

PR #3061, opened 2023-02-27
remove unneeded `reference &&` assertion, convert to switch

B2o5T:


packages/graphiql-react/src/editor/query-editor.ts
```diff
@@ -111,14 +111,27 @@ export function useQueryEditor(
         return;
       }
       plugin.setVisiblePlugin(DOC_EXPLORER_PLUGIN);
-      if (reference && reference.kind === 'Type') {
```
B2o5T, line 114: since we don't verify `reference && ` for other `if` cases it's unnecessary

acao (approved):

PR #3064, opened 2023-02-27
do not suggest `ReactDOM.render(..., document.body)` as it's anti pattern!

B2o5T:
https://github.com/facebook/react/blob/6ff1733e63fdb948ae973a713741b4526102c73c/CHANGELOG.md?plain=1#L1678


acao (approved):

acao:
@B2o5T this caused a prettier issue :/

PR #3067, opened 2023-03-01
[graphql-language-service-cli] use a valid command in the readme example

ryantm:
The command originally there did not work, because "schema" is not a valid option to the CLI.


ryantm:
No I'm not signing a CLA for this.

PR #3069, opened 2023-03-02
fix: add graphql-language-service as direct dep to cm6-graphql

sergeichestakov:
Just tried using the `cm6-graphql` package but was seeing the following error:
> Module not found: Can't resolve 'graphql-language-service'

It looks like `graphql-language-service` needs to be a direct dependency of `cm6-graphql` in order for this package to work properly.

Additionally, I noticed that the `@codemirror/*`, `@lezer/highlight`, and `graphql` packages need to be declared as peer dependencies as their absence would lead to the same errors for anyone trying to use this package.

Lastly, I went ahead and upgraded all of the above packages to the latest versions while I was at it.


imolorhe:
Thanks! I think I already had these dependencies in my project, which would explain why I hadn't noticed this issue. I've resolved the merge conflicts.

Regarding the other `@codemirror/*` dependencies that you updated, codemirror 6 has an issue with object references at runtime if you have multiple instances of `@codemirror/state` which happens when you use differing versions of the different `@codemirror/*` packages, so setting a specific version in the package.json helps maintain the same version. Alternatively this can be resolved by using the [selective dependency resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/).

sergeichestakov:
right but that's why they're just peer dependencies. so that the package just consumes which ever version that you have installed (we've been doing this in all of our open source CM repos at Replit and it's been working great for us). I made sure to deduplicate the CM and Lezer packages after upgrading to ensure we weren't using two versions but increasing the dev dependency versions shouldn't affect the final package.

sergeichestakov:
but yeah also not sure why CI is still failing 😕

acao:
@sergeichestakov not your fault, my bad! just some unrelated hiccups as we transition to `pnpm`. your next commit should pass green. thanks for this!

sergeichestakov:
no prob! just curious: any word on when this might be published?

PR #3082, opened 2023-03-06
use writeFile from `fs/promises` instead `promisify(writeFile)`

B2o5T:


acao (approved):

acao:
thanks for this!

why is there a changeset if this is not a feature or a bugfix?

B2o5T:
@acao removed!

PR #3084, opened 2023-03-06
use promise version of fs.readFile

B2o5T:
just saw a piece of code where `fs.readFile` was used with `new Promise`

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLCache.ts
+++ b/packages/graphql-language-service-server/src/GraphQLCache.ts
@@ -19,6 +19,7 @@ import type {
 } from 'graphql-language-service';
 
 import * as fs from 'node:fs';
+import { readFile } from 'node:fs/promises';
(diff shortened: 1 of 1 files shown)


packages/graphql-language-service-server/src/GraphQLCache.ts
```diff
@@ -789,57 +790,50 @@ export class GraphQLCache implements GraphQLCacheInterface {
...
+          filePath,
+          content,
+          asts,
+          queries,
+          mtime: 0,
+          size: 0,
+        };
+      } catch {
+        // If query has syntax errors, go ahead and still resolve
+        // the filePath and the content, but leave ast empty.
```
TallTed, line 826:
```suggestion
        // If query has syntax errors, resolve the filePath
        // and the content anyway, but leave ast empty.
```

thomasheyenbrock (approved):
Looks good to me in terms of code changes 👍

PR #3088, opened 2023-03-06
[graphql-language-service-server] remove nowhere used `node-fetch` dependency

B2o5T:


packages/vscode-graphql-execution/package.json
```diff
@@ -95,7 +95,6 @@
     "@types/dotenv": "8.2.0",
     "@types/mocha": "5.2.7",
     "@types/node": "^16.18.4",
-    "@types/node-fetch": "3.0.3",
```
B2o5T, line 98: this types also seems nowhere used

acao (approved):
Thanks, I think I left it because I wasn't sure what would happen next. We have switched server fetch implementations so many times, but i think before long we can just use native node.js fetch in vscode, not sure which node version vscode ships with currently.

PR #3093, opened 2023-03-07
monaco example fixes, fix missing worker bug

acao:
I _think_ this will fix a bug on `_doValidate()` that shows up in some contexts


acao:
@jonathanawesome confirmed this fixed the bug!

PR #3097, opened 2023-03-10
fix: html examples

chenaski:
I tried to use `packages/graphiql-plugin-code-exporter/examples/index.html` and `packages/graphiql-plugin-explorer/examples/index.html` examples, but it turned out that they don't work.

- Updated integrity attribute of `https://unpkg.com/graphiql/graphiql.min.js`
- Added `regenerator-runtime` script in `graphiql-plugin-code-exporter`

### DIFF
--- a/packages/graphiql-plugin-code-exporter/examples/index.html
+++ b/packages/graphiql-plugin-code-exporter/examples/index.html
@@ -44,8 +44,14 @@
     ></script>
 
     <script
+      src="https://unpkg.com/regenerator-runtime@0.13.11/runtime.js"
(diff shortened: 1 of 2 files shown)


acao:
thank you for this @chenaski ! hard to keep these examples working, so many to keep track of, no automatic validation

chenaski:
Yeah, I can understand. Actually I was thinking about writing some tests to check if HTMLs open without errors, but it takes time, so at least this will fix them for now.

thomasheyenbrock (changes requested):
Seems like the integrity hashes are outdated yet again 😓 
CleanShot 2023-04-14 at 17 13 42@2x

chenaski:
Updated them again :)

B2o5T:
`integrity` was removed with update all examples to React 18 here https://github.com/graphql/graphiql/pull/3185/files

thomasheyenbrock:
@B2o5T makes sense, as we don't have a way to automatically update these hashes on releases, it's just too much overhead at the moment. Thanks @chenaski!

Issue #2897, opened 2022-11-12
Promise in try/catch


hinogi:
Consider using 'await' for the promise inside this 'try' or replace it with 'Promise.prototype.catch(...)' usage

https://github.com/hinogi/graphiql/blob/1a3d034c02114c4a164e94626beb840260f05359/packages/graphql-language-service-server/src/MessageProcessor.ts#L994

B2o5T:
Your are right!

@acao future mistakes could be avoided by using https://typescript-eslint.io/rules/no-floating-promises/ rule, but since that rule also requires type information CI in followed PR should be fixed first https://github.com/graphql/graphiql/pull/2878

hinogi:
you could also add sonarqube to the repo, that's how I found it and it is free for OSS

hinogi:
> Your are right!
> 
> @acao future mistakes could be avoided by using https://typescript-eslint.io/rules/no-floating-promises/ rule, but since that rule also requires type information CI in followed PR should be fixed first #2878

I guess there is a js version also
https://github.com/SebastienGllmt/eslint-plugin-no-floating-promise

B2o5T:
> I guess there is a js version also https://github.com/SebastienGllmt/eslint-plugin-no-floating-promise

The above plugin is not maintained and it always better to use rule that requires type information

PR #3109, opened 2023-03-22
[ESLint] enable `@typescript-eslint/no-floating-promises`, update typescript-eslint to v6

B2o5T:
added changeset explicitly

fixes https://github.com/graphql/graphiql/issues/2897


packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -963,7 +963,7 @@ export class MessageProcessor {
         );
       } else {
         try {
-          this._cacheSchemaFile(uri, project);
+          await this._cacheSchemaFile(uri, project);
         } catch {
           // this string may be an SDL string even, how do we even evaluate this?
```
TallTed, line 968:
```suggestion
          // This string may even be an SDL string. How do we evaluate this?
```

PR #3110, opened 2023-03-22
fix: Webpack dev server config before and after middleware and renderExample path

mylesmmurphy:
**Current Behavior:**
Upon a fresh pull and install of the project and following the dev setup instructions for GraphiQL only, you will receive the following error when trying to run `yarn run start-graphiql`:
```
$ yarn workspace graphiql dev
$ cross-env NODE_ENV=development webpack-dev-server --config resources/webpack.config.js
[webpack-cli] Invalid options object. Dev Server has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'after'. These properties are valid:
   object { allowedHosts?, bonjour?, client?, compress?, devMiddleware?, headers?, historyApiFallback?, host?, hot?, http2?, https?, ipc?, liveReload?, magicHtml?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, port?, proxy?, server?, setupExitSignals?, setupMiddlewares?, static?, watchFiles?, webSocketServer? }
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 2
```

After resolving that error, the graphiql interface does not load when opening `http://localhost:8080/` or `http://localhost:8080/dev.html`, and instead hangs on the "Loading..." screen, with the error:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
```

**Desired Behavior:**
`start-graphiql` should start a webpack development server with the correct middlewares and no errors. You should be able to navigate to graphiql ui via `http://localhost:8080/dev.html`

**Solution:**
Following webpack dev server's [documentation](https://webpack.js.org/configuration/dev-server/#devserversetupmiddlewares) for setting up middlewares, we should use the `setupMiddlewares` in place of `before` and `after`.

`index.html.ejs` was pointing to the wrong path for `renderExample.js`. Updated to correct path.

**Notes:**
We could use `onBeforeSetupMiddleware` and `onAfterSetupMiddleware` instead of the newer `setupMiddlewares`, which behaves closer to the original behavior than what is in this PR, but that has a deprecation warning.


acao:
@mylesmmurphy hey there! thanks for this! we are in the midst of overhauling build tooling and it looks like this example was broken

acao:
Hmm, seems to cause the exact same error I get when I make this change - something about it breaks the cypress build

packages/graphiql/package.json
```diff
@@ -101,7 +101,7 @@
     "webpack": "5.76.0",
     "webpack-bundle-analyzer": "^3.6.1",
     "webpack-cli": "^5.0.1",
-    "webpack-dev-server": "^4.11.1",
+    "webpack-dev-server": "^4.13.1",
```
thomasheyenbrock, line 104: Can we leave that version unchanged? Otherwise we'd be installing two versions for `webpack-dev-server` right now. Seems to me that the fix also works with `4.11.1`.

thomasheyenbrock (commented):
Thanks @mylesmmurphy 🙌  the main workflow for developing GraphiQL is indeed broken at the moment. Just a nit.

mylesmmurphy:
@acao Good catch, my apologies, I forgot to check that. I reworked this PR slightly and everything appears to be working correctly now :)

B2o5T (approved):

acao:
It looks like you're getting the same failure on netlify deploy previews that other PRs are getting now. I think this has to do with a non-pinned resolution. Hope we can get this fixed!


Screenshot 2023-04-04 at 12 59 22

acao (changes requested):
lets get netlify deploy preview builds working before merging, the current failure is the same across all PRs

acao:
also to note - here the webpack NODE_ENV=development build is working for the cypress suite, but the NODE_ENV=production build we do for netlify is what is failing.

acao:
I found the source of the netlify issue, this is good to merge as is! thanks for the fix.

acao (approved):

PR #3120, opened 2023-03-31
[ESLint] enable `promise/prefer-await-to-then` for non React packages

B2o5T:
why for non-React - because it's ugly have `async IIFE` inside `useEffect`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -366,5 +366,12 @@ module.exports = {
         'import/no-unresolved': ['error', { ignore: ['^node:', 'vscode'] }],
       },
     },
+    {
+      files: ['packages/**'],
+      excludedFiles: ['packages/graphiql/**', 'packages/graphiql-react/**'],
+      rules: {
+        'promise/prefer-await-to-then': 'error',
+      },
+    },
   ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -368,6 +368,7 @@ module.exports = {
     },
     {
       files: ['packages/**'],
+      // ignore React packages because it's ugly to have `async IIFE` inside `useEffect`
       excludedFiles: ['packages/graphiql/**', 'packages/graphiql-react/**'],
       rules: {
         'promise/prefer-await-to-then': 'error',
--- a/packages/graphiql-toolkit/src/async-helpers/index.ts
+++ b/packages/graphiql-toolkit/src/async-helpers/index.ts
@@ -52,42 +52,34 @@ export function isAsyncIterable(
   );
 }
 
-function asyncIterableToPromise<T>(
+async function asyncIterableToPromise<T>(
   input: AsyncIterable<T> | AsyncIterableIterator<T>,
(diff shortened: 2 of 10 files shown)


packages/graphiql-toolkit/src/async-helpers/index.ts
```diff
@@ -52,42 +52,34 @@ export function isAsyncIterable(
...
-        iteratorReturn?.();
-      })
-      .catch(err => {
-        reject(err);
-      });
-  });
+  const result = await iteratorNext();
+  // ensure cleanup
+  iteratorReturn?.();
+  return result.value;
```
B2o5T, line 71: the only thing is that after refactoring to `async/await` we call `iteratorReturn?.()` before returning values, previously we did following, does it ok?
```js
        resolve(result.value);
        // ensure cleanup
        iteratorReturn?.();
```

.eslintrc.js
```diff
@@ -361,5 +361,12 @@ module.exports = {
         'import/no-unresolved': ['error', { ignore: ['^node:', 'vscode'] }],
       },
     },
+    {
+      files: ['packages/**'],
+      excludedFiles: ['packages/graphiql/**', 'packages/graphiql-react/**'],
```
B2o5T, line 366:
```suggestion
      // ignore React packages because it's ugly to have `async IIFE` inside `useEffect`
      excludedFiles: ['packages/graphiql/**', 'packages/graphiql-react/**'],
```

packages/vscode-graphql-execution/src/providers/exec-content.ts
```diff
@@ -108,11 +108,10 @@ export class GraphQLContentProvider implements TextDocumentContentProvider {
       enableScripts: true,
     };
 
-    this.loadProvider()
-      .then()
```
B2o5T, line 112: I guess calling `.then` is redundant

packages/graphiql-toolkit/src/async-helpers/index.ts
```diff
@@ -52,42 +52,34 @@ export function isAsyncIterable(
...
-    // so every AsyncIterable must be implemented using AsyncGenerator.
-    const iteratorReturn = (
-      'return' in input ? input : input[Symbol.asyncIterator]()
-    ).return?.bind(input);
-    const iteratorNext = (
-      'next' in input ? input : input[Symbol.asyncIterator]()
-    ).next.bind(input);
+  // Also support AsyncGenerator on Safari iOS.
+  // As mentioned in the isAsyncIterable function there is no Symbol.asyncIterator available
+  // so every AsyncIterable must be implemented using AsyncGenerator.
```
TallTed, line 60:
```suggestion
  // As mentioned in the isAsyncIterable function, there is no Symbol.asyncIterator available,
  // so every AsyncIterable must be implemented using AsyncGenerator.
```

TallTed (changes requested):
Minor tweak of comment for clarity

PR #3126, opened 2023-04-01
[ESLint] Prefer `KeyboardEvent#key` over `KeyboardEvent#keyCode`

B2o5T:


packages/graphiql-react/src/schema.tsx
```diff
@@ -317,10 +318,11 @@ export function SchemaContextProvider(props: SchemaContextProviderProps) {
    */
   useEffect(() => {
     function triggerIntrospection(event: KeyboardEvent) {
-      if (event.keyCode === 82 && event.shiftKey && event.ctrlKey) {
+      if (event.ctrlKey && event.key === 'R') {
```
B2o5T, line 321: don't need to check `event.shiftKey` because `key: 'R'` is always with shift

packages/graphiql-react/src/explorer/components/search.tsx
```diff
@@ -50,10 +50,11 @@ export function Search() {
 
   useEffect(() => {
     function handleKeyDown(event: KeyboardEvent) {
-      if (event.metaKey && event.keyCode === 75 && inputRef.current) {
-        inputRef.current.focus();
+      if (event.metaKey && (event.key === 'K' || event.key === 'k')) {
```
B2o5T, line 53: catch both lowercase and uppercase (as current behaviour)

thomasheyenbrock: I'd be fine to changing this to just lowercase, I'd guess that's how most users use this shortcut

thomasheyenbrock (approved):

.eslintrc.js
```diff
@@ -279,6 +284,8 @@ module.exports = {
     'unicorn/prefer-node-protocol': 'error',
     'import/no-unresolved': ['error', { ignore: ['^node:'] }],
     'unicorn/prefer-string-replace-all': 'error',
+    // doesn't catch a lot of cases we use ESLint builtin `no-restricted-syntax` to forbid `.keyCode`
```
TallTed, line 287:
```suggestion
    // doesn't catch a lot of cases; we use ESLint builtin `no-restricted-syntax` to forbid `.keyCode`
```

B2o5T: @TallTed ty!

packages/graphiql-react/src/explorer/components/search.tsx
B2o5T, line 53:
```suggestion
      if (event.metaKey && event.key === 'k') {
```

Issue #3115, opened 2023-03-28
Vue support is broken when script tag is not on top


SeanDylanGoff:
After some debugging it Vue VSCode Intellisense works if the setup scripts comes first (before template & script), but fails in any other order. If the setup script comes second, GraphQL errors are overlaid on whatever is actually there as if the script came first. So it looks like the setup script offset isn't added somewhere.   

I'm not familiar with this codebase, so maybe this is an upstream issue with the vue sfc compiler.

_Originally posted by @flotothemoon in https://github.com/graphql/graphiql/issues/2827#issuecomment-1333707395_

This issue still is still present in GraphQL: Language Feature Support v0.8.7

KammererTob:
Can confirm. No errors in log. If `<script setup lang="ts">` is below the template, Intellisense isn't working. As soon as i move the whole script block to the top it works again.

PR #3135, opened 2023-04-14
fix: script offset if vue-sfc contains template above

KammererTob:
closes https://github.com/graphql/graphiql/issues/3115

Related:
https://github.com/graphql/graphiql/pull/2827#issuecomment-1333707395

Description:
This fixes Intellisense for Vue SFC files which contain template code above the script (both setup and normal). The issue was that the line locations inside the ASTs returned from `compileScript` where relative to the start of the script tag and not the start of the file. This lead to intellisense and errors working with the wrong line information when the file contained a template above the script.
The fix returns a `scriptOffset` from `parseVueSFC`, which should contain the correct offset of the script tag
I added some tests and also tested it visually with a project for which i had problems with.

### DIFF
--- a/.changeset/thin-poems-search.md
+++ b/.changeset/thin-poems-search.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-server': patch
+---
+
+fixed wrong script tag offset for vue-sfc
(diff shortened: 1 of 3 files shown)


acao:
hey @KammererTob, awesome work on this one! thank you for adding a test case as well.

upstream there were some changes that broke typescript when I rebased, probably something minor. if you like, I can fix the issue myself today!

KammererTob:
@acao Sure. Go ahead. Thanks! If you need anything from my side just let me know.

acao:
@KammererTob ok! for the error case, the `scriptOffset` returns 0 which I think makes sense?

PR #3140, opened 2023-04-17
simplify `monaco-graphql-react-vite` example, fix sending variables

B2o5T:


examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -1,69 +1,67 @@
...
 import { Uri, editor, KeyMod, KeyCode, languages } from 'monaco-editor';
 import { initializeMode } from 'monaco-graphql/src/initializeMode';
-import { createGraphiQLFetcher } from '@graphiql/toolkit';
+import { createGraphiQLFetcher, SyncFetcherResult } from '@graphiql/toolkit';
 import * as JSONC from 'jsonc-parser';
 import { debounce } from './debounce';
 
 const fetcher = createGraphiQLFetcher({
-  url: 'https://api.spacex.land/graphql/',
+  url: 'https://countries.trevorblades.com',
```
B2o5T, line 10: previous endpoint is no longer works

examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -1,69 +1,67 @@
...
-const execOperation = async function () {
+async function execOperation(): Promise<void> {
   const variables = editor.getModel(Uri.file('variables.json'))!.getValue();
   const operations = editor.getModel(Uri.file('operation.graphql'))!.getValue();
   const resultsModel = editor.getModel(Uri.file('results.json'));
-  // @ts-expect-error
   const result = await fetcher({
     query: operations,
-    variables: JSON.stringify(JSONC.parse(variables)),
+    variables: JSONC.parse(variables),
```
B2o5T, line 56: here was a mistake, we should pass object and not string

examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -156,52 +148,50 @@ export default function App() {
    * Handle the initial schema load
    */
   useEffect(() => {
-    if (!schema && !loading) {
```
B2o5T, line 159: refactored to `earlier return` to avoid spagetti code
```js
    if (schema || loading) {
      return;
    }
```

examples/monaco-graphql-webpack/src/index.ts
```diff
@@ -143,36 +143,34 @@ async function render() {
...
-          body: JSON.stringify(body, null, 2),
-        });
-
-        const resultText = await result.text();
-        resultsEditor.setValue(JSON.stringify(JSON.parse(resultText), null, 2));
-      } catch (err) {
-        if (err instanceof Error) {
-          resultsEditor.setValue(err.toString());
-        }
+  const getOperationHandler = () => async () => {
```
B2o5T, line 146: I only removed `return` to have less indent on left side

examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -156,52 +148,50 @@ export default function App() {
...
+            allowComments: true,
+            trailingCommas: 'ignore',
+          },
+        },
+        schemas: [{ introspectionJSON, uri: 'myschema.graphql' }],
+      });
 
-          setSchema(data.data);
-        })
-        .then(() => setLoading(false));
```
B2o5T, line 193: here `.then(...)` is redundant, we can call `setLoading(false)` right after `setSchema(...)`

examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -1,69 +1,67 @@
...
-`;
-
-const getSchema = async () =>
-  fetcher({
+  `{
+  // 'code' will appear here as autocomplete,
+  // and because the default value is 0, will
+  // complete as such
+  $1
+}`;
```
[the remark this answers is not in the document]
acao: @B2o5T my bad, I can't remember what went wrong here, it has to do with an option that needs to be enabled somewhere, let's remove for now

examples/monaco-graphql-react-vite/src/App.tsx
acao: hm,,, this used to work! but perhaps monaco-graphql changed

examples/monaco-graphql-react-vite/src/App.tsx
acao: true, it's not even proper async haha. maybe if we returned setSchema() but it's not necessary

examples/monaco-graphql-react-vite/src/App.tsx
acao: oh ok! let's add a default query for this new schema perhaps?

we can also use our own to make sure it continues working

B2o5T: I changed the default query too!
https://github.com/graphql/graphiql/blob/bf99ea8e8d4b50c58e934ce3a1bf3b51e0991cdf/examples/monaco-graphql-react-vite/src/App.tsx#L19-L25

examples/monaco-graphql-react-vite/src/App.tsx
```diff
@@ -1,7 +1,7 @@
 import { ReactElement, useEffect, useRef, useState } from 'react';
 import { getIntrospectionQuery, IntrospectionQuery } from 'graphql';
 import { Uri, editor, KeyMod, KeyCode, languages } from 'monaco-editor';
-import { initializeMode } from 'monaco-graphql/src/initializeMode';
+import { initializeMode } from 'monaco-graphql/dist/initializeMode';
```
B2o5T: @acao this is big mistake 🥲

acao: i couldn't get the dist bundling working before, but this example is from last year, it seems monaco graphql bundling had changed since

PR #3143, opened 2023-04-18
[ESLint] enable `sonar/prefer-promise-shorthand` and `sonar/no-dead-store` rules

B2o5T:


packages/vscode-graphql-execution/src/providers/exec-content.ts
```diff
@@ -37,10 +37,8 @@ export class GraphQLContentProvider implements TextDocumentContentProvider {
 
   timeout = (ms: number) => new Promise(res => setTimeout(res, ms));
 
-  getCurrentHtml(): Promise<string> {
-    return new Promise(resolve => {
-      resolve(this.html);
-    });
+  async getCurrentHtml(): Promise<string> {
```
B2o5T, line 40: I think this should be sync too
```suggestion
  async getCurrentHtml(): string {
```

packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -353,11 +353,6 @@ export class MessageProcessor {
         await this._updateGraphQLConfig();
         return { uri, diagnostics: [] };
       }
-      // update graphql config only when graphql config is saved!
-      const cachedDocument = this._getCachedDocument(uri);
-      if (cachedDocument) {
-        contents = cachedDocument.contents;
```
B2o5T, line 359: there is no need to reassign to `contents` since this variable is declared inside current function

PR #3148, opened 2023-04-21
feat: Use native LSP logger

mskelton:
The current LSP logger implementation fails in Neovim with the following error.

```
[ERROR][2023-04-20 22:36:19] .../vim/lsp/rpc.lua:734    "rpc"   "/Users/mark/.local/share/npm/bin/graphql-lsp"  "stderr"        '4/20/2023, 10:36:19 PM [3] (pid: 84902) graphql-language-service-usage-logs: {"type":"usage","messageType":"initialize"}\n'
```

Digging into this a little, I notice that the GraphQL LSP is using a custom logger implementation that is writing files to disk rather than the built-in capabilities of LSP for logging messages.

This PR switches the logger to use the native LSP capabilities by using `connection.console` helpers. This is how many other language servers handle log messages.  For example:

- https://github.com/microsoft/vscode-eslint/blob/0608fb403df8553ae75ae3ac4ad71fb9baf22ad7/server/src/eslint.ts#L890
- https://github.com/microsoft/azuredatastudio/blob/f83815cecde9789413382c008485c03be6609199/extensions/html-language-features/server/src/utils/validation.ts#L55

I renamed the old logger to `ConsoleLogger` since it's being used in unit tests.  Likely that could be replaced with a noop logger, but I figured I'd get your thoughts first on the rest of the PR.

Also, part of this change involves no longer exporting `Logger` from `graphql-language-service-server`. While technically a breaking change, does that warrant being a breaking change when I'm guessing the only usage was in `graphql-language-service-cli` as it seems pretty unlikely that consumers of `graphql-language-service-server` would ever import and use `Logger`. 🤷‍♂️


acao (changes requested):
awesome! I believe this was an artefact from the pre-LSP days (yes this server was originally implemented without  `vscode-languageserver` and the LSP spec)!

I don't think we need to use what is now ConsoleError anymore, like you say, a no-op logger we can use for asserting errors is all we need!

packages/graphql-language-service-server/src/parseDocument.ts
```diff
@@ -48,7 +49,7 @@ export function parseDocument(
   uri: string,
   fileExtensions: string[] = DEFAULT_SUPPORTED_EXTENSIONS,
   graphQLFileExtensions: string[] = DEFAULT_SUPPORTED_GRAPHQL_EXTENSIONS,
-  logger: Logger = new Logger(),
+  logger: Logger = new ConsoleLogger(),
```
acao, line 52: I don't think we even need to instantiate a logger as a default here, not sure why it was added to begin with! again i think this should be fixed in the test suite

mskelton: The default value is there due to tests calling `parseDocument` with only two args since `fileExtensions` and `graphQLFileExtensions` have default values.  I added a simple noop logger for tests.

mskelton:
Thanks for the feedback, I've updated the PR with a noop logger.

acao:
thank you @mskelton for modernizing this! so happy to get rid of the file logger

acao:
Screenshot 2023-05-08 at 13 56 46

😎

mskelton:
Got main merged in and tests fixed.

acao (approved):
Excellent work, this looks read to go!

Issue #2874, opened 2022-11-08
[lsp-server] Off by one error in hover information location


DanielPower:
graphql-lsp version 3.3.11

When hovering an identifier, there appears to be an off by one error, which results in the first character of an identifier not being hoverable, but the first space after the identifier is hover-able, and recognized as part of the identifier.

I've reproduced this on both VS Code and Neovim.

Neovim

https://user-images.githubusercontent.com/12166320/200630624-a6362b41-4cba-4407-9a8b-fa5413d5c067.mov

VS Code

https://user-images.githubusercontent.com/12166320/200630840-a2b0b366-49ca-4ef7-bb50-866a677365f9.mov

## Desired Behavior

Hovering the first character of an identifier should provide hover information.

acao:
@DanielPower thank you! i will take a look at this bug

PR #3149, opened 2023-04-21
fix: Hover docs were offset one position

mskelton:
Fixes https://github.com/graphql/graphiql/issues/2874

The hover docs were off by one position. This was due to autocomplete suggestions and hover docs using the same `getTokenAtPosition` function. For autocomplete, the behavior of the function was expected since your cursor is _after_ the token you want to get when requesting autocomplete results, but with hover docs you want the cursor _at_ the cursor position.

### DIFF
--- a/.changeset/metal-cats-hammer.md
+++ b/.changeset/metal-cats-hammer.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service': patch
+---
+
+Fix hover docs being off by one position.
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -165,7 +165,7 @@ export function getAutocompleteSuggestions(
     schema,
   };
   const token: ContextToken =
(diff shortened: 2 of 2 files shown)


acao:
@mskelton thanks for this! can you check that the hover positions for `codemirror-graphql` and `monaco-graphql` are still correct after this change? I know in `monaco-graphql` we re-use this method for certain, and the way that monaco and LSP server offsets work are slightly different, the former being base 1 and the latter base 0 if I recall offhand.  you could test with the `examples/monaco-graphql-webpack` example

mskelton:
@acao I tried running the monoco-graphql-webpack example, but hovering didn't display any sort of information, so I'm not exactly sure how to validate it worked/didn't. I noticed the same behavior on main as well, so I'm a little confused.

acao:
ohhh it's off by one _character_ not by one _line_. sorry it took me so long to look closely at this. ok, we don't need to worry as much about the other implementations

acao:
@mskelton in case you're wondering, this will be released in an alpha first, before we return to stable state, because graphiql@3 is coming!

https://github.com/graphql/graphiql/pull/3219

PR #3152, opened 2023-04-25
chore(graphiql): refactor Short Keys table

B2o5T:


acao:
hey thanks for this! i suggested this change in the initial peer review but we decided to come back to it later.

sorry, looks like this deploy preview issue is still happening, I will look into it further. last time I solved this bug a few weeks ago, the issue was because someone had removed `yarn build` from one root script, i can't remember which..

Issue #2836, opened 2022-10-23
[lsp-server] svelte & vue issues with LSP server


tobiasdiez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following error is logged:
```
10/23/2022, 4:15:03 PM [1] (pid: 1704) graphql-language-service-usage-logs: Could not parse the Vue file at file:///workspace/components/JabrefLogo.vue to extract the graphql tags:
10/23/2022, 4:15:03 PM [1] (pid: 1704) graphql-language-service-usage-logs: Error: [@vue/compiler-sfc] SFC contains no <script> tags.
```

### Expected Behavior

No error is logged, and the vue file is simply ignored.

### Steps To Reproduce

Just have one vue file without a script block, then the above error is logged in the vscode console.
Example:
```
<template>
  <div>
    Hello!
  </div>
</template>
```

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version:
- OS:
- Browser:
- Bundler:
- `react` Version:
- `graphql` Version:
```

### Anything else?

The error is thrown here: https://github.com/vuejs/core/blob/b0b74a160c08941fa9a7a5460f36a1f2fccbf423/packages/compiler-sfc/src/compileScript.ts#L195-L198 So I guess it should be sufficient to check if the sfc descriptor has a script or script setup field, before calling the compileScript method.

acao:
hey @tobiasdiez thanks for reporting, I will take a look at this, I can see why this is an issue!

acao:
bah.  i think i will need to just move this to using the svelte compiler as such, we can only support a few formats with the vue compiler I think, it was a bit naïve of me to think this would work out haha. I will work on this and another mid-sized LSP bug this weekend.

https://github.com/sveltejs/svelte/blob/master/src/compiler/parse/index.ts

KammererTob:
If helpful: For vue this error is also thrown when the script tag is present but empty.

jycouet:
I just gave a try, let me know if I'm missing something 👍

PR #3157, opened 2023-04-27
fix: vue & svelte wo script tag

jycouet:
fixes: [#2836](https://github.com/graphql/graphiql/issues/2836)

When there is no `<script>` tag in the file, it's ok. We will not have any `graphql`.
Let's not log an error in this situation.

### DIFF
--- a/.changeset/eight-swans-destroy.md
+++ b/.changeset/eight-swans-destroy.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-server': patch
+---
+
+fix: vue and svelte files doesn't log errors anymore when parsing with no script tag (#2836)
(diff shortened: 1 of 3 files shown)


acao:
brilliant, thank you!

to answer your question, the lockfile will resolve new versions as long as dependencies or internal child dependencies have unpinned versions as usual! i think i had to disable esbuild bundling because we had bugs. i will take a closer look later in the week, but would love to get this out tomorrow to start the work week!

acao:
I trust you've tested this locally! If you can fix the eslint issues, such as the duplicate test title, we can get this merged today!

acao (approved):

Issue #2989, opened 2023-01-09
[lsp-server] 🐞 Goto definition of a Scalar defined in separate SDL file does not work


jogly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Goto-definition does not work ("definition not found") for scalars defined in other schema files. It does work for scalars defined in the same file.

### Expected Behavior

I expect Goto-definition of a scalar type to navigate to the location of definition for that scalar, regardless of which file it is declared in.

### Steps To Reproduce

Repro repo: https://github.com/jogly/graphql-goto-definition

1. Open the repo in VSCode with the GraphQL LSP extension installed with no custom settings.
1. In the file `query.gql`, goto does not work on the scalar reference `AScalar`, but does work on `ATypeDef`.
1. In the file `typedef.gql`, goto does work on the scalar reference `World`

### Environment

- LSP Server Version: v2.9.4
- OS: Mac OS 
- LSP Client: `vscode` v1.74.2 & `graphql.vscode-graphql` v0.8.4

PR #3163, opened 2023-05-07
Fix GraphQLCache to read both documents and schema

AaronMoat:
Resolves: https://github.com/graphql/graphiql/issues/2989 (and perhaps others)


acao:
thank you @AaronMoat

Issue #3080, opened 2023-03-05
[graphiql] move tests files outside src everywhere


B2o5T:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

to avoid publishing them since we publish `src` directory
![image](https://user-images.githubusercontent.com/7361780/222956198-477ecfa0-e62d-4329-b231-4d3198d03851.png)

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version:
- OS:
- Browser:
- Bundler:
- `react` Version:
- `graphql` Version:
```

Ronnie-personal:
@B2o5T  I will work on this issue shortly! Thanks.

acao:
ohhh this is about test files getting published. seems fine to me. i'm guessing there is no way to exclude files from `files` config?

PR #3177, opened 2023-05-16
Fixes graphql/graphiql#3080

Ronnie-personal:
Fixes graphql/graphiql#3080


B2o5T (commented):
I don't know if we want to apply these changes or just avoid bundling `src` directory with source typescript files

personally, I never saw that somebody bundle them, I would prefer just remove them from the bundle and find an alternative to source maps

cc @thomasheyenbrock @acao @jonathanawesome

acao:
this has been proposed before, but it breaks sourcemaps

PR #3054, opened 2023-02-25
pnpm mode WIP

acao:
- [x] `tsc --build` builds for all workspaces
- [ ] vite build
- [ ] webpack build
- [x] eslint errors
- [ ] eslint warnings
- [ ] jest resolution bugs
- [ ] cypress?

PR #3180, opened 2023-05-22
fix duplicated vite config

thomasheyenbrock:
I just noticed that my fix from #3017 got undone and discovered that we have duplicate vite configs for `@graphiql/react`. @acao do you have context on why this was added?

Anyways, I removed the JS config and updated the TS one with the latest changes from the JS one (which was the one that was used by Vite).


packages/graphiql-react/vite.config.js
```diff
@@ -1,42 +0,0 @@
-import { defineConfig } from 'vite';
-import react from '@vitejs/plugin-react';
-import reactSvgPlugin from 'vite-plugin-react-svg';
-import postCssNestingPlugin from 'postcss-nesting';
-// import noBundlePlugin from 'vite-plugin-no-bundle';
-export default defineConfig({
-  plugins: [
-    // TODO:  disable bundling dependencies
-    // noBundlePlugin(),
```
thomasheyenbrock, line 9: Not sure why we'd need this, we already exclude dependencies that shouldn't be bundled with `build.rollupOptions.external` 🤔

B2o5T (approved):
updated `.gitignore` to avoid this in the future https://github.com/graphql/graphiql/pull/3054#discussion_r1183541842 😃

packages/graphiql-react/vite.config.ts
```diff
@@ -11,12 +12,18 @@ export default defineConfig({
...
+      plugins: [postCssNestingPlugin()],
+    },
+  },
   esbuild: {
     // We use function names for generating readable error messages, so we want
     // them to be preserved when building and minifying.
     keepNames: true,
   },
   build: {
+    sourcemap: true,
```
B2o5T, line 26: should we still include `.ts` files in the bundle?https://discord.com/channels/625400653321076807/966768858402816020/1080508965768667236

PR #3183, opened 2023-05-23
3️⃣ replace `@reach/listbox` and  `@reach/menu-button` by `@radix-ui/react-dropdown-menu`

B2o5T:


packages/graphiql-react/src/toolbar/listbox.tsx
```diff
@@ -1,36 +0,0 @@
...
-      </Tooltip>
-      <Listbox.Popover>{children}</Listbox.Popover>
-    </Listbox.Input>
-  );
-});
-ToolbarListboxRoot.displayName = 'ToolbarListbox';
-
-export const ToolbarListbox = createComponentGroup(ToolbarListboxRoot, {
-  Option: Listbox.Option,
-});
```
B2o5T, line 36: I completely removed this component because it was nowhere used and I couldn't reproduce his proper usage to refactor by Radix

thomasheyenbrock (commented):
Changes look good so far, only have the comment about the removal of listbox

packages/graphiql-react/src/toolbar/listbox.tsx
B2o5T: Thanks @thomasheyenbrock for the helpful playground link

Can you explain what's the difference between `<ToolbarListbox />` with the current `<Menu.List />` ?
https://github.com/graphql/graphiql/blob/2b212941628498957d95ee89a7a5a0623f391b7a/packages/graphiql-react/src/toolbar/execute.tsx#L37-L61

I don't see any differences, for me, they do the same job, just show a dropdown 😅

https://github.com/graphql/graphiql/assets/7361780/96bbbc53-d0b1-45b3-8893-ad922dbc869d

thomasheyenbrock: The difference is that `Listbox` has state whereas `Menu.List` has not. (I imagine `Listbox` like being a radio button hidden in a dropdown, whereas `Menu.List` is just a list of buttons in a dropdown.) If you would hover over the listbox, you'd see that it says `Fruits: Apple` because that one is currently selected. Hovering over the execution button just shows `Execute`, it doesn't keep track of any "selected operation".

PR #3184, opened 2023-05-23
4️⃣ replace `@reach/combobox` with `Combobox` from `@headlessui/react`

B2o5T:
currently, radix doesn't have `Combobox` unstyled component (a related issue https://github.com/radix-ui/primitives/issues/1342)

After a long googling I found a perfect alternative a lightweight [`cmdk`](https://github.com/pacocoursey/cmdk) that uses under the hood only 1 dependency `@radix-ui/react-dialog` but it has peerDependency of React 18 because it uses React.useId https://github.com/pacocoursey/cmdk/blob/db1e29aa7ee173bf3ac269f9f815ee45ebbabd56/cmdk/src/index.tsx#L165

So I propose to release GraphiQL v3 with replaced Reach UI with Radix/CMDK with React 18 and bump GraphiQL v4 with Monaco instead Codemirror

### DIFF
--- a/packages/graphiql-react/src/explorer/components/search.tsx
+++ b/packages/graphiql-react/src/explorer/components/search.tsx
@@ -1,11 +1,4 @@
 import {
-  Combobox,
-  ComboboxInput,
-  ComboboxPopover,
-  ComboboxList,
-  ComboboxOption,
-} from '@reach/combobox';
-import {
   GraphQLArgument,
   GraphQLField,
   GraphQLInputField,
@@ -15,6 +8,7 @@ import {
   isObjectType,
 } from 'graphql';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { Command } from 'cmdk';
 import { MagnifyingGlassIcon } from '../../icons';
 import { useSchemaContext } from '../../schema';
 import debounce from '../../utility/debounce';
@@ -31,8 +25,6 @@ export function Search() {
   });
 
   const inputRef = useRef<HTMLInputElement>(null);
-  const popoverRef = useRef<HTMLDivElement>(null);
-
   const getSearchResults = useSearchResults();
   const [searchValue, setSearchValue] = useState('');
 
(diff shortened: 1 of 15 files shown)


packages/graphiql-react/src/explorer/components/search.tsx
```diff
@@ -61,124 +53,99 @@ export function Search() {
...
-            </ComboboxOption>
-          ))}
-          {results.within.length +
-            results.types.length +
-            results.fields.length ===
-          0 ? (
-            <div className="graphiql-doc-explorer-search-empty">
-              No results found
-            </div>
-          ) : null}
```
B2o5T, line 177: cmdk exports abstract `<Command.Empty />` so this code was removed and replace by 

```ts
         <Command.Empty className="graphiql-doc-explorer-search-empty">
            No results found
          </Command.Empty>
```

jonathanawesome:
_Thank you_ @B2o5T for all of the excellent work.

We should be cautious about moving to React 18 before soliciting opinions from the community...it could be a deal-breaker for certain folks. While we will have to make the move eventually, there are alternatives to the combobox problem. 

Last year I started to feel a bit sketchy about the situation with Radix after the acquisition and went looking for alternatives. The folks over at Tailwind have put together a great set of React components with the [headlessui](https://github.com/tailwindlabs/headlessui) package. The API felt familiar enough to the Radix implementation that I was able to migrate my existing Tabs and Dialog components with very little fuss. Their combobox component looks great, but I, admittedly, haven't yet had to implement. Maybe it's a non-react-18 option for us here?

B2o5T:
> We should be cautious about moving to React 18 before soliciting opinions from the community...it could be a deal-breaker for certain folks. While we will have to make the move eventually, there are alternatives to the combobox problem.

you are right, support react 17 will be better

> The folks over at Tailwind have put together a great set of React components with the [headlessui](https://github.com/tailwindlabs/headlessui) package.

I don't have a strong opinion on this, I used radix in graphql-hive and headlessui in Nextra, and they do mostly the same job 😅

> Their combobox component looks great, but I, admittedly, haven't yet had to implement. Maybe it's a non-react-18 option for us here?

I didn't know that they have a combobox component, if they have so it will solve issue with supporting both React 17/18 ❤️

thomasheyenbrock:
Haven't done much research on this, but what are the parts of `reach-ui` that don't work with React 18? Does `combobox` maybe work with React 18?

In any case, if we find a solution that works with both React 17 _and_ 18 that would be awesome! Since we only use this component for the search functionality in the docs explorer which is fairly simple I could also imagine us doing a hand-rolled solution with "good enough" a11y support to unblock this.

B2o5T:
> Haven't done much research on this, but what are the parts of reach-ui that don't work with React 18? Does combobox maybe work with React 18?

reach-ui technically works with react 18, but it has downside
1. he is no longer maintained https://github.com/reach/reach-ui/issues/972
2. he doesn't have react 18 as peerDependency range https://github.com/reach/reach-ui/blob/43f450db7bcb25a743121fe31355f2294065a049/packages/tooltip/package.json#L35 so users have warnings while using it with React 18

> In any case, if we find a solution that works with both React 17 and 18 that would be awesome! Since we only use this component for the search functionality in the docs explorer which is fairly simple I could also imagine us doing a hand-rolled solution with "good enough" a11y support to unblock this.

Building combobox component from scratch will be brainstorm, I think we should go with @jonathanawesome proposal of using headlessui combobox https://headlessui.com/react/combobox

acao:
@thomasheyenbrock we have already decided to replace reach ui, users are having lots of issues with it downstream

packages/graphiql/cypress/e2e/docs.cy.ts
```diff
@@ -21,42 +21,43 @@ describe('GraphiQL DocExplorer - button', () => {
...
+      'have.text',
+      'Other results',
+    );
+    cy.dataCy('doc-explorer-option').contains('hasArgs');
   });
 
-  it('Closes popover when blurring input', () => {
-    cy.get('[data-reach-combobox-input]').blur();
-    cy.get('[data-reach-combobox-popover]').should('have.attr', 'hidden');
+  it(' Closes popover when blurring input', () => {
```
B2o5T, line 48:
```suggestion
  it('Closes popover when blurring input', () => {
```

PR #3203, opened 2023-05-31
fix: make tooltips work when Graphiql is not full page and window has scrolled

### LINKED ISSUE #3202: [graphiql] Info popups lose position on page scroll

### Current Behavior

If not using the editor as full page (have other divs above or below and scroll enabled by turning off the body css overflow: hidden), the info popups will lose their place on the page when you scroll. I think it's coming from codemirror-graphql > src > utils > info-addon > onMouseOver > info on line 104. I wasn't sure if there's a way to try re-calculating the info if it's not found based on where the editor has scrolled to.

<table>
  <tr>
      <th>initial position person tooltip works</th>
      <th>scolled position person tooltip broke</th>
  </tr>
  <tr>
      <td>initial position works</th>
      <td>scrolled position broke</th>
  </tr>
</table>

### Expected Behavior

Should be able to see the hover tooltips

### Steps To Reproduce

```Markdown
1. Temporarily adjust grapiql > src > components > Graphiql.tsx to 
    <div>
       <div style={{height: '200px'}}/>
       ...GraphiQLProvider element that's there
       <div style={{height: '200px'}}/>
    </div>
2. graphiql > resources > index.html.ejs temporarily comment out the body css "overflow: hidden"
```

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version:
- OS:
- Browser:
- Bundler:
- `react` Version:
- `graphql` Version:
```

### Anything else?

_No response_

lesleydreyer:
Makes tooltips work when not using Graphiql full page and window has scrolled.

Closes #3202

### DIFF
--- a/.changeset/sour-monkeys-film.md
+++ b/.changeset/sour-monkeys-film.md
@@ -1,0 +1,5 @@
+---
+'codemirror-graphql': patch
+---
+
+fix info tooltips to work when Graphiql is not used as full page
--- a/packages/codemirror-graphql/src/utils/info-addon.ts
+++ b/packages/codemirror-graphql/src/utils/info-addon.ts
@@ -90,10 +90,13 @@ function onMouseOver(cm: CodeMirror.Editor, e: MouseEvent) {
 }
 
 function onMouseHover(cm: CodeMirror.Editor, box: DOMRect) {
-  const pos = cm.coordsChar({
-    left: (box.left + box.right) / 2,
-    top: (box.top + box.bottom) / 2,
-  });
+  const pos = cm.coordsChar(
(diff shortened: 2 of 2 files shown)


thomasheyenbrock (commented):
Amazing, thanks for catching that and already implementing a fix 🙌 👏 

Can you please also add a changeset file so that we can release this fix with a version bump? You can create one by running `yarn changeset` in the repository root and follow through the CLI prompts.

B2o5T (approved):

PR #3204, opened 2023-05-31
fix: use Svelte parser for Svelte files

tadeokondrak:
Using the Vue parser here leads to broken behavior when files use more
than basic Svelte syntax.


packages/graphql-language-service-server/package.json
```diff
@@ -41,6 +41,9 @@
     "@babel/types": "^7.21.2",
     "@graphql-tools/load": "^7.5.3",
     "@vue/compiler-sfc": "^3.2.41",
+    "typescript": "^4.8",
```
B2o5T, line 44: why put typescript in `dependency` and not in `devDependencies`?

tadeokondrak: `svelte2tsx` or one of its dependencies calls methods in `typescript` that were added in 4.8.

B2o5T:
could you sign CLA and add some tests to https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service-server/src/__tests__/findGraphQLTags-test.ts so there will be no regression in the future?

tadeokondrak:
Can't get the CLA site to work unfortunately

acao:
@tadeokondrak just make sure your git commits use the same email you signed the CLA with!

tadeokondrak:
That isn't the issue - downloading the PDF of the agreement gives me a .pdf file containing only the following:

```
{"Message":"applying watermark failed : pdfcpu: validateStringEntry: dict=FreeText entry=DA invalid type","x-request-id":"c618f34a-b516-4fe5-9e77-5bb5af850624"}
```

Sorry, but I'd rather not go through the CLA support process to make this contribution.

jorydotcom:
@tadeokondrak I have filed an issue with the EasyCLA team to see if we can get this resolved for you.

jorydotcom:
@tadeokondrak the CLA team confirmed your error and has pushed a fix; would you be willing to give it another go? We appreciate your contribution!

acao:
@tadeokondrak reviving this PR as a new PR now, can you give me some test cases where svelte crashes?

tadeokondrak:
> reviving this PR as a new PR now

Great!

> can you give me some test cases where svelte crashes?

Not sure why, but this breaks with the old parser:

```svelte
<script>
	graphql(`
		mutation Test {
			echo(m: "message")
		}
	`)
</script>

<div>
	<div>Text</div>
</div>
```

acao:
ah!  my guess is that is because vue expects `<template>` tags around html, whereas svelte in this case would parse anything outside of `<script/>` as html

acao:
ah, this introduced a bug with typescript module

acao:
@tadeokondrak unfortunately using svelte2tsx introduces bundling issues with the typescript peer dependency. this isn't revealed with our unit tests, or by the local dev runner or even the vscode test runner, because it requires manually testing with the bundled vsix extension

I am going to have to revert this change so that we can get the LSP server working again. You are welcome to re-create this PR again either without using a dependency with a peer dependency on typescript, or using svelte2tsx in a way that you can prove works with the bundled vscode-graphql vsix

Issue #3187, opened 2023-05-23
[graphiql] Run dropdown list


jetsstarplus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With multiple queries and mutations on the web editor, when I hit the run button a list of the commands are shown and if they exceed a certain number, some of them are hidden from the dropdown list.
Its only until I zoom in the UI for me to see the remaining commands.
The Z-Index of the dropdown is also overridden by those of the editors left pane overflow bars.

### Expected Behavior

The dropdown, should have a maximum height that should have a scroll behavior incase the items in it exceed it's height.
The z-index of the dropdown should also be higher than those of the editor pane scrollbars.

### Steps To Reproduce

```Markdown
1. Use the CDN provided for both the minified css and js for the HTML File.
```

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version:latest
- OS:Windows 10
- Browser: Chrome 106
- Bundler:
- `react` Version:18
- `graphql` Version:16
```

B2o5T:
could you attach screenshot or video?

jetsstarplus:
Ofcourse, Here is normal view. 
As you can see, the scroll bars are on top of the dropdown
![graphiql](https://github.com/graphql/graphiql/assets/52789494/d5f00455-314d-46d1-87ec-ecebb7c9a762)

The dropdown also does not show all commands, here is the view with 33% zoom
![graphiql2](https://github.com/graphql/graphiql/assets/52789494/24b1e36c-cd7f-450a-9441-e40419b221fa)

acao:
I remember this bug in 0.x. We set the max height dynamically using vh if i recall, but that caused issues on resize of course, so we set a fixed pixel height

B2o5T:
i think it's should be fixed with https://github.com/graphql/graphiql/pull/3183 will try it once it will be merged!

jetsstarplus:
Thats Ok

B2o5T:
@jetsstarplus you can try a fix in `graphiql@3.0.0-alpha.0`

jetsstarplus:
> @jetsstarplus you can try a fix in `graphiql@3.0.0-alpha.0`

Yes It is Ok now, It has a scroll behavior.
Though the scrollbar is still on top of the dropdown as in the picture below.
I'm using the latest chrome browser on windows
[...]

PR #3206, opened 2023-06-01
fixed long list items in dropdown were hidden, rename `<Menu />` to `<DropdownMenu />`

B2o5T:
fixes https://github.com/graphql/graphiql/issues/3187 cc @jetsstarplus

https://github.com/graphql/graphiql/assets/7361780/f17569a5-8ac0-4dac-99dc-be4f695ffee3

### DIFF
--- a/packages/graphiql-react/src/ui/dropdown.css
+++ b/packages/graphiql-react/src/ui/dropdown.css
@@ -8,7 +8,7 @@
   padding: var(--px-4);
   font-family: var(--font-family);
   color: hsl(var(--color-neutral));
-  max-height: calc(var(--radix-dropdown-menu-content-available-height) - 10px);
+  max-height: min(calc(var(--radix-dropdown-menu-content-available-height) - 10px), 400px);
   overflow-y: scroll;
 }
 
--- a/packages/graphiql-react/src/ui/dropdown.css
+++ b/packages/graphiql-react/src/ui/dropdown.css
@@ -1,4 +1,4 @@
-[data-radix-menu-content] {
+.graphiql-dropdown-content {
   background-color: hsl(var(--color-base));
   border: var(--popover-border);
   border-radius: var(--border-radius-8);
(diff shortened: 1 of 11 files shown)


packages/graphiql-react/src/ui/dropdown.tsx
```diff
@@ -1,40 +1,52 @@
...
+      >
         {children}
-      </DropdownMenu.Content>
-    </DropdownMenu.Portal>
+      </DM.Content>
+    </DM.Portal>
   );
 }
 
-export const Menu = createComponentGroup(DropdownMenu.Root, {
```
B2o5T, line 36: @thomasheyenbrock I renamed Menu to DropdownMenu because

1. for me menu, listbox are confused names
2. we follow radix convention
3. even this file is name dropdown 😅
4. major release is coming so we can do some adjustments :)

packages/graphiql-react/src/ui/dropdown.css
```diff
@@ -1,4 +1,4 @@
-[data-radix-menu-content] {
+.graphiql-dropdown-content {
```
B2o5T, line 1: to avoid conflicts with users _maybe_ radix components

packages/graphiql-react/src/ui/dropdown.css
```diff
@@ -8,9 +8,11 @@
   padding: var(--px-4);
   font-family: var(--font-family);
   color: hsl(var(--color-neutral));
+  max-height: calc(var(--radix-dropdown-menu-content-available-height) - 10px);
```
B2o5T, line 11: https://www.radix-ui.com/docs/primitives/components/dropdown-menu#constrain-the-contentsub-content-size

saihaj (approved):

packages/graphiql-react/src/ui/dropdown.tsx
thomasheyenbrock: Sounds good 👍 

Since we're still on `0.x`, I'd even be fine with doing this in a minor. The expectations around `@graphiql/react` always were that it's not yet stable.

packages/graphiql-react/src/ui/dropdown.css
thomasheyenbrock: Using all the available height seems too much for me, having the dropdown fill the whole height of the page looks weird. I'd propose to go with a smaller value here like `400px`.

B2o5T: hardcoding 400px would provoke issues on small `vh`, we should use css' `min` function with it

https://github.com/graphql/graphiql/assets/7361780/703a8a3f-d227-475b-884a-e6d89bdf2cd7

PR #3216, opened 2023-06-07
fix: reload schema when a change to the schema file is detected

simowe:
# Summay
Currently, the schema is cached permanently and never reloaded. This makes it very awkward when you're developing the graphql api and the front-end at the same time, as the auto completion and validation will rely on incorrect data.

## Notes
I found a function named `_invalidateSchemaCacheForProject` that did what I needed, however it's not actually used anywhere. I assume this was removed by accident, so I found somewhere I think is appropriate to invoke it.

## Changes
* Invalidate the schema cache when changed in handleWatchedFilesChangedNotification
* Add .json to watched files so it will pick up changes to introspection files

### DIFF
--- a/packages/vscode-graphql/src/extension.ts
+++ b/packages/vscode-graphql/src/extension.ts
@@ -76,7 +76,7 @@ export async function activate(context: ExtensionContext) {
         // TODO: load ignore
         // These ignore node_modules and .git by default
         workspace.createFileSystemWatcher(
-          '**/{*.graphql,*.graphqls,*.gql,*.js,*.mjs,*.cjs,*.esm,*.es,*.es6,*.jsx,*.ts,*.tsx,*.vue,*.svelte,*.cts,*.mts}',
+          '**/{*.graphql,*.graphqls,*.gql,*.js,*.mjs,*.cjs,*.esm,*.es,*.es6,*.jsx,*.ts,*.tsx,*.vue,*.svelte,*.cts,*.mts,*.json}',
         ),
       ],
     },
--- a/.changeset/soft-dingos-report.md
+++ b/.changeset/soft-dingos-report.md
@@ -1,0 +1,6 @@
+---
+'graphql-language-service-server': patch
+'vscode-graphql': patch
+---
+
+fix: reload schema when a change to the schema file is detected
--- a/packages/graphql-language-service-server/src/GraphQLCache.ts
+++ b/packages/graphql-language-service-server/src/GraphQLCache.ts
@@ -671,7 +671,7 @@ export class GraphQLCache implements GraphQLCacheInterface {
     return schema;
   };
 
-  _invalidateSchemaCacheForProject(projectConfig: GraphQLProjectConfig) {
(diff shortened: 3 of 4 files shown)


acao:
@simowe awesome work! have you tested this with both URL schema and file schema?

also, please add a changeset summarizing the work as per the bot comment

lastly, there is one failing test, if you can try to fix it locally

packages/vscode-graphql/src/extension.ts
```diff
@@ -76,7 +76,7 @@ export async function activate(context: ExtensionContext) {
         // TODO: load ignore
         // These ignore node_modules and .git by default
         workspace.createFileSystemWatcher(
-          '**/{*.graphql,*.graphqls,*.gql,*.js,*.mjs,*.cjs,*.esm,*.es,*.es6,*.jsx,*.ts,*.tsx,*.vue,*.svelte,*.cts,*.mts}',
+          '**/{*.graphql,*.graphqls,*.gql,*.js,*.mjs,*.cjs,*.esm,*.es,*.es6,*.jsx,*.ts,*.tsx,*.vue,*.svelte,*.cts,*.mts,*.json}',
```
B2o5T, line 79:
```suggestion
          '**/*.{graphql,graphqls,gql,js,mjs,cjs,esm,es,es6,jsx,ts,tsx,vue,svelte,cts,mts,json}',
```

@acao I think this can be simplified to ?

B2o5T: also `esm` and `es6` extension should be removed 😅

B2o5T:
```suggestion
          '**/*.{graphql,graphqls,gql,js,mjs,cjs,jsx,ts,tsx,vue,svelte,cts,mts,json}',
```

simowe:
> @simowe awesome work! have you tested this with both URL schema and file schema?
> 
> also, please add a changeset summarizing the work as per the bot comment
> 
> lastly, there is one failing test, if you can try to fix it locally

Thanks for the feedback! I've fixed the failing test and added the changeset.

This PR will only reload the schema when the local schema file itself is changed, no other events trigger a reload. Therefore a URL schema will be unaffected by this. 

I'm not sure how to best handle a URL schema. Possibly polling on a set interval or reloading the schema whenever any file changes?

packages/vscode-graphql/src/extension.ts
acao: @B2o5T these unusual seeming extensions were requested by users earlier this year and last, and there were lots of emoji reacts :shrug: anecdotal expderience isn't always enough to evaluate userland haha

acao:
@simowe this is the tricky part about URL schemas - we have to (hope) assume the introspection request is configured with cache headers, or just make a fresh request every time we need the schema. More intricate polling mechanisms would be nice but there are no standards for this, just many userland implementations

acao:
@simowe I think this is fine for now. we can solve the problem for url schemas later

Issue #3225, opened 2023-06-13
broken link in docs


ChristophHoyer:
Hej there!
In packages/graphiql#readme the first link in the "Live Demos" is broken.
According to its description it should link to a "latest stable version" - but instead I only get an **Ooops! Page not found!**

B2o5T:
fixed, thx!

PR #3227, opened 2023-06-13
fix broken graphql.org links Update README.md

B2o5T:
fixes https://github.com/graphql/graphiql/issues/3225


acao:
Thanks for patching this! 

We should have this link resolving though, for historical reasons (it's in many docs and training modules in the wild)

https://graphql.org/swapi-graphql

let's open a ticket over there. i previously had a _redirects rule for this for netlify, but something must have changed on the graphql.org side of things

B2o5T:
@acao yeah! there is already a ticket https://github.com/graphql/graphql.github.io/issues/1427

PR #3228, opened 2023-06-14
[@graphiql/react]: exclude peer dependencies and dependencies from bundle

B2o5T:
| before                                                                                  | after |
|:---------------------------------------------------------------------------------------------------:|:--------------:|
| image |     image         |

also I renamed `main`/`module` entry to simple `index`


packages/graphiql-react/vite.config.ts
```diff
@@ -26,11 +27,18 @@ export default defineConfig({
...
       formats: ['cjs', 'es'],
     },
     rollupOptions: {
-      external: ['graphql', 'react', 'react-dom', 'react/jsx-runtime'],
+      external: [
+        'react/jsx-runtime',
+        // Exclude peer dependencies and dependencies from bundle
+        ...Object.keys(packageJSON.peerDependencies),
+        ...Object.keys(packageJSON.dependencies).filter(
+          dependency => dependency !== 'codemirror',
```
acao, line 39: is there a reason we leave codemirror?

acao (approved):
much needed! i can sleep at night now, knowing that our users can security patch `markdown-it` for example

packages/graphiql-react/vite.config.ts
B2o5T: yeah there were some errors from `graphiql` without this filter, anyway we'll remove it with Monaco migration

acao: I think when we replace webpack with vite in graphiql we can get rid of this beforehand

PR #3229, opened 2023-06-14
[plugin-code-exporter/plugin-explorer]: exclude peer dependencies and dependencies from bundle

B2o5T:
## `@graphiql/plugin-explorer`
| before                                                                                  | after |
|:---------------------------------------------------------------------------------------------------:|:--------------:|
| image |   image        |

## `@graphiql/plugin-code-exporter`

| before                                                                                  | after |
|:---------------------------------------------------------------------------------------------------:|:--------------:|
| image |     image      |

### DIFF
--- a/.changeset/little-stingrays-sin.md
+++ b/.changeset/little-stingrays-sin.md
@@ -1,0 +1,6 @@
+---
+'@graphiql/plugin-code-exporter': patch
+'@graphiql/plugin-explorer': patch
+---
+
+exclude peer dependencies and dependencies from bundle
--- a/packages/graphiql-plugin-code-exporter/package.json
+++ b/packages/graphiql-plugin-code-exporter/package.json
@@ -7,8 +7,8 @@
     "directory": "packages/graphiql-plugin-code-exporter"
   },
   "author": "LekoArts",
-  "main": "dist/graphiql-plugin-code-exporter.cjs.js",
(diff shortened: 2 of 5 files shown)


acao:
@B2o5T one issue here is - how do users use this plugin if they are using the CDN version of react?

B2o5T:
@acao how this was changed with the previous implementation?

acao:
@B2o5T there are a few examples where users were able to use the bundled CDN build of this plugin with the CDN graphiql. Each plugin should provide an unbundled module version and a bundled cdn version, as with GraphiQL, @graphiql/toolkit and @graphiql/react. We don't have to solve this problem now, but it will lead to a regression until we fix it

B2o5T:
@acao Since a new major is coming, we can go back to this problem in case of reporting it

thomasheyenbrock:
@B2o5T Too late with my review, sorry 🙇  @acao has a good point, and I don't think the new major will solve this. Removing the bundling of dependencies breaks the CDN examples, in particular this one: https://github.com/graphql/graphiql/blob/main/packages/graphiql-plugin-explorer/examples/index.html

At least the UMD build has to bundle up all dependencies, with the exception of the ones where we define globals: https://github.com/graphql/graphiql/blob/main/packages/graphiql-plugin-explorer/vite.config.ts#L23-L28

B2o5T:
@thomasheyenbrock ok! I'll fix it

PR #3230, opened 2023-06-14
replace unmaintained `vite-plugin-react-svg` with `vite-plugin-svgr`

B2o5T:
@thomasheyenbrock I improved your `generateIcon()` function 😅

## before (2 unneeded nested components)

image

## now, only 1

image


packages/graphiql-react/src/icons/index.ts
```diff
@@ -0,0 +1,77 @@
+import { ComponentProps, FC } from 'react';
```
acao, line 1: hmm... I'm used to doing something like this in a vite plugin in my proprietary work, is there a way we can configure the plugin to do this, or copy the plugin source and tweak it as such?

B2o5T: svgr plugin incorrectly shows the type `string` for imported svg files, even though we set up `exportAsDefault: true` in vite config, if you know how to fix it feel free

PR #3233, opened 2023-06-14
remove root's package.json `@types/react` and `@types/react-dom`

B2o5T:


acao:
probably should do this with most @types packages for the sake of typescript users

PR #3238, opened 2023-06-17
remove `/// <reference path='../../node_modules/monaco-editor/monaco.d.ts'/>` and simple import it

B2o5T:


acao:
please when doing a major monaco-editor upgrade, do regression testing and check the changelog. because monaco-editor is 0.x still, every minor release can have breaking changes!

B2o5T:
> please when doing a major monaco-editor upgrade, do regression testing and check the changelog. because monaco-editor is 0.x still, every minor release can have breaking changes!

@acao This monaco upgrade is used in `devDependency`, also different versions of monaco in yarn monorepo when everything is hoisted in an unexpected way produce broken styles as I mentioned in discord 

image 
image
https://discord.com/channels/@me/1113381551019925594/1118806612904247336

acao:
@B2o5T should we update `peerDependency` resolutions then?

B2o5T:
@acao I guess peer dependency already includes all possible `monaco-editor`  ?

https://github.com/B2o5T/graphiql/blob/6133657dbddc8939235b90e6f8aa67ab61b55542/packages/monaco-graphql/package.json#L39

acao:
@B2o5T I personally don't think we should support all versions of monaco editor in peer dependencies. I thought we had changed the range to a higher a minimum sometime earlier this year because of compatibility issues

acao (approved):
lgtm! we can deal with the peerDependency issue for graphiql@4.0.0 perhaps

B2o5T:
> @B2o5T I personally don't think we should support all versions of monaco editor in peer dependencies. I thought we had changed the range to a higher a minimum sometime earlier this year because of compatibility issues

you have a point! this will silently produce bugs since we can't support big range of monaco-editors

PR #3246, opened 2023-06-21
fix: tests without build

acao:
- [x] resolve monorepo tests without build
- [x] resolve `@graphiql/react` in tests without build
- [ ] mock icons


package.json
```diff
@@ -33,7 +33,7 @@
     "build": "yarn build:clean && yarn build:cm6-graphql && yarn build:packages && yarn build:graphiql-react && yarn build:graphiql-plugin-explorer && yarn build:graphiql-plugin-code-exporter && yarn build:graphiql",
     "build-bundles": "yarn prebuild-bundles && wsrun -p -m -s build-bundles",
     "build-bundles-clean": "rimraf '{packages,examples,plugins}/**/{bundle,cdn,webpack}' && yarn workspace graphiql run build-bundles-clean",
-    "build-clean": "wsrun -m build-clean ",
+    "build-clean": "wsrun -m build-clean && rimraf packages/**/{esm,cjs,dist,types}",
```
B2o5T, line 36: this will remove  directories from `node_modules` too

packages/graphiql/cypress/e2e/docs.cy.ts
```diff
@@ -1,7 +1,7 @@
 import { version } from 'graphql/version';
 
 beforeEach(() => {
-  cy.visit('/');
+  cy.visit('');
```
B2o5T, line 4: why this change? cypress doesn't recommend this, an recommend use `/`, is much more understandable and previously worked fine

acao: has to do with using the dev build, pushing a fix that satisfies this reccomendation now

PR #3250, opened 2023-06-21
Add basic support for 'go to references'

AaronMoat:


acao (changes requested):
Looks fantastic! Just a few minor nits. Also, can you add a test case for null result just so we can have a net positive coverage? Also it needs a rebase it looks like.

acao:
@AaronMoat thanks for this awesome PR! I have created a new PR based off of this, #3254. It works with fragments and will work for _all documents_ in a project when I'm finished,like definition lookup, after some refactoring of the LSP.

AaronMoat:
Thanks for the review, as well as cleaning up and expanding on this @acao! I'll close one off given you've extended off it (will have a look through your changes on the other PR)

PR #3252, opened 2023-06-21
[`@graphiql/plugin-explorer`] `@graphiql/react` should be in `peerDependencies` not in `dependencies`

B2o5T:


acao:
this makes sense! we will need to add a bare script import at least for the cdn example

acao (approved):

PR #3256, opened 2023-06-22
Delete index.js.flow

B2o5T:
saw with a new prettier setup https://github.com/graphql/graphiql/pull/3259 prettier wanted to prettify these files, I guess they are no longer used?


acao:
they are around for backwards compatibility for users. @benjie what do you think we should do with these?

acao:
@B2o5T the scripts can be removed in a separate PR like you had before, but the legacy flow-typed files are a different question. Can we do seperate PRs again please?

B2o5T:
@acao Yeah, I didn't see that the base PR is this one! will remove them in improvement prettier PR

acao:
@benjie @dwwoelfel are you or your users still using `flow-typed` with the language services? A cursory search of code on github showed only projects 4+ years old were still using `flow-typed` with our libraries

benjie:
I believe that Facebook is the main place where Flow and GraphiQL were used together. Maybe @mjmahone will have some clue as to whether this is still needed or not. Personally I'm in favour of removing it - no types is better than out of date types, and maintenance of them is a burden.

acao (approved):

PR #3257, opened 2023-06-22
Delete buildFlow.js

B2o5T:
I guess it is outdated scripts, we no longer build flow.


acao:
yes these are artifacts from when the repo used flow, and when we were transitioning from flow-typed to typescript a few years ago

PR #3259, opened 2023-06-22
simplify prettier checks commands, remove `script/pretty.js`

B2o5T:
I always wanted to simplify this homemade overcomplicated solution 🫠
- [x] prettier should not relay on `.eslintignore` but on `.gitignore` and `.prettierignore`
        add `--ignore-path .gitignore --ignore-path .prettierignore` (starting from prettier v3 he supports array of ignore paths)
- [x] remove `.eslintignore` and add `--ignorepath .gitgnore` instead for eslint too


acao:
@B2o5T looks good, another artifact from a time when the tooling possibilities were different

can you remove the redundant prettier check step in Lint job now?

acao:
image

image

it only adds a few seconds, but avoids confusion, no need to run it twice in the same workflow

B2o5T:
@acao nice catch! about few seconds - because I made setup `--cache` flag recently :)

PR #3261, opened 2023-06-22
Update CONTRIBUTING.md

B2o5T:
outdated


acao (approved):
awesome! yes we dropped commitlint with changesets but there is a way we can conventional commits with changesets, so perhaps it will come back. my biggest gripe with conventional commits is that it requires users to re-write commit messages, which many talented contributors don't always know how to do, or are excited about. many users have an aversion to force pushing I've noticed, and will sometimes even abandon PRs, if they aren't used to the rebase approach

next up, `Development.md`

PR #3263, opened 2023-06-22
some fixes for prettier/cspell

B2o5T:


packages/graphql-language-service/benchmark/fixtures/github.graphql
```diff
@@ -273,7 +273,7 @@ type BlameRange {
   startingLine: Int!
 }
 
-type Blob implements Node, GitObject {
+type Blob implements Node & GitObject {
```
B2o5T, line 276: this is not valid syntax that fails in prettier v3

acao: indeed, this is not spec valid syntax, I'm not sure where the author got it from

Issue #2343, opened 2020-11-13
proposal: introduce VSCode Textmate grammar test for testing grammar


yoshiakis:
Currently grammar files aren't being tested so it's likely to happen a regression when grammar files are edited in order to fix bugs or add new functionalities. I've looked for a node module that can test grammar files and I've found [VSCode Textmate grammar test](https://github.com/PanAeon/vscode-tmgrammar-test) which is being used in some of vscode extensions, such as [Angular Language Service](https://github.com/angular/vscode-ng-language-service) and [Scala Syntax](https://github.com/scala/vscode-scala-syntax), providing syntax highlighting feature. Both extensions are using this module more than 10 months now and there are almost no issues on that as far as I can see.

This module can do unit test and snapshot test. I think it's better to use mainly snapshot test since it needs less manual code than unit test, though sometimes unit test is also needed when changed part of snap file is large and you think you're likely to overlook regression.

What do you think?

If it looks good, I can write code for grammar test using this module.

acao:
this is a great idea @yoshiakis !

Issue #3217, opened 2023-06-08
[vscode-graphql-syntax] Problem with GraphQL Syntax Highlighting


B2o5T:
### Is there an existing issue for this?

- [X] I have searched the existing issues

acao:
yes this is where the issue would be. we badly need to add tests for the syntax extension, because minor improvements sometimes cause regressions, or we overlook lots of edge cases like these. i can pull up the vscode docs for writing tests for the grammar configs

B2o5T:
Yeah! I would like to add image snapshots tests before fixing some issues

PR #3267, opened 2023-06-24
Add test scaffolding for vscode-graphql-syntax

AaronMoat:
Resolves: #2343

Saw the discussion in #3217 and wanted to revive the older discussion in #2343. Main question - do we think this approach will be helpful, or should we look into something more style/visual based?

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,9 +74,9 @@
     "start-graphiql": "yarn workspace graphiql dev",
     "start-monaco": "yarn workspace example-monaco-graphql-webpack start",
     "t": "yarn run test",
-    "test": "yarn jest",
-    "test:ci": "yarn build:cm6-graphql && jest",
-    "test:coverage": "yarn jest --coverage",
+    "test": "yarn workspace vscode-graphql-syntax run test && yarn jest",
+    "test:ci": "yarn build:cm6-graphql && yarn test",
+    "test:coverage": "yarn test --coverage",
     "test:watch": "yarn jest --watch",
     "tsc": "tsc --build"
   },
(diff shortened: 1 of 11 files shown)


AaronMoat:
Ok, immediate problem, prettier autoformatting indents comments, whereas the package doesn't support unit tests unless the line literally starts with #

acao:
@AaronMoat this is awesome! I find it visual enough for our needs. and also, prettier should not be formatting snapshots anyways!

AaronMoat:
Hi @acao, I was actually talking about the `unit` tests. You can do stuff like... 

```graphql
query {
  something
# ^ do some assertions here
}
```

But prettier adds two spaces before the `#` and the "unit test" assertion no longer runs. I suppose we could ignore just the plain .graphql files too 🤔 

I'm tempted to see what this stuff looks by-hand with jest, etc, anyway, I'll have a play and put an alternate PR up (if I get something sensible)

acao:
we shouldn't ignore all plain .graphql files, but if you need them for this, you can set specific ignore rules for these unit tests. this is the only syntax extension so it would make sense.

sure, jest sounds nice! @B2o5T wants to move to vitest. I just made it so that jest doesn't need a `build` step anymore, except for `cm6-graphql`, so you can run `yarn test` without and build steps being necessary, enjoy!

AaronMoat:
Have put an alternative in https://github.com/graphql/graphiql/pull/3273 😄

AaronMoat:
Closing in favour of https://github.com/graphql/graphiql/pull/3273

PR #3273, opened 2023-06-24
Add test scaffolding for vscode-graphql-syntax (option 2)

AaronMoat:
Resolves: https://github.com/graphql/graphiql/issues/2343 

Alternative to #3267 using much more custom code, but I think it gives us more control. Very keen to see what people think!


acao:
@AaronMoat oh this is very nice, and very easy to read!

In general, writing tests for anything vscode related will involve complicated test tooling, but this seems like it will really do the trick!

acao:
I wonder why this doesn't seem to trigger a codecov report?

AaronMoat:
image

I guess it did? but there's no new "code" tested by this change technically, "just" the package.json (from the eyes of coverage)

acao:
weird, i don't see the comments for some reason !

I guess it makes sense, as istanbul (or whatever jest uses now) wouldn't have a way to instrument JSON for coverage in any way it could recognize.

That's ok though, coverage data is only a rough metric for code quality, I think we can all agree how important this is!

AaronMoat:
@acao no comment - I was clicking through to the github actions 'details' :) 

image

AaronMoat:
So @acao seems you're a fan of this approach over the other one. Is that right - should I close off the other one and prefer this? I might flesh it out with a few more test examples. 

Do you prefer inline snapshots or proper snapshots? 

There's a few options

1. `expect(fileName).toMatchSnapshot()`
2. `expect(fileName).toMatchInlineSnapshot(stuff)`
3. `expect(/* GraphQL */"blah").toMatchSnapshot()`
4. `expect(/* GraphQL */"blah").toMatchInlineSnapshot(stuff)`

Any thoughts on that as well?

acao:
all very good questions! I'm not sure I am ready to decide yet, and I'd be interested in what my colleagues think as well

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+
+async function getGrammar(scopeName: string) {
+  if (grammarCache[scopeName]) {
+    return grammarCache[scopeName];
+  }
+
+  const grammars = (await loadConfiguration()).map(grammar => {
+    return {
+      grammar,
+      content: fs.readFileSync(grammar.path).toString(),
```
B2o5T, line 52:
```suggestion
      content: fs.readFileSync(grammar.path, 'utf8'),
```

B2o5T:
@AaronMoat Nicely job, thanks for putting effort into this. I prefer this PR rather than the first one.

But to be honest it's not exactly what I am searching for. The goal of this PR is to test tokenizing right? We can still keep it, it's a good job. (let me know if I am right in terms I am still new to this project and still learning 😅)

But I have in mind to have `image syntax highlighting` snapshots tests, and not just snapshots tests. I wanna test the right syntax highlighting and throw an error when % of pixels is has a difference.

I was inspired by these image tests in satori [https://github.com/vercel/satori/tree/main/test/__image_snapshots__](https://github.com/vercel/satori/tree/main/test/__image_snapshots__)
and I added tests for our open graph images https://github.com/the-guild-org/docs/blob/main/packages/og-image/src/__image_snapshots__/handler-test-ts-packages-og-image-src-handler-test-ts-handler-should-align-title-and-have-container-padding-1-snap.png and it fails (sometimes) when a new version of satori is out.

B2o5T:
> There's a few options
> 
> expect(fileName).toMatchSnapshot()
> expect(fileName).toMatchInlineSnapshot(stuff)
> expect(/* GraphQL */"blah").toMatchSnapshot()
> expect(/* GraphQL */"blah").toMatchInlineSnapshot(stuff)

I prefer the first one `expect(fileName).toMatchSnapshot()`
- snapshots can be long
- the same filename we can reuse to test mentioned previously `image snapshots` tests

acao:
@B2o5T great idea! I've built some custom image snapshot tooling before, for a geospatial project. that might be the most readable option. the underlying libraries for image diffing used by percy and these jest plugins are quite surprisingly simple and powerful

the problem herein is that we have to find a way to render an image of the syntax highlighting in the first place. monaco for example uses a different grammar than vscode, but I'm sure we can find a library to render vscode grammars somewhere

B2o5T:
> that might be the most readable option.

exactly! I found image snapshots, the most maintainable option

> the problem herein is that we have to find a way to render an image of the syntax highlighting in the first place

you have a point, this is most trickiest part :)

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+  scopes: string[];
+};
+
+export async function tokenizeFile(
+  file: string,
+  grammarScopeName: string,
+): Promise<Token[]> {
+  const grammar = await getGrammar(grammarScopeName);
+  const lines = (await readFile(path.join(__dirname, file), 'utf8'))
+    .toString()
```
B2o5T, line 20:
```suggestion
```
you don't need to stringify, since you already specified the encoding

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+    },
+    createOnigString(s: any) {
+      return new oniguruma.OnigString(s);
+    },
+  };
+}
+
+async function loadConfiguration() {
+  const json = JSON.parse(
+    (await readFile(path.join(ROOT, 'package.json'))).toString(),
```
B2o5T, line 111:
```suggestion
    await readFile(path.join(ROOT, 'package.json'), 'utf8'),
```

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const grammars = (await loadConfiguration()).map(grammar => {
+    return {
+      grammar,
+      content: fs.readFileSync(grammar.path).toString(),
+    };
+  });
+
+  const grammarMap: { [key: string]: tm.IRawGrammar } = {};
+  const injections: { [scopeName: string]: string[] } = {};
```
B2o5T, line 57:
```suggestion
  const grammarMap: { [key: string]: tm.IRawGrammar } = Object.create(null);
  const injections: { [scopeName: string]: string[] } = Object.create(null);
```

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+      });
+    }
+
+    ruleStack = result.ruleStack;
+  }
+
+  return tokens;
+}
+
+const grammarCache: Record<string, tm.IGrammar> = {};
```
B2o5T, line 42:
```suggestion
const grammarCache: Record<string, tm.IGrammar> = Object.create(null);
```

packages/vscode-graphql-syntax/jest.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../../jest.config.base')(__dirname, 'node');
```
B2o5T, line 1: can you run these tests with vitest? you need to make the same as I made in another PR https://github.com/graphql/graphiql/pull/3285

1. modify root `package.json`

`"test:ci": "yarn build && jest --coverage && yarn workspace vscode-graphql-syntax test"`

https://github.com/B2o5T/graphiql/blob/ee52a1be7d9f99abf8884907a122126f8777d261/package.json#L80

2. modify this action
image

3. and add test command
image

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (grammar.injectTo) {
+      for (const injectScope of grammar.injectTo) {
+        (injections[injectScope] ||= []).push(grammar.scopeName);
+      }
+    }
+  }
+
+  const registry = new tm.Registry({
+    onigLib: vscodeOnigurumaLib(),
+    loadGrammar: scope => Promise.resolve(grammarMap[scope]),
```
B2o5T, line 72: does promise here is mandatory? we can do too
```suggestion
    loadGrammar: async scope => grammarMap[scope],
```

packages/vscode-graphql-syntax/tests/utilities.ts
```diff
@@ -0,0 +1,140 @@
+import * as fs from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import * as path from 'node:path';
+import * as oniguruma from 'vscode-oniguruma';
+import * as tm from 'vscode-textmate';
+
+const ROOT = path.join(__dirname, '..');
```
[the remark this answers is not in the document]
acao: if these are executed with jest from the monorepo root, then the former option may be preferrable

packages/vscode-graphql-syntax/tests/utilities.ts
AaronMoat: Yeah, it is

`loadGrammar(scopeName: ScopeName): Promise<IRawGrammar | undefined | null>;`

packages/vscode-graphql-syntax/jest.config.js
AaronMoat: done

AaronMoat:
Fully agree that images would be better. The how is the big question :)

B2o5T (approved):

acao:
I think I'll just build this library and a set of runners (cypress, playwright, etc) for fun 😜

## Shared Steps
1. render dom with codemirror, codemirror textmate mode and load our grammars
2. domToSvG(), then

## test assert 
pixelmatch against each SVG, exit non 0 if pixelalmatch doesn't match, write out failing pixelmatch results with res error highlighting for the diff 

## snap update
save svg output

acao:
Here is another upcoming issue, you may have mentioned already, is that there are multiple issues regarding embedded languages:
https://github.com/microsoft/vscode-textmate/issues/207

acao:
looking to carbon for inspiration on code -> image front, and I [found this](https://github.com/carbon-app/carbon/blob/2cbdcd0cc23d2d2f23736dd3cfbe94134b141191/package.json#L38)! haha!

acao:
here is a nice rust library that seems to use modern stdout features to print code without a browser at all. the problem I anticipate is that this rendering will be inconsistent across various local and CI environments

https://github.com/trishume/syntect#screenshots

acao:
after some research, vscode is one of the few web runtimes that supports `injectionSelector` textmate rules, so it would be very difficult to do visual snapshot testing. @AaronMoat 's approach here is probably the best approach, of using the tokenizer that vscode provides

acao:
just added more tests cases and organized things a bit. lets go with this strategy for now so we have something to make improvements on and to use as a regression suite until we figure out a good solution for using images. i think shiki supports injection rules, which is heavily based on vscode-textmate

PR #3281, opened 2023-06-24
simplify Next.js example + add symlinks to React Vite to maintain 2 examples from 1 place

B2o5T:


examples/monaco-graphql-nextjs/src/pages/index.tsx
```diff
@@ -1,23 +1,18 @@
 import Head from 'next/head';
 import dynamic from 'next/dynamic';
 
-const DynamicEditor = dynamic(() => import('../components/editor'), {
-  suspense: true,
```
B2o5T, line 5: image

acao:
just noting that using symlinks in the past brought us a lot of issues for contributors using windows, but that was before WSL 2. this is why cross-env and such is used everywhere

B2o5T:
@acao I guess it's fine for the moment, in the case of reports, we can always refactor to pre/post script to copy file-1 file-2 inside vitejs from nextjs haha

take a look how I refactored this editor, it's much easy than before 😜

acao:
sure, do whatever you like! i created the sample vite editor in a few hours to demo the monaco graphql improvements to a user. it seemed ideal to keep it to one component. then with next I just copy pasta'd it. I was hoping someone would make improvements, but also let's try to keep it simple as in as close to a single file as possible, because I don't know if splitting the example editor into multiple files makes it easier to demonstrate? a matter of preference I guess

B2o5T:
@acao For me it’s still simple, I tried to make it as simple as possible, maybe let’s move debounce.ts inside editor

acao:
sure, the only goal was to show in 2020 that we could use vite and render with monaco. if a multi-file example seems easier then that's fine, I come from the era of single-file TODO examples so we tend to prefer everything in a single file, haha!

acao (approved):
I know there are more changes upcoming, but I just want to make it clear that I approve of this approach, and my approval is not needed to merge this!

PR #3288, opened 2023-06-26
attempt to get vscode working in an e2e runner

acao:
this is the closest I can get - if you follow the console log statements through the call stack, we are able to find the extension but not activate it. I then went so far as to disable most of the logic in extension#activate() to no avail.

image

going to try some alternative test runners tomorrow, such as [`vscode-extension-tester`](https://github.com/redhat-developer/vscode-extension-tester), however these don't clearly document how to test LSP functionality


acao:
made more progress by switching to insiders build, dist, adding fixture files that will trigger activation events, and it worked!
the extension activated, but now the language-service client is causing this error

image

package.json
```diff
@@ -6,9 +6,7 @@
   "workspaces": {
     "packages": [
       "packages/*",
-      "examples/monaco-graphql-webpack",
-      "examples/monaco-graphql-nextjs",
-      "examples/monaco-graphql-react-vite"
```
B2o5T, line 11: I added it to test removing unused languages and avoid regressions

acao: no worries, this PR will take a long while before it will be stable enough for review even! I expect to do a significant rebase next week

PR #3294, opened 2023-06-26
simplify imports from `graphql`, restrict some with eslint

B2o5T:


acao (approved):
Nice to see that these `graphql` imports have improved!
Speaking of, we should probably update the versions we test in the `graphql` regression test workflow

Also related, we may also need to change `peerDependencies` ranges here, because these imports won't work for older versions

B2o5T:
no type errors while changing `resolutions` of graphql to v15, merging as is, without modifying peerdependency range since no breaking changes where introduced, all imports works for v15 too

PR #3296, opened 2023-06-26
update regression test matrix

acao:
we only need to worry about old 15.x versions and latest 16.x that we support.

### DIFF
--- a/.github/workflows/pr-graphql-compat-check.yml
+++ b/.github/workflows/pr-graphql-compat-check.yml
@@ -12,8 +12,6 @@ on:
       - 'examples'
       - '!examples/monaco-graphql-webpack'
 
-# TODO: test matrix?
-
 permissions:
   contents: read # to fetch code (actions/checkout)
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-graphql-compat-check.yml
```diff
@@ -23,7 +21,8 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        release: ['15.5.3', '^15.8.0', '16.1.0', '16.2.0', '16.3.0']
+        # test all peer ranges
+        release: ['^15.5.0', '^15.8.0', 'latest']
```
B2o5T, line 25: should we really test 15.5.0? or only 15.8.0

acao: @B2o5T I'm testing the lowest of the peer ranges, in `codemirror-graphql` we set that as the lowest

B2o5T (approved):

PR #2609, opened 2022-07-31
refactor: extract create diagnostics

Foo-x:
Extracted the logic that creating diagnostics to a method.

PR #3301, opened 2023-06-28
Ensure diagnostics get regenerated on schema save

yasamoka:
First of all, thank you for all of this great work!

I just installed the `GraphQL.vscode-graphql` extension yesterday and noticed that when I update the GraphQL schema and save it, diagnostics for queries / mutations do not get updated. For example, say I have a mutation using a field that no longer exists in the new schema; diagnostics would not show up for that mutation until I switch to the file containing the mutation and edit it.

I was also thinking that it may be a good idea to separate the cached documents according to project since with the current way cached documents are stored, there seems to be no way to differentiate between documents belonging to different projects.

I am not very familiar with the overall structure of the project or the API as I just learned about this extension yesterday, so any advice regarding potentially better caching behavior, or any existing implementation details, would be appreciated.

Thank you again!

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -633,7 +633,7 @@ export class MessageProcessor {
 
   async handleWatchedFilesChangedNotification(
     params: DidChangeWatchedFilesParams,
-  ): Promise<Array<PublishDiagnosticsParams | undefined> | null> {
+  ): Promise<Array<PublishDiagnosticsParams> | null> {
     if (
       this._isGraphQLConfigMissing ||
       !this._isInitialized ||
@@ -642,7 +642,7 @@ export class MessageProcessor {
       return null;
     }
 
-    return Promise.all(
+    const allDiagnosticsParams = await Promise.all(
       params.changes.map(async (change: FileEvent) => {
         if (
           this._isGraphQLConfigMissing ||
@@ -650,7 +650,7 @@ export class MessageProcessor {
           !this._graphQLCache
         ) {
           this._logger.warn('No cache available for handleWatchedFilesChanged');
-          return;
+          return [];
         }
         if (
           change.type === FileChangeTypeKind.Created ||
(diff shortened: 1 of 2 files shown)


yasamoka:
Just realized that I should run the diagnostics only once after and schema invalidation has happened. I can push this change once the initial change has been confirmed to be useful and practical.

yasamoka:
> I was also thinking that it may be a good idea to separate the cached documents according to project since with the current way cached documents are stored, there seems to be no way to differentiate between documents belonging to 
different projects.

Found the `getProjectForFile()` function so I used that to filter by project.

> Just realized that I should run the diagnostics only once after any schema invalidation has happened. I can push this change once the initial change has been confirmed to be useful and practical.

Fixed that as well.

acao:
awesome stuff! thank you!

yasamoka:
> awesome stuff! thank you!

Thanks for the kind words!

I actually have an incoming pull request that revamps diagnostics entirely. It's already in my fork ("diagnostics" branch) but I still need to fix a test and add a change that is currently local.

acao:
@yasamoka awesome! I was in the midst of doing something similar this week when discovering some of these issues while working on a branch that adds end to end tests for `vscode-graphql`. the more tests you can add the better. i plan on overhauling the unit tests in language server to make it easier to test multi-file configs as well.

acao:
See #2609 for some additional prior art,

As  I did with the new GraphiQL 2 architecture, I've created a draft proposal I will be working on here

https://github.com/graphql/graphiql/discussions/3314

yasamoka:
Thank you for pointing me in the right direction. I am still very new to this extension so I might need a bit of guidance. I'm wrapping up a project where this extension was a big boost to keeping everything sane so maybe in a week or so, I should be done with that. Do you think it's a good idea to keep the pull requests small and incremental? In other words, do I push what I already have regarding revamping diagnostics then we base further work on top of that? Let me know.

acao:
@yasamoka I'm elated to hear it was helpful for you! happy to help guide you. can you and your colleauges leave a review on the vscode marketplace because our reviews have plummeted lately

yes, incremental is better, and I plan to carry out the re-write as such as well. 

if you can add unit test coverage for the code you add here, that would be super helpful! codecov highlights which lines are missing coverage in the file diff

acao:
for some reason, this only seems to work for me if the file tab is closed and re-opened. in another PR I have it working so that the diagnostics reload immediately on schema file change, so that the files highlight immediately. maybe it's just a merge conflict issue but I can't tell

PR #3306, opened 2023-06-29
pnpm + turbo (no cjs + no ts project refs)

acao:
this one is for real, not just for reference!

- [x] drop cjs
- [x] drop typescript project references
- [x] add turbo
- [x] replace yarn with pnpm
- [x] add `build` and `dev` packages to every workspace
- [x] prettier
- [x]  drop `build-bundles`
- [x] vscode compile -> build
- [x] proper `exports` for codemirror-graphql
- [x] eslint
- [ ] cypress failures - something with the minor bundling tweaks is causing `codemirror-graphql` jsonParser exceptions to not be caught as linter errors??
- [ ] webpack dev server - build works but dev server 404s
- [ ] changsets upgrade
- [ ] typedoc

Notes
- we may need to leave project references in place just so that we can resolve projects for the latest typedoc. I don't think there is a way around this for typescript. we don't have to use typescript project references for the build of course
- running `pnpm dev` alongside `cypress open` or `jest`, a common workflow for me when working across packages, seems to crash my PC
- for some reason, the `@graphiql/react` tsconfig.json was configured to compile `__test__` directory. this was not needed for `jest-ts` but seemed to make eslint happy, but i found a way to resolve this

What can hopefully come in  later PRs:
- migrate graphiql webpack bundler to vite
- monaco language bundling fix
- replace tsc usage with bob, tsup, etc
- typescript upgrade
- replace jest with vitest
- import react properly
- etc

it's a more lean version of #3303 with some agreed upon tooling changes


acao:
the biggest issue I've had so far is that if i run pnpm dev with cypress-open while tweaking codemirror-graphql to try to get this e2e test to pass, my system just freezes up. in general pnpm dev just eats up resources

B2o5T:
why open 3 similar PRs and put everything here, migration to pnpm should be done in separated PR otherwise impossible to do a review

examples/monaco-graphql-nextjs/src/constants.ts
```diff
@@ -1,5 +1,5 @@
 import { editor, Uri } from 'monaco-editor';
-import { initializeMode } from 'monaco-graphql/esm/initializeMode';
+import { initializeMode } from 'monaco-graphql/dist/initializeMode';
```
B2o5T, line 2: why this change? everything worked well before, so the whole js ecosystem is witching to ESM so I'll expected to keep esm imports and avoid cjs (`dist`)

package.json
```diff
@@ -29,58 +21,42 @@
     }
   },
   "engines": {
-    "npm": "please_use_yarn_instead"
+    "npm": "please_use_pnpm_instead"
```
B2o5T, line 24: this should be removed, and `pnpm` engine should be set

packages/cm6-graphql/package.json
```diff
@@ -15,7 +15,7 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "dependencies": {
-    "graphql-language-service": "^5.1.7"
+    "graphql-language-service": "workspace:"
```
B2o5T, line 18: where did you find this syntax?

packages/graphiql-react/src/editor/query-editor.ts
```diff
@@ -143,11 +143,11 @@ export function useQueryEditor(
...
-      import('codemirror-graphql/esm/lint'),
-      import('codemirror-graphql/esm/info'),
-      import('codemirror-graphql/esm/jump'),
-      import('codemirror-graphql/esm/mode'),
+      import('codemirror-graphql/dist/hint'),
+      import('codemirror-graphql/dist/lint'),
+      import('codemirror-graphql/dist/info'),
+      import('codemirror-graphql/dist/jump'),
+      import('codemirror-graphql/dist/mode'),
     ]).then(CodeMirror => {
```
B2o5T, line 151: this is not fixing the problem but AVOID problem

.github/workflows/license-check.yml
```diff
@@ -2,7 +2,7 @@ name: License Check
 on:
   pull_request:
     paths:
-      - 'yarn.lock'
+      - 'pnpm.lock'
```
B2o5T, line 5: I already wrote a big review about similar mistakes here https://github.com/graphql/graphiql/pull/3054 so I will not repeat

packages/cm6-graphql/package.json
acao: not sure what you're speaking of, it's `workspace:` here?
image

.github/workflows/license-check.yml
acao: my local is not the same as the commit you are reviewing, there must be a mistake

package.json
acao: https://nodejs.org/api/packages.html#packagemanager

acao:
@B2o5T please note that when a PR is in draft state, it is not ready to review. I appreciate the early review, but things are subject to change drastically

acao:
yay! cypress now runs using something closer to the reccomended config. we had to move cypress back into `devDependencies` as per the[ cypress reccomendation](https://github.com/bahmutov/cypress-gh-action-monorepo/blob/master/package.json)

acao:
@B2o5T I will include instructions soon on how to try this out and start giving very specifically scoped feedback and/or contribution if you are interested. It has somehow introduced a few detectable regressions, see notes. The vscode extension tests will take precedence as they are of a greater concern

Noodleontour (approved):
Cool

dimaMachina:
closed by https://github.com/graphql/graphiql/pull/3975

anowardear062-svg (approved):

PR #3310, opened 2023-07-01
replace `express-graphql` with `graphql-http`

acao:
`express-graphql` has been deprecated. `graphql-http` is now the transport wg reference implementation. because we are a reference implementation of a reference implementation, we should be using this library instead, or whatever the wg is using.

### DIFF
--- a/functions/schema-demo.js
+++ b/functions/schema-demo.js
@@ -1,7 +1,7 @@
 /* example using https://github.com/awslabs/aws-serverless-express */
 const express = require('express');
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { createHandler } from 'graphql-http/lib/use/express';
+const { createHandler } = require('graphql-http/lib/use/express');
 const awsServerlessExpress = require('aws-serverless-express');
 const schema = require('../packages/graphiql/test/schema');
 const cors = require('cors');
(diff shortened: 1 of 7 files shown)


acao:
I think it's also worth considering whether we should even be using express.js anymore, as nowadays we can use native node http createServer() instead

acao:
@enisdenjo @benjie is there any work happening on an graphql-over-http wg reference client as well? currently we are maintaining one ourselves for graphiql (`@graphiql/toolkit` `createGraphiQLFetcher()`, and it would be nice to use a reference implementation that the spec wg is maintaining instead

enisdenjo:
@acao [graphql-http](https://github.com/graphql/graphql-http) is both a client and a server. Check out the ["use the client" section in the readme](https://github.com/graphql/graphql-http#use-the-client).

acao:
@enisdenjo oof, I didn't look closely enough, awesome news, thank you!

functions/schema-demo.js
```diff
@@ -1,7 +1,7 @@
 /* example using https://github.com/awslabs/aws-serverless-express */
 const express = require('express');
 // eslint-disable-next-line import/no-extraneous-dependencies
-const { graphqlHTTP } = require('express-graphql');
+import { createHandler } from 'graphql-http/lib/use/express';
```
B2o5T, line 4: This should use `require`, or every import should use es `import`

B2o5T (approved):

PR #3313, opened 2023-07-01
[tooling] use cspell `--no-must-find-files` on git hooks, simplify output

acao:
- fixes a false positive non 0 exit error on husky git commit hooks when no files are found with [--no-must-find-files](https://github.com/streetsidesoftware/cspell/issues/2594#issuecomment-1072751914)
- `--no-progress` makes output easier to read locally and in CI

image

Notes:
- it would be nice to add these to the config file, but reading the code for `cspell`, it appears these flags are used by the CLI only, thus why they aren't configurable in the config file
- in another PR, we should consider replacing husky with [lefthook](https://www.npmjs.com/package/lefthook), or upgrading husky. this version is not compatible with `pnpm`, and the compatible upgrade will require config changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,11 @@
       "eslint --fix",
       "prettier --write",
       "jest --passWithNoTests",
-      "yarn cspell lint --gitignore --no-progress --no-must-find-files"
+      "yarn cspell lint . --gitignore --no-progress --no-must-find-files"
     ],
     "*.{md,html,json,css}": [
       "prettier --write",
-      "yarn cspell lint --gitignore --no-progress --no-must-find-files"
+      "yarn cspell lint . --gitignore --no-progress --no-must-find-files"
     ]
   },
   "husky": {
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "jest": "jest --testPathIgnorePatterns cm6-graphql",
     "license-check": "jsgl --local ./",
     "lint": "yarn eslint && yarn pretty-check && yarn lint-cspell",
-    "lint-cspell": "cspell --unique --gitignore --no-progress --no-must-find-files",
+    "lint-cspell": "cspell --unique --no-progress --no-must-find-files",
     "lint-fix": "yarn eslint --fix",
     "postbuild": "yarn workspace codemirror-graphql run postbuild",
     "prebuild-bundles": "yarn build-bundles-clean",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "license-check": "jsgl --local ./",
     "lint": "yarn eslint && yarn lint-check && yarn pretty-check && yarn lint-cspell",
     "lint-check": "eslint-config-prettier .eslintrc.js",
-    "lint-cspell": "cspell --unique",
+    "lint-cspell": "cspell --unique --gitignore --no-progress --no-must-find-files",
```
B2o5T, line 63: Gitignore already specified in config file https://github.com/graphql/graphiql/blob/main/cspell.json

package.json
```diff
@@ -16,11 +16,11 @@
       "eslint --fix",
       "prettier --write",
       "jest --passWithNoTests",
-      "yarn cspell --gitignore"
+      "yarn cspell lint --gitignore --no-progress --no-must-find-files"
```
B2o5T, line 19: image

```suggestion
      "yarn lint-cspell"
```

package.json
```diff
@@ -16,11 +16,11 @@
...
     ],
     "*.{md,html,json,css}": [
       "prettier --write",
-      "yarn cspell --gitignore"
+      "yarn cspell lint --gitignore --no-progress --no-must-find-files"
```
B2o5T, line 23:
```suggestion
      "yarn lint-cspell"
```

package.json
```diff
@@ -16,11 +16,11 @@
       "eslint --fix",
       "prettier --write",
       "jest --passWithNoTests",
-      "yarn cspell --gitignore"
+      "cspell lint . --gitignore --no-progress --no-must-find-files"
```
B2o5T, line 19: image

you don't need 
- `lint`
- `gitignore` already specified in cspell config
- `.` already specified in https://github.com/graphql/graphiql/blob/2f15e202b194387a1bd78fc778268fb9b8c0fa12/cspell.json#L17

why not just call `yarn lint-cspell`?

acao: oops, i meant to remove the `.`.  but yes you are correct, the the commands are the same. are you aware that husky passes only the files that changed to the command as stdin?

acao:
here is an example of the bug on `main` branch
image

acao:
merging because this has been very annoying and impacting contributors

PR #3316, opened 2023-07-02
replace `createGraphiQLFetcher` with `graphql-http`

### LINKED ISSUE #3312: [graphiql] `@graphiql/toolkit` `createGraphiQLFetcher` should use `graphql-http`
`@graphiql/toolkit` `createGraphiQLFetcher()` should use `graphql-http` client!

https://github.com/graphql/graphql-http#use-the-client

this would offload the concern of staying up with the exciting `graphql-over-http` spec to the working group who is happily doing this for us, instead of us maintaining a spec implementation in duplicate.

acao:
resolves #3312

creating a draft PR because I couldn't even get the root `graphiql-start` or `yarn dev` command in the graphiql workspace working 👀 

I doubt this will work but it was neat to see how easily I could get the typescript inference to approve

We will likely need to simplify types as well, and encapsulate further for all the related behavior in `@graphiql/react` to continue working, and probably will need to normalize our legacy props to the `graphql-http` props to some degree

Todo:
- [x] works for HTTP get/post ops? at least it makes the request properly
- [ ] works for websockets and incremental delivery
- [ ]  add simple unit tests for the encapsulation
- [ ] simplify types
- [ ] simplify dependencies
- [ ] simplify docs


acao:
as per the advice of the `graphql-over-http` group, we will keep our implementation until IncrementalDelivery is part of the official spec and available in `graphql-http`

https://discord.com/channels/625400653321076807/863141924126588958/1125454270691545168

PR #3319, opened 2023-07-03
fix(graphiql-plugin-explorer): Use named `Explorer` import from `graphiql-explorer`

LekoArts:
Hi!

While working on https://github.com/gatsbyjs/gatsby/pull/38279 I saw failures in our E2E testing suite.

We use the GraphiQL explorer plugin and when opening that tab the page just got blank and the console threw an error:

image

It errors at this call:

https://github.com/graphql/graphiql/blob/2f15e202b194387a1bd78fc778268fb9b8c0fa12/packages/graphiql-plugin-explorer/src/index.tsx#L130

Because the import is an object with `Explorer` and `default`.

If I compare https://unpkg.com/browse/@graphiql/plugin-explorer@0.1.22/dist/index.mjs with https://unpkg.com/browse/@graphiql/plugin-explorer@0.1.20/dist/graphiql-plugin-explorer.es.js you can see that older versions had helper functions to pick `.default` if available.

### DIFF
--- a/.changeset/shaggy-eyes-melt.md
+++ b/.changeset/shaggy-eyes-melt.md
@@ -1,0 +1,5 @@
+---
+"@graphiql/plugin-explorer": patch
+---
+
+fix(graphiql-plugin-explorer): Use `preferDefault` helper to correctly import `graphiql-explorer`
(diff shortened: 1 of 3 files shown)


LekoArts:
@B2o5T PR should be good to go, thanks for the review!

B2o5T (approved):

alokeutpal (approved):

PR #3324, opened 2023-07-05
[tooling] retire CJS for libraries

acao:
It's 2023 and we don't need commonjs anymore.
`graphql-js` will drop cjs for 16

This PR:

- removes all `main` entrypoints
- removes any unneeded commonjs related config and tools for libraries
- leaves the build directory as `esm` for now, for legacy import support? what do we think of this choice? or should we standardize on `dist`? I only change this for `codemirror-graphql`, but this keeps `monaco-graphql/esm` imports intact. if only `imports` key in `package.json` accepted `*` 😢 
- simplifies the project references build
- if you're using `graphql-lsp`, upgrade
- transparently builds all packages with a `build` script in order of dependency

CJS still exists for:

- vscode projects because vscode requires commonjs targer -`@graphiql/react` and `graphiql`, where legacy consumers may need them
- TODO i think we will need it for cm6-graphql still for tests

After this:

- turbopack
- drop `build-bundles`?
- targeted `dev` scripts


B2o5T:
> or should we standardize on dist?
I prefer `dist`, also you can add `type: module` in package.json to avoid mjs extensions

acao:
@B2o5T it seems to work fine without mjs extensions

acao:
@B2o5T I agree with your point that `dist` makes more sense... and while we are making breaking changes and bumping versions... we might as well document this in the changelog and examples

B2o5T:
@acao it’s per this recommendation https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm

and about `monaco-graphql` - it's already a major release so we could introduce breaking changes, also I would keep only the following exports:
- `monaco-graphql` with initialize mode, not ugly `monaco-graphql/esm/initializeMode`
- some exports for workers

so it simplifies things without `monaco-graphql/{dist,esm}/something`

Also, I would setup this rule https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md for consistency `kebab-case` filenames that we use in supported The Guild projects

and please hold this PR to merge until I approve it

acao:
already one step ahead 😆 
image

acao:
I will get vitest working again, and make a few more tweaks, this is almost ready for a final review!

acao:
adding `"type": "module"` will take some more time, and will require converting a bunch of scripts and configs from require to import, I will have to come to that later. this will not be merged without much more review don't worry, haha

acao:
We will also need to figure out why this effort regressed before this can have the final review

image

B2o5T:
> and will require converting a bunch of scripts and configs from require to import

you can temporarily renames their extension to `cjs` to avoid such as converting

acao:
@B2o5T indeed, but the jest side of things will be tricky, I will need to spend a few more days on this to finish the `esm` conversion

B2o5T:
Also tests fail with the following error

AssertionError: expected '✓ 969 modules transformed.' to include '✓ 1092 modules transformed.

nice 👍 we have fewer converted modules 😂

acao:
@B2o5T haha nice! thus why i suggest using the "less than" approach suggested in the comment above. it seems github comments are slow for us today 😆 

I think this may be caused by a regression where `monaco-graphql/dist/typings/monaco-editor.d.ts` is not being picked up. the branch in it's current state fails the example builds

Speaking of, I'm considering whether we should make the example monaco build and one of the graphiql examples part of the netlify build again. This allowed us to validate consuming types previously, and then we had a `/monaco` or `/webpack` path to demonstrate the example build output on netlify. it's slower but worth it

B2o5T:
> maybe ensure it has less than n modules?

I guess it is good to keep the exact number of dependencies, I think we'll adjust this number very rarely when `vite` version changes or such as this migration

B2o5T:
> I think this may be caused by a regression where

I can take a look later this week, currently on vacation

acao:
@B2o5T no worries, I will figure it out! enjoy your holiday

acao:
> I guess it is good to keep the exact number of dependencies, I think we'll adjust this number very rarely when `vite` version changes or such as this migration

This is what I'm worried about. This can be caused by either changes to `vite` or `monaco-editor`, and as a rule of thumb, I prefer to not write tests that test internal details of third party dependencies. I still think `less than n` will be more ideal, otherwise this might frustrate contributors as just small patch resolution changes could break this

acao:
@B2o5T ah i found the issue, had overlooked your script that patches types in renaming from `esm` to `dist` paths

acao:
and now yarn 1 hosting is breaking the vite monaco plugin 😢  pnp mode can't come soon enough haha

packages/graphiql-react/package.json
```diff
@@ -11,19 +11,24 @@
...
     ".": {
-      "import": "./dist/index.mjs",
-      "require": "./dist/index.js",
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs",
       "types": "./types/index.d.ts"
     },
     "./font/roboto.css": "./font/roboto.css",
     "./font/fira-code.css": "./font/fira-code.css",
-    "./dist/style.css": "./dist/style.css"
```
B2o5T, line 25: I guess it worked well before?

acao: it did,  but this didnt work with webpack-cli serve

acao:
@B2o5T i think this is good to merge now, and then your ` type": "module"` effort can be a follow up PR, just to make the effort more manageable

packages/graphiql-plugin-code-exporter/package.json
```diff
@@ -7,7 +7,6 @@
     "directory": "packages/graphiql-plugin-code-exporter"
   },
   "author": "LekoArts",
-  "main": "dist/index.js",
```
B2o5T, line 10: This is needed, we still export cjs

packages/graphiql-plugin-explorer/package.json
```diff
@@ -6,7 +6,6 @@
     "url": "https://github.com/graphql/graphiql",
     "directory": "packages/graphiql-plugin-explorer"
   },
-  "main": "dist/index.js",
```
B2o5T, line 9: This is needed, we still export cjs

examples/monaco-graphql-react-vite/vite.config.ts
```diff
@@ -1,6 +1,6 @@
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
-import monacoEditorPlugin from 'vite-plugin-monaco-editor';
+import { default as monacoEditorPlugin } from 'vite-plugin-monaco-editor';
```
B2o5T, line 3: This is not needed

packages/graphiql-react/package.json
```diff
@@ -13,18 +13,22 @@
   "license": "MIT",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
+  "types": "dist/types/index.d.ts",
```
B2o5T, line 16:
```suggestion
  "types": "dist/index.d.ts",
```

I wanna types lives just in `dist`

packages/graphiql/package.json
```diff
@@ -16,8 +16,7 @@
     "url": "https://github.com/graphql/graphiql/issues?q=issue+label:graphiql"
   },
   "license": "MIT",
-  "main": "dist/index.js",
```
B2o5T, line 19: Graphiql still export cjs?

packages/graphiql-plugin-code-exporter/package.json
acao: see the PR description

packages/graphiql-react/package.json
acao: I couldn't get this working with the vite plugin for whatever reason. i think keeping them at `/types` as before is fine for now

acao:
just to clarify @B2o5T, the purpose of this PR is not to move the whole monorepo to `type: module`. it is *just to remove CJS exports for libraries,  but not for consuming packages*. 

the effort of moving to `"type": "module"` is much greater and not required for this, as we can see with vite and webpack (including next js) happily consuming the esm only exports

the effort to move the repo to _use_ esm only will require migration away from jest which does not support `type: module`, and changes to many files and tooling. these tooling changes are _not required_ to only publish esm, despite how nonsensical it seems to use cjs to publish esm only libraries, this is sufficient for our users. the rest of the tooling migration will happen, but it's not the greatest priority right now.

I'm trying to keep these PRs small, towards simplifying the repo tooling so we can make other improvements. As I mention in the last comment, the effort to move monaco-graphql to `"type": "modules"` can happen seperately, as it is not required anymore by bundlers, unlike when sindre published that gist

acao:
I've learned from shipping other libraries to webpack consumers that commonjs exports are still required, so this is probably a bad idea. type: module is one thing, but dropping main and commonjs exports is another.

PR #3327, opened 2023-07-07
make `monaco-graphql` true esm with `"type": "module"` and `"exports"` field, simplify his exports

B2o5T:


examples/monaco-graphql-webpack/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "1.1.1",
   "private": true,
   "license": "MIT",
-  "description": "A simple monaco example with webpack and typescript",
+  "description": "A simple monaco example with webpack and TypeScript",
   "scripts": {
-    "build": "cross-env NODE_ENV=production webpack-cli",
-    "start": "cross-env NODE_ENV=development webpack-cli serve"
+    "bundle": "cross-env NODE_ENV=production webpack-cli",
```
B2o5T, line 8: renamed to `bundle` because `build` from root fails since I added this example to `workspace` and `postbuild` are executed too late for this example

packages/monaco-graphql/README.md
```diff
@@ -447,11 +447,11 @@ import type { worker as WorkerNamespace } from 'monaco-editor';
 // @ts-expect-error - ignore missing types
 import * as worker from 'monaco-editor/esm/vs/editor/editor.worker';
 
-import { GraphQLWorker } from 'monaco-graphql/dist/GraphQLWorker';
+import { GraphQLWorker } from 'monaco-graphql/GraphQLWorker';
 
 import { myValidationRules } from './custom';
 
-self.onmessage = () => {
+globalThis.onmessage = () => {
```
B2o5T, line 454: monaco-editor uses `globalThis` in latest version instead of `self`

packages/monaco-graphql/package.json
```diff
@@ -45,7 +65,7 @@
   },
   "peerDependencies": {
     "graphql": "^15.5.0 || ^16.0.0",
-    "monaco-editor": ">= 0.20.0 < 1",
+    "monaco-editor": ">= 0.39.0 < 1",
```
B2o5T, line 68: update as per Rikki suggestion for next major

B2o5T:
@acao something was broken with your latest updates, `build` succeeded and all examples worked well until I merged with your PR

acao:
@B2o5T yes, there is a race condition where the graphiql build begins before the `@graphiql/react` build is complete. previously, this worked because the tsc --noEmit was fast enough, but by adding the dts plugin to vite and dropping that, it has to wait for the full vite build which errors

acao:
@B2o5T try rebasing again, it should work better now

.changeset/chilled-trains-laugh.md
```diff
@@ -9,7 +9,16 @@
 '@graphiql/toolkit': minor
 ---
 
-_BREAKING CHANGE:_ drop commonjs exports in all libraries except for `graphiql` and `@graphiql/react`
+_BREAKING CHANGE:_ drop commonjs exports in all libraries except for `graphiql`
+and `@graphiql/react`
```
acao, line 13: why is this on a new line? it's harder to read

packages/monaco-graphql/package.json
acao: is there a reason we need `0.39.0` specifically? is there a more forgiving range we can support? we should document what we use from the `0.39.0` release that deprecates support for `0.38.0` etc

.changeset/chilled-trains-laugh.md
B2o5T: Some locale prettier changes, will revert 👍

packages/monaco-graphql/package.json
B2o5T: No reason, which version you propose to bump?

B2o5T: @acao I was rethinking this, since in `monaco-editor` every `minor` version is `major`, according to semver policy (since major is `0`) let's bump it as highest as possible (`0.40.0` in this case)

PR #3333, opened 2023-07-11
`.prettierignore` => `.eslintignore`, mutual ignore rules, cleanup prettier config

acao:
this fixes a bug that only appears when changesets issues a release, and finally lints the changelogs. we decided last year to not lint or prettier fix changelogs or changesets

Fixes this bug on pre-commit:
https://github.com/graphql/graphiql/actions/runs/5519491714/jobs/10064813611#step:5:153

Follow Up:
the docs say that `metadata` cache strategy is faster, though the default when passing `--cache` is `content`
https://prettier.io/docs/en/cli.html#--cache-strategy


package.json
```diff
@@ -55,7 +55,7 @@
     "dev-graphiql": "yarn workspace graphiql dev",
     "e2e": "yarn run e2e:build && yarn workspace graphiql e2e",
     "e2e:build": "yarn build && yarn workspace graphiql build-bundles",
-    "eslint": "NODE_OPTIONS=--max-old-space-size=4096 eslint --max-warnings=0 --ignore-path .gitignore --cache .",
+    "eslint": "NODE_OPTIONS=--max-old-space-size=4096 eslint --max-warnings=0 --ignore-path .gitignore --ignore-path .prettierignore --cache .",
```
B2o5T, line 58: Eslint doesn't accept multiple ignore-path as prettier v3 do

acao: ok! then it makes sense to switch to rename the file to `.eslintignore`, and to add it to the prettier `--ignore-path`

acao: @B2o5T do you approve of this strategy? anything that can go wrong? trying to keep it simple

PR #3344, opened 2023-07-13
patch changesets

acao:
from @B2o5T 's patch!


package.json
```diff
@@ -135,6 +136,8 @@
     "js-green-licenses": "3.0.0",
     "lint-staged": "^10.1.2",
     "mkdirp": "^1.0.4",
+    "patch-package": "^7.0.2",
+    "postinstall-postinstall": "^2.1.0",
```
B2o5T, line 140: I guess you don't need it?

acao: why? only yarn 3 has it's own `patch` command

acao: @B2o5T https://www.npmjs.com/package/patch-package#why-use-postinstall-postinstall-with-yarn

B2o5T: never used this dep even with yarn v1, you can leave it if it's required

acao: @B2o5T good point, i'm just nerdily following the docs to a T, this is only needed for `yarn remove` which doesn't issue `postinstall`, but then that's not used in CI where this is needed. if a user runs `yarn remove` locally and then `yarn changeset add`, the patch would be regressed in their local, but that command won't trigger this library even, I don't think

PR #3376, opened 2023-07-29
feat: Add support for custom scalar validation

bboure:
This PR fixes the following issues:

- In GraphQL, optional input variables and types are also nullable.
- in the JSON Schema, `enum` should not be combined with `type` (removed)
- custom scalars should only accept basic primitives (i.e. no objects, arrays, etc)

### DIFF
--- a/packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
+++ b/packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
@@ -88,6 +88,34 @@ function renderType(into: string[], t: GraphQLInputType | GraphQLInputField) {
   }
 }
 
+function renderDefinitionDescription(
+  t: GraphQLInputType | GraphQLInputField,
+  useMarkdown?: boolean,
+  description?: string | undefined | null,
+) {
+  const into: string[] = [];
+
+  text(into, renderTypeToString(t, useMarkdown));
+
+  if (description) {
+    text(into, '\n');
+    text(into, description);
+  } else if (!isScalarType(t) && 'description' in t && t.description) {
+    text(into, '\n');
+    text(into, t.description);
+  } else if (
+    isNonNullType(t) &&
+    !isScalarType(t.ofType) &&
+    'description' in t.ofType &&
+    t.ofType.description
+  ) {
+    text(into, '\n');
+    text(into, t.ofType.description);
+  }
+
+  return into.join('');
+}
+
 function renderTypeToString(
   t: GraphQLInputType | GraphQLInputField,
   useMarkdown?: boolean,
@@ -242,20 +270,11 @@ function getJSONSchemaFromGraphQLType(
         properties: {},
         required: [],
       };
-      if (type.description) {
-        fieldDef.description =
-          type.description + '\n' + renderTypeToString(type);
-        if (options?.useMarkdownDescription) {
-          // @ts-expect-error
-          fieldDef.markdownDescription =
-            type.description + '\n' + renderTypeToString(type, true);
-        }
-      } else {
-        fieldDef.description = renderTypeToString(type);
-        if (options?.useMarkdownDescription) {
-          // @ts-expect-error
-          fieldDef.markdownDescription = renderTypeToString(type, true);
-        }
+
+      fieldDef.description = renderDefinitionDescription(type);
+      if (options?.useMarkdownDescription) {
+        // @ts-expect-error
+        fieldDef.markdownDescription = renderDefinitionDescription(type, true);
       }
 
       for (const fieldName of Object.keys(fields)) {
@@ -276,18 +295,18 @@ function getJSONSchemaFromGraphQLType(
           ...fieldDefinition,
         } as JSONSchema6;
 
-        const renderedField = renderTypeToString(field.type);
+        // prepend field description to type description
         fieldDef.properties[fieldName].description = field.description
-          ? field.description + '\n' + renderedField
-          : renderedField;
+          ? field.description + '\n\n' + typeDefinition.description
+          : typeDefinition.description;
+
         if (options?.useMarkdownDescription) {
-          const renderedFieldMarkdown = renderTypeToString(field.type, true);
-          fieldDef.properties[
-            fieldName
-            // @ts-expect-error
-          ].markdownDescription = field.description
-            ? field.description + '\n' + renderedFieldMarkdown
-            : renderedFieldMarkdown;
+          // @ts-expect-error
+          fieldDef.properties[fieldName].markdownDescription = field.description
+            ? // @ts-expect-error
+              field.description + '\n\n' + typeDefinition.markdownDescription
+            : // @ts-expect-error
+              typeDefinition.markdownDescription;
         }
 
         if (fieldRequired) {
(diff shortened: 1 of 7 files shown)


packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
```diff
@@ -139,22 +140,48 @@ function getJSONSchemaFromGraphQLType(
...
+    // default scalars
+    if (scalarTypesMap[type.name]) {
+      if (isNonNull) {
+        definition.type = scalarTypesMap[type.name];
+      } else {
+        definition.type = [scalarTypesMap[type.name], 'null'];
+      }
+    } else {
+      // custom scalar can be of any type
+      // FIXME: allow custom scalars to be configured somewhere?
```
acao, line 161: these are determined from the custom scalar spec https://graphql.org/blog/2023-01-14-graphql-scalars/

so the mapping should be based on the provided scalars in the official repo

acao:
awesome work! can you add a changeset and another test case for the null case that isn't covered? the coverage info you can see on the files tab

packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
bboure: Good point, we probably could implement validation for "official" specs.

But one might want custom or non-spec'ed scalars, so I'd still like a way to inject custom validation.
(example, [AWS AppSync scalars.](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html)). 

Also, one might want to override the community specs (e.g. custom `DateTime` that does not follow the specs for some reason).

I'd be happy to work on that, but it's kinda out of the scope of this PR.

acao: indeed, i agree it's out of scope, just adding a meta-comment for your comment haha

acao:
@bboure why don't you go ahead and add the custom mappings in this PR then, I agree it makes sense. 

Others have expressed this need, and even introspecting the scalar info doesn't give us a deterministic solution. perhaps we can even provide a default mapping that can be overridden by users? Or perhaps we should provide none to avoid confusion.

Bonus if you can add the new scalar mapping to monaco-graphql as well, but if not I can. The main thing to remember is that the inter process boundary between main and worker process means the mapping will need to be json serializeable, though you don't need to perform the serialization.

it just means we will need a flat object mapping rather than a callback function or some such

Also, be sure to update the changelog and readme to reflect the changed behaviors with scalars

acao:
by flat mapping, it could look like this:

```js
{ 
  DateTime: { type: "string", format: "datetime" }
}
```

come to think of it, i made a mapping just like this for a proprietary project where we decoratively generated openapi3 specs from graphql schema, I'm embarassed I didn't provide it sooner!

bboure:
@acao  ok, I'll give it a shot.

I'll give a shot at adding it to `monaco-graphql` since this is something that I'll need, but I might need some guidance.

I was thinking in that direction too.

I think one good option would be to add an option somewhere in `DiagnosticSettings`. We'll probably need a config per model uri.

Something like this

```ts
monacoGraphQLAPI.setDiagnosticSettings({
  validateVariablesJSON: {...},
  customScalarSchemaMappings: {
    [Uri.file('operation.graphql').toString()]: {
      DateTime: { type: "string", format: "date-time" },
      // maybe a short-hand for type could be useful but not strictly necessary as it might bring confusion too.
      MyCustomScarlar: 'string',
    }
  }
});
```

acao:
@bboure awesome!  I think this could work fine, but I wonder if configuring it per schema makes even more sense?

```ts
initializeMode({
  schemas: [
    {
      uri: 'https://my-schema.com',
      fileMatch: ['**/*.graphql'],
      // note: not sure if ^ works if the graphql model is using urls for uris?
      customScalarMappings: {
          DateTime: { type: "string", format: "date-time" },
          // maybe a short-hand for type could be useful but not strictly necessary as it might bring confusion too.
          MyCustomScarlar: 'string',
      }
    },
  ],
});
```

acao:
also, in case you're wondering why this lives on `graphql-language-service`, I have plans to use this logic for variables validation/completion in the vscode extension as well :)

bboure:
@acao I added custom scalar schemas support and `monaco-graphql` config as suggested.

I also took the opportunity to improve the schema descriptions a bit.
Descriptions will now look as follow:

```
[field description] (for object types and if field description is present)
(two new lines)
Type (scalar or custom)
[type description] (If present, except for default scalars)
```

Note about type descriptions: I kept backward compatibility and do not render the description of default scalars (String, Int, etc). (They were probably removed for a reason, probably because they are obvious?)

For custom scalars, we use the description from the custom schema, if present.

Examples
```
The birth date of the user.

Date!
An ISO_8601 date
```

```
The name of the user

String!
```

lmk if you have comments

acao:
@bboure nice, and input objects retain their formatting as well?

bboure:
the above applies to all types. Objects might look like this


```
A field description

MyObjectType!
My object description
```

acao:
hm,,. testing this out in vscode and monaco-graphql, it seems that something breaks and the hover no longer renders, I will need to see if something has regressed elsewhere

acao:
another issue with monaco is the schema url, the warning blocks validation messages

can you change the meta schema to draft 4? this is what package.json and many standards use, and is mostly all we need for the simple schemas we generate from variable definitions
`'http://json-schema.org/draft-04/schema'`

acao:
I have some other changes I was planning to add today, for example to add optional anchor links for the graphiql@4 release with monaco coming soon, but I will wait until this PR is complete

bboure:
> update: the issue is with regressions in the example projects from what I can tell

Where is the issue exactly? 

did I break something?

bboure:
I tried this in one of my projects. it seems to work fine

image

acao:
> did I break something?

no, there is a regression in the demos, where someone changed the mode instantiation pattern, so when the schema is fetched, the validation doesn't happen automatically for queries and variables, so you have to modify the query first before variables are validated again

bboure:
> no, there is a regression in the demos, where someone changed the mode instantiation pattern, so when the schema is fetched, the validation doesn't happen automatically for queries and variables, so you have to modify the query first before variables are validated again

Ah! I noticed this too and drove me crazy. It was next on my list to have a look at and fix. 
You'll probably know better than me how to fix, but if you point me in the right direction, I can also have a look

acao:
@bboure I have some more fixes on the way, as you can see in #3384

packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
```diff
@@ -99,14 +141,14 @@ function renderTypeToString(
...
-  Boolean: 'boolean',
+const defaultScalarTypesMap: { [key: string]: JSONSchema6 } = {
+  Int: { type: 'integer' },
+  String: { type: 'string' },
+  Float: { type: 'number' },
+  ID: { type: 'string' },
+  Boolean: { type: 'boolean' },
   // { "type": "string", "format": "date" } is not compatible with proposed DateTime GraphQL-Scalars.com spec
-  DateTime: 'string',
+  DateTime: { type: 'string' },
```
acao, line 151: is there a reason we don't specify `format: 'date-time'` here?
I think these are compatible:
https://scalars.graphql.org/andimarek/date-time
https://json-schema.org/draft/2020-12/json-schema-validation.html#section-7.3.1

bboure:
@acao  I can rebase as soon as you merge the other.

packages/graphql-language-service/src/utils/getVariablesJSONSchema.ts
bboure: After double checking, there is no spec for `Date`.
Also, JSONSchema `date-time` is not fully compatible with the specs
e.g. offset is required

I might be able to come with a regex that follows the specs.

acao:
ok, in that case, I will just open my PR against this one tonight, because that PR will take a while to get reviews before we can merge

acao:
@bboure I think this is fine to merge now actually, it won't be released until we merge the Release Packages PR. Otherwise it would be hard for me to make a PR against this branch

acao:
@bboure ok! this looks good to merge, thanks for all of this!

PR #3384, opened 2023-08-02
fix: `monaco-graphql` exports, validate on load

acao:
- validates both graphql and json variables on load
- reduces bundle size further, allowing users to specify which features they want, instead of importing all of them
- adds a typescript mode to the vite/next.js demos to ensure our implementation doesn't break `typescript` language mode & worker
- fix variables json validation by pointing to a meta schema
- add `exports` aliases that are backwards compatible, but allow `monaco-graphql/initializeMode` and `monaco-graphql/graphql.worker` to be imported directly
- add a `monaco-graphql/lite` import for minimal features
- import the json language mode only if variables json validation is used


packages/monaco-graphql/src/monaco-editor.ts
```diff
@@ -19,4 +19,14 @@
 import 'monaco-editor/esm/vs/basic-languages/graphql/graphql.contribution.js';
 import 'monaco-editor/esm/vs/language/json/monaco.contribution.js';
 
-export * from 'monaco-editor/esm/vs/editor/edcore.main.js';
```
acao, line 22: `edcore.main.js` imports `editor.all.js`, which includes many features we don't even use yet!

instead, we implement only the minimum features we need for our mode to work, and users can opt in to the rest

packages/monaco-graphql/test/monaco-editor.test.ts
```diff
@@ -13,7 +13,7 @@ describe('monaco-editor', () => {
     // expect(lines[0]).toBe('$ vite build');
     // expect(lines[1]).toMatch(' building for production...');
     // expect(lines[2]).toBe('transforming...');
-    expect(lines[3]).toMatch('✓ 970 modules transformed.');
+    expect(lines[3]).toMatch('✓ 839 modules transformed.');
```
acao, line 16: reduces the total modules by replacing `edcore.main.js`, even though we add the typescript mode to the  vite bundle!

packages/monaco-graphql/README.md
```diff
@@ -235,15 +236,41 @@ MonacoGraphQL.setCompletionSettings({
 });
 ```
 
-You can also experiment with the built-in I think `jsonc`? (MSFT json5-ish
-syntax, for `tsconfig.json` etc.) and the 3rd party `monaco-yaml` language modes
+You can also experiment with the built-in `jsonc`? (JSON
+syntax that allows comments and trailing commas, for `tsconfig.json` etc.) and the 3rd party `monaco-yaml` language modes
```
TallTed, line 240:
```suggestion
syntax that allows comments and trailing commas, for `tsconfig.json`, etc.) 
and the 3rd party `monaco-yaml` language modes
```

acao:
I was able to manually make a canary release for this:

`1.3.1-canary-21ee6a0e.0`

I'm not sure why the automated canary release is failing, as I ran the same exact script and it worked locally...

acao:
@bboure feel free to give the canary release here a try!

bboure:
Here are some issues I found:

`import { initializeMode } from 'monaco-graphql/full';` does not work with error `Module not found: Error: Default condition should be last one`

Same error with `import { initializeMode } from 'monaco-graphql/esm/lite';` and `import { initializeMode } from 'monaco-graphql/esm/initializeMode';`

`import { initializeMode } from 'monaco-graphql/esm/full';` does not work either. `full` does not export anything

`import { initializeMode } from 'monaco-graphql/lite';` does not work with error `Module not found: Error: Package path ./lite is not exported from package`

`import { initializeMode } from 'monaco-graphql/initializeMode';` does not error at runtime but VScode says `Cannot find module 'monaco-graphql/initializeMode' or its corresponding type declarations.`

I think all these errors are related to the `exports` in `package.json`

With the following config in my project:

```ts
import * as monaco from 'monaco-editor'; // force fully featured monaco
import { initializeMode } from 'monaco-graphql/initializeMode';

initializeMode();
```

```ts
new MonacoWebpackPlugin({
      languages: ['json', 'typescript', 'graphql'],
      features: ['!readOnlyMessage'],
      customLanguages: [
        {
          entry: undefined,
          label: 'graphql',
          worker: {
            id: 'graphql',
            entry: require.resolve('monaco-graphql/esm/graphql.worker.js'),
          },
        },
      ],
    }),
```

Graphql features don't work: autocomplete, validation, formatting...
Other languages work. e.g. typescript

Speaking of typescript, at some point, I also faced [that issue](https://github.com/microsoft/monaco-editor/issues/2738) that I never had before. 

These are just a few issues I met.
tbh, I don't know what is the root cause of all these issues yet. I tried many combinations by upgrading to monaco- editor (from 0.38.0 to 0.40.0 and 0.41.0) and the webpack plugin (from 7.01 to 7.1.0), importing fully -featured monaco vs just the API, etc. 

I'll try to do more tests and investigate what causes them later this weekend.

And thanks for the great work!

acao:
there is no /full export if that helps!

strange though, so many of these are working for me with vite and webpack

bboure:
@acao I did see the typescript fix, thanks. and it worked. I was just saying that it used to work for me before.
I could no determine what caused it to not work anymore. It's probably not this package.

Another thing I noticed, but minor. The canary release does not include the updates in `graphql-language-service`, so I was seeing the schema error too.

FYI, The README shows [an example](https://github.com/graphql/graphiql/pull/3384/files#diff-1cccda2d79e24f09552d5efe1bd9d2af68ffe365e37923abaa122c0316c3c000R155) exporting full

acao:
@bboure i fixed the readme and added some `require` exports for cjs, which may be needed if you are using node 16 or so.

packages/monaco-graphql/package.json
```diff
@@ -16,6 +16,41 @@
...
+      "import": "./esm/graphql.worker.js",
+      "require": "./dist/graphql.worker.js",
+      "types": "./esm/graphql.worker.d.ts"
+    },
+    "./lite": {
+      "import": "./esm/lite.js",
+      "types": "./esm/lite.d.ts",
+      "require": "./dist/lite.js"
+    },
+    "./graphql.worker": {
```
bboure, line 41: I'm still getting the following error.

```
Module not found: Error: Default condition should be last one
```

After some research, it turns out the solution is to move this property last in the object.

https://github.com/mpociot/chatgpt-vscode/issues/1#issuecomment-1337384976

Doing so removed the error.

acao: that is... so weird 😆

acao: @bboure are you using `vsce package`  as your bundler?

bboure: I'm using webpack.

scripts/canary-release.js
```diff
@@ -70,10 +70,10 @@ async function updateVersions() {
   const packages = await getPackages(cwd);
   const config = await readConfig(cwd, packages);
   const modifiedChangesets = getRelevantChangesets(config.baseBranch);
+  console.log({ modifiedChangesets });
```
bboure, line 73: you might want to remove this :)

packages/monaco-graphql/package.json
acao: what version? and which node version? because this is working with webpack 5 in next.js on node 20

bboure: node 18.12.0
webpack 5.80.0

bboure: It might be an issue specific to my project and/or configuration. It's an electron app

It seems related to `moduleResolution` but no matter how I change it, I'm having issues in one way or another.

Importing `import { initializeMode } from 'monaco-graphql/esm/initializeMode';` works fine in all cases.

acao: ah ok, that makes sense, we are still using moduleResolution: node and not setting type: module yet, so this probably confuses some bundler and/or node versions

acao: to note, next.js uses their own swc loader, whereas you may be using ts loader, and each of these will interpret exports config differently, so i will add default at the end if vite and next are happy with that as well. we should re-add monaco examples to the netlify build to confirm this works in node 16 as well

bboure: FWIW, I confirm I'm using ts-loader

acao:
@B2o5T this is important as it includes several key bugfixes, and is a useful preceeding step to `type: module`

PR #3397, opened 2023-08-08
 Fix incorrectly using 0-indexed number where 1-indexed number is expected

u9g:
This pr causes console spam in our case for when we hover a comment on the first line. See https://github.com/obi1kenobi/trustfall/issues/285 for more context.

What does this pr do?

Fixes an off-by-one bug that caused console spam when hovering things in the first line of a graphql file. This occurs because monaco's `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.

Why is this the right fix?
We have a 0-indexed row/col. To see this for myself, I put a `console.log()` [after this line](https://github.com/graphql/graphiql/pull/3397/files#diff-864a2605362ad715bfa1b09cd9cbf18396a38b11c4a09b28a05dd1e831aed2aaR87). I then hovered the first character on the first line. I see 0,0 printed, and this is after we subtract 1 from both line/col in `toGraphQLPosition()`, so we definitely have a 0-indexed number.

`getRange()` takes in a `SourceLocation`, however that's an interface, so it's not really obvious whether the line/col is 0-based or 1-based. However, there is a function `getLocation()` that returns a `SourceLocation` declared in the same `location.d.ts` from within the `graphql/language` repo. So taking a look at their `getLocation()` function, it looks like this:
```js
function getLocation(source, position) {
  let lastLineStart = 0;
  let line = 1;

  for (const match of source.body.matchAll(LineRegExp)) {
    typeof match.index === 'number' || (0, _invariant.invariant)(false);

    if (match.index >= position) {
      break;
    }

    lastLineStart = match.index + match[0].length;
    line += 1;
  }

  return {
    line,
    column: position + 1 - lastLineStart,
  };
}
```

When I saw that, it became clear that this line is 1-indexed, and the `+ 1` in the column leads me to believe this number is 1-based too. (if this isn't, please tell me!)

### DIFF
--- a/.changeset/little-hats-explain.md
+++ b/.changeset/little-hats-explain.md
@@ -2,4 +2,4 @@
 'monaco-graphql': patch
 ---
 
-Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
+Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco, because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
--- a/.changeset/little-hats-explain.md
+++ b/.changeset/little-hats-explain.md
@@ -2,4 +2,4 @@
 'monaco-graphql': patch
 ---
 
-Fixes an off-by-one bug that caused console spam when hovering things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
+Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
--- a/.changeset/little-hats-explain.md
+++ b/.changeset/little-hats-explain.md
@@ -1,0 +1,5 @@
+---
+'monaco-graphql': patch
+---
+
+Fixes an off-by-one bug that caused console spam when hovering things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
--- a/packages/monaco-graphql/src/GraphQLWorker.ts
+++ b/packages/monaco-graphql/src/GraphQLWorker.ts
@@ -74,13 +74,16 @@ export class GraphQLWorker {
     }
   }
 
-  public async doHover(uri: string, position: monaco.Position) {
+  public async doHover(uri: string, position: Position) {
+    // `position` is 1-indexed
     try {
       const documentModel = this._getTextModel(uri);
       const document = documentModel?.getValue();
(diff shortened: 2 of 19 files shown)


acao:
@u9g thanks, can you add a changeset?

u9g:
> @u9g thanks, this also fixes a bug in `monaco-graphql` iirc, can you add a changeset?

I added a changeset for the bug I fixed, should I add a changeset for [fix error range handling](https://github.com/graphql/graphiql/pull/3397/commits/5cdb733e7d543eed66056f7eff5287c3e912361a) bug too if it only fixed bugs in the test as far as I know?

acao:
huh. seeing the change you made to the language service makes me want to check for regressions in `monaco-graphql` and other places where we don't have tests yet

acao:
from what i can tell, this is a breaking change, so we need to give the graphql-language-service a major bump

acao:
yeah, looks from the e2e tests that youll need to handle the breaking change in `codemirror-graphql` as well

.changeset/little-hats-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'monaco-graphql': patch
+---
+
+Fixes an off-by-one bug that caused console spam when hovering things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
```
TallTed, line 5:
```suggestion
Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
```

u9g:
> yeah, looks from the e2e tests that youll need to handle the breaking change in `codemirror-graphql` as well

I actually took another look at my though process and remade this pr, take a look at the pr description for why I made the changes.

packages/monaco-graphql/src/GraphQLWorker.ts
```diff
@@ -94,8 +97,9 @@ export class GraphQLWorker {
         range: toMonacoRange(
           getRange(
             {
-              column: graphQLPosition.character,
-              line: graphQLPosition.line,
+              // we add one here because `getRange()` expects a 1-indexed position
```
acao, line 100: just a nit but this would be very useful to have in this file as `toMonacoPosition()` in `../utils`? I think it's the only place where we need to call it.
https://github.com/graphql/graphiql/blob/0f38050953a4158170034b94429c0429a229e653/packages/monaco-graphql/src/utils.ts#L42

.changeset/little-hats-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'monaco-graphql': patch
+---
+
+Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
```
TallTed, line 5:
```suggestion
Fixes an off-by-one bug that caused console spam when hovering over things in the first line of a graphql file with monaco, because `toGraphQLPosition()` from `monaco-graphql` returns 0-based line/col numbers and `getRange()` from `graphql-language-service` expects 1-based line/col numbers.
```

TallTed (changes requested):
to improve human clarity

trevor-scheer:
@u9g sorry this went by the wayside, but it should be fixed in current versions via https://github.com/graphql/graphiql/pull/4225

PR #3380, opened 2023-07-31
Installable PWA w/ simple server management plugin

acao:
In the interest of playground parity, this would be an installable, auto-updating replacement for the electron build

Because we deploy the webpack example build with netlify again, we can just go ahead and make it an installable PWA

image

Demo:

https://deploy-preview-3380--graphiql-test.netlify.app/webpack

click "install" in the URL bar on chrome

## Additional Features
- simple example SelectServer plugin that allows you to provide a graphql server, saves previously requested URLs and shows schema state
- working copy as cURL using the code-exporter plugin

PR #3402, opened 2023-08-12
Use the prod distribution of React instead of dev in the CDN example

kiendang:


acao:
we are already doing this in #3380

acao:
thanks, though!

kiendang:
Maybe I don't understand it fully, but that PR is for the `graphiql-webpack` example not the `graphiql-cdn` example no?

The reason for this PR is there are GraphQL frameworks out there like [strawberry](https://github.com/strawberry-graphql/strawberry) and [graphene-django](https://github.com/strawberry-graphql/strawberry) that serve GraphiQL and use the CDN example here as reference. Because of this sometimes they just copy paste and ended up shipping with react development which is 10 times the size of `production.min.js`. [Example](https://github.com/strawberry-graphql/strawberry/blob/4eb629ebc99f1ddfa82fa7ff0f2f369f1d3b5112/strawberry/static/graphiql.html#L46).

 Maybe it's worth changing here too?

tyranron:
@acao ping... I do agree with @kiendang here. Would you mind to reconsider this PR?

acao:
does https://graphiql-dev.netlify.app/webpack not work? what is the purpose in particular?

tyranron:
@acao it seems to me that @kiendang stated the problem quite clear:

> The reason for this PR is there are GraphQL frameworks out there like [strawberry](https://github.com/strawberry-graphql/strawberry) and [graphene-django](https://github.com/strawberry-graphql/strawberry) that serve GraphiQL and use the CDN example here as reference.

The choice of the `graphiql-cdn` example is obvious, because they do not want to bother with webpack and/or other toolchains just to serve the GraphiQL page. So, they mostly use it "as is", adjusting it [only in few bits to work with the framework](https://github.com/strawberry-graphql/strawberry/blob/4eb629ebc99f1ddfa82fa7ff0f2f369f1d3b5112/strawberry/static/graphiql.html#L134), but do not re-check optimality of the dependencies. And then things go viral, because every framework has its users.

> Because of this sometimes they just copy paste and ended up shipping with react development which is 10 times the size of `production.min.js`.

acao:
oh this is for the cdn example! that makes sense

acao:
really, we should be showing how to use `<script type="module">` as modern browsers don't need UMD anymore but that's another matter

kiendang:
Thanks a lot @tyranron for the clarification 👍 

@acao let me know if there's anything else needed to get this merged.

dimaMachina (approved):
I updated other examples too

PR #3405, opened 2023-08-12
relint when schema changes

imolorhe:
Relint when schema changes. Codemirror 6 normally only runs linting whenever the document changes. However this means whenever the GraphQL schema changes, the linting is not run again until the document changes. This can be lead to confusing behaviors.

A new option was added to the linter [recently](https://github.com/codemirror/lint/commit/2da5dbdae05e28af5b6b52334f7b10e70fd1111e) that allows configuring a predicate to determine when the linter will be ran when the state changes. We are leveraging the same here to trigger linting whenever the schema or passed option changes.

Note: linting always runs on document changes irrespective of what we define in this predicate, _which is what we want_.

Related issue: https://github.com/altair-graphql/altair/issues/2242

### DIFF
--- a/.changeset/blue-rice-arrive.md
+++ b/.changeset/blue-rice-arrive.md
@@ -1,0 +1,5 @@
+---
+"cm6-graphql": patch
+---
+
+relint when schema changes
--- a/packages/cm6-graphql/package.json
+++ b/packages/cm6-graphql/package.json
@@ -19,12 +19,12 @@
     "graphql-language-service": "^5.1.7"
   },
   "devDependencies": {
-    "@codemirror/autocomplete": "^6.0.0",
+    "@codemirror/autocomplete": "6.2.0",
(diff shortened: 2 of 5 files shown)


acao:
is it possible that we could update the example to show a simple schema change as well? perhaps just with a drop down?

acao (approved):

imolorhe:
@acao I was trying to get the example package working but I was running into some issues with the duplicate @codemirror/state dependency instances which is why I haven't done that yet. We may have to move the example into the package itself like we did in codemirror-json-schema

acao:
i got around that, issue. big pr coming, haha

imolorhe:
Oh nice. Can't wait to see what you come up with!

acao:
@imolorhe i was able to make this error go away by moving away from `basicSetup` package, whether it came from `codemirror` or `@codemirror/basic-setup`

Issue #3373, opened 2023-07-26
[graphiql] allow disabling tabs


n1ru4l:
In version 2 tabs were enabled by default, however, for some use-cases you might not need tabs at all. It seems like there is currently no possibility of disabling them.

acao:
good idea! this should be doable

PR #3408, opened 2023-08-22
GraphiQL component: `disableTabs` prop

TuvalSimha:
Closes issue: #3373 

Description:

- Now we can control the use of tabs and allow disabling tabs.

Result UI:
- No tabs at all:
Screenshot 2023-08-22 at 16 27 41


acao:
@TuvalSimha awesome, thank you!

acao:
I think there was a bug introduced with this PR so we will have to revert it

Issue #3372, opened 2023-07-26
[graphiql] Sending extensions in request


dondonz:
Hello, it can be useful to send extensions in a GraphQL request and I'd like to know if you're interested in supporting this feature.

At work, I have a use case where I want to send metadata via the extensions key in a GraphQL request. This metadata is then used to toggle behaviour in the GraphQL service, which is very useful for developers wanting to quickly iterate and test ideas in staging.

I saw a [previous discussion about request extensions](https://github.com/graphql/graphiql/issues/375) from 2017. Given that GraphiQL has changed significantly since then, I thought it could be easier to implement this now.

Would you consider this as a feature? I am happy to help out.

acao:
@dondonz yes, for sure! considering it's part of the spec, I would say we **should** support it as a requirement. Technically it is supported in that the fetcher prop supports any promise resolved exec request you like, but would you suggest additional features somehow? createGraphiQlFetcher () we've added to make creating a fetcher implementation easier, and additional headers can be supplied, but should we give users the ability to extens the request payload with this abstraction as well? and do we need to allow more keys in the response payload outside data as well? how do you see this working?

dondonz:
Thanks @acao for considering this feature!

I would like to add an extra tab for extensions in the UI, next to the tabs for variables and HTTP headers. In terms of the code, my plan was to essentially mirror what is already there for query variables, and insert the values into the `extensions` key of the request payload. Hopefully I can do that with a small change, as you say the fetcher technically supports it.

I think only adding the `extensions` key will be enough to be useful and also remain compliant with the [GraphQL over HTTP spec](https://graphql.github.io/graphql-over-http/draft/#sec-Request-Parameters).

The `extensions` key is a map without any restrictions, so I think that would be the best place to add any extra metadata, instead of enabling additional non-spec top level keys.

Additionally, I think we should restrict ourselves to the top level request keys defined in the spec, otherwise the GraphQL engine in the backing service may not recognise or read the extra top level keys. For example, I help out with GraphQL Java and the engine's representation of a request (`ExecutionInput`) would only include the top level keys defined in the spec.

And I'll admit, I have a bias for keeping the PR's scope small so I can get it done and merged faster!

Of course, this is your project so I am very open to ideas and advice. Thanks also for your time spent on maintaining this project!!

If you're happy to go ahead, I can give this a try and open a PR.

acao:
Here is a quick example for simple HTTP queries/mutations if you have static extensions at runtime

```ts
[...]

PR #3409, opened 2023-08-23
Add extensions to request payload & UI

dondonz:
This PR adds `extensions` to the GraphQL request payload and adds a new Extensions tab to the UI (alongside the Variables and Headers tabs). This PR addresses #3372.

Extensions

For more background on `extensions` in the request payload, see the GraphQL over HTTP specification regarding [Request Parameters](https://graphql.github.io/graphql-over-http/draft/#sec-Request-Parameters). It is a map of values with no restrictions on content. It's a useful place to send metadata such as development flags.

I tried to keep Extensions UI functionality as close as possible to the existing Variables and Headers behaviour. This PR also adds extensions to history and enables prettifying for extensions.

**What happens to the request if the Extensions tab is empty?** If the Extensions tab is empty, the corresponding GraphQL request payload will **not contain** the `extensions` key. Although the [spec](https://graphql.github.io/graphql-over-http/draft/#sec-Request-Parameters) states:

> Specifying null for optional request parameters is equivalent to not specifying them at all.

I prefer to be consistent with the prior behaviour of not setting an `extensions` key, rather than setting `extensions` to null, in case a service somewhere has not covered the null case.

### DIFF
--- a/.changeset/tricky-lions-try.md
+++ b/.changeset/tricky-lions-try.md
@@ -1,0 +1,7 @@
+---
+'graphiql': minor
+'@graphiql/react': minor
+'@graphiql/toolkit': minor
+---
+
+Add extensions to GraphQL request payload and add Extensions tab to UI
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -44,7 +44,7 @@ this repo._
    If you are focused on GraphiQL development, you can run —
 
    ```sh
-   yarn start-graphiql
+   yarn dev-graphiql
    ```
 
 5. Get coding! If you've added code, add tests. If you've changed APIs, update
@@ -89,7 +89,7 @@ First, you'll need to `yarn build` all the packages from the root.
 
 Then, you can run these commands:
 
-- `yarn start-graphiql` — which will launch `webpack` dev server for graphiql
+- `yarn dev-graphiql` — which will launch `webpack` dev server for graphiql
   from the root
 
 > The GraphiQL UI is available at http://localhost:8080/dev.html
--- a/packages/graphiql-react/src/editor/context.tsx
+++ b/packages/graphiql-react/src/editor/context.tsx
@@ -19,6 +19,7 @@ import {
 import { useStorageContext } from '../storage';
 import { createContextHook, createNullableContext } from '../utility/context';
 import { STORAGE_KEY as STORAGE_KEY_HEADERS } from './header-editor';
+import { STORAGE_KEY as STORAGE_KEY_EXTENSIONS } from './extension-editor';
 import { useSynchronizeValue } from './hooks';
 import { STORAGE_KEY_QUERY } from './query-editor';
 import {
@@ -97,6 +98,10 @@ export type EditorContextType = TabsState & {
    */
   variableEditor: CodeMirrorEditor | null;
   /**
+   * The CodeMirror editor instance for the extensions editor.
+   */
+  extensionEditor: CodeMirrorEditor | null;
+  /**
    * Set the CodeMirror editor instance for the headers editor.
    */
   setHeaderEditor(newEditor: CodeMirrorEditor): void;
@@ -112,6 +117,10 @@ export type EditorContextType = TabsState & {
    * Set the CodeMirror editor instance for the variables editor.
    */
   setVariableEditor(newEditor: CodeMirrorEditor): void;
+  /**
+   * Set the CodeMirror editor instance for the extensions editor.
+   */
+  setExtensionEditor(newEditor: CodeMirrorEditor): void;
 
   /**
    * Changes the operation name and invokes the `onEditOperationName` callback.
@@ -138,6 +147,11 @@ export type EditorContextType = TabsState & {
    * component.
    */
   initialVariables: string;
+  /**
+   * The contents of the extensions editor when initially rendering the provider
+   * component.
+   */
+  initialExtensions: string;
 
   /**
    * A map of fragment definitions using the fragment name as key which are
@@ -256,6 +270,14 @@ export type EditorContextProviderProps = {
   variables?: string;
 
   /**
+   * This prop can be used to set the contents of the extensions editor. Every
+   * time this prop changes, the contents of the extensions editor are replaced.
+   * Note that the editor contents can be changed in between these updates by
+   * typing in the editor.
+   */
+  extensions?: string;
+
+  /**
    * Headers to be set when opening a new tab
    */
   defaultHeaders?: string;
@@ -274,6 +296,9 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
   const [variableEditor, setVariableEditor] = useState<CodeMirrorEditor | null>(
     null,
   );
+  const [extensionEditor, setExtensionEditor] = useState<CodeMirrorEditor | null>(
+      null,
+  );
 
   const [shouldPersistHeaders, setShouldPersistHeadersInternal] = useState(
     () => {
@@ -288,6 +313,7 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
   useSynchronizeValue(queryEditor, props.query);
   useSynchronizeValue(responseEditor, props.response);
   useSynchronizeValue(variableEditor, props.variables);
+  useSynchronizeValue(extensionEditor, props.extensions);
 
   const storeTabs = useStoreTabs({
     storage,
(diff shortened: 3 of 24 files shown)


dondonz:
@acao could you please grant me permission to run the workflow? I'd like to run through all the build checks

packages/graphiql-toolkit/src/storage/history.ts
```diff
@@ -28,6 +28,7 @@ export class HistoryStore {
   private shouldSaveQuery(
     query?: string,
     variables?: string,
+    // extensions?: string, // dz todo: ask about saving strategy for extensions
```
dondonz, line 31: In this method, there is logic to determine whether a query should be saved depending on the difference to the last saved query.

I would prefer to not add more logic for extensions here, but I wanted to highlight this in case you had something in mind.

.changeset/tricky-lions-try.md
```diff
@@ -0,0 +1,7 @@
+---
+'graphiql': minor
+'@graphiql/react': minor
+'@graphiql/toolkit': minor
```
dondonz, line 4: I set these as minor changes but I am happy to be corrected. I chose minor as 
* an empty Extensions tab will result in the same behaviour as before (no `extensions` key in the request payload)
* `FetcherParams` are backwards compatible, it is fine to not set `extensions`
```
export type FetcherParams = {
  query: string;
  operationName?: string | null;
  variables?: any;
  extensions?: any;
};
```

DEVELOPMENT.md
```diff
@@ -44,7 +44,7 @@ this repo._
    If you are focused on GraphiQL development, you can run —
 
    ```sh
-   yarn start-graphiql
+   yarn dev-graphiql
```
dondonz, line 47: Updating the documentation

packages/graphiql-toolkit/src/create-fetcher/types.ts
```diff
@@ -28,6 +28,7 @@ export type FetcherParams = {
   query: string;
   operationName?: string | null;
   variables?: any;
+  extensions?: any;
```
dondonz, line 31: This is the key line which adds `extensions` to the payload. Everything else in this PR is for the Extensions UI tab

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -655,11 +703,7 @@ export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
                   <div ref={editorToolsResize.secondRef}>
                     <section
                       className="graphiql-editor-tool"
-                      aria-label={
-                        activeSecondaryEditor === 'variables'
-                          ? 'Variables'
-                          : 'Headers'
-                      }
```
dondonz, line 662: I replaced this ternary with a basic method. I tried to not be clever and go for a plain method rather than a .capitalize for the tab name

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -318,6 +338,18 @@ export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
...
+    if (currentTab === 'variables') {
+      return 'Variables';
+    }
+    if (currentTab === 'extensions') {
+      return 'Extensions';
+    }
+    if (currentTab === 'headers') {
+      return 'Headers';
+    }
+  };
```
dondonz, line 351: This looks like a pretty basic method, I thought I'd make it as simple as possible and not be clever with a .capitalize

Previously this was a ternary in the component, which was fine when there were only 2 tabs.

packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -267,19 +280,26 @@ export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
...
+      props.defaultEditorToolsVisibility === 'extensions' ||
       props.defaultEditorToolsVisibility === 'headers'
     ) {
       return props.defaultEditorToolsVisibility;
     }
-    return !editorContext.initialVariables &&
-      editorContext.initialHeaders &&
-      isHeadersEditorEnabled
-      ? 'headers'
-      : 'variables';
```
dondonz, line 282: I had to remove the ternary and use plain old if statements when introducing the extensions tab (the third tab)

dondonz: Well I guess I could have kept going with nested ternaries but nobody wants to see that 😄

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -233,6 +234,23 @@ export function usePrettifyEditors({ caller }: UsePrettifyEditorsArgs = {}) {
...
+          null,
+          2,
+        );
+        if (prettifiedExtensionEditorContent !== extensionEditorContent) {
+          extensionEditor.setValue(prettifiedExtensionEditorContent);
+        }
+      } catch {
+        /* Parsing JSON failed, skip prettification */
+      }
+    }
```
dondonz, line 252: Support for prettifying

packages/graphiql-react/src/history/components.tsx
```diff
@@ -112,10 +112,11 @@ export function HistoryItem(props: QueryHistoryItemProps) {
...
     });
-  const { headerEditor, queryEditor, variableEditor } = useEditorContext({
-    nonNull: true,
-    caller: HistoryItem,
-  });
+  const { headerEditor, queryEditor, variableEditor, extensionEditor } =
+    useEditorContext({
+      nonNull: true,
+      caller: HistoryItem,
+    });
```
dondonz, line 119: Add extensions to history items so they can be retrieved later

dondonz:
@acao I'm ready for a review, please let me know what you think. Thanks!

dondonz:
Hello, checking in, is there anything else you need from me?

dondonz:
Hi @acao could you please review this PR?

acao:
hi @dondonz, yes I can take a look today!

.changeset/tricky-lions-try.md
acao: @dondonz agreed on the minor release, and users will appreciate the backwards compatibility!

packages/graphiql-react/src/editor/context.tsx
```diff
@@ -288,6 +312,7 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
   useSynchronizeValue(queryEditor, props.query);
   useSynchronizeValue(responseEditor, props.response);
   useSynchronizeValue(variableEditor, props.variables);
+  useSynchronizeValue(extensionEditor, props.extensions);
```
acao, line 315: awesome stuff! i would prefer this to be possible via improvements to the plugin ecosystem, however `extensions` are already spec and 100% be a built-in feature of the execution context, though it will be tricky because as you can see there are many touchpoints. eventually each of these tabs will be provided by plugins, so someone could replace with a yaml and/or json schema react hook form implementation or some such

packages/graphiql-toolkit/src/storage/history.ts
acao: ah so - this will be important for extensions because it's how unique tabs are determined, and how extension input values can be saved in history and/or active tabs. you could leave this out for now, but then the lack of persistence will be reported as a bug knowing the expectations of users

packages/graphiql/src/components/GraphiQL.tsx
acao: nice that works. and it's still in the component scope so we can readily translate it later

acao (approved):
awesome work! looking forward to getting this shipped!

the only thing I want to add, as you can see from the comment, is persistence and query param support, if that's ok?

acao:
I would also like to add an extensions response example to our test schema, just so we can validate that extension payload responses are coming through, I think they are already yeah? at work we've been using apollo studio so i haven't tried graphiql with a schema with extensions in a while and I've forgotten haha. i would assume it works or there would be a bug haha

packages/graphiql-react/src/editor/context.tsx
acao: Here's an example to illustrate:

say you have 4 tabs, each with the exact same query/variables/headers combination, but each with unique extension values. when you reload the browser, you'll only have one tab, because the persistence only picked up one (I haven't tried this but I assume this will happen b/c of the unique operation matching logic)

dondonz:
Thanks for the review!! I was a bit busy with other things at work but I also want to get this merged in as soon as I can. I'll work on putting through your changes.

acao:
@dondonz thank you!

dondonz:
Hello it's been a while and I'd like to return to this work, as request extensions are now formally included in the spec https://github.com/graphql/graphql-spec/pull/976

However this branch is quite stale and there are now a lot of merge conflicts. I'll start a new branch and new PR

dimaMachina:
@dondonz Hey, would you like to rebase your PR on v5? I think it's nice to have this feature, saw it in Inigo Explorer https://explorer.inigo.io/

dondonz:
Hi @dimaMachina thanks, yes I would be very keen to. Given the major refactors, there's a number of breaking changes, so I'll start from scratch but the ideas will be the same as this PR

Also nice work on GraphiQL!

PR #3411, opened 2023-08-24
feat: support queries for multiple schemas in the same file

frandiox:
Hi 👋  First of all, thanks so much for the active maintenance of these packages!

While working on [shopify/hydrogen](https://github.com/shopify/hydrogen) I noticed that the GraphQL Language Service Server so far only supports 1 schema per file. However, we are introducing multiple GraphQL APIs and it would be nicer for DX to combine queries in the same file instead of separating them in different files.

This is an attempt to implement that. I was just trying to see if it was possible at all to open an issue, but it actually works so I'm just opening a PR instead 😅  -- feel free to discuss this feature completely!

The idea is that you need a config like the following with 2 projects pointing to the same files. Then, you can pass an option to set a "gql annotation suffix" in `extensions.languageService.gqlTagOptions.annotationSuffix`:

image

Then, in a given file that matches the pattern for both projects, you can choose to which GraphQL API/schema each of the queries points to with a modifier of the GraphQL annotation: `#graphql:<suffix>`

image
image

---

This whole thing also works with GraphQL Codegen by using [`documentTransforms`](https://the-guild.dev/graphql/codegen/docs/advanced/document-transform). Also, noticed that syntax highlights already works due to how the [match](https://github.com/graphql/graphiql/blob/57b5fcd8749bd1c54f4a6a04a9ab07316fd05e71/packages/vscode-graphql-syntax/grammars/graphql.js.json#L75) happens in that package (doesn't end in `\n`).

This feature at the moment only works when using the in-query comment `#graphql`. It doesn't work with `gql` tag or the leading comment `/* GraphQL */` because there's no access to that information where it's needed (maybe that could be changed? Unsure).
Another option is to just read a different comment inside the query, like `#project:storefront` and make it independent from the `#graphql` comment.

Thanks for checking it!

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
@@ -56,6 +56,10 @@ import {
   SymbolKind,
 } from 'vscode-languageserver-types';
 
+import { fileURLToPath } from 'node:url';
+
+export const EXTENSION_NAME = 'languageService';
+
 const KIND_TO_SYMBOL_KIND: { [key: string]: SymbolKind } = {
   [Kind.FIELD]: SymbolKind.Field,
   [Kind.OPERATION_DEFINITION]: SymbolKind.Class,
@@ -96,11 +100,38 @@ export class GraphQLLanguageService {
     this._logger = logger;
   }
 
-  getConfigForURI(uri: Uri) {
-    const config = this._graphQLCache.getProjectForFile(uri);
-    if (config) {
-      return config;
+  getAllProjectsForFile(uri: Uri) {
+    const filePath = uri.startsWith('file:') ? fileURLToPath(uri) : uri;
+    const projects = Object.values(this._graphQLConfig.projects).filter(
+      project => project.match(filePath),
+    );
+
+    return projects.length > 0
+      ? projects
+      : // Fallback, this always finds at least 1 project
+        [this._graphQLConfig.getProjectForFile(filePath)];
+  }
+
+  getProjectForQuery(
+    query: string,
+    uri: Uri,
+    projects?: GraphQLProjectConfig[],
+  ) {
+    if (!query.startsWith('#graphql')) {
+      // Query is not annotated with #graphql.
+      // Skip suffix check and return the first project that matches the file.
+      return projects?.[0] ?? this._graphQLConfig.getProjectForFile(uri);
     }
+
+    return (projects || this.getAllProjectsForFile(uri)).find(project => {
+      const ext = project.hasExtension(EXTENSION_NAME)
+        ? project.extension(EXTENSION_NAME)
+        : null;
+
+      const suffix = ext?.gqlTagOptions?.annotationSuffix;
+
+      return query.startsWith(`#graphql${suffix ? ':' + suffix : ''}\n`);
+    });
   }
 
   public async getDiagnostics(
@@ -111,7 +142,7 @@ export class GraphQLLanguageService {
     // Perform syntax diagnostics first, as this doesn't require
     // schema/fragment definitions, even the project configuration.
     let documentHasExtensions = false;
-    const projectConfig = this.getConfigForURI(uri);
+    const projectConfig = this.getProjectForQuery(document, uri);
     // skip validation when there's nothing to validate, prevents noisy unexpected EOF errors
     if (!projectConfig || !document || document.trim().length < 2) {
       return [];
@@ -218,7 +249,7 @@ export class GraphQLLanguageService {
     position: IPosition,
     filePath: Uri,
   ): Promise<Array<CompletionItem>> {
-    const projectConfig = this.getConfigForURI(filePath);
+    const projectConfig = this.getProjectForQuery(query, filePath);
     if (!projectConfig) {
       return [];
     }
@@ -255,7 +286,7 @@ export class GraphQLLanguageService {
     filePath: Uri,
     options?: HoverConfig,
   ): Promise<Hover['contents']> {
-    const projectConfig = this.getConfigForURI(filePath);
+    const projectConfig = this.getProjectForQuery(query, filePath);
     if (!projectConfig) {
       return '';
     }
@@ -272,7 +303,7 @@ export class GraphQLLanguageService {
     position: IPosition,
     filePath: Uri,
   ): Promise<DefinitionQueryResult | null> {
-    const projectConfig = this.getConfigForURI(filePath);
+    const projectConfig = this.getProjectForQuery(query, filePath);
     if (!projectConfig) {
       return null;
     }
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -288,6 +288,39 @@ export class MessageProcessor {
     );
   }
 
+  async _getDiagnosticsForAllFileProjects(
+    contents: CachedContent[],
+    uri: Uri,
+    projects: GraphQLProjectConfig[],
+  ): Promise<Diagnostic[]> {
+    return (
+      await Promise.all(
+        contents.map(async ({ query, range }) => {
+          const project = this._languageService.getProjectForQuery(
+            query,
+            uri,
+            projects,
+          );
+
+          if (
+            project?.extensions?.languageService?.enableValidation !== false
+          ) {
+            const results = await this._languageService.getDiagnostics(
+              query,
+              uri,
+              this._isRelayCompatMode(query),
+            );
+            if (results && results.length > 0) {
+              return processDiagnosticsMessage(results, query, range);
+            }
+          }
+
+          return [];
+        }),
+      )
+    ).reduce((left, right) => left.concat(right), []);
+  }
+
   async handleDidOpenOrSaveNotification(
     params: DidSaveTextDocumentParams | DidOpenTextDocumentParams,
   ): Promise<PublishDiagnosticsParams | null> {
@@ -359,35 +392,28 @@ export class MessageProcessor {
       return { uri, diagnostics };
     }
     try {
-      const project = this._graphQLCache.getProjectForFile(uri);
-      if (
-        this._isInitialized &&
-        project?.extensions?.languageService?.enableValidation !== false
-      ) {
-        await Promise.all(
-          contents.map(async ({ query, range }) => {
-            const results = await this._languageService.getDiagnostics(
-              query,
-              uri,
-              this._isRelayCompatMode(query),
-            );
-            if (results && results.length > 0) {
-              diagnostics.push(
-                ...processDiagnosticsMessage(results, query, range),
-              );
-            }
-          }),
+      const projects = this._languageService.getAllProjectsForFile(uri);
+
+      if (this._isInitialized) {
+        diagnostics.push(
+          ...(await this._getDiagnosticsForAllFileProjects(
+            contents,
+            uri,
+            projects,
+          )),
         );
       }
 
-      this._logger.log(
-        JSON.stringify({
-          type: 'usage',
-          messageType: 'textDocument/didOpenOrSave',
-          projectName: project?.name,
-          fileName: uri,
-        }),
-      );
+      for (const project of projects) {
+        this._logger.log(
+          JSON.stringify({
+            type: 'usage',
+            messageType: 'textDocument/didOpenOrSave',
+            projectName: project?.name,
+            fileName: uri,
+          }),
+        );
+      }
     } catch (err) {
       this._handleConfigError({ err, uri });
     }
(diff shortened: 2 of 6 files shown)


packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -288,6 +288,39 @@ export class MessageProcessor {
     );
   }
 
+  async _getDiagnosticsForAllFileProjects(
```
frandiox, line 291: This method is extracted from the code below to avoid repetition, and it's adapted to run for every project that matches the file.

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -254,8 +254,7 @@ export function findGraphQLTags(
       }
     },
     TemplateLiteral(node: TemplateLiteral) {
-      const hasGraphQLPrefix =
-        node.quasis[0].value.raw.startsWith('#graphql\n');
+      const hasGraphQLPrefix = node.quasis[0].value.raw.startsWith('#graphql');
```
frandiox, line 257: This makes the matching similar to the [syntax highlight here](https://github.com/graphql/graphiql/blob/57b5fcd8749bd1c54f4a6a04a9ab07316fd05e71/packages/vscode-graphql-syntax/grammars/graphql.js.json#L75), allowing `#graphql:suffix`.

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -96,11 +100,38 @@ export class GraphQLLanguageService {
     this._logger = logger;
   }
 
-  getConfigForURI(uri: Uri) {
-    const config = this._graphQLCache.getProjectForFile(uri);
-    if (config) {
-      return config;
+  getAllProjectsForFile(uri: Uri) {
+    const filePath = uri.startsWith('file:') ? fileURLToPath(uri) : uri;
```
frandiox, line 104: Passing an Uri that starts with the `file://` protocol doesn't work properly in `grahiQLConfig.getProjectForFile`. Internally that calls `minimatch` and it always returns false.

acao:
never thought to implement this because graphql-config didn't seem to support it, but here we are and it's a great idea.
as you may have noticed, isolated projects end up accidentally sharing schema as a bug in many contexts, so I'm glad you've added projects to the test mock as I'm planning to fix that as well.

I will focus on a few bugfixes this week, but with my current day job workload, I can't promise I'll have time to fully review this until next week. The lack of test coverage makes things precarious and there are a number of manual scenarios I like to check, but it seems you've already fixed a few known bugs yourself! And added coverage, thanks for that!

CC @dotansimha @Urigo

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
acao: yes there was a breaking change in vscode some years ago, and there are a few places that we didn't handle this, thanks for the bugfix! I will come back around look at the implementations because they should all be passing a file url

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -96,11 +100,38 @@ export class GraphQLLanguageService {
...
+  getProjectForQuery(
+    query: string,
+    uri: Uri,
+    projects?: GraphQLProjectConfig[],
+  ) {
+    if (!query.startsWith('#graphql')) {
+      // Query is not annotated with #graphql.
+      // Skip suffix check and return the first project that matches the file.
+      return projects?.[0] ?? this._graphQLConfig.getProjectForFile(uri);
     }
```
frandiox, line 124: Note: I didn't want to modify the existing behavior of other patterns so I didn't fix the `file://` protocol bug here. Since it's acknowledged that it's a bug we should fix, this line should be updated to remove `file://` as well.

benjaminsehl:
Any chance you could take a look at this @acao? 🙏

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
frandiox: It now removes `file://` here as well to prevent the bug.

frandiox:
I've merged main branch and resolved conflicts. Tests are still passing 🙏

acao:
unfortunately this impacts a lot of behavior for which there is no test coverage, so an exhaustive set of manual tests will be necessary. 

If it makes sense to introduce this after review with the spec committees, I should be able to get to it in the coming months. I hope to present it at the January WG, or perhaps you can present it at the next graphql working group? There is some high priority stability & test work that needs to come before major changes, as requested by the TSC and other stakeholders.

Though the LSP is only at about 50% coverage, even 100% coverage won't account for all the scenarios we need to cover before we can make major changes.

frandiox:
Understandable! It would be great if you could mention this at the January WG 🙌 
In the meantime, is there any public list of use cases that are normally tested manually? That would help us and the community to contribute with more unit/e2e test cases whenever possible.

acao:
@frandiox the readme for the language server and the graphql config website document many combinations of test cases we don't test for

also if you can update the PR, i refactored tag parsing and improved types in the language server some. I think we can get this improvement out there sooner than later!

acao:
oh and, can you add documention of this feature to the language server and vscode-graphql readmes?

frandiox:
>also if you can update the PR, i refactored tag parsing and improved types in the language server some. I think we can get this improvement out there sooner than later!
>oh and, can you add documention of this feature to the language server and vscode-graphql readmes?

Done! Let me know if you prefer docs in some other way.

Also, are we OK with the suggested `languageService.gqlTagOptions.annotationSuffix` public API? Or prefer to rename it / unnest it?

packages/graphql-language-service-server/README.md
```diff
@@ -237,18 +241,37 @@ via `initializationOptions` in nvim.coc. The options are mostly designed to
...
-| `vscode-graphql.cacheSchemaFileForLookup` | `false`                         | generate an SDL file based on your graphql-config schema configuration for schema definition lookup and other features. useful when your `schema` config are urls |
+| Parameter                                       | Default                         | Description                                                                                                                                                                                                                   |
+| ----------------------------------------------- | ------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `graphql-config.load.baseDir`                   | workspace root or process.cwd() | the path where graphql config looks for config files                                                                                                                                                                          |
```
TallTed, line 246:
```suggestion
| `graphql-config.load.baseDir`                   | workspace root or `process.cwd()` | the path where graphql config looks for config files                                                                                                                                                                          |
```

packages/graphql-language-service-server/src/findGraphQLTags.ts
```diff
@@ -128,7 +128,9 @@ export function findGraphQLTags(
       }
     },
     TemplateLiteral(node: TemplateLiteral) {
+      // check if the template literal is prefixed with #graphql
       const hasGraphQLPrefix = node.quasis[0].value.raw.startsWith('#graphql');
+      // check if the template expression has
```
TallTed, line 133:
```suggestion
      // check whether the template literal is prefixed with #graphql
      const hasGraphQLPrefix = node.quasis[0].value.raw.startsWith('#graphql');
      // check whether the template expression has
```

packages/vscode-graphql/README.md
```diff
@@ -137,6 +142,24 @@ module.exports = {
 Notice that `documents` key supports glob pattern and hence `["**/*.graphql"]`
 is also valid.
 
+Normally, you want to point your `documents` to different files per project to ensure only one schema is used for the queries. However, you can also mix queries for different schemas in the same file by adding a `#graphql:<suffix>` comment to the query that matches the `languageService.gqlTagOptions.annotationSuffix` for the project:
```
TallTed, line 145:
```suggestion
Normally, you would point your `documents` to multiple-files-per-project to ensure that only one schema is used for the queries. However, you can also mix queries for different schemas into the same file by adding a `#graphql:<suffix>` comment to each query, matching the `languageService.gqlTagOptions.annotationSuffix` for the project:
```

packages/vscode-graphql/README.md
```diff
@@ -300,14 +323,14 @@ further!
...
-1.  Run "VScode Extension" launcher in vscode
-1.  This will open another VSCode instance with extension enabled
-1.  Open a project with a graphql config file - ":electric_plug: graphql" in
-    VSCode status bar indicates that the extension is in use
-1.  Logs for GraphQL language service will appear in output section under
-    GraphQL Language Service
+1. Clone the repository - <https://github.com/graphql/graphiql>
+1. `yarn`
+1. Run "VScode Extension" launcher in vscode
+1. This will open another VSCode instance with extension enabled
```
TallTed, line 329:
```suggestion
1. Run "VScode Extension" launcher in vscode. This will open another VSCode instance with extension enabled
```

packages/vscode-graphql/README.md
```diff
@@ -300,14 +323,14 @@ further!
...
+1. Open a project with a graphql config file - ":electric_plug: graphql" in
+   VSCode status bar indicates that the extension is in use
+1. Logs for GraphQL language service will appear in output section under
```
TallTed, line 332:
```suggestion
1. Open a project with a graphql config file. `:electric_plug: graphql` in the
   VSCode status bar indicates that the extension is in use
1. Logs for GraphQL language service will appear in the output section, under
```

TallTed (changes requested):
A little human-facing language clarification. If it seems I've gotten any wrong, that's a good sign more clarification is needed. :-)

packages/vscode-graphql/README.md
frandiox: I think `multiple` here is not correct 🤔
What I wanted to say is that each project in `.graphqlrc` should point to different sets of files -- i.e. no overlap between projects.

frandiox:
>A little human-facing language clarification. If it seems I've gotten any wrong, that's a good sign more clarification is needed. :-)

Hi 👋  Thanks for the suggestions. It looks like most of the changes are not related to this PR, though. Should we move it to a different PR? Not sure how things strict things should be with PRs  cc @acao

packages/vscode-graphql/README.md
acao: with graphql config, this is kind of true, but not always. you could have both some shared schema files and seperate schema files between two projects. you could have shared fragments declared in documents config but seperate queries. i will show some examples soon

packages/graphql-language-service-server/src/GraphQLLanguageService.ts
```diff
@@ -111,7 +149,7 @@ export class GraphQLLanguageService {
     // Perform syntax diagnostics first, as this doesn't require
     // schema/fragment definitions, even the project configuration.
     let documentHasExtensions = false;
-    const projectConfig = this.getConfigForURI(uri);
+    const projectConfig = this.getProjectForQuery(document, uri);
```
acao, line 152: can you rename this to `getProjectForDocument` ? just so people don't assume this only works with queries, as it should also work with SDL

frandiox: Updated! I've also resolved the new conflicts but fyi, there's a problem with `cspell` on commit because it finds unknown words in `.vscode/settings.json`.

frandiox:
@acao Hi! I see there was a major refactoring in `main` branch, and that there's [another PR on the way](https://github.com/graphql/graphiql/pull/3556) with more changes.

Do you think it makes sense to update this PR now and resolve conflicts or wait until new notice?

TallTed:
> > [@TallTed] A little human-facing language clarification. If it seems I've gotten any wrong, that's a good sign more clarification is needed. :-)
> 
> [@frandiox] Hi 👋 Thanks for the suggestions. It looks like most of the changes are not related to this PR, though. Should we move it to a different PR? Not sure how strict things should be with PRs in this repo cc @acao

GitHub's change request interface includes an "add commit to batch" or similar button for you to use. If you *don't* use that button (i.e., you apply the change manually), it's difficult to be sure that change requests have been fully applied, and people like me have to review the current PR character-by-character. 

As it stands, it appears that you've rejected all the change requests that are now "marked resolved". There's no argument in place, saying why some change request has been rejected, so there's no argument I can make for applying them, nor any way for me to revise my request.

If you *do* have good reason to reject any request, please state it inline, or use that change request to start a new issue, such that the work I've already done here is not wasted, and all interested parties can weigh in.

(As to whether my change requests are relevant to this PR -- I generally confine my suggestions to the lines changed by any given PR, though sometimes they extend to a nearby line, which is displayed in GitHub's "Files changed" pane. My change requests are generally human-focused and non-normative. There shouldn't be a problem with including such changes in *any* PR, even if totally unrelated. However, if you feel that any given change request *does* deserve to be in a separate PR, you can always use that change request as the start of a new issue, as described earlier.)

Issue #3155, opened 2023-04-27
[graphiql] Don't render newlines in documentation strings


leonardehrenfried:
Long documentation strings that contain a newline are rendered as is in GraphiQL. This makes reading them a bit confusing and disjointed.

#### Doc pane
![Screenshot from 2023-04-27 12-39-23](https://user-images.githubusercontent.com/151346/234838868-39bcb8eb-209a-4574-8e97-9089e4ca9b08.png)

#### Tooltip
![Screenshot from 2023-04-27 12-31-46](https://user-images.githubusercontent.com/151346/234838872-a99f447b-6e17-46e7-81fb-5b894441c666.png)

In the original schema this looks like this:

```graphqls
    """
    Bearing, in degrees, clockwise from North, i.e., 0 is North and 90 is East. This can be the
    compass bearing, or the direction towards the next stop or intermediate location.
    """
    heading: Float
```

I would love it if these doc strings, which in our case can be multiple paragraphs, be treated like markdown. This would mean that a single newline is ignored and a double newline treated as a new paragraph.

Thank you for making this wonderful piece of software.

acao:
thanks @leonardehrenfried ! this is a great suggestion. `packages/graphiql-react/src/markdown.ts` is where we keep the configuration for `markdown-it`, and iirc, this is just a simple config tweak or plugin away from happening!

dimaMachina:
will be fixed in graphiql v4

PR #3414, opened 2023-08-28
style: Don't convert single \n to <br>

leonardehrenfried:
As described in #3155 we (OpenTripPlanner) tend to have quite long, multi-paragraph descriptions of our fields.

In Markdown a single `\n` is ignored and only two `\n` should lead to a new paragraph being started.

markdown-it makes this fairly easy to achieve, which is why this PR is tiny.

### Before

Screenshot from 2023-08-28 12-59-50

### After

Screenshot from 2023-08-28 13-00-02

I've also taken the liberty to update the dev instructions.

Many thanks for this wonderful tool!


acao:
@leonardehrenfried thank you for this, this is exactly how it should behave, I agree. I recall there was disagreement some years back and again with the redesign about this, I will look up some issues/discussions/? when I have more time. @thomasheyenbrock or @jonathanawesome what do you think?

one last thing, can you add a changeset? then it'll be ready to go!

acao:
awesome, thanks @leonardehrenfried ! I think we would make this a minor release for @graphiql/react as well

leonardehrenfried:
Changeset was updated.

leonardehrenfried:
Could I perhaps have another review for this?

acao:
hey thanks for this! i'll come back to this later today or tomorrow and release it with some other new features for this `@graphiql/react` minor release

leonardehrenfried:
Lovely thanks!

leonardehrenfried:
Could I perhaps have another review for this, please?

leonardehrenfried:
@acao Isn't this good to merge?

acao:
sorry I hope to have more time for this in the coming months. my current commitments are for my day job and for a grant to work on the GraphQL LSP server that I'm wrapping up

acao:
i think the only thing still stumping me is whether this constitutes a breaking change for some use cases, and whether there is some kind of markdown it config that solves the issue this intended to solve? maybe it was a workaround for the previous markdown library, was it marked? I need to look at the original history from, say, tag 0.7.y era or earlier to determine what the original purpose was

dimaMachina:
I agree with @acao that this should not affect the current stable version of GraphiQL, let's fix it in GraphiQL v4

Issue #2985, opened 2023-01-05
When using large schema files, you can not jump quickly every time


stfujnkk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm doing work about GraphQL API of shopify on VsCode.
However, the graphql definition of shopify is very large, resulting  jump to the definition is slow.
At first, I thought it was caused by not setting the jump related index. 
Later, I found that  I can jump very quickly when the first time. 
Therefore, I thought it was because the large file opened after the jump caused the plug-in to get stuck. 
I hope to cache this large file in the background instead of reloading it every time.

### Module pattern

- [ ] graphiql-umd
- [ ] graphiql-esm
- [ ] graphiql-commonjs

### Environment

```Markdown
- GraphiQL Version:
- OS: Windos
- Browser:
- Bundler:
- `react` Version:
- `graphql` Version:
```

stfujnkk:
I have temporarily solved this problem by splitting the file.
![image](https://user-images.githubusercontent.com/52458236/210795846-d5be3be6-535c-428b-a3e1-e7c660020ef8.png)

Brookke:
@stfujnkk also experiencing the same problem, do you have a script/tool to split up the schema file?

PR #3430, opened 2023-10-03
Disable document symbol request for files over 1MB

yepitschunked:
I ran into a problem similar to https://github.com/graphql/graphiql/issues/2985 with a massive schema graphql file (~2MB file). The performance was fine when the schema is not opened (e.g, hovering for type definintions and autocomplete), but the LSP was stuck at 100% as soon as I opened the schema. It looks like the language server was stuck in `getOutline`, which is called by `textDocument/documentSymbol`. I'm not sure why this is so slow, but I generally don't care about the outline in such a large file anyway, so it seems OK to disable that feature. 

This PR adds a configuration option `largeFileThreshold`  which defaults to 1000000. At the moment, it just disables document symbols computation, but there might be other expensive operations we could disable with this option as well. 

I also noticed there was no logging for this LSP message, so I added it.

### DIFF
--- a/packages/vscode-graphql/package.json
+++ b/packages/vscode-graphql/package.json
@@ -97,7 +97,7 @@
           "type": [
             "number"
           ],
-          "description": "Disables outlining and other expensive operations for files larger than this threshold (in bytes). Defaults to 1000000.",
+          "description": "Disables outlining and other expensive operations for files larger than this threshold (in bytes). Defaults to 1000000 (one million).",
           "default": 1000000
         },
         "vscode-graphql.rejectUnauthorized": {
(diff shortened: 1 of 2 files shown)


packages/vscode-graphql/package.json
```diff
@@ -93,6 +93,13 @@
           ],
           "description": "Use a cached file output of your graphql-config schema result for definition lookups, symbols, outline, etc. Disabled by default."
         },
+        "vscode-graphql.largeFileThreshold": {
+          "type": [
+            "number"
+          ],
+          "description": "Disables outlining and other expensive operations for files larger than this threshold (in bytes). Defaults to 1000000.",
```
TallTed, line 100:
```suggestion
          "description": "Disables outlining and other expensive operations for files larger than this threshold (in bytes). Defaults to 1000000 (one million).",
```

acao (approved):
looks great! thank you @yepitschunked !

acao:
Oops, overlooked the missing changeset comment. I will have to fix this for the release to work.

Issue #2896, opened 2022-11-12
Unexpected duplicate "display"


hinogi:
https://github.com/hinogi/graphiql/blob/1a3d034c02114c4a164e94626beb840260f05359/packages/graphiql-plugin-code-exporter/src/index.css#L52-L57

B2o5T:
Your are right!

@acao I think we can setup stylint for css files to avoid future mistakes

PR #3439, opened 2023-10-16
FIX:Unexpected duplicate "display".

xonx4l:
ISSUE-:#2896


B2o5T (commented):
I think you should remove first `display`

acao:
@xonx4l thanks for this, three things!

1. i think you need to remove display: block above instead as @dimaMachina noted
2. can you add a changeset with patch as per the comment?
3. can you sign the CLA, and make sure your commit is signed with the same email you signed the CLA with?

acao:
@xonx4l thanks so much! good to go!

PR #3441, opened 2023-10-28
fix: use the default query on adding a new tab

cimdalli:
When a new tab is added, it sets the query value as an empty string instead of `defaultQuery` value. Tab default query value should be `defaultQuery` variable if it's set on the component. Otherwise, it should be DEFAULT_QUERY constant.


acao:
makes sense to me. do any other @graphql/graphiql-maintainers have feelings about this?

benjie:
I'm for this change 👍

acao:
two things:

1. the cypress suite is failing on some tests you added
2. the `headers` prop is a bit nuanced - however if it is set, then `defaultHeaders` should always be overridden. `query`, `variables` and `headers` prop, and soon `extensions`, should drive a controlled component state, at least I think? For 4.x I'm considering dropping all the controlled component props with a migration path to the `@graphiql/react` hooks instead, because it leads to a much more perfomant and easier to implement controlled ccomponent implementation

acao:
@cimdalli do you plan on overriding `defaultHeaders` with provided `headers` prop?

acao:
@cimdalli so it seems this clashes a bit with previously expected behavior:

image

when `headers` are supplied, they should only set the first tab, and not be used for new tabs apparently?

so we need to revert this for headers and do the same for query/defaultQuery. otherwise, it's a change in functionality and requires a new minor version

fieldsco:
Any update on this?

.changeset/brown-tigers-nail.md
```diff
@@ -0,0 +1,6 @@
+---
+"@graphiql/react": patch
+"graphiql": patch
+---
+
+fix: use the default query on adding a new tab
```
dimaMachina, line 6:
```suggestion
fix: set query editor to `defaultQuery` while adding a new tab or GraphiQL's default query

```graphql
# Welcome to GraphiQL
#
# GraphiQL is an in-browser tool for writing, validating, and
# testing GraphQL queries.

....
```

```

**Review by dimaMachina — APPROVED** [review]

**Comment by klippx** [comment]
Is everyone happy with this change? I find it quite a bit annoying to have the "welcome message" shown every time you open a new tab... 😞

**Comment by klippx** [comment]
I decided to keep the default query as we assert on text from it to ensure the user is logged in and page is fully loaded and that the test can be started. `defaultQuery=''` is an option, but makes it harder to know that the page is loaded. To get things working in playwright we simply clear the editor using a <select all> + <cut> directive before filling it with input:

```
  const editorTextbox = page.getByLabel('Query Editor').getByRole('textbox')
  await editorTextbox.press('ControlOrMeta+a')
  await editorTextbox.press('ControlOrMeta+x')
```

It is more a matter of the UX: Each time a new tab is opened the defaultQuery is filled, which previously was only shown on the initial tab (something I have gotten used to). Imagine if your browser showed a tutorial screen every time you opened a new tab, or your favorite text editor showed a whole page of documentation every time you hit CMD+N. 

When creating a new tab - you want a blank screen, at least that's my opinion. However it is fine if it is shown on inital tab on initial load. Perhaps an option `defaultQueryBehaviour = 'initial' | 'always'` is what I would consider, and maybe this extra amount of complexity would be merited.

PR #3451, opened 2023-11-05
add header variable for v0.13.2

chunhui2001:
add header variable for v0.13.2


acao:
@chunhui2001 we are currently on 3.x releases, and cannot cut a 0.x release, sorry!

PR #3452, opened 2023-11-05
add header variable for v0.13.2

chunhui2001:
add header variable for v0.13.2


acao:
I'm sorry but this is still against a very very old version of graphiql. are you able to check out the latest `main` branch? perhaps you have a very old cloned copy of graphiql and havent run `git pull`?  we converted to typescript in 2019/20

PR #3461, opened 2023-11-16
Wrap cm6-graphql lint logic in try..catch

imolorhe:
It turns out that `getDiagnostics` may throw errors which codemirror 6 doesn't handle in the lint. Wrapping the logic in try..catch to handle errors and just return an empty array instead.

### DIFF
--- a/.changeset/silver-lions-build.md
+++ b/.changeset/silver-lions-build.md
@@ -1,0 +1,5 @@
+---
+"cm6-graphql": patch
+---
+
+Wrap cm6-graphql lint logic in try..catch
--- a/packages/cm6-graphql/src/lint.ts
+++ b/packages/cm6-graphql/src/lint.ts
@@ -8,44 +8,48 @@ const SEVERITY = ['error', 'warning', 'info'] as const;
 
 export const lint: Extension = linter(
   view => {
-    const schema = getSchema(view.state);
-    if (!schema) {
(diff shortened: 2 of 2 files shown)


acao (approved):
makes sense for me! now I think i need to add this to our other mode, haha!

imolorhe:
@acao I was thinking about this a bit more. The error that gets thrown is usually because of an invalid schema:
- `getDiagnostics()` [calls](https://github.com/graphql/graphiql/blob/main/packages/graphql-language-service/src/utils/validateWithCustomRules.ts#L84) `validate()` in graphql
- `validate()` [throws an error](https://github.com/graphql/graphql-js/blob/main/src/validation/validate.ts#L52) if the schema is not valid

With this current try..catch, we are ignoring this type of error, which you can argue is fine since it's not a validation error in the GraphQL query (which the linter is meant for), but should we also report the invalid GraphQL schema from the linter as well? (The position of the error can just be line 1, col 1). I feel like that might be more useful to devs. We could also consider making devs choose the behavior they want here with a config option.

What do you think?

acao:
Hmm yeah, that makes a lot of sense to me!
Perhaps we can even write schema errors to an StateField? Or maybe that's not needed

imolorhe:
Yeah writing to a state field may not be needed

PR #3463, opened 2023-11-18
Create a lint diagnostic from invalid schema

imolorhe:
Create a lint diagnostic from invalid schema
Follow up from https://github.com/graphql/graphiql/pull/3461

### DIFF
--- a/.changeset/two-doors-shave.md
+++ b/.changeset/two-doors-shave.md
@@ -1,0 +1,5 @@
+---
+"cm6-graphql": patch
+---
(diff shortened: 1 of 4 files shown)


acao (approved):
this all makes sense to me, thanks @imolorhe !

yes I would assume the schema error should be present by default. this should only need to be disabled when, for example, someone is making a new spec proposal, or perhaps when building a schema server from scratch and experimenting with the introspection response and you want to reduce the noise

PR #3480, opened 2024-01-05
Update esbuild.js

### LINKED ISSUE #3171: vscode-graphql-execution not running
### Discussed in https://github.com/graphql/graphiql/discussions/3170

<div type='discussions-op-text'>

<sup>Originally posted by **PixeledCode** May 10, 2023</sup>
Attaching the screenshot. I have the latest extension installed, noticed I was not getting any codelens in `.graphql` files. Language Feature and Syntax Highlighting are working as expected.
I also do not see any log for `vscode-graphql-execution`. Is that normal?

Screenshot 2023-05-10 at 2 25 37 PM
</div>

craig-riecke:
This is probably a naive fix, but it does make the extension work again in my copy of VSCode.  

Fixes #3171

### DIFF
--- a/.changeset/smooth-clocks-try.md
+++ b/.changeset/smooth-clocks-try.md
@@ -1,0 +1,6 @@
+---
+"vscode-graphql-execution": patch
+"vscode-graphql": patch
+---
+
+Fix execution extension esbuild bundling
--- a/packages/vscode-graphql-execution/esbuild.js
+++ b/packages/vscode-graphql-execution/esbuild.js
@@ -46,10 +46,13 @@ build({
     'atpl',
     'liquor',
     'twig',
-    'graphql-config',
   ],
   format: 'cjs',
   sourcemap: true,
+  define: { 'import.meta.url': '_importMetaUrl' },
+  banner: {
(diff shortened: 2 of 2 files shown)


acao:
oh great, this wasn't working before. did you test with URL schema or sdl file schema or both? and you tested by installing the locally generated vsix bundle?

craig-riecke:
Turns out I was naive - merely making graphql-config an external package stopped that particular error, and worked in VSCodes extension runner, but failed when I bundled the extension and installed it on its own.  

This fix actually does work, inspired by https://github.com/evanw/esbuild/issues/1492.  I exported it as a .vsix bundle, installed it locally, and tried it with both URL and file-based Graphql Schemas, and it worked in every case.

acao:
@craig-riecke awesome, thank you for looking it into me! this is the last piece we needed. now i can add an e2e test suite hopefully!

acao:
@craig-riecke also can you add a changeset so we can kick off a patch release for this? perhaps even a minor release now that it's working again (new feature: it works 😆 )

acao:
works for me as well on vsix bundle! awesome to see

now i see a bug though where it reports a successful response as an error, so I can fix that in another PR

Issue #3515, opened 2024-01-25
[vscode-graphql-syntax] (v1.2.3) triple double-quote comment syntax broken for TypeScript 


TiE23:
With v1.2.3 (possibly originating in #3490, though I am not certain) support for triple double-quote comments, which are used to generate doc comments for the schema (whether it's a built-in feature for Prisma or an undocumented feature for graphql-codegen I'm not sure).

### Behavior in v1.2.2
![Screenshot 2024-01-25 at 12 33 34 AM](https://github.com/graphql/graphiql/assets/2433505/4d79158a-f91d-4311-babb-ba41435e3235)

### Behavior in v1.2.3
![Screenshot 2024-01-25 at 12 32 57 AM](https://github.com/graphql/graphiql/assets/2433505/9016e35e-3ef2-43c0-89d2-6ab2b603b12d)

It appears that if the `"""` comment is placed inside a curly brace, it works fine. But in the root scope (?), it appears to break.

How this is useful to me:
<img width="359" alt="image" src="https://github.com/graphql/graphiql/assets/2433505/bea2e8be-c53d-44c4-bbb2-21b776f73221">

<img width="294" alt="image" src="https://github.com/graphql/graphiql/assets/2433505/b9a22dc0-608d-43d8-8a1e-0e81429d0518">

Current work-around is to install v1.2.2
![image](https://github.com/graphql/graphiql/assets/2433505/f5398e86-5e53-4462-8cde-64a47bb54f7b)

acao:
ah yes, it appears to be this:

https://github.com/graphql/graphiql/pull/3490/files#diff-89a0bf9dd1dfa9498b6e3f566257d012e1bfff0cb6b663735aac011eaae74dbcR50

which was introduced to fix inline double quote queries like `graphql("#query { id }")`

can you open a PR that adds this failing case to the test suite? I can take a look after work today

TiE23:
I hope this is what you needed!
* #3517

acao:
@TiE23 thank you! I will finish fixing this for now, there is another huge PR someone has proposed for the textmate grammars which does not solve this problem, so I will solve this one first!

acao:
`1.3.2` will be your fix version!

PR #3517, opened 2024-01-25
add failing tests for triple-double-quote comments for js

TiE23:
This PR is not intended to be merged, but rather to provide tests to work with in fixing my reported bug in #3515 - for @acao and [their comment](https://github.com/graphql/graphiql/issues/3515#issuecomment-1909438594).

I'm new to this project so this is simply my educated guess as to what they need.

I can imagine these tests can be reduced, but given that the error affected these multi-line doc comments differently if they were outside or inside a block, I wanted to cover my bases by placing these test comments in various locations. 

_I can imagine the last one for scalar is not necessary but hey, it's there in case it's needed._

### DIFF
--- a/packages/vscode-graphql-syntax/tests/__fixtures__/test.js
+++ b/packages/vscode-graphql-syntax/tests/__fixtures__/test.js
@@ -86,3 +86,50 @@ const queryWithLeadingAboveComment =
       }
     }
   `;
+
+// TODO: fix this
+const queryWithDocString = `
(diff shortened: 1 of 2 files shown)


acao:
@TiE23 thanks for this, but I don't see the graphql delimiters though? as it stands, we don't have any way to know these are graphql strings.

acao:
fixed by #5329, thank you!

PR #3518, opened 2024-01-26
fix: Fix vscode-graphql-syntax’s grammar to support string literals on separate lines

kitten:
## Summary

We're currently working on [`gql.tada`](https://github.com/0no-co/gql.tada), which is _mostly_ compatible with the expected syntax of `graphql` calls.
However, while getting this in users hands we noticed that a specific grammar that Prettier formats `graphql()` calls into fails to be matched by the Textmate grammar in `vscode-graphql-syntax`.

Specifically, when a GraphQL string literal is placed on a new line, the `begin` pattern of the rules can't match. This is a limitation in Textmate grammars, as it turns out, and we can't even manually match `\n` patterns.

I'm not an expert at Textmate grammars, but, in theory, a grammar rule that matches GraphQL calls as a lookbehind, then matches the inner string literal separately, should do the trick.

<table>
<thead><tr>
<td><strong>Before</strong></td>
<td><strong>After</strong></td>
</tr></thead>
<tr>
<td>Before</td>
<td>After</td>
</tr>
</table>

## Follow-up questions

- Is there a test suite / sample files for this we can check for regressions?
- Does anyone here know more about Textmate grammars to proof read this change?

## Set of changes

- Try out newline matches in `begin` rule (**DELETED CHANGE**)
- Add separate rule for matching `graphql(...)` calls then matching string literals inside
- Combining redundant rules into one (move generic over)
- Update snapshots (remove `TODO: fix this`) and add more grammar checks for JS

### DIFF
--- a/.changeset/famous-games-begin.md
+++ b/.changeset/famous-games-begin.md
@@ -2,4 +2,4 @@
 'vscode-graphql-syntax': patch
 ---
 
-Fix Textmate grammar to support string literals that don’t immediately follow a function call paranthesis (`(`).
+Fix TextMate grammar to support string literals that don’t immediately follow a function call's left-parenthesis (`(`).
--- a/.changeset/famous-games-begin.md
+++ b/.changeset/famous-games-begin.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql-syntax': patch
+---
+
+Fix Textmate grammar to support string literals that don’t immediately follow a function call paranthesis (`(`).
--- a/packages/vscode-graphql-syntax/grammars/graphql.js.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.js.json
@@ -4,7 +4,7 @@
   "patterns": [
     {
       "contentName": "meta.embedded.block.graphql",
-      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental)|(/\\* GraphQL \\*/))\\s*\\(?\\s*(`|'|\")",
+      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental)|(/\\* GraphQL \\*/))\\s*(\\(\\s*[\\n\\s]*)?(`|'|\")",
       "beginCaptures": {
         "1": {
           "name": "variable.other.class.js"
--- a/packages/vscode-graphql-syntax/grammars/graphql.js.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.js.json
@@ -36,7 +36,7 @@
     },
     {
       "contentName": "meta.embedded.block.graphql",
-      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental))\\s*(?:<.*>)\\s*\\(?\\s*(`|'|\")",
+      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental))\\s*(?:<.*>)\\s*(\\(\\s*[\\n\\s]*)?(`|'|\")",
       "beginCaptures": {
         "1": {
           "name": "variable.other.class.js"
--- a/packages/vscode-graphql-syntax/grammars/graphql.js.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.js.json
@@ -15,6 +15,11 @@
               "name": "punctuation.definition.string.template.begin.js"
             }
           },
+          "endCaptures": {
+            "0": {
+              "name": "punctuation.definition.string.template.end.js"
+            }
+          },
           "patterns": [
             {
               "include": "source.graphql"
--- a/packages/vscode-graphql-syntax/grammars/graphql.js.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.js.json
@@ -36,7 +36,7 @@
     },
     {
       "contentName": "meta.embedded.block.graphql",
-      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental))\\s*\\(?\\s*(?:<.*>)(`|'|\")",
+      "begin": "\\s*+(?:(?:(Relay)\\??\\.)(QL)|(gql|graphql|graphql\\.experimental))\\s*(?:<.*>)\\s*\\(?\\s*(`|'|\")",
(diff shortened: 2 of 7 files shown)


acao:
@kitten thanks for this! to answer your questions, see the syntax extension readme

acao:
also, can you add to the test fixtures the test case you originally hoped to cover? I don't see it present in your PR

kitten:
@acao: Cheers! I forgot to add some more cases for the original failing matches ✌️ Those should be added in: https://github.com/graphql/graphiql/pull/3518/commits/057f22e378bd74d13928917b1541c81704aa94e7
Regarding the readme, I did test the extension manually as well, so the screenshots should cover that 🤔

.changeset/famous-games-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+'vscode-graphql-syntax': patch
+---
+
+Fix Textmate grammar to support string literals that don’t immediately follow a function call paranthesis (`(`).
```
TallTed, line 5:
```suggestion
Fix TextMate grammar to support string literals that don’t immediately follow a function call's left-parenthesis (`(`).
```

acao:
@kitten amazing work here! looking forward to getting this released

I had to release a quick hotfix for the triple double quote syntax `""" comment """`, I hope you don't mind merging that change. for now we aren't as concerned with `graphql(" query { id }")` type scenarios because they are much less common than the `""" comment """` syntax. if you can make both of those worlds happy, then heck yeah that's awesome! but if not, it's no bother

also, one last nice to have - maybe confirm a few weird pre-auto-formatted variations just to be sure we don't deprecate it somehow in the future. it's nice to have the strings always working no matter whether auto-formatting (prettier, etc) is used or has been applied

for example, cases like this

```js
const query = graphql(
` query {
    something
    }
 `);
```

our LSP server uses normal AST parsing ofc, but with the textmate regexes so many things can go wrong haha

kitten:
@acao Merged that back 👍 I'm not too concerned either. No idea what the syntax currently says for single line `graphql(" \"\"\"comment\"\"\" query { id }")` cases, but the original case that this PR addresses is Prettier putting a template literal (or `'query'` string) onto their own lines, which broke the previous syntax definition.

Given that double-quotes `"` inside a GraphQl string would have to be quoted, Prettier likely would reformat the string with single quotes. And that even assumes they're single line. So, either way, we should be fine there.

re. corner cases, I'm not too concerned about adding all kinds of newline cases to the tests. There's one, which basically represents all possible cases; i.e. `\n[ ]{2}` before the start of a string. That's equivalent to “all” cases, so to speak, since Textmate grammar can't match past newlines, so the call syntax (e.g. `graphql(\n ... \n)`) is what we ultimately have to check against.
So, all good from my end 👍

kitten:
Just came back from some holidays, so coming back to this, let me know if there's anything else we want to do here, or if we're good to go ✌️❤️

acao:
> since Textmate grammar can't match past newlines
I'm not sure if this is true, if I understand?

otherwise though it looks great! i will take a closer look and see about merging it soon

acao:
actually, can you add some more test cases to `test-js.md` as well? it should work transparently

kitten:
@acao: re. not matching past multiple lines, I found a more detailed discussion [here](https://github.com/microsoft/vscode-textmate/issues/41), which is where I initially ended up for this PR. Basically, what I meant was that `begin` and `end` can't literally include matches for a newline character, since the matches are separated line by line and that's why the new rules match the function call first.

I applied some of the additions to `test-js.md`. I don't think the fixtures there test for code matches, but it's now covering the relevant cases for future changes to the syntax 👍

xamir82:
Just stumbled into this issue. Would love to see this merged soon.

acao (approved):
awesome work, this is really comprehensive!

the regression snapshots will help tremendously if someone tries a "bugfix" for their corner case that breaks this scenario, which is exactly why we set up this snapshot suite!

this is a fantastic fix contribution to our ecosystem and many others that use our extension, so thank you @kitten !

acao:
There appears to be something awry with github actions, where it's not reporting the run status (for me at least), and the log output formatting is broken, but manually checking each job, it looks fine! merging now, thanks again!

acao:
bad news - it appears this github actions event is not triggering the changeset bot?

good news - once the issue recovers, we will still be able to release the changes without a problem with the next merge to `main`

aradalvand:
@acao Sorry for being annoying but when will this be released?Thank you.

kitten:
It has already been released: https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql-syntax

VSCode extension, as per the link showing a release time of  	
01/03/2024, 18:31:18

aradalvand:
Has caused a regression, showcased in #3542.

acao:
no worries @kitten this happens! lets double check the snapshot suite and make sure we have the right fixtures for this case and/or see if the snapshot change reveals this bug

acao:
@aradalvand 0.9.2 is now released, and removes this regression by reverting the 0.9.1 release. thanks for reporting!

acao:
@kitten I think the issue we can see from the snapshot side is that the end ` and ; i think are being tokenized by graphql somehow? So i think the query remains unterminated and attempts to highlight the rest of the file from the first graphql tag to the end of the file. i would look closely at the serializer results, and follow the readme instructions for performing manual builds to visually confirm the outcomes, where you open the fixture files and can confirm that everything is as expected

kitten:
I'll take a look! I've tested this before in VSCode, but I'm guessing there's either a case that isn't covered or I made a mistake while resolving conflicts 🤔

acao:
yeah, that's what I guessed as well! also, my bad for not manually confirming everything immediately pre-release myself, that's my job 😆

kitten:
Not your job and not your job solely if you do consider it your job ❤️ I did confirm and double-check this, but it was a subtle mistake where the pattern wasn't quite matching, because the positive lookbehind failed, and was neither applying ts/js/etc highlighting inside the function call expression, nor was it then matching correctly after the pattern, since the pattern causes an error.

At least that's my assumption.
It's easy to miss because the tests don't contain the VSCode tmLanguage patterns for JS/TS themselves, which makes it hard to see that this is happening, because it does at first glance all look correct.

This does make me think that while this may sound silly, visual snapshots would be really helpful in the future, if the VSCode team is willing to provide such a tool 🤔 

I found the issue and will open a follow-up PR

kitten:
Fix is in a new PR here ❤️ https://github.com/graphql/graphiql/pull/3545
Sorry for messing this up!

PR #3523, opened 2024-01-31
fix: Do not override existing json diagnostic schemas

bboure:
When working with different editors, the JSON schemas are always overridden by new ones


acao (approved):
awesome work! thanks for all the multi-editor fixes

PR #3526, opened 2024-02-01
Fix bug in Explorer plugin where characters are dropped when typing quickly

benjie:
To see a reproduction of this issue, check out: https://github.com/graphile/crystal/issues/1845

The issue stems from some delays interfering with the reactivity model of React - when you type a character into Explorer, that change gets transformed into a GraphQL document string which is sent upstream into GraphiQL, GraphiQL then updates everything and sends the result back down stream to GraphiQL Explorer, which renders the new data. The issue is that if the user has typed another character in the interrim, their keypress will be overwritten by this flow because there is (I think!) a delay somewhere, i.e. the state is not updated synchronously. If the user types faster than this update loop, characters will be lost.

To solve this, we could use the "throttle" or "debounce" patterns to only send updates every so often or when the user is paused typing, but this leads to a less-than-instant experience for the user. Instead, I've opted to send the update upstream and then not send any further updates upstream until we receive back what we sent. If we get back what we sent then we send the next update (skipping all the intermediary keypresses and just jumping right to the current state), giving the user near-instant feedback on their typing. If what we receive is not what we sent then something else must have happened, and thus we overwrite our local state (which is the status quo). Through intense and rapid typing into both the explorer plugin and the operations text I was not able to trigger this branch, but nonetheless it is present to handle this situation.

The interface now feels buttery smooth to me, so I think this issue is solved.

To make the solution nice and clean I've put it in its own hook.

### DIFF
--- a/packages/graphiql-plugin-explorer/src/index.tsx
+++ b/packages/graphiql-plugin-explorer/src/index.tsx
@@ -9,7 +9,7 @@ import {
   Explorer as GraphiQLExplorer,
   GraphiQLExplorerProps,
 } from 'graphiql-explorer';
-import React, { useCallback } from 'react';
+import React, { useCallback, useState, useEffect, useRef } from 'react';
 
 import './graphiql-explorer.d.ts';
 import './index.css';
@@ -139,7 +139,8 @@ function ExplorerPlugin(props: GraphiQLExplorerPluginProps) {
   );
 
   // load the current editor tab state into the explorer
-  const [operationsString, handleEditOperations] = useOperationsEditorState();
+  const [operationsString, handleEditOperations] =
+    useOptimisticOperationsEditorState();
 
   return (
     <GraphiQLExplorer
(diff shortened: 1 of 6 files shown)


packages/graphiql-plugin-explorer/src/index.tsx
```diff
@@ -159,6 +166,80 @@ function ExplorerPlugin(props: GraphiQLExplorerPluginProps) {
...
+        lastStateRef.current.pending = null;
+        if (upstreamOperationsString !== operationsString) {
+          // Change has occurred; upstream it
+          lastStateRef.current.pending = operationsString;
+          upstreamHandleEditOperations(operationsString);
+        }
+      } else {
+        // They got a different update; overwrite our local state (!!)
+        lastStateRef.current.pending = null;
+        setOperationsText(upstreamOperationsString);
```
benjie, line 213: I am unable to trigger this branch; but it is present for completeness.

acao:
awesome work!! i wonder if we should add this logic to the hooks in graphiql/react actually though? perhaps with a denounce option and arguments? what do you think? i could see other plugins needing this

benjie:
Sure; I've moved it into the react hooks as a hook result wrapper, so you can combine it e.g. `const [state, setState] = useOptimisticState(useOperationsEditorState());`

I considered making it an option (`useOperationsEditorState({optimistic: true})`) but since it requires a number of additional hooks (and branching is not allowed in hooks, so we couldn't opt out of them) I thought that this would be an unnecessary performance overhead for most usage of this hook - the wrapper approach seemed more efficient and clearer that there are performance side-effects (both good and bad).

acao:
awesome work!!!

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -388,3 +388,76 @@ export const useHeadersEditorState = (): [
...
 };
+
+/**
+ * Implements an optimistic caching strategy around a useState-like hook in
+ * order to prevent loss of updates when the hook has an internal delay and the
+ * update function is called again before the updated state is sent out.
+ *
+ * Use this as a wrapper around `useOperationsEditorState`,
+ * `useVariablesEditorState`, or `useHeadersEditorState` if you anticipate
+ * calling them with great frequency (e.g. due to mouse, keyboard or network
```
TallTed, line 399:
```suggestion
 * calling them with great frequency (e.g., due to mouse, keyboard or network
```

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -396,8 +396,8 @@ export const useHeadersEditorState = (): [
  *
  * Use this as a wrapper around `useOperationsEditorState`,
  * `useVariablesEditorState`, or `useHeadersEditorState` if you anticipate
- * calling them with great frequency (e.g. due to mouse, keyboard or network
- * events).
+ * calling them with great frequency (due to, e.g., mouse, keyboard or
+ * network events).
```
TallTed, line 400:
```suggestion
 * calling them with great frequency (due to, for instance, mouse, keyboard, or
 * network events).
```

acao (approved):
looks great!

PR #3545, opened 2024-03-02
fix: Fix vscode-graphql-syntax’s grammar to support string literals on separate lines [Reapply & Fix]

### LINKED ISSUE #3542: [lsp-server] 🐞 regression: syntax highlighting messed up

### Current Behavior

Most likely caused by #3518

in v1.3.3:
image

in v1.3.2:
image

### Expected Behavior

_No response_

### Steps To Reproduce

The example screenshotted above (just make sure you have the Svelte extension + the GraphQL extension installed in VS Code, and then create a `Test.svelte` file, pasting the following content into it):
```svelte
<script
	lang="ts"
	context="module"
>
	export const cartEventFragment = graphql(`
		fragment cartEvent on CartEvent @_unmask {
			__typename
			occurredAt
			guestToken
			user {
				fullName
				mobilePhoneNumber
			}
			course {
				id
				miniTitle
			}
		}
	`);
</script>

<script lang="ts">
	// Imports:
	import { routeMap } from '$common/routeMap';
	import { toPersianDigits } from '$common/misc';
	import type { ComponentProps } from 'svelte';
	import { graphql, type ResultOf } from '$graphql';
	// Imports | Components:
	import EventBox from '$components/EventBox.svelte';
	import type EventBoxEvent from '$components/EventBoxItem.svelte';
	// Imports | Icons:
	import CartPlusIcon from '$assets/icons/cart-plus.svg?component';

	// Props | Required:
	export let events: readonly ResultOf<typeof cartEventFragment>[];

	// Maps:
	const typeToTheme = new Map<
		ResultOf<typeof cartEventFragment>['__typename'],
		ComponentProps<EventBoxEvent>['theme']
	>([
		['CartItemAddedEvent', 'cyan'],
		['CartItemDeletedEvent', 'red'],
	]);
</script>

<EventBox
	title="آخرین *رویدادهای سبد خرید*"
	icon={CartPlusIcon}
	{events}
	fetchMore={async currentCount => {
		const { events } = await graphql(
			`
				query getLatestCartEvents($skip: Int!, $take: Int!) {
					events: latestCartEvents(skip: $skip, take: $take) {
						...cartEvent
					}
				}
			`,
			[cartEventFragment]
		).execute({ skip: currentCount, take: 10 });
		return events;
	}}
	let:event
	let:EventItem
>
	<EventItem
		occurredAt={event.occurredAt}
		theme={typeToTheme.get(event.__typename)}
		let:Highlighted
		let:Link
	>
		{#if event.user}
			<Highlighted textToCopy={event.user.fullName}>
				{event.user.fullName}
			</Highlighted>
			به شماره همراه
			<Highlighted textToCopy={event.user.mobilePhoneNumber}>
				{toPersianDigits(event.user.mobilePhoneNumber)}
			</Highlighted>
		{:else}
			کاربر میهمان به شناسه
			<Highlighted textToCopy={event.guestToken}>
				{event.guestToken.substring(0, 6)}
			</Highlighted>
		{/if}
		دوره‌ی
		<Link href={routeMap.coursePage(event.course)}>{event.course.miniTitle}</Link>
		را
		{#if event.__typename == 'CartItemAddedEvent'}
			به سبد خرید خود افزود.
		{:else}
			از سبد خرید خود حذف کرد.
		{/if}
	</EventItem>
</EventBox>
```

### Environment

* LSP Server Version:
* OS:
* LSP Client:

### Anything else?

_No response_

kitten:
Resolves #3542
Reapplies: #3518
Related: #3543

The previous implementation for GraphQL template string literals inside function expression caused a subtle error in syntax highlighting.

Because it used a positive lookbehind that contained whitespace matches, this caused an issue with syntax matching, which breaks subsequent TS/JS rules from matching. In other words, it caused a conflict between built-in highlighting and the embedded GraphQL highlighting.

Instead of this, we now match the function expressions explicitly, and include JS/TS rules inside the match to also fix arguments that live alongside the GraphQL template string.

Fixed syntax highlighting with this PR applied

### DIFF
--- a/packages/vscode-graphql-syntax/tests/__snapshots__/markdown-grammar.spec.ts.snap
+++ b/packages/vscode-graphql-syntax/tests/__snapshots__/markdown-grammar.spec.ts.snap
@@ -109,50 +109,6 @@ const Component = () => {                            |
 };                                                   | 
 \`\`\`                                                  | 
                                                      | 
-\`\`\`js                                                | 
-const variable = 1;                                  | 
-                                                     | 
-const queryA = graphql(\`                             | 
-  query {                                            | 
-    something(arg: \${variable})                      | 
-  }                                                  | 
-\`);                                                  | 
-                                                     | 
-const queryB = graphql(                              | 
-  \`                                                  | 
-    query {                                          | 
-      something(arg: \${variable})                    | 
-    }                                                | 
-  \`                                                  | 
-);                                                   | 
-                                                     | 
-const queryC = graphql(                              | 
-  'query { something(arg: \${variable}) }'            | 
-);                                                   | 
-\`\`\`                                                  | 
-                                                     | 
-\`\`\`ts                                                | 
-const variable: number = 1;                          | 
-                                                     | 
-const queryA = graphql(\`                             | 
-  query {                                            | 
-    something(arg: \${variable})                      | 
-  }                                                  | 
-\`);                                                  | 
-                                                     | 
-const queryB = graphql(                              | 
-  \`                                                  | 
-    query {                                          | 
-      something(arg: \${variable})                    | 
-    }                                                | 
-  \`                                                  | 
-);                                                   | 
-                                                     | 
-const queryC = graphql(                              | 
-  'query { something(arg: \${variable}) }'            | 
-);                                                   | 
-\`\`\`                                                  | 
-                                                     | 
 ### svelte                                           | 
                                                      | 
 \`\`\`svelte                                            | 
--- a/packages/vscode-graphql-syntax/tests/__snapshots__/markdown-grammar.spec.ts.snap
+++ b/packages/vscode-graphql-syntax/tests/__snapshots__/markdown-grammar.spec.ts.snap
@@ -109,6 +109,50 @@ const Component = () => {                            |
 };                                                   | 
 \`\`\`                                                  | 
                                                      | 
+\`\`\`js                                                | 
+const variable = 1;                                  | 
+                                                     | 
+const queryA = graphql(\`                             | 
+  query {                                            | 
+    something(arg: \${variable})                      | 
+  }                                                  | 
+\`);                                                  | 
+                                                     | 
+const queryB = graphql(                              | 
+  \`                                                  | 
+    query {                                          | 
+      something(arg: \${variable})                    | 
+    }                                                | 
+  \`                                                  | 
+);                                                   | 
+                                                     | 
+const queryC = graphql(                              | 
+  'query { something(arg: \${variable}) }'            | 
+);                                                   | 
+\`\`\`                                                  | 
+                                                     | 
+\`\`\`ts                                                | 
+const variable: number = 1;                          | 
+                                                     | 
+const queryA = graphql(\`                             | 
+  query {                                            | 
+    something(arg: \${variable})                      | 
+  }                                                  | 
+\`);                                                  | 
+                                                     | 
+const queryB = graphql(                              | 
+  \`                                                  | 
+    query {                                          | 
+      something(arg: \${variable})                    | 
+    }                                                | 
+  \`                                                  | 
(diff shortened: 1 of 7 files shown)


kitten:
Zip containing a `vsix` file for testing these changes: [vscode-graphql-syntax-1.3.4.vsix.zip](https://github.com/graphql/graphiql/files/14470071/vscode-graphql-syntax-1.3.4.vsix.zip)

acao:
thank you @kitten !!! looks awesome!

packages/vscode-graphql-syntax/tests/__snapshots__/markdown-grammar.spec.ts.snap
```diff
@@ -107,6 +107,50 @@ graphql\`                                             |
 const Component = () => {                            | 
   return <div> </div>;                               | 
 };                                                   | 
+\`\`\`                                                  | 
+                                                     | 
+\`\`\`js                                                | 
+const variable = 1;                                  | 
+                                                     |
```
acao, line 114: huh, these aren't reporting any graphql tokens here in the spec file...

acao: wait... i wonder if our changes in a previous PR mean they won't show up as tokens here anymore. I wonder if there is even a point to this spec anymore. I will need to check. obviously it's already regressed for the rest of the file

acao: I will check it out manually as well, but my guess is that it's working with markdown, but with the new markdown injection strategy we were granted by some user with incredible TM knowledge in a vscode issue the team had trouble answering, I don't think our serializer picks up on it's tokenization correctly, because you problably noticed this outlandish injection selector:

```
  "injectionSelector": "L:(meta.embedded.block.javascript | meta.embedded.block.typescript | source.js | source.ts | source.tsx | source.vue | source.svelte | source.astro) -source.graphql -inline.graphql -string -comment",
```

i think now it can universally applying to every embedded js/ts block it's pointed at including... html <script tags? 
i think i would just need to add that here:

```
 "injectTo": [
          "source.js",
          "source.ts",
          "source.js.jsx",
          "source.tsx",
          "source.vue",
          "source.svelte",
          "source.astro",
          "text.html.markdown"
          /// other text.html values??
        ],
        ```

acao: I wonder if at this point, we are testing vscode's `text.html.markdown` injector then? I think the contributor said as much, that the markdown fixture and spec is not longer necessary, except for ```graphql ``` tags

kitten: I'm not sure I'm 100% following on what needs to be done 🤔
For what it's worth, I did open the fixtures manually and they seem fine

But yea, while I don't understand the test setup 100%, and it seems that they don't necessarily include any syntax definitions except for the ones defined in this repo, this may need to be fixed up in the future! Not sure if it's that pressing though?

Screenshot 2024-03-05 at 14 21 20

acao: hey sorry, nothing needs to be done here, I am musing on another issue unrelated that I just noticed while reviewing this PR, I digress!

but if you _want_, you can remove the update to the markdown snap/fixture, because I'm about to remove them!

kitten:
Rebased & reverted changes to `tests/__fixtures/test-js.md`

acao (approved):
Ok let's do this again!

If there are any more edge case bugs this time, instead of reverting we will create new tests and solve the issues as they arise

acao (approved):

Issue #3516, opened 2024-01-25
[lsp-server] 🐞 Error: Cannot find module 'astrojs-compiler-sync'


M1CK431:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Server crash with error:
```
Error: Cannot find module 'astrojs-compiler-sync'
```

### Expected Behavior

Server starts without error

### Steps To Reproduce

Simply install the latest v0.9.0 of "GraphQL: Language Feature Support" vscode extension.

### Environment

* LSP Server Version: v0.9.0
* OS: Archlinux
* LSP Client: "GraphQL: Syntax Highlighting" vscode extension v1.3.0

### Anything else?

Rollback on v0.8.25 => no error

acao:
@M1CK431 oh noo... i manually tested with vscode-graphql vsix and everything!

did you see this error with vscode-graphql or another LSP client?

acao:
huh, are you using it with astro files? i cannot reproduce it from the vscode marketplace. i will try again after work

M1CK431:
Extra info: I'm using [Code OSS](https://wiki.archlinux.org/title/Visual_Studio_Code#Installation), again dunno if it may have any impact.

Dalichowski:
![image](https://github.com/graphql/graphiql/assets/45388509/d1d20805-a245-4ec2-b5ab-932136d2f2d1)
I have the same issue here :sweat:

joeiw:
I'm also experiencing this error on v0.9.1

`[Error - 11:48:13] vscode-graphql client: couldn't create connection to server.
  Message: Pending response rejected since connection got disposed
  Code: -32097 
node:internal/modules/cjs/loader:966
  throw err;
  ^

Error: Cannot find module 'astrojs-compiler-sync'
Require stack:
- /home/joe/.vscode-oss/extensions/graphql.vscode-graphql-0.9.1-universal/out/server/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:963:15)
    at Module._load (node:internal/modules/cjs/loader:811:27)
    at f._load (node:electron/js2c/asar_bundle:2:13330)
    at Module.require (node:internal/modules/cjs/loader:1035:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at /home/joe/.vscode-oss/extensions/graphql.vscode-graphql-0.9.1-universal/out/server/index.js:4617:6454
[...]

Issue #3547, opened 2024-03-05
[lsp-server] 🐞 Pending response rejected since connection got disposed


M1CK431:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Server crash with stacktrace:
```
Node.js v18.17.1
[Error - 14:12:42] The vscode-graphql server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
Initialization failed
Pending response rejected since connection got disposed
Error: Pending response rejected since connection got disposed
	at Object.dispose (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:29:4460)
	at Object.dispose (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:39:22314)
	at Ll.handleConnectionClosed (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:39:12275)
	at Ll.handleConnectionClosed (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:39:50478)
	at t (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:39:11956)
	at ia.invoke (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:2:14331)
	at n.fire (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:2:15100)
	at Lt (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:3:6456)
	at ia.invoke (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:2:14331)
	at n.fire (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:2:15100)
	at Pa.fireClose (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:2:17355)
	at ChildProcess.<anonymous> (/home/user/.vscode-oss/extensions/graphql.vscode-graphql-0.9.2-universal/out/extension.js:29:13076)
	at ChildProcess.emit (node:events:526:35)
[...]

PR #3549, opened 2024-03-05
fix OpenVSX dependency bug by tweaking openvsx publish

acao:
I think this will fix the open VSX issue, because as it turns out, `ovsx` cli just uses `vsce` to re-build your project if you don't pass `extensionFile` on `publish`! oops!


M1CK431:
closes #3516 #3547

PR #3554, opened 2024-03-06
Fix autocomplete suggestion

### LINKED ISSUE #3553: [lsp-server] 🐞 Adding whitespaces\newlines causes autocompletion to move up a level

### Current Behavior

Consider the following code:

```dart
mutation createComment2 @auth(is: PUBLIC) {
  comment_insert2(data: {  
  
  })
}
```

where `data` expects an `input` type.

In this scenario, adding whitespaces when trying to get autocompletion seems to change the result. Here's a video showcasing the issue:

https://github.com/graphql/graphiql/assets/20165741/50786f0c-faf1-4fac-8fdf-93805d8ada2e

### Expected Behavior

Adding whitespaces/newlines should not have any impact on the suggestion. 

### Steps To Reproduce

_No response_

### Environment

* LSP Server Version: latest
* OS: macos
* LSP Client: vscode-graphql

### Anything else?

I've tried debugging this by adding logs on a fork of the repository, and the issue appears to be `getAutocompletionSuggestions.ts#getTokenAtPosition`.

I added a `console.log(token)` inside `getAutocompleteSuggestions` right before the call to `getTypeInfo` here
https://github.com/graphql/graphiql/blob/ece99f63f5d8d01057b735e90a6957edea3e42b9/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts#L181

Then compared the log outputs based on where the cursor is.

Given `comment_insert2(data: {|  })`, the token returned is:
```
{
  start: 0,
  end: 0,
  string: '{',
  state: { ... },
  style: 'punctuation'
}
```

But for `comment_insert2(data: { | })` it is:
```
{
  start: 0,
  end: 0,
  string: '}',
  state: { ... },
  style: 'punctuation'
}
```

And lastly for `comment_insert2(data: {  |})` it is:
```
{
  start: 0,
  end: 0,
  string: ')',
  state: { ... },
  style: 'punctuation'
}
```

This doesn't make sense to me. I would expect all of them to return the token with `string: '{'`

rrousselGit:
fixes #3553

I'll add tests later, assuming that those changes are the correct way to go. I'm not familiar enough with the codebase to know for sure :)

### DIFF
--- a/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
+++ b/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
@@ -66,6 +66,9 @@ import {
   RuleKinds,
   RuleKind,
   ContextTokenForCodeMirror,
+  ParserOptions,
+  LexRules,
+  ParseRules,
 } from '../parser';
 
 import {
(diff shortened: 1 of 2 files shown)


.changeset/early-ties-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'graphql-language-service': minor
+---
+
+Fixes incorrect suggestions when adding whitespace/newlines. Fixes #3553
```
rrousselGit, line 5:
```suggestion

`runOnlineParser` now optionally receives a `ParserOptions`.
```

.changeset/early-ties-bake.md
```diff
@@ -0,0 +1,6 @@
+---
+'graphql-language-service': minor
+---
+
+
+`runOnlineParser` now optionally receives a `ParserOptions`.
```
rrousselGit, line 6:
```suggestion
Fixes incorrect suggestions when adding whitespace/newlines. Fixes #3553

`runOnlineParser` now optionally receives a `ParserOptions`.
```

acao:
awesome, thank you @rrousselGit! can you add some unit tests for the failing cases?

rrousselGit:
Sure! But I'd like to know whether this is the right approach first.

There are a few failing tests, and I'm not necessarily sure why. Do you have any clue on what could possibly be going wrong with this approach?

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -968,18 +971,70 @@ export function getTokenAtPosition(
...
   let stateAtCursor = null;
   let stringAtCursor = null;
-  const token = runOnlineParser(queryText, (stream, state, style, index) => {
-    if (
-      index !== cursor.line ||
-      stream.getCurrentPosition() + offset < cursor.character + 1
-    ) {
-      return;
+
+  /** Returns whether a token is before, at or after the cursor */
```
TallTed, line 975:
```suggestion
  /** Returns whether a token is before, at, or after the cursor */
```

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -968,18 +971,70 @@ export function getTokenAtPosition(
...
+  const token = runOnlineParser(
+    queryText,
+    (stream, state, style, index) => {
+      const compare = compareTokenPositionToCursor(
+        index,
+        stream.getCurrentPosition() + offset,
+      );
+
+      if (compare === 'before') {
+        // The token is before the cursor, keep track of it
```
TallTed, line 1008:
```suggestion
        // The token is before the cursor; keep track of it
```

packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts
```diff
@@ -968,18 +971,70 @@ export function getTokenAtPosition(
...
+        lastState = state;
+        lastStyle = style;
+        lastString = stream.current();
+        return;
+      }
+
+      // The token is at or after the cursor, we return the last token before the cursor
```
TallTed, line 1015:
```suggestion
      // The token is at or after the cursor; we return the last token before the cursor
```

TallTed (changes requested):
human-facing punctuation tweaks, for clarity of comments

acao:
@rrousselGit I will have to get back to you on that once I have time to build up a bit more context. Going into the parser takes great delicacy as you can see haha!

I would suggest a TDD approach at first to learn how the parser works, and reading the tests, you can use yarn test --watch at the workspace root and it will run only the impacted/changed unit tests

acao:
this is causing a lot of regressions it seems. perhaps you can find another place to fix this issue?

PR #3563, opened 2024-03-19
Allow GraphiQL apps control over closing tabs

klippx:
### Allow GraphiQL apps control over closing tabs

This feature is about closing tabs.

Current functionality:
- When closing a tab, it just happens, the tab is removed and all state is gone

Requested functionality
- Have a possibility to at give control back to the user (i.e. the app that mounts the GraphiQL component) if they are OK with closing a particular tab
- This could for instance be an dialogue "Are you sure you want to close the tab?"

In our case, we are working on a Collection plugin that you can click to read saved queries / variables / headers from a DB, and this opens it in a new tab. So we need to sync tabState with the list of Collection items. And if a user starts to change a Collection item, we want to warn the user that changes will be lost.

It is the same as if the user selected a file in a code editor, started to change it, and then tries to close the tab. 

#### Mock screenshots

Screenshot 2024-03-19 at 16 32 30

Screenshot 2024-03-19 at 16 32 41

### DIFF
--- a/packages/graphiql-react/src/editor/context.tsx
+++ b/packages/graphiql-react/src/editor/context.tsx
@@ -56,6 +56,14 @@ export type EditorContextType = TabsState & {
    */
   changeTab(index: number): void;
   /**
+   * When the user clicks a close tab button, this function is invoked with
+   * the index of the tab that is about to be closed. It returns a promise
+   * that should resolve to `true` (meaning the tab may be closed) or `false`
+   * (meaning the tab may not be closed).
+   * @param index The index of the tab that should be closed.
+   */
+  closeTabConfirmation(index: number): Promise<boolean>;
+  /**
    * Move a tab to a new spot.
    * @param newOrder The new order for the tabs.
    */
@@ -212,6 +220,14 @@ export type EditorContextProviderProps = {
    * @param operationName The operation name after it has been changed.
    */
   onEditOperationName?(operationName: string): void;
+  /**
+   * When the user clicks a close tab button, this function is invoked with
+   * the index of the tab that is about to be closed. It returns a promise
+   * that should resolve to `true` (meaning the tab may be closed) or `false`
+   * (meaning the tab may not be closed).
+   * @param index The index of the tab that should be closed.
+   */
+  confirmCloseTab?(index: number): Promise<boolean>;
   /**
    * Invoked when the state of the tabs changes. Possible triggers are:
    * - Updating any editor contents inside the currently active tab
@@ -422,6 +438,19 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
     [onTabChange, setEditorValues, storeTabs],
   );
 
+  const closeTabConfirmation = useCallback<
+    EditorContextType['closeTabConfirmation']
+  >(
+    async index => {
+      if (props.confirmCloseTab) {
+        const confirmation = await props.confirmCloseTab(index);
+        return confirmation;
+      }
+      return true;
+    },
+    [props.confirmCloseTab],
+  );
+
   const closeTab = useCallback<EditorContextType['closeTab']>(
     index => {
       setTabState(current => {
@@ -497,6 +526,7 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
       addTab,
       changeTab,
       moveTab,
+      closeTabConfirmation,
       closeTab,
       updateActiveTabValues,
 
--- a/packages/graphiql/src/components/GraphiQL.tsx
+++ b/packages/graphiql/src/components/GraphiQL.tsx
@@ -102,6 +102,7 @@ export type GraphiQLProps = Omit<GraphiQLProviderProps, 'children'> &
 
 export function GraphiQL({
   dangerouslyAssumeSchemaIsValid,
+  confirmCloseTab,
   defaultQuery,
   defaultTabs,
   externalFragments,
@@ -138,6 +139,7 @@ export function GraphiQL({
 
   return (
     <GraphiQLProvider
+      confirmCloseTab={confirmCloseTab}
       getDefaultFieldNames={getDefaultFieldNames}
       dangerouslyAssumeSchemaIsValid={dangerouslyAssumeSchemaIsValid}
       defaultQuery={defaultQuery}
@@ -545,11 +547,15 @@ export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
                             {tab.title}
                           </Tab.Button>
                           <Tab.Close
-                            onClick={() => {
-                              if (editorContext.activeTabIndex === index) {
-                                executionContext.stop();
+                            onClick={async () => {
+                              if (
+                                await editorContext.closeTabConfirmation(index)
+                              ) {
+                                if (editorContext.activeTabIndex === index) {
+                                  executionContext.stop();
+                                }
+                                editorContext.closeTab(index);
                               }
-                              editorContext.closeTab(index);
                             }}
                           />
                         </Tab>
--- a/packages/graphiql-react/src/provider.tsx
+++ b/packages/graphiql-react/src/provider.tsx
@@ -22,6 +22,7 @@ export type GraphiQLProviderProps = EditorContextProviderProps &
 
 export function GraphiQLProvider({
   children,
+  confirmCloseTab,
   dangerouslyAssumeSchemaIsValid,
   defaultQuery,
   defaultHeaders,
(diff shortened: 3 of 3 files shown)


packages/graphiql/src/components/GraphiQL.tsx
```diff
@@ -545,11 +547,15 @@ export function GraphiQLInterface(props: GraphiQLInterfaceProps) {
                             {tab.title}
                           </Tab.Button>
                           <Tab.Close
-                            onClick={() => {
-                              if (editorContext.activeTabIndex === index) {
-                                executionContext.stop();
+                            onClick={async () => {
+                              if (
```
acao, line 551: love this! however, i think it would make more sense to add this to the context methods in the sdk rather than in graphiql reference implementation itself. so we will add this to `closeTab()` instead, so it can be available to both `graphiql` and @graphiql/react` users and graphiql plugin developers as well
in `editor/context.tsx`, you can update the method to this
```ts
  const closeTab = useCallback<EditorContextType['closeTab']>(
    async index => {
      if (await closeTabConfirmation(index)) {
        setTabState(current => {
          const updated = {
            tabs: current.tabs.filter((_tab, i) => index !== i),
            activeTabIndex: Math.max(current.activeTabIndex - 1, 0),
          };
          storeTabs(updated);
          setEditorValues(updated.tabs[updated.activeTabIndex]);
          onTabChange?.(updated);
          return updated;
        });
      }
    },
    [onTabChange, setEditorValues, storeTabs, closeTabConfirmation],
  );

```

acao, line 551: i think we should move all of this into `closeTab` itself, even what was directly in the handler here before, making it async. i see no reason to let this be an implementation detail for sdk users, as if the context they are using has this logic they would expect this as as an internal detail of the method if it's exposed to the user yeah?

```ts
  const closeTab = useCallback<EditorContextType['closeTab']>(
    async index => {
      if (await closeTabConfirmation(index)) {
      // this could be a breaking change for users who are already stopping 
      // execution like graphiql does here:
        if (index === tabState.activeTabIndex) {
          stop();
        }
        setTabState(current => {
          const updated = {
            tabs: current.tabs.filter((_tab, i) => index !== i),
            activeTabIndex: Math.max(current.activeTabIndex - 1, 0),
          };
          storeTabs(updated);
          setEditorValues(updated.tabs[updated.activeTabIndex]);
          onTabChange?.(updated);
          return updated;
        });
      }
    },
    [
      onTabChange,
      setEditorValues,
      storeTabs,
      closeTabConfirmation,
      tabState.activeTabIndex,
    ],
  );
  ```
and then here in graphiql:

```tsx

 <Tab.Close
    onClick={async () => this.closeTab(index)}
/>

```

acao (approved):
this is awesome, and makes a lot of sense! i'm glad you're able to achieve this with your app. just the one change,,I hope that's ok!

klippx:
> this is awesome, and makes a lot of sense! i'm glad you're able to achieve this with your app. just the one change,,I hope that's ok!

Absolutely, I don't mind a little boy scouting, looks much cleaner now overall. I think this looks good to go now.

PS:

> i'm glad you're able to achieve this with your app. 

Yea, it totally works - but... Since the plugin is handling its own state management and its own data fetching, it is a bit tricky, since I need the top level component that mounts <GraphiQL> to essentially get the confirmation from a component further down the component tree. I got it to work using event emitters and listeners, which I think is not React-kosher but still fine, since I don't want to lift state further up, that price is much higher to pay since I want my plugin to manage its own state.

- Just sharing my solution FYI, like I said I am fine with this.

klippx:
@acao Are we good to go?

acao (approved):
I think this is awesome, I would love to hear from my colleauges first as well, but if not I think we can merge soon!

klippx:
@thomasheyenbrock can you take a look at this as well? 🙏

packages/graphiql/resources/renderExample.js
```diff
@@ -48,6 +48,14 @@ function onTabChange(tabsState) {
   updateURL();
 }
 
+async function confirmCloseTab(_index) {
+  // eslint-disable-next-line no-alert
+  if (window.confirm('Are you sure you want to close this tab?')) {
+    return true;
+  }
+  return false;
+}
```
dimaMachina, line 57: this now adds confirmation on netlify preview, so it will be annoying
image

packages/graphiql/cypress/e2e/tabs.cy.ts
```diff
@@ -64,7 +82,13 @@ describe('Tabs', () => {
...
     });
 
-    // Close tab
+    // Close tab (this will get rejected)
+    cy.get('#graphiql-session-tab-1 + .graphiql-tab-close').click();
+
+    // Tab is still visible
+    cy.get('#graphiql-session-tab-1 + .graphiql-tab-close').should('exist');
+
+    // Close tab (this will get accepted)
```
dimaMachina, line 91: it is better to add independent test for confirmation, and not merge it with existent one

packages/graphiql/cypress/e2e/tabs.cy.ts
```diff
@@ -1,5 +1,23 @@
 describe('Tabs', () => {
   it('Should store editor contents when switching between tabs', () => {
```
dimaMachina, line 2:
```suggestion
  describe('confirmCloseTab()', () => {
    it('should keep tab when `Cancel` was clicked', () => {
      cy.on('window:confirm', () => false);
      cy.visit('/');

      cy.get('.graphiql-tab-add').click();

      cy.get('.graphiql-tab-button + .graphiql-tab-close').eq(1).click();

      cy.get('.graphiql-tab-button').should('have.length', 2);
    });

    it('should close tab when `OK` was clicked', () => {
      cy.on('window:confirm', () => true);
      cy.visit('/');

      cy.get('.graphiql-tab-add').click();

      cy.get('.graphiql-tab-button + .graphiql-tab-close').eq(1).click();

      cy.get('.graphiql-tab-button').should('have.length', 1);
    });
  })

  it('Should store editor contents when switching between tabs', () => {
```

add this test instead, it's much clear to understand

packages/graphiql/resources/renderExample.js
dimaMachina, line 57:
```suggestion
function confirmCloseTab(index) {
  // eslint-disable-next-line no-alert
  return confirm(`Are you sure you want to close tab with index ${index}?`);
}

const searchParams = Object.fromEntries(
  new URLSearchParams(location.search).entries(),
);
```

packages/graphiql/resources/renderExample.js
```diff
@@ -101,6 +109,7 @@ root.render(
     isHeadersEditorEnabled: true,
     shouldPersistHeaders: true,
     inputValueDeprecation: GraphQLVersion.includes('15.5') ? undefined : true,
+    confirmCloseTab,
```
dimaMachina, line 112:
```suggestion
    confirmCloseTab: searchParams.confirmCloseTab === 'true' ? confirmCloseTab : undefined,
```

.changeset/empty-lobsters-breathe.md
```diff
@@ -0,0 +1,6 @@
+---
+'graphiql': minor
+'@graphiql/react': minor
+---
+
+Allow control of closing tabs
```
dimaMachina, line 6:
```suggestion
Add new prop `confirmCloseTab` to allow control of closing tabs
```
let's be more informative

packages/graphiql/src/components/GraphiQL.tsx
dimaMachina: the reason why it was not inside `closeTab` in `ExecutionContextProvider` is that `ExecutionContextProvider` is children of `EditorContextProvider` and can be accessible from `EditorContextProvider`
image

packages/graphiql-react/src/editor/context.tsx
```diff
@@ -423,20 +441,45 @@ export function EditorContextProvider(props: EditorContextProviderProps) {
...
-        };
-        storeTabs(updated);
-        setEditorValues(updated.tabs[updated.activeTabIndex]);
-        onTabChange?.(updated);
-        return updated;
-      });
+    async index => {
+      if (await closeTabConfirmation(index)) {
+        if (index === tabState.activeTabIndex) {
+          executionContext?.stop();
```
dimaMachina, line 461: this will never be called, because we can't execution context inside the editor context, because we already use editor context inside execution context. to reproduce it simple change 
```suggestion
          executionContext!.stop();
```

and use `useExecutionContext({ notNull: true });`

.changeset/empty-lobsters-breathe.md
```diff
@@ -0,0 +1,6 @@
+---
+'graphiql': minor
+'@graphiql/react': minor
```
dimaMachina, line 3:
```suggestion
```

dimaMachina (approved):
I added e2e tests, some refactoring of moving `executionContext.stop()` can be done after moving to `zustand`

PR #3567, opened 2024-03-25
fix: Stop a tag with href of 'javascript:void 0' from navigating

### LINKED ISSUE #3565: [graphiql] Grafana Embedded GraphiQL Editor CodeMirror info tooltips links navigate away from page

### Current Behavior

When embedding a `GraphiQLInterface` inside of a Grafana query editor, any link to a part of the documentation will instead take the user to `http://localhost:3000/javascript:void%200`, rather than actually executing whatever JavaScript is supposed to execute.

Screenshot to show what I mean by "info tooltip" - the div at the bottom of the inspect window pops up and I don't really have a way to debug it because when I stop hovering, it disappears.

Screenshot from 2024-03-22 19-08-34

### Expected Behavior

The links in the tooltip should open up the documentation, rather than navigate away from the current page.

### Steps To Reproduce

I don't exactly know how to ask others to reproduce this since I have such a unique setup. I mostly would just like pointers on how to debug this myself. If anyone is willing to debug this:

1. Clone https://github.com/wildmountainfarms/wild-graphql-datasource/tree/6f2557995f59e4cb64d65427247522d753528b3b
2. `npm run dev`
3. `mage -v build:linux && npm run server`
4. Open http://localhost:3000
5. Navigate to the Star Wars Film Info dashboard
6. Edit one of the panels
7. In the GraphiQL editor, hover over one of the fields in the query to make a tooltip pop up
8. Click either the name of the field or the type of the field
9. You are then navigated to http://localhost:3000/javascript:void%200

### Environment

* GraphiQL Version: 3.1.0
* OS: Pop!_OS 22.04
* Browser: Firefox 123.0.1
* Bundler: webpack 5.86.0
* `react` Version: 18.2.0
* `graphql` Version: 16.8.1

### Anything else?

This issue is a weird one to reproduce, but I would appreciate any guidance in to why this might be happening. Embedding GraphiQL into Grafana is likely something that hasn't been done before, and Grafana itself uses a fair bit of React code, so my guess is that something is conflicting with GraphiQL in a way no one expected.

I would appreciate any guidance to get this fixed, as if you navigate away, you lose whatever progress you have made while editing the dashboard.

retrodaredevil:
I created #3565 to document a bug that I was only able to reproduce with GraphiQL embedded inside of Grafana. Some behavior or bug in Grafana was causing clicking on an `a` tag with an href of 'javascript:void 0' to navigate away from the page. Although this isn't supposed to happen when the href is 'javascript:void 0', this PR is a one-liner fix to this issue by calling `event.preventDefault()`, which prevents the default behavior of trying to navigate to `javascript:void 0`.

Closes #3565

### DIFF
--- a/.changeset/sharp-rules-hide.md
+++ b/.changeset/sharp-rules-hide.md
@@ -1,0 +1,5 @@
+---
+'codemirror-graphql': patch
+---
+
+tooltip a tag's click listener calls event.preventDefault() to stop navigating away from page
--- a/packages/codemirror-graphql/src/info.ts
+++ b/packages/codemirror-graphql/src/info.ts
@@ -299,6 +299,9 @@ function text(
       // want clicking the node to navigate anywhere.
       node.href = 'javascript:void 0'; // eslint-disable-line no-script-url
       node.addEventListener('click', (e: MouseEvent) => {
+        // Although an href of 'javascript:void 0' should never navigate away from the page,
+        //   that is not always the case: https://github.com/graphql/graphiql/issues/3565
+        e.preventDefault();
         onClick(ref, e);
       });
     } else {


acao:
makes sense to me!

acao (approved):

PR #3580, opened 2024-04-10
implement correct merging of incremental respones (@defer/@stream)

thomasheyenbrock:
Hello fellow GraphiQL friends 👋 

We recently stumbled over the fact that GraphiQL does not yet implement the latest incremental response format for defer/stream, concretely the `incremental` payload property is not handled.

This PR cleans up the merge logic and makes it work for any incremental response payload. (Big inspiration taken from `graphql-tools/utils`, but if felt overkill to add the whole package just for this one utility, and since this is the official reference implementation for a GraphQL IDE it felt appropriate to have that utility as first-class citizen in our code.)

Some before-after illustrations:

| Feature | Before | After |
| --- | --- | --- |
| Defer | CleanShot 2024-04-10 at 11 16 14 | CleanShot 2024-04-10 at 11 16 55 |
| Stream | CleanShot 2024-04-10 at 11 16 30 | CleanShot 2024-04-10 at 11 17 13 |

### DIFF
--- a/.changeset/clean-pets-dance.md
+++ b/.changeset/clean-pets-dance.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': minor
+---
+
+Implement correct merging of incremental responses (@defer/@stream)
--- a/packages/graphiql-react/src/execution.tsx
+++ b/packages/graphiql-react/src/execution.tsx
@@ -1,13 +1,17 @@
 import {
   Fetcher,
-  FetcherResultPayload,
   formatError,
   formatResult,
   isAsyncIterable,
   isObservable,
   Unsubscribable,
 } from '@graphiql/toolkit';
-import { ExecutionResult, FragmentDefinitionNode, print } from 'graphql';
+import {
+  ExecutionResult,
+  FragmentDefinitionNode,
+  GraphQLError,
+  print,
+} from 'graphql';
 import { getFragmentDependenciesForAST } from 'graphql-language-service';
 import { ReactNode, useCallback, useMemo, useRef, useState } from 'react';
 import setValue from 'set-value';
@@ -183,7 +187,7 @@ export function ExecutionContextProvider({
     });
 
     try {
-      let fullResponse: FetcherResultPayload = { data: {} };
+      const fullResponse: ExecutionResult = {};
       const handleResponse = (result: ExecutionResult) => {
         // A different query was dispatched in the meantime, so don't
         // show the results of this one.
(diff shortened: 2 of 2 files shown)


yaacovCR:
Neat!

Something I've been wondering...

The [new response format](https://github.com/graphql/defer-stream-wg/discussions/69) has each payload containing supplemental data/items, but also metadata about the incremental stream in the form of pending/completed notifications.

With the UI strategy of merging incremental data into the initial response, that information is not currently shown. It also makes it impossible to tell from the final result what actually happened; the user has to watch carefully as the process unfolds.

Might we want to model display of incremental results similar to subscriptions in which we just display the different payloads separately.

Maybe as a toggle-able separate useful option?

thomasheyenbrock:
I've been pondering about that as well. GraphiQL currently has no good interface for long-running requests that return more than one payload.

- For subscriptions you only ever see the latest payload (GraphiQL does replace the result contents each time a payload is received)
- For defer/stream you only see the latest combined result, but not the increments (like you mentioned @yaacovCR)

This would need some bigger re-architecting for how the internal state is managed as well as the UI. I see that as outside of the scope of this PR which is just intending to fix the most pressing issues with defer/stream.

yaacovCR:
Oh => I suppose I am simply mistaken about subscriptions. Sorry about that!

acao:
yes this UI change to results for subscriptions has been something we've been hoping to execute for a while. And incremental delivery with stream creates another UI case, where we want to see the final result but also a log of the requests that formed that result. We hope to expose plugins for custom tabs in results, and to allow overriding the results pane, and the other idea we had years ago was for these results plugins to register and display for specific response types automatically.

acao:
@thomasheyenbrock looks awesome! just need to fix this ts bug:

```
src/execution.tsx(364,7): error TS2357: The operand of an increment or decrement operator must be a variable or a property access.
```

and get the unit and related cypress tests passing. if you like, i can show you how to enable to e2e tests for defer and stream, if this has been merged into `graphql` by now? previously, the e2e tests for defer and stream were designed to be run against the special release tag version for years, so it's just a small tweak to cypress to get them to run for the stable `graphql` release version

jerofasion (approved):

thomasheyenbrock:
Fixed the TS and linter issues 👍 (sorry for taking so long to come back to this). Will have another look at e2e tests tomorrow. The current ones are passing but I also remember that we had some for defer/stream, so worth a look.

acao:
@thomasheyenbrock awesome! yes the e2e tags are designed to reference the `graphql` tag in package.json, and were previously run in the graphql version test matrix which i think we disabled. if defer and stream are in the stable `graphql` release we can get rid of that logic

thomasheyenbrock:
Yes, just found the same thing 👍 will merge this for now and look into testing defer/stream reliably a bit more in detail separately

PR #3590, opened 2024-04-25
perf: use cdn for fonts

neo773:
This PR makes the fonts load via Google font CDN reducing bundle CSS size bundle by 360KB causing rendering bottlenecks.


acao:
our users need a bundle that works offline unfortunately!

we have plans to seperate the theme into a seperate, optional package which would help with this!

if you need an editor optimized for high performance applications, I highly recommend checking out our cm5 and cm5 modes!

PR #3597, opened 2024-05-10
fix TS type of Tooltip.label prop

thomasheyenbrock:
The tooltip can accept any React Node as prop, no need to limit it to only `string`.


acao:
thanks @thomasheyenbrock !

PR #3601, opened 2024-05-14
WIP: incremental payloads UI

thomasheyenbrock:
How it looks like for `@defer`/`@stream`: https://twitter.com/heyenbrock/status/1790134466960093476

How it looks like for subscriptions:

https://github.com/graphql/graphiql/assets/22288634/4af491f9-838a-4b7c-aa57-d0e4dfc5a369

TODOs:
- [ ] make sure the tests pass
- [ ] fix alignment of the new section with the secondary editors
- [x] check that the layout works when there is a footer

### DIFF
--- a/packages/graphiql-react/src/editor/components/increments-editors.tsx
+++ b/packages/graphiql-react/src/editor/components/increments-editors.tsx
@@ -1,0 +1,141 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+import { ChevronDownIcon, ChevronUpIcon } from '../../icons';
+import { UnStyledButton } from '../../ui';
+import {
+  commonKeys,
+  DEFAULT_EDITOR_THEME,
+  DEFAULT_KEY_MAP,
+  importCodeMirror,
+} from '../common';
+import { useSynchronizeOption } from '../hooks';
+import { IncrementalPayload } from '../tabs';
+import { CodeMirrorEditor, CommonEditorProps } from '../types';
+
+import '../style/codemirror.css';
+import '../style/fold.css';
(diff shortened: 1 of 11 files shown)


packages/graphiql-react/src/editor/components/increments-editors.tsx
```diff
@@ -0,0 +1,141 @@
...
+) {
+  const [isOpen, setIsOpen] = useState(false);
+  const incrementEditor = useIncrementsEditor(props);
+
+  const toggleEditor = useCallback(() => setIsOpen(current => !current), []);
+
+  return (
+    <div
+      className="graphiql-increment-editor"
+      style={isOpen ? { height: '30vh' } : {}}
```
thomasheyenbrock, line 102: For the editors to show properly I needed to give the wrapper a fixed height, which is not ideal. But I don't see a better way right now other than starting to do JS magic to calculate the container height on-runtime.

packages/graphiql-react/src/editor/components/increments-editors.tsx
```diff
@@ -0,0 +1,141 @@
...
+    >
+      <UnStyledButton
+        className="graphiql-increment-editor-toggle"
+        onClick={toggleEditor}
+      >
+        {props.isInitial ? 'Initial payload' : 'Increment'} (after{' '}
+        {props.increment.timing / 1000}s)
```
thomasheyenbrock, line 109: TODO: Show in ms if the timing is <1s, otherwise show seconds

TallTed (changes requested):
trivial, for humans

ljnkns (approved):

PR #3604, opened 2024-05-15
remove unused useState import from plugin-explorer example docs

Yahkob:


acao:
good catch, thanks @Yahkob !

PR #3607, opened 2024-05-20
fix: improve handling of tabs scrolling

TPXP:
The scrollbar for tabs would usually end up being shown behind other UI elements. This change brings the following enhancements:
* reduce tabs height when the scrollbar is shown to fit the scrollbar
* use a transparent scrollbar to look nicer, especially when using dark mode

### DIFF
--- a/.changeset/tabs-enhancements.md
+++ b/.changeset/tabs-enhancements.md
@@ -1,0 +1,7 @@
+---
+"@graphiql/react": patch
+"graphiql": patch
+---
+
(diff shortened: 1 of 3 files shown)


dimaMachina (commented):
Thank you, could you add screenshots, before/after?

dimaMachina:
closed because there will be no scroll for tabs in graphiql v4 https://github.com/graphql/graphiql/pull/3644

PR #3618, opened 2024-05-29
make props optional to avoid passing empty object

Yahkob:


packages/graphiql-plugin-explorer/src/graphiql-explorer.d.ts
```diff
@@ -52,7 +52,7 @@ declare module 'graphiql-explorer' {
       buttonStyle?: CSSProperties;
       actionButtonStyle?: CSSProperties;
     } | null;
-    showAttribution: boolean;
+    showAttribution?: boolean;
```
Yahkob, line 55: I did a global search for this prop but couldn't really find anything so sadly I will admit I just made it optional without fully understanding the implications 😓

acao: i think it's fine, as these are manually maintained types for the external `graphql-explorer` flow library, I can show a PR where i convert it to ts and adopt it into the repo

dimaMachina:
ty, was fixed by me in https://github.com/graphql/graphiql/pull/3633

PR #3631, opened 2024-06-25
fix: remove schema cache invalidation

### LINKED ISSUE #3622: [lsp-server] vscode-graphql triggers large amounts of requests to my graphql server for bulk file change events in 0.11.0
## Current Behavior (if applicable)

When I use `graphql-codegen` to generate types from my schema whilst VSCode is open, `vscode-graphql` automatically sends tens or hundreds of requests to my backend server to get the introspection schema.

In this video, I have two terminals running. Top terminal is my backend server which prints a log everytime it receives a request. Bottom terminal is me just generating types with `graphql-codegen`. You will see on `0.11.0` my backend server gets spammed, but on `0.9.3` it does not.

https://github.com/graphql/graphiql/assets/3223500/758b8773-8cc6-4df9-bfa0-d3ac5911b6c4

## Desired Behavior

A change to my generated types should not cause `vscode-graphql` to need to re-request introspection information. If it does, it should certainly not need to do it more than once.

My `graphql.config.cjs`:

```js
module.exports = {
  projects: {
    default: {
      schema: ['http://localhost:8080/graphql'],
      documents: ['src/**/!(*.d).{ts,tsx}', '!src/types/graphql.ts'],
      extensions: {
        codegen: {
          hooks: {
            afterOneFileWrite: ['prettier --write --ignore-path=.prettierignore'],
          },
          config: {
            withHooks: true,
            withHOC: false,
            withComponent: false,
            avoidOptionals: {
              field: true,
              inputValue: false,
              object: true,
              defaultValue: true,
            },
            scalars: {
              Time: {
                input: 'Date | string',
                output: 'string',
              },
            },
          },
          generates: {
            'src/types/graphql.ts': {
              plugins: ['typescript', 'typescript-operations', 'typescript-react-apollo', 'fragment-matcher'],
            },
          },
        },
      },
    },
  },
};
```

acao:
the intention is to ensure that type definitions update, and that the schema cache is not completely invalidated when handling every (watched) file change. the schema cache should now only be invalidated by the lru-cache ttl itself

this should fix #3622 but requires further testing, as there could be other problems with my lru cache implementation here

### DIFF
--- a/.changeset/spotty-cameras-unite.md
+++ b/.changeset/spotty-cameras-unite.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-server': patch
+---
+
+Fix schema cache invalidation for network schemas
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -706,7 +706,6 @@ export class MessageProcessor {
       await this._invalidateCache({ uri, version }, uri, contents);
       await this._updateFragmentDefinition(uri, contents);
       await this._updateObjectTypeDefinition(uri, contents, project);
-      await this._updateSchemaIfChanged(project, uri);
       return { contents, version };
     } catch {
       return { contents: [], version: 0 };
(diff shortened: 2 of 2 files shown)


acao:
oof, and now a test fails that shouldn't... probably because lru cache introduces a timing condition

acao:
closed for #3647

PR #3634, opened 2024-06-27
use `hsl` instead `hsla` when alpha is `1`

dimaMachina:


.changeset/fair-cups-cry.md
```diff
@@ -0,0 +1,7 @@
+---
+'graphiql': patch
+'@graphiql/plugin-code-exporter': patch
+'@graphiql/react': patch
+---
+
+use `hsl` instead `hsla` when alpha is `1`
```
TallTed, line 7:
```suggestion
when alpha is `1`, use `hsl` instead of `hsla`
```

packages/graphiql-react/src/ui/markdown.css
```diff
@@ -1,6 +1,6 @@
 /**
  * We render markdown in the following places:
- * - In the hint tooltip when typing in the query editor (field description 
+ * - In the hint tooltip when typing in the query editor (field description
  *   and optionally deprecation reason).
  * - In the info tooltip when hovering over a field in the query editor
  *   (field description any optionally deprecation reason).
```
TallTed, line 6:
```suggestion
 * - In the hint tooltip when typing in the query editor (field description
 *   and, optionally, deprecation reason).
 * - In the info tooltip when hovering over a field in the query editor
 *   (field description and, optionally, deprecation reason).
```

TallTed (changes requested):

Issue #3385, opened 2023-08-02
[graphiql] custom formatting


mfulton26:
I'd like to be able to specify a custom formatter for GraphiQL (e.g. prettier or something else).

Can we make `formatQuery` an option to `graphiql-react`?

dimaMachina:
Will be possible in next major https://github.com/graphql/graphiql/releases/tag/%40graphiql%2Freact%401.0.0-alpha.4

PR #3640, opened 2024-07-02
Add support for onPrettifyQuery callback to enable customised query formatting

### LINKED ISSUE #1517: Proposal: Code Formatting, Prettier
**Currently**: We aren't using prettier, instead a lightweight custom solution that has issues, and is not configurable.

**Proposed**: 
- Use prettier for graphql, json, yaml, etc. 
- make it fully configurable by users
- make it available via command palette, right click context menu, and key commands
- allow formatting on paste, format range, etc

**WIP:**: This PR implements all of the above using monaco and a new language service: https://github.com/acao/graphiql/pull/17

i-like-robots:
This adds a new optional callback function named `onPrettifyQuery` to enable the use of customised query formatting, such as integrating Prettier. This is a very quick implementation but I'm happy to invest more time if the feature is accepted as being valuable.

When provided this callback will be invoked when the user formats the current contents of the query editor. It has no affect on the header or variable editors. The callback will receive the current value of the query editor and should return a string containing the formatted query. When the callback is not provided it will fallback to the current parse/print logic.

Intended to resolve #1517 and #3385.


dimaMachina:
@i-like-robots can you sign CLA?

dimaMachina:
superseded by https://github.com/graphql/graphiql/pull/3733

i-like-robots:
Thanks @dimaMachina 🙏

PR #3655, opened 2024-07-24
[graphiql] remove unused dependencies `graphql-language-service` and `markdown-it`

dimaMachina:
they are used in `@graphiql/react` but not in `graphiql`


acao:
@dimaMachina some of these are used here
`packages/graphiql/src/cdn.ts`

dimaMachina:
@acao tests failed due dedup of `markdown-it`, was fixed by adding `overrides` in package.json
```
[webpack-dev-server] ERROR in ../../../node_modules/markdown-it/lib/common/entities.js 6:0-59
Module not found: Error: Package path ./lib/maps/entities.json is not exported from package /Users/dmytro/Desktop/GUILD/graphiql/node_modules/entities (see exports field in /Users/dmytro/Desktop/GUILD/graphiql/node_modules/entities/package.json)
```

PR #3656, opened 2024-07-25
[vite] set `build.minify: false` for cjs/esm builds

dimaMachina:


.changeset/tidy-fans-reflect.md
```diff
@@ -0,0 +1,7 @@
+---
+"@graphiql/plugin-code-exporter": patch
+"@graphiql/plugin-explorer": patch
+"@graphiql/react": patch
+---
+
+set `build.minify: false` for cjs/esm builds since minified variable names changes every time
```
dimaMachina, line 7:
```suggestion
set `build.minify: false` for cjs/esm builds since minified variable names change every build time
```

PR #3670, opened 2024-07-31
Parse SDL definitions and extensions with optional bodies, per spec

trevor-scheer:
The online parser required a body (`{ ... }`, or `= ...` for `union`) in several places where the GraphQL spec marks the body as optional, causing otherwise-valid schema documents to fail to tokenize cleanly — especially when a body-less definition or extension was followed by another definition.

This PR aligns the parser with the spec for each construct:

| Construct | Spec |
|---|---|
| `extend schema` (no root operation defs) | [§Schema Extension](https://spec.graphql.org/draft/#sec-Schema-Extension) |
| `type` / `extend type` (no fields body) | [§Objects](https://spec.graphql.org/draft/#sec-Objects), [§Object Extensions](https://spec.graphql.org/draft/#sec-Object-Extensions) |
| `interface` / `extend interface` (no fields body) | [§Interfaces](https://spec.graphql.org/draft/#sec-Interfaces), [§Interface Extensions](https://spec.graphql.org/draft/#sec-Interface-Extensions) |
| `union` / `extend union` (no `=` member list) | [§Unions](https://spec.graphql.org/draft/#sec-Unions), [§Union Extensions](https://spec.graphql.org/draft/#sec-Union-Extensions) |
| `enum` / `extend enum` (no values body) | [§Enums](https://spec.graphql.org/draft/#sec-Enums), [§Enum Extensions](https://spec.graphql.org/draft/#sec-Enum-Extensions) |
| `input` / `extend input` (no fields body) | [§Input Objects](https://spec.graphql.org/draft/#sec-Input-Objects), [§Input Object Extensions](https://spec.graphql.org/draft/#sec-Input-Object-Extensions) |

Each construct gets its own commit for easy review.

### Out of scope

- The spec also requires extensions to have at least one of `{ implements, directives, body }` (and `extend scalar` to have at least one directive). The parser is already permissive around such constraints (e.g. zero-or-more `Directive` lists), so this PR doesn't try to enforce them — that's validator territory, not a syntax-highlighting parser.


trevor-scheer:
@acao @yaacovCR would either of you be the right person to ping to get eyes on this? I've just updated this PR from a reproduction to a fix and would love to get this reviewed. Thanks in advance!

trevor-scheer:
Hey @acao, is there anything I can do to help move this along?

.changeset/afraid-oranges-clean.md
```diff
@@ -0,0 +1,9 @@
+---
+'graphql-language-service': patch
+---
+
+fix: Correctly parse schema extensions with no root operations
+
+Previously, the parser gave schema extensions the same treatment as schema definitions. The requirements are slightly different, however, since a schema extension does not require a list of root operations according to the spec: https://spec.graphql.org/draft/#sec-Schema-Extension.
+
+The rule for parsing a schema extension is now distinct from schema definition, allowing the root operations list to be omitted.
```
TallTed, line 9:
```suggestion
The rule for parsing a schema extension is now distinct from that for a schema definition, allowing the root operations list to be omitted.
```

benjie (commented):
This looks correct to me, the tests look sensible.

However; it looks like the same issue applies to object, interface, enum and input object extensions?

Actually... for those, it looks like the `{...}` _should_ be optional in all cases (extension _and_ original definition), though they _appear_ to be required in this parser definition? It's not immediately obvious to me why `schema` doesn't follow that same pattern, e.g. if you can have

```graphql
input Foo @oneof
scalar Bar
extend input Foo {
  int: Int
  bar: Bar
}
```
why could you not also have:

```graphql
schema @blah
type Query {
  int: Int
}
extend schema {
  query: Query
}
```

Perhaps this warrants a spec tweak. That's out of scope for this PR though!

I'm unfamiliar with the parser, so I defer to someone more familiar.

benjie:
Regarding my above comment, I did raise a spec tweak to allow `schema` without operation types:

https://github.com/graphql/graphql-spec/pull/1166

However, Lee doesn't see value in this:

https://youtu.be/Lo0OhLoMBII?t=2754

If you have a strong use case for it, please let me know via that pull request!

dimaMachina (changes requested):
needs rebase

PR #3675, opened 2024-08-02
[graphiql] move `@graphiql/toolkit` to `devDependecies` because umd build is bundled with all dependencies in one file

dimaMachina:
after https://github.com/graphql/graphiql/pull/3655#issuecomment-2262788115 I realized that there is no need to have `@graphiql/toolkit` in `dependencies` too


acao:
works for me!

PR #3685, opened 2024-08-07
new tabs look for graphiql v4

dimaMachina:


acao:
@dimaMachina what is the rationale behind removing the `disableTabs` option?

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -204,14 +204,23 @@ export function useMergeQuery({ caller }: UseMergeQueryArgs = {}) {
...
   /**
    * This is only meant to be used internally in `@graphiql/react`.
    */
   caller?: Function;
+  /**
+   * Invoked when the prettify callback is invoked
+   * @param query The current value of the query editor.
+   * @returns {string} The formatted query
+   */
+  onPrettifyQuery?: (query: string) => string;
```
acao, line 217: 😎

acao: @dimaMachina along the lines of our previous discussion, i think we should discontinue exposing handlers and state via props, and expose them via hooks instead, then people can provide formatters as plugins. in this case it's simply a callback and no state, but for something like this that is more likely to be used by plugins, we should expose that instead, and provider a prettier plugin

acao, line 217:
```suggestion
  onPrettifyQuery?: async (query: string) => string;
```
and the required changes as such, as `prettier.format` returns a promise, for example

acao: we already expose the use editor hooks and users can use the refs, so a compositional `@graphiql/react` experience already exists (as of course GraphiQL has been!), so we should just provide a simple example in `examples` of a custom editor?

.changeset/rich-jobs-kick.md
```diff
@@ -0,0 +1,23 @@
+---
+'graphiql': major
+---
+
+migrate from `webpack` to `vite`
```
dimaMachina, line 5:
```suggestion
```

.changeset/good-vans-refuse.md
```diff
@@ -0,0 +1,12 @@
+---
+'@graphiql/react': major
```
dimaMachina, line 2:
```suggestion
'@graphiql/react': minor
```

PR #3346, opened 2023-07-13
`@graphiql/plugin-utils` & docs, redo #3328

acao:
see #3328 for details & discussion

update:

the canary releases from the previous PR must be being auto-scanned by a vulnerability detection system of some kind? these libraries have no dependents, and we haven't consumed any of these canaries, so I don't understand what's going on here 😆 

TODO:
- fix `umd` build, currently react jsx runtime is inlined and tries to read `process`?

PR #3702, opened 2024-08-08
[v4] generate types with `vite-plugin-dts`

dimaMachina:


acao:
if you're interested, there is a previous PR where i introduced a shared vite config plugin to use for graphiql plugins, I can look it up if you like

dimaMachina:
Do you talk about https://github.com/graphql/graphiql/pull/3346? I will take a look

I don't think it's a good approach to publish shared vite config as an npm module, but we could use it internally in monorepo to DRY in plugins / `graphiql-react` and `graphiql`

PR #3705, opened 2024-08-09
[v4] use `vite build --watch` instead of `vite` for `dev` script because we don't need development server for them

dimaMachina:


packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -333,8 +333,6 @@ export function useAutoCompleteLeafs({
   }, [getDefaultFieldNames, queryEditor, schema]);
 }
 
-export type InitialState = string | (() => string);
```
dimaMachina, line 336: unused

PR #3707, opened 2024-08-09
[v4] Remove `toolbar.additionalContent` and `toolbar.additionalComponent` props in favor of `GraphiQL.Toolbar` render props.

dimaMachina:


.changeset/weak-dancers-jog.md
```diff
@@ -0,0 +1,79 @@
...
+    {({ prettify, copy }) => (
+      <>
+        {copy /* Copy button will be first instead of default last */}
+        {/* Merge button is removed from toolbar */}
+        {prettify}
+      </>
+    )}
+  </GraphiQL.Toolbar>
+</GraphiQL>
+```
```
dimaMachina, line 79: @acao wdyt?

PR #3728, opened 2024-08-14
[v4] polish `tabs.cy` tests, remove `.graphiql-session` class

dimaMachina:
<s>I firstly did mistake and remove `#graphiql-session` id, I was thinking it nowhere used</s> but it was used by `aria-controls="graphiql-session"`

### DIFF
--- a/packages/graphiql/cypress/e2e/tabs.cy.ts
+++ b/packages/graphiql/cypress/e2e/tabs.cy.ts
@@ -2,8 +2,8 @@ describe('Tabs', () => {
   it('Should store editor contents when switching between tabs', () => {
     cy.visit('/?query=');
 
-    // Assert that no tab visible when there's only one session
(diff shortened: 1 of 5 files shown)


packages/graphiql/cypress/e2e/tabs.cy.ts
```diff
@@ -2,8 +2,8 @@ describe('Tabs', () => {
   it('Should store editor contents when switching between tabs', () => {
     cy.visit('/?query=');
 
-    // Assert that no tab visible when there's only one session
-    cy.get('#graphiql-session-tab-0').should('not.exist');
+    // Assert that tab visible when there's only one session
+    cy.get('.graphiql-tab-button').eq(0).should('exist');
```
dimaMachina, line 6: was false negative after https://github.com/graphql/graphiql/pull/3713 merged, if we add `cy.wait(1000)` right before it failed

PR #3750, opened 2024-08-24
[graphiql/toolkit] support graphql-sse

dimaMachina:


packages/graphiql-toolkit/src/create-fetcher/createFetcher.ts
```diff
@@ -40,7 +41,13 @@ export function createGraphiQLFetcher(options: CreateFetcherOptions): Fetcher {
           graphQLParams.operationName || undefined,
         )
       : false;
+
     if (isSubscription) {
+      if (options.sseUrl) {
```
dimaMachina, line 46: or maybe instead of new argument `sseUrl` check by `ws` protocol like in `graphql-helix`?
```suggestion
      if (!options.subscriptionUrl.startsWith('ws')) {
```
https://github.com/contra/graphql-helix/blob/f8e9b1acde278d3526bb43a6d573383d037e4993/packages/graphiql/src/index.tsx#L119

packages/graphiql-toolkit/src/create-fetcher/create-sse-fetcher.ts
```diff
@@ -0,0 +1,87 @@
...
+export async function createSseFetcher(opts: ClientOptions): Promise<Fetcher> {
+  const { createClient } =
+    process.env.USE_IMPORT === 'false'
+      ? (require('graphql-sse') as { createClient: typeof createClientType })
+      : await import('graphql-sse');
+
+  const sseClient = createClient({
+    retryAttempts: 0,
+    // @ts-expect-error
+    singleConnection: true, // or use false if you have an HTTP/2 server
```
derekwilling, line 20: How is this supposed to be configured by the user?

dimaMachina: Nice catch, we need to pass options in
https://github.com/graphql/graphiql/blob/2936a79bf3d8d2a7723987bdd6d268d3a627293e/packages/graphiql-toolkit/src/create-fetcher/createFetcher.ts#L50-L52

acao:
@dimaMachina can you update the docs in toolkit as well?

acao:
@dimaMachina needs a very minor rebase it seems, main release branch is now unblocked!

jaredtbates:
Just saw this PR - any chance this is still on the roadmap? We're debating if we want to use SSE or WebSockets and GraphiQL is the only dependency that still needs support for SSE for us to go that route. Appreciate the project and all the work you all put into it!

Issue #3608, opened 2024-05-22
[graphiql] Cant assign name for the tab


Kowjako:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I cant change name when new tab is added

### Expected Behavior

I want to give specific name for each newly added tab

### Steps To Reproduce

1. Go to graphiql example like this: https://graphql.github.io/swapi-graphql/?
2. Add new tab using "+" button
3. Try to change tab name

### Environment

* GraphiQL Version: 3.2.2
* OS: Windows 10
* Browser: Chrome
* Bundler:
* `react` Version: 18
* `graphql` Version: 16

peterldowns:
Running into this on [the github graphql explorer](https://docs.github.com/en/graphql/overview/explorer), would be really nice to be able to name my queries instead of seeing them all as `<untitled>`:

![image](https://github.com/graphql/graphiql/assets/824173/e7cbf4ae-39e1-40fe-98a0-d0417484a50d)

seeruk:
It's because it's based on the name of the query in that tab, you have to give a name to the query:

![image](https://github.com/graphql/graphiql/assets/2083033/0b0c04d3-cf2b-45c4-a064-91ff0f3b3dc2)

It took me a few moments to figure this out when I first tried it out too...

mfulton26:
Using the name of the first query isn't a great experience when a tab contains multiple queries in. I also sometimes want to name the tab differently than the first query or leave the query unnamed.

acao:
@mfulton26 agreed, it is not ideal, do you want to create a feature request?

mfulton26:
> @mfulton26 agreed, it is not ideal, do you want to create a feature request?

I thought I did. It got marked as a duplicate of this issue: https://github.com/graphql/graphiql/issues/3406#issuecomment-2308288268

PR #3753, opened 2024-08-24
[graphiql v4] Allow reassign tab name

dimaMachina:
fixes https://github.com/graphql/graphiql/issues/3608

TODO:
- [ ] add cypress tests
- [ ] disallow updating tab.title if it was reassigned
- [ ] move new component `input` to `@graphiql/react`

[Preview](https://deploy-preview-3753--graphiql-test.netlify.app)

### DIFF
--- a/.changeset/lovely-bobcats-lie.md
+++ b/.changeset/lovely-bobcats-lie.md
@@ -1,0 +1,5 @@
+---
+'graphiql': minor
+---
+
+allow reassign name of the tabs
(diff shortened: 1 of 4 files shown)


acao:
I see it's still WIP so maybe persistence is already planned, i think we should provide hooks for handling and setting tab name (so for example explorer could change the tab name), but i want to improve the hooks design. we already provide hooks for controlling tabs. we should provide tabs for reading and controlling tab state, including tab name. ideally we could combine some of this into something like `useTabState()`, perhaps this can be the beginning of that? this is also important for projects that need to persist the full editor state remotely via the storage API override, so either way it's ideal to persist this state and perhaps make it manageable via context

there was something else I wanted to reccomend here but i forgot

PR #3761, opened 2024-09-01
feat(graphql-language-service): Support for experimental fragment arguments

JoviDeCroock:
This is a work in progress branch for:

- Spec: https://github.com/graphql/graphql-spec/pull/1081
- Reference implementation: https://github.com/graphql/graphql-js/pull/4015

Being able to finish auto-complete and validation will depend on getting an alpha published as we need `parse` to behave well. The steps included in finishing this would be to add an auto-complete step for fragment arguments where we look up all fragments and check the available arguments so we can auto-complete.

### DIFF
--- a/packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
+++ b/packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
@@ -397,6 +397,7 @@ describe('getAutocompleteSuggestions', () => {
         },
       ]);
     });
+
     const metaArgs = [
       {
         label: '__DirectiveLocation',
@@ -409,6 +410,7 @@ describe('getAutocompleteSuggestions', () => {
           'An enum describing what kind of type a given `__Type` is.',
       },
     ];
+
     it('provides correct input type suggestions', () => {
       const result = testSuggestions(
         'query($exampleVariable: ) { ',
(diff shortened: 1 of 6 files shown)


packages/graphql-language-service/src/parser/getTypeInfo.ts
```diff
@@ -178,6 +178,9 @@ export function getTypeInfo(
               argDefs =
                 directiveDef && (directiveDef.args as GraphQLArgument[]);
               break;
+            case RuleKinds.FRAGMENT_SPREAD:
```
JoviDeCroock, line 181: Not sure whether getting the type-info is actually needed here

acao: it is required, otherwise autocompletion for existing arguments or types will not work. see how `getAutocompleteSuggestions` uses the typeInfo generated here with fragment arguments, for example. You probably don't even need to anything to the AllTypeInfo parser response, as you an re-use the empty `arguments` array. it is a streaming context parser, so typeInfo is re-built for each token

packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
```diff
@@ -423,6 +425,45 @@ describe('getAutocompleteSuggestions', () => {
...
+        new Position(0, 46),
+        externalFragments,
+      );
+
+      expect(result).toEqual([
+        {
+          label: 'someVariable',
+          insertText: 'someVariable: ',
+          command: suggestionCommand,
+          insertTextFormat: 2,
```
acao, line 462: if this test is not working yet, even with your graphql-js changes, it may be because we aren't outputting the arguments in `getTypeInfo`

packages/graphql-language-service/src/parser/getTypeInfo.ts
acao: so you are lexing properly with the changes in `Rules`, but you still need to track the `arguments` here

acao: actually, I think I'm incorrect, because once we reach the Argument tokens, we're dealing with RuleKinds.ARGUMENT already. so I think you are correct, you might not need to change anything here, but I need to look more closely to give a definite answer

PR #3682, opened 2024-08-07
upgrade to graphql v17 including support for new incremental delivery format

yaacovCR:
This PR provides support for the next iteration of the incremental delivery response format. It is based on [the proposed changes to `@graphql-tools/utils`](https://github.com/ardatan/graphql-tools/pull/6243), that library being the inspiration for the most recent changes within this repository in #3580.

I am not sure what the testing story will be for this PR as I am not too familiar with development with this repository. Feel free to point me in the right direction or to take this PR over. :)

PR #3769, opened 2024-09-08
language support for `@oneOf`

### LINKED ISSUE #3678: [graphiql] Allow passing header options to legacy WS client
Some time ago, the ability to pass headers from a tab to a websocket client was added in the following PR:
https://github.com/graphql/graphiql/pull/2719

We have a similar use case, but are forced into using a legacy WS client. Is there a way of updating the toolkit to allow the headers to passed to the legacy client in the same way?

acao:
Resolves #3678

Requires reverting https://github.com/graphql/graphiql/pull/3682

- [x] switch `devDependecies` to use `graphql@16.9.0` for full `@oneOf` support
- [ ] diagnostics support - I can't seem to get the rule to work :( - caused by #3682 
- [ ] autocompletion support - works in unit tests, but not yet in `graphiql`  because of #3682 
- [ ] hover support
- [x] fix all unit tests failures caused by adding a new input type to the test schema

### DIFF
--- a/packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
+++ b/packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
@@ -683,9 +683,14 @@ describe('getAutocompleteSuggestions', () => {
     });
 
     it('provides correct oneOf input type field suggestions', () => {
+      const args = [
+        { ...inputArgs[0], detail: 'String' },
+        inputArgs[1],
+        inputArgs[2],
+      ];
       expect(
         testSuggestions('{ oneOfInputTypeTest(oneOf: {', new Position(0, 29)),
-      ).toEqual(inputArgs);
+      ).toEqual(args);
     });
 
     it('provides no more field suggestions once a oneOf field is chosen', () => {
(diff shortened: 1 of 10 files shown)


packages/graphql-language-service/src/interface/__tests__/getAutocompleteSuggestions-test.ts
```diff
@@ -664,6 +682,37 @@ describe('getAutocompleteSuggestions', () => {
...
+        testSuggestions(
+          '{ oneOfInputTypeTest(oneOf: { value: 2 ',
+          new Position(0, 40),
+        ),
+      ).toEqual([]);
+    });
+
+    // TODO: decide if we want this. Discussing with @benjie, we might want to actually give the user flexibility here,
+    // instead of being strict
+    it('provides no more field suggestions once a oneOf field is chose and a user begins typing another field', () => {
```
TallTed, line 707:
```suggestion
    it('provides no more field suggestions once a oneOf field is chosen and a user begins typing another field', () => {
```

TallTed (changes requested):
Tiny, human-facing typo fix

PR #3772, opened 2024-09-10
fix(cdn): export exact desired version

yaacovCR:


acao (changes requested):
`main` branch should always have the latest stable `graphql` as `devDependency`

It is not necessary to use an alpha version here, as the test matrix workflow will take care of that

PR #3773, opened 2024-09-10
move back to v16

yaacovCR:


acao:
@yaacovCR just need a minor version changeset  bump for `graphiql` and `vscode-graphql`

acao:
@yaacovCR there is a test that was added that counts number of modules from vite for some reason, and this changes when dependencies change, so you just need to change the number

PR #3774, opened 2024-09-10
switch back to graphql-http

yaacovCR:
since we have rolled back to graphql@16 we are no longer testing incremental delivery and do not need to use graphql-helix for incremental delivery.

### DIFF
--- a/packages/graphiql/test/e2e-server.js
+++ b/packages/graphiql/test/e2e-server.js
@@ -9,38 +9,14 @@
 const { createServer } = require('node:http');
 const express = require('express');
 const path = require('node:path');
-const {
-  getGraphQLParameters,
-  processRequest,
-  sendResult,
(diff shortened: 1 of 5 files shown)


packages/graphiql/test/e2e-server.js
```diff
@@ -9,38 +9,14 @@
 const { createServer } = require('node:http');
 const express = require('express');
 const path = require('node:path');
-const {
-  getGraphQLParameters,
-  processRequest,
-  sendResult,
-} = require('graphql-helix'); // update when `graphql-http` is upgraded to support multipart requests for incremental delivery https://github.com/graphql/graphiql/pull/3682#discussion_r1715545279
+const { createHandler } = require('graphql-http/lib/use/express');
```
acao, line 12: the `eslint-disable` directive above just needs removal!

PR #3800, opened 2024-10-29
docs: remove path(/dev.html) from local development GraphiQL URL

hxrxchang:
When I first tried to set up the development environment for GraphiQL, I noticed that the documentation in DEVELOPMENT.md mentioned that GraphiQL is hosted at `http://localhost:8080/dev.html`
However, it was actually hosted at `http://localhost:8080`, so I updated the documentation to reflect this.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -92,7 +92,7 @@ Then, you can run these commands:
 - `yarn dev-graphiql` — which will launch `webpack` dev server for graphiql
   from the root
 
-> The GraphiQL UI is available at http://localhost:8080/dev.html
+> The GraphiQL UI is available at http://localhost:8080
 
 ### Developing Monaco GraphQL
 


acao:
@hxrxchang it looks like a dev tooling regression, as the /dev path is meant to resolve the unminified bundle for debugging

hxrxchang:
@acao 
Thanks for your review.
If you want to use dev.html as the entry point when in the development environment (i.e., when the server is started with webpack-dev-server), then it seems that the part at https://github.com/hxrxchang/graphiql/blob/fix-local-development-url/packages/graphiql/resources/webpack.config.js#L91 should be isHMR instead of !isHMR. isHMR refers to process.env.WEBPACK_SERVE, which is an environment variable that webpack-dev-server sets automatically. Refer to https://github.com/webpack/webpack-dev-server/blob/a11d81f063585a80a427d2d9dfc79f58799278cc/lib/Server.js#L241-L243.

acao:
@hxrxchang yes, it seems the regression was introduced the last time I upgraded webpack

Issue #3581, opened 2024-04-12
[graphiql] Upgrade to @graphiql/react breaks unit tests


cwoolum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We are getting the error

``` console
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
 - /local/myproject/node_modules/@graphiql/react/src/utility/context.ts:24:58
 - /local/myproject/node_modules/@graphiql/plugin-explorer/dist/index.js:1:2531
 - /local/myproject/node_modules/react-dom/cjs/react-dom.development.js:16305:18
 - /local/myproject/node_modules/react-dom/cjs/react-dom.development.js:20074:13
 - /local/myproject/node_modules/react-dom/cjs/react-dom.development.js:21587:16
 - /local/myproject/node_modules/react-dom/cjs/react-dom.development.js:4164:14
 - /local/myproject/node_modules/happy-dom/src/event/EventTarget.ts:223:42
 - /local/myproject/node_modules/happy-dom/src/nodes/element/Element.ts:1171:29
 - /local/myproject/node_modules/happy-dom/src/event/EventTarget.ts:132:9
 - /local/myproject/node_modules/happy-dom/src/nodes/element/Element.ts:1171:29
  ```

We had previously been using 

```
"@graphiql/plugin-explorer": "1.0.3",
"graphiql": "3.1.0",
```

Upon upgrading our packages, `@graphiql/plugin-explorer` upgraded to `1.0.4` and `@graphiql/react` to `0.20.4`, the above error started occurring.

### Expected Behavior

Out tests were previously been passing so I would expect them to continue with no changes

### Steps To Reproduce

Open this environment

https://stackblitz.com/edit/vitejs-vite-kkpoqd?file=vite.config.ts

Run `npx vitest`. 

### Environment

* GraphiQL Version: 3.1.0
* OS: Linux
* Browser: N/A
* Bundler: ESBuild
* `react` Version: N/A
* `graphql` Version: 3.1.0

ethan-tam33:
Hi, did you ever resolve this issue? What versions of graphiql/react, graphiql, and graphiql/plugin-explorer don't cause this issue?

ethan-tam33:
Oops never mind! Just use the versions of graphiql and graphiql/plugin-explorer above and upgrade graphiql/react to 0.22.4. That worked for me.

Issue #3509, opened 2024-01-21
[graphiql] - Upgrading to 3.1.0 causes runtime error


oferitz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

https://github.com/graphql/graphiql/issues/2763 - Already mention this issue, maybe a regression?

### Current Behavior

After upgrading `graphiql` from 3.0.5 to 3.1.0 i'm getting the following runtime error in the browser console

<img width="1188" alt="Screenshot 2024-01-21 at 14 01 54" src="https://github.com/graphql/graphiql/assets/1668687/bdeb734e-cbf6-42e6-b78d-c0ec97a8025d">

### Expected Behavior

Upgrading error should not throw error

### Environment

* GraphiQL Version: 3.1.0
* OS: macOS 14.1.2
* Browser: Chrome 120
* Bundler: Vite
* `react` Version: 18.2.0
* `graphql` Version: 16.8.1

mutefiRe:
Happens for us as well when using toolbar: additionalContent with a plugin which uses useEditorContext.

E.g. with using the plugin from this pull request: https://github.com/graphql/graphiql/pull/3055 

I think if that is expected to happen then I feel like some documentation for the toolbar and adding things there is missing.

vibdev:
If you add a no-op for the caller you can get a new error:
```
useEditorContext({
    nonNull: true,
    caller: () => {},
  });
 ```
<img width="976" alt="Screenshot 2024-07-25 at 3 27 33 PM" src="https://github.com/user-attachments/assets/878c0a74-62a9-42e1-8f8b-c6ba207216a1">

oferitz:
https://github.com/graphql/graphiql/issues/3509

ShiraDC:
We still have an issue with updating from 3.0.5 to 3.8.3.

<img width="517" alt="Image" src="https://github.com/user-attachments/assets/e07a60e1-488f-4766-9809-205520e8d743" />

Issue #2763, opened 2022-09-08
GraphQL Explorer Plugin throwing error


faizan-siddiqui:
I have added GraphQL Explorer Plugin, following the same steps in the readme here and the examples: https://github.com/graphql/graphiql/tree/main/packages/graphiql-plugin-explorer

The plugin button shows up but when i click it, i get this error:
`'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
`

Is there anything I'm missing, has anyone else come across this?

thomasheyenbrock:
Hey @faizan-siddiqui 👋 thanks for opening an issue! Can you give more context about the setup you're using? I copy-pasted the example from the readme into a Codesandbox but wasn't able to reproduce the issue: https://codesandbox.io/s/agitated-wright-m56pb2?file=/src/App.tsx

tom-quiltt:
I have the same problem and I am using Next.js.

jonathanawesome:
@tom-quiltt @faizan-siddiqui additional information about your setup or, even better, a running example would help with troubleshooting.

AndKiel:
In my remix, when I try to upgrade graphiql to v2 and use the explorer plugin, I get the following error/stack after clicking the button to open explorer:

```
useGivenContext@http://localhost:3000/build/_shared/chunk-6QWELR4S.js:675:60
ExplorerPlugin@http://localhost:3000/build/routes/explorer-GBKJ37U4.js:15383:23
renderWithHooks@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:13563:35
mountIndeterminateComponent@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:16292:21
beginWork@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:17247:22
beginWork$1@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:21048:22
performUnitOfWork@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:20496:20
workLoopSync@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:20432:30
renderRootSync@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:20411:15
recoverFromConcurrentError@http://localhost:3000/build/_shared/chunk-XVUW4YK6.js:20033:42
[...]

PR #3806, opened 2024-11-05
fix: remove deprecated Function.prototype.caller reference

simmerer:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_caller_or_arguments_usage

Screenshot 2024-11-05 at 3 08 01 PM

While it's helpful during development to have as much information as possible about which component doesn't have the necessary context, it's very not helpful to have the error message itself throw a blocking error due to this deprecated usage 😅 

In a future PR, it might be nice to rename the `caller` argument found throughout GraphiQL packages to something like `callerFn` for clarity, making it clear at a glance that it's a custom arg and not an attempt to access `Function.prototype.caller`.

### DIFF
--- a/.changeset/four-dogs-swim.md
+++ b/.changeset/four-dogs-swim.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': minor
+---
+
+Fix: removed deprecated usage of Fn.prototype.caller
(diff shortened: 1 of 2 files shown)


simmerer:
Related issues:
- https://github.com/graphql/graphiql/issues/3509
- https://github.com/graphql/graphiql/issues/3581
- https://github.com/graphql/graphiql/issues/2763

acao:
thanks @simmerer ! I am currently in the process of migrating most of this logic to a lower service level, and will be removing this usage of `caller`. this will be very helpful for the time being, thank you!

acao:
I'm going to call this good to go for now, as the full resolution will be introduced in 4.x soon! thank you!

PR #3821, opened 2024-12-11
[1/2] [@graphiql/react] compile source code with react-compiler, remove `useMemo` and `useCallback` usages

dimaMachina:


acao:
@dimaMachina any chance this can happen against the v4 branch? will it be hard to adopt these changes to the v4 branch?

TallTed:
So many commits with meaningless comments. `upd`? `aa`? `fix`? Might as well just leave the comment blank. *Much* better to put some hint about what the commit actually contains!

packages/graphiql-react/src/explorer/components/search.tsx
```diff
@@ -60,20 +50,16 @@ export function Search() {
...
-          ? { name: def.field.name, def: def.field }
-          : { name: def.type.name, def: def.type },
-      );
-    },
-    [push],
-  );
-  const isFocused = useRef(false);
-  const handleFocus: FocusEventHandler = useCallback(e => {
-    isFocused.current = e.type === 'focus';
-  }, []);
```
emmenko, line 76: Hi 👋

Was there a reason to remove this? This might be related to the following regression: https://github.com/graphql/graphiql/issues/3842

dimaMachina: Hi, it still present https://github.com/graphql/graphiql/blob/1d598a9439a0cb265a72b830e68aac71e79337f1/packages/graphiql-react/src/explorer/components/search.tsx#L60-L62

I will take a look at your issue soon!

PR #3826, opened 2024-12-14
[v3] migrate graphiql to vite and react compiler

dimaMachina:
- [x] I took my changes from GraphiQL v4 https://github.com/graphql/graphiql/pull/3685
- [x]  added `prepublishOnly` with creating "graphiql.js", "graphiql.min.js", "graphiql.min.js.map", "graphiql.css", "graphiql.min.css"


packages/graphiql/index.html
```diff
@@ -0,0 +1,45 @@
...
+    <!--vite-replace-start-->
+    <script
+      crossorigin
+      src="https://unpkg.com/react@18/umd/react.production.min.js"
+    ></script>
+    <script
+      crossorigin
+      src="https://unpkg.com/react-dom@18/umd/react-dom.production.min.js"
+    ></script>
+    <script src="/dist/index.umd.js"></script>
```
acao, line 35: this should be graphiql.js or else it will break all the unpkg implementations

packages/graphiql/test/e2e-server.js
```diff
@@ -24,8 +24,14 @@ app.use(express.json());
...
 app.get('/graphql', handler);
 
-app.use(express.static(path.resolve(__dirname, '../')));
-app.use('index.html', express.static(path.resolve(__dirname, '../dev.html')));
+// On CI we test the UMD build
+if (process.env.CI === 'true') {
+  app.use(express.static(path.join(__dirname, '..')));
+} else {
+  app.get('/', (req, res) => {
+    res.redirect('http://localhost:5173');
```
acao, line 32: this is fine as long as it's serving the unminified umd version

acao:
this looks good so far, but we still need `graphiql.min.js` and `graphiql.js` as the UMD versions or it will break all the many unpkg/jsdelivr implementations, just do a github code search and you'll see thousands. the benefits of consistent naming here are much less important than this pivotal backwards compatibility for our majority of users.

packages/graphiql/package.json
```diff
@@ -17,37 +17,39 @@
...
     "graphiql.min.js.map",
     "graphiql.css",
-    "graphiql.min.css",
-    "graphiql.min.css.map"
+    "graphiql.min.css"
   ],
+  "exports": {
+    "./package.json": "./package.json",
+    "./style.css": "./dist/style.css",
+    "./graphiql.css": "./dist/style.css",
```
acao, line 33: exports don'tt work with jsdelivr/unpkg unfortunately, we must publish to npm with these exact files

dimaMachina: I added `prepublishOnly` with exact files https://github.com/graphql/graphiql/blob/ad75cdaacd0f0043dce436d6869299af0a618094/packages/graphiql/package.json#L48

packages/graphiql/index.html
dimaMachina: what do you mean by `it will break all the unpkg implementations`?

/dist/index.umd.js === /graphql.js
/dist/index.umd.js === /graphql.min.js

packages/graphiql/test/e2e-server.js
dimaMachina: I made to use umd only on CI, and you can see that all tests pass

locally I replace umd import with https://github.com/graphql/graphiql/blob/ad75cdaacd0f0043dce436d6869299af0a618094/packages/graphiql/vite.config.mts#L102-L111

dimaMachina:
> elivr implementations, just do a github code search and you'll see thousands. the benefits of consistent naming here are much less important than this pivotal backwards compatibility for our majority of users.

see https://github.com/graphql/graphiql/pull/3826#discussion_r1885488383

dimaMachina:
> this looks good so far, but we still need `graphiql.min.js` and `graphiql.js` as the UMD versions or it will break all the many unpkg/jsdelivr implementations, just do a github code search and you'll see thousands. the benefits of consistent naming here are much less important than this pivotal backwards compatibility for our majority of users.

see https://github.com/graphql/graphiql/pull/3826#discussion_r1885488383

packages/graphiql-react/src/editor/hooks.ts
```diff
@@ -459,7 +458,7 @@ export function useOptimisticState([
       lastStateRef.current.pending === null &&
       lastStateRef.current.last !== newState
     ) {
-      // No pending updates and change has occurred... send it upstream
+      // No pending updates and change have occurred... send it upstream
```
TallTed, line 461: The active subject is the singular `change`, and the verb should match.
```suggestion
      // No pending updates and change has occurred... send it upstream
```

PR #3829, opened 2024-12-14
migrate `@graphiql/react` to vitest

dimaMachina:
brings to improvements

- [x] use the same svg loader used for build
- [x] use react-compiler for components


packages/graphiql-react/__mocks__/svg.jsx
```diff
@@ -1,7 +0,0 @@
-export default function MockedIcon(props) {
-  return (
-    <svg {...props}>
-      <title>mocked icon</title>
-    </svg>
-  );
-}
```
dimaMachina, line 7: as an improvement we no longer provide mocks for tests, but process them with same svg loader

PR #3848, opened 2025-01-06
fix(language-service): improve types, nullable member access

forivall:
In VSCode, I noticed that the editor sticky scroll headers were showing the wrong headers, and I tracked it down to the GraphQL language support. This also fixes the locations when you navigate to graphql sections using breadcrumbs or the outline panel.

I also emit outlines for all embedded graphql documents, rather than just the first one. This only works properly when the location data is correct (i tried this first, and it lead to some interesting results)

This PR also includes some additional typing and performance improvements, but I saw some mentions of an ongoing refactoring in the commit history, so I have just the targeted changes in a branch: https://github.com/graphql/graphiql/compare/main...forivall:graphiql:outline-embedded-offset-simple

I only added tests for my performance improvement, but let me know if you need more tests!

### DIFF
--- a/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
@@ -379,14 +379,12 @@ export class GraphQLLanguageService {
       }
 
       output.push({
-        // @ts-ignore
         name: tree.representativeName ?? 'Anonymous',
         kind: getKind(tree),
         location: {
           uri: filePath,
           range: {
             start: tree.startPosition,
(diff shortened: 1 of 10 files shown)


acao:
awesome work! my outline impl was originally quite half baked, thank you for fixing it!

acao:
rebasing this now - one spec file in the LSP server will fail, because of a test bug, but I know this was already passing when you first submitted it, so it should work out fine!

acao:
oof, yeah there were some upstream changes I think, let me take a closer look here within the week if you don't have time to

PR #3854, opened 2025-01-17
feat(graphiql-toolkit): accept HeadersInit

jasonkuhrt:
This PR enhances `graphiql-toolkit` package fetcher utility to accept `HeadersInit` type instead of `Record<string, string>` in places where headers are accepted.

This is a backwards compatible change because `Record<string, string>` is a sub-type of `HeadersInit`. As well, care in this PR has been taken to preserve the way in which same-name headers from different sources would overwrite, not merge, together.

The benefit of this change is that consumers can now pass values of `Headers` or `[string,string][]` type which permits the possibility of a header repeated multiple, which HTTP supports.

In practice a place this issue became relevant was in my work on Hive Console's Laboratory Preflight Script feature:

https://github.com/graphql-hive/console/pull/6378#discussion_r1919125548

### DIFF
--- a/packages/graphiql-toolkit/src/create-fetcher/lib.ts
+++ b/packages/graphiql-toolkit/src/create-fetcher/lib.ts
@@ -25,6 +25,23 @@ const errorHasCode = (err: unknown): err is { code: string } => {
 };
 
 /**
+ * Merge two Headers instances into one.
+ *
+ * Returns a new Headers instance (does not mutate).
+ *
+ * Headers are merged by having a copy of the first headers argument apply its `set`
+ * method to assign each header from the second headers argument. This means that headers
+ * from the second Headers instance overwrite same-named headers in the first.
+ */
+const mergeHeadersWithSetStrategy = (headersA: Headers, headersB: Headers) => {
+  const newHeaders = new Headers(headersA);
+  for (const [key, value] of headersB.entries()) {
+    newHeaders.set(key, value);
+  }
+  return newHeaders;
+};
+
+/**
  * Returns true if the name matches a subscription in the AST
  *
  * @param document {DocumentNode}
@@ -57,14 +74,18 @@ export const isSubscriptionWithName = (
 export const createSimpleFetcher =
   (options: CreateFetcherOptions, httpFetch: typeof fetch): Fetcher =>
   async (graphQLParams: FetcherParams, fetcherOpts?: FetcherOpts) => {
+    const headers = [
+      new Headers({
+        'content-type': 'application/json',
+      }),
+      new Headers(options.headers ?? {}),
+      new Headers(fetcherOpts?.headers ?? {}),
+    ].reduce(mergeHeadersWithSetStrategy, new Headers());
+
     const data = await httpFetch(options.url, {
       method: 'POST',
       body: JSON.stringify(graphQLParams),
-      headers: {
-        'content-type': 'application/json',
-        ...options.headers,
-        ...fetcherOpts?.headers,
-      },
+      headers,
     });
     return data.json();
   };
@@ -141,17 +162,19 @@ export const createMultipartFetcher = (
   httpFetch: typeof fetch,
 ): Fetcher =>
   async function* (graphQLParams: FetcherParams, fetcherOpts?: FetcherOpts) {
+    const headers = [
+      new Headers({
+        'content-type': 'application/json',
+        accept: 'application/json, multipart/mixed',
+      }),
+      new Headers(options.headers ?? {}),
+      new Headers(fetcherOpts?.headers ?? {}),
+    ].reduce(mergeHeadersWithSetStrategy, new Headers());
+
     const response = await httpFetch(options.url, {
       method: 'POST',
       body: JSON.stringify(graphQLParams),
-      headers: {
-        'content-type': 'application/json',
-        accept: 'application/json, multipart/mixed',
-        ...options.headers,
-        // allow user-defined headers to override
-        // the static provided headers
-        ...fetcherOpts?.headers,
-      },
+      headers,
     }).then(r =>
       meros<Extract<ExecutionResultPayload, { hasNext: boolean }>>(r, {
         multiple: true,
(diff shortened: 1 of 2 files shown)


packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -187,9 +210,21 @@ export async function getWsFetcher(
...
   if (options.subscriptionUrl) {
+    const fetcherOptsHeaders = new Headers(fetcherOpts?.headers ?? {});
+    // @ts-expect-error: Current TS Config target does not support `Headers.entries()` method.
+    // However it is reported as "widely available" and so should be fine to use. This could
+    // would be more complicated without it.
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries
+    const fetcherOptsHeadersEntries: [string, string][] = [
+      ...fetcherOptsHeaders.entries(),
+    ];
+    // todo: If there are headers with multiple values, they will be lost. Is this a problem?
```
jasonkuhrt, line 221: I am not familiar with `graphql-ws` and have not investigated what it does with these key-values yet.

Can someone advise what the correct logic would be here? I am currently apply a lossy approach, but I do not know if that is ok here.

packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -187,9 +210,21 @@ export async function getWsFetcher(
...
   }
   if (options.subscriptionUrl) {
+    const fetcherOptsHeaders = new Headers(fetcherOpts?.headers ?? {});
+    // @ts-expect-error: Current TS Config target does not support `Headers.entries()` method.
+    // However it is reported as "widely available" and so should be fine to use. This could
+    // would be more complicated without it.
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries
+    const fetcherOptsHeadersEntries: [string, string][] = [
+      ...fetcherOptsHeaders.entries(),
+    ];
```
jasonkuhrt, line 220: Ideally this isn't needed. However I don't want to open a pandora's box by changing the `tsconfig.json` file.

jasonkuhrt:
Hey @acao @dimaMachina it looks like you're the current active maintainers. Could you let me know if you'll have time to look at this PR in the coming days?

packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -57,14 +74,18 @@ export const isSubscriptionWithName = (
 export const createSimpleFetcher =
   (options: CreateFetcherOptions, httpFetch: typeof fetch): Fetcher =>
   async (graphQLParams: FetcherParams, fetcherOpts?: FetcherOpts) => {
+    const headers = [
+      new Headers({
+        'content-type': 'application/json',
+      }),
+      new Headers(options.headers ?? {}),
+      new Headers(fetcherOpts?.headers ?? {}),
+    ].reduce(mergeHeadersWithSetStrategy, new Headers());
```
dimaMachina, line 83:
```suggestion
const headers = new Headers({
  'content-type': 'application/json',
  ...options.headers,
  ...fetcherOpts?.headers,
})
```
produce the same result? you can pass multiple key-values
image

dimaMachina (commented):
is it possible to add some test to avoid regression in the future?

jasonkuhrt:
I assume so, I haven't looked at the existing test suit yet. I wanted to get a general thumbs up first. From what you see, this makes sense and could get merged?

dimaMachina:
> I assume so, I haven't looked at the existing test suit yet. I wanted to get a general thumbs up first. From what you see, this makes sense and could get merged?

in my perspective yes once you add a test which fix current behaviour

jasonkuhrt:
@dimaMachina thanks for confirming. I will proceed this or next week with adding a test and fixing anything else CI is failing on.

packages/graphiql-toolkit/src/create-fetcher/lib.ts
jasonkuhrt: The type of `fetcherOpts?.headers` is `undefined | HeadersInit`, do you agree? If so, can you see how spreading fails when the given value for `HeadersInit` is case of `Headers` for example:


```ts
> new Headers({ ...new Headers([['set-cookie','one'],['set-cookie','two']]) })
Headers {}
```

Am I missing something?

dimaMachina: sorry I missed the type of `options.headers` and `fetcherOpts.headers` was changed

jasonkuhrt: All good, should we consider this thread resolved?

dimaMachina: is following can be used instead?

```js
  const headers = new Headers(
    Object.fromEntries([
      [['content-type'], ['application/json']],
      ...(options.headers || []),
      ...(fetcherOpts.headers || [])
    ])
  )
```

jasonkuhrt: I have applied the refactor. Thanks @dimaMachina.

packages/graphiql-toolkit/src/create-fetcher/lib.ts
jasonkuhrt: I discovered that same-named headers are eagerly joined by Headers instance, so this is a non issue.

jasonkuhrt:
@dimaMachina I've added tests and changeset. Waiting for your feedback.

.changeset/early-pears-remember.md
```diff
@@ -0,0 +1,12 @@
...
+'@graphiql/react': patch
+---
+
+`graphiql-toolkit` now accepts `HeadersInit` input.
+
+`graphiql-react` has internal type changes to support this.
+
+BREAKING CHANGE:
+
+Because `graphiql-toolkit` functions now accept HeadersInit where previously a partially wider type of `Record<string, unknown>` was accepted, there is a technical backwards incompatibility. This new stricter type could for example cause your project to fail type checking after this upgrade. At runtime, nothing should change since if you weren't already using `string` typed value headers already then they were being coerced implicitly. In practice, this should only serve to marginally improve your code with trivial effort.
```
TallTed, line 12:
```suggestion
Because `graphiql-toolkit` functions now accept HeadersInit where previously a partially-wider type of `Record<string, unknown>` was accepted, there is a technical backwards incompatibility. This new stricter type could, for example, cause your project to fail type-checking after this upgrade. At runtime, nothing should change, since if you weren't already using `string` typed value headers, then they were being coerced implicitly. In practice, this should only serve to marginally improve your code, with trivial effort.
```

packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -1,3 +1,9 @@
+// todo
+// Current TS Config target does not support `Headers.entries()` method.
+// However it is reported as "widely available" and so should be fine to use.
```
TallTed, line 3:
```suggestion
// However, it is reported as "widely available", and so should be fine to use.
```

packages/graphiql-toolkit/src/create-fetcher/lib.ts
```diff
@@ -1,3 +1,9 @@
...
+// @see https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries
+// Several places we currently use ts-expect-error to allow this.
```
TallTed, line 5:
```suggestion
// We currently use ts-expect-error at several places to allow this.
```

TallTed (changes requested):
Tiny, human-facing

saihaj (approved):

jasonkuhrt:
CI is failing for some reason after recent commits here by not me. Should I get involved here again or?...

TallTed (commented):
lgtm, fwiw

Issue #3066, opened 2023-03-01
[graphql-lsp] Template string fragment get `Unknown fragment` error


VinceBT-BG:
When inserting a fragment inside a template string, the fragment does not get recognized when it gets called in the rest of the gql string

![2023-03-01 11 27 00](https://user-images.githubusercontent.com/117643868/222113242-a0cfdcfb-8c76-47e6-9686-e8aa6b0971c0.gif)

IDE: VSCode `1.75.1`

Extensions: 
- GraphQL: Language Feature Support `v0.8.5`
- GraphQL: Syntax Highlighting `v1.0.6`

avlaguta:
The related issue:
https://github.com/graphql/graphiql/issues/2064

acao:
there must have been a change that broke this use case. it works if the fragments are in a separate file, or are defined in the same graphql string, which is what the tests cover

acao:
Huh, I can't seem to reproduce the bug now, is it fixed for you in the latest version?

![fragment-validation-bug](https://github.com/graphql/graphiql/assets/1368727/8e2791fc-660a-49cf-9ac7-b789de7de27c)

VinceBT-BG:
Can you show your config and schema ? Here is my config, I am using a different root if that helps
![2023-06-26 14 22 32](https://github.com/graphql/graphiql/assets/117643868/a459898c-6d2e-49e5-a387-a073ccfad863)

VinceBT-BG:
I tried to reproduce your example and I don't have the same result, can you show your config/schema ?

![2023-06-26 15 49 31](https://github.com/graphql/graphiql/assets/117643868/0cb0f438-4f82-4369-be15-6f8468faed1e)

acao:
@VinceBT-BG here is the config;

```ts
export default {
   schema: "src/schema/schema.graphql",
   documents: "src/queries/**/*.ts",
};
```
```graphql
# for example
type ExampleType {
   shelter: String
   something: String
   however: String
}

type Query {
    exampleField: ExampleType
    love: String
    message: String
}
```

VinceBT-BG:
I was indeed using `schemaPath` instead of `schema` in the configuration! I tried on the example and it worked, I also tried in my project and I don't have the errors anymore, thanks !

acao:
@VinceBT-BG music to my ears! another happy user. I will add something to the FAQ section for users using legacy `schemaPath`

awesome gif tool by the way, i like the keyboard icons, maybe I will use this!

[...]

PR #3861, opened 2025-02-05
fix "Unknown fragment" validation error

### LINKED ISSUE #3620: [lsp-server] 🐞 VSCode validation error Unknown fragment in v0.11.0

### Current Behavior

I get errors after updating “GraphQL: Language Feature Support” to v0.11.0.

image

* `FooFragment` is defined another file.
* No errors if defined in the same file.

### Expected Behavior

No errors when using fragments from another files.

### Steps To Reproduce

1. define fragment
2. define query or fragment  in another file.
3. use fragment defined in step 1

### Environment

* LSP Server Version: latest
* OS: macOS 13.5.1
* LSP Client: VSCode v1.89.1
* graphql: v16.8.1

### Anything else?

_No response_

motemen:
fixes #3620

Above error occurs because it tries to parse already-parsed GraphQL documents as languages like TypeScript or so. This PR fixes this behavior.

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -1223,25 +1223,29 @@ export class MessageProcessor {
   private async _cacheDocumentFilesforProject(project: GraphQLProjectConfig) {
     try {
       const documents = await project.getDocuments();
-      return Promise.all(
-        documents.map(async document => {
-          if (!document.location || !document.rawSDL) {
-            return;
-          }
+      const documentLocations = new Set(
+        documents
+          .filter(doc => doc.location && doc.rawSDL)
+          .map(doc => doc.location!),
+      );
 
-          let filePath = document.location;
+      return Promise.all(
+        Array.from(documentLocations).map(async loc => {
+          let filePath = loc;
           if (!path.isAbsolute(filePath)) {
-            filePath = path.join(project.dirpath, document.location);
+            filePath = path.join(project.dirpath, loc);
           }
 
           // build full system URI path with protocol
           const uri = URI.file(filePath).toString();
 
+          const fileContent = await readFile(filePath, 'utf-8');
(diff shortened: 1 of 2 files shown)


VinceBT-BG:
@wenchienchen-asana Can you take a look at the two check errors ?

acao:
I am also getting test failures as well, it seems that there were some changes to to the integration tests since I last worked on them and they are failing for me locally and in CI here

The change makes sense, but I need to get these tests working, because these tests are already designed to test for this bug, if you see the other fragment related tests in this file!

acao:
@motemen the vscode github PR extension has things mixed up again for pushing to your branch, so can you just make this one change in `MessageProcessor`:

change `from 'node:fs'` to `from 'fs'` on the first line, and add a comment to not change the import to `node:fs` as it will break `mock-fs`.

it's annoying to make runtime changes to prevent test breakages but this integration suite is important!

acao:
no dice on that fixing CI as you can see 😭 
I hope to fix this over the weekend. hopefully it's just a node version issue. we may just replace the `mock-fs` impl with something more stable, as this should not be this hard 😆 

I would otherwise just skip the test because this bug is so important to fix, but this integration suite is very important, so I need to give this my due diligence!

acao:
I am going to go ahead and merge this with a failing test, because this bugfix is more important than the test currently, and I can confirm that the new test works and confirms the bugfix. I will take care of solving this test bug as a follow up after the next release. thanks everyone for your patience!

acao:
oddly enough, the test passed here
https://github.com/graphql/graphiql/actions/runs/13848401899/job/38751280571?pr=3881
```
PASS graphql-language-service-server packages/graphql-language-service-server/src/__tests__/MessageProcessor.spec.ts (44.561 s)
```

either way, enjoy the latest release! thanks again!

motemen:
Thank you!

VinceBT-BG:
@acao Should it fix https://github.com/graphql/graphiql/issues/3066#issuecomment-2451017684 as well ?

PR #3878, opened 2025-03-13
Upgrade wgutils; note AI summary tools

benjie:
AI summaries are now enabled... It might take us a while to actually use them, but they're enabled. We should note this at the start of each meeting.


benjie:
I don't think anything in this PR caused CI to fail?

acao:
i think this was caused by a related dev dependency update, or node version change for these tests. so they have started failing everywhere :(

PR #3882, opened 2025-03-16
fix(graphql-language-service): fix off by one on hover

bensengupta:
Hey, thanks for making this! This PR fixes a small off-by-one issue when hovering.

<details>
  <summary>Details</summary>
I am using graphql-language-service-cli in Neovim. The issue is illustrated below:

Hover, no information (incorrect, should show info)
Screenshot 2025-03-16 at 6 37 17 PM
Hover, shows information (correct)
Screenshot 2025-03-16 at 6 37 22 PM
Hover, shows information (incorrect, should **not** show info)
Screenshot 2025-03-16 at 6 37 27 PM
Hover, no information (correct)
Screenshot 2025-03-16 at 6 37 31 PM

</details>


.changeset/dry-queens-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'graphql-language-service': patch
```
bensengupta, line 2: I realize that `getContextAtPosition` is exported, this might be a minor change

bensengupta:
Seems like there is a race condition in the tests. Am unable to reproduce test failure locally.

Maybe replace this
https://github.com/graphql/graphiql/blob/b495159382465a2cab6d8592b8bce66fece066fe/packages/graphql-language-service-server/src/MessageProcessor.ts#L142-L144

with this?

```ts
    try {
      void mkdirSync(this._tmpDirBase);
    } catch (err) {
      if (err instanceof Error && 'code' in err && err.code === 'EEXIST') {
        return;
      }
      throw err;
    }

```

packages/graphql-language-service/src/parser/api.ts
```diff
@@ -176,14 +176,15 @@ export function getContextAtPosition(
   schema: GraphQLSchema,
   contextToken?: ContextToken,
   options?: { mode?: GraphQLDocumentMode; uri?: string },
+  offset = 0,
```
bensengupta, line 179: Just to explain this change: `getContextAtPosition` is used in 2 places:

In `getHoverInformation`, the offset needs to be changed to 0 to fix the off-by-one problem on hover.
https://github.com/graphql/graphiql/blob/b495159382465a2cab6d8592b8bce66fece066fe/packages/graphql-language-service/src/interface/getHoverInformation.ts#L39


In `getAutocompleteSuggestions`, the offset needs to be kept at 1 as the current behavior is correct.
https://github.com/graphql/graphiql/blob/b7e0a66af7ddc2cac30c943e55963a3e9745f376/packages/graphql-language-service/src/interface/getAutocompleteSuggestions.ts#L145-L152

johnaAr555 (approved):

trevor-scheer (approved):
Thanks for the fix!

PR #3891, opened 2025-04-25
[lsp] Fix parsing of relay style locations in locateCommand results

hayes:
Currently `locateCommand` is described as accepting "relay style" locations, which are interpreted as `uri:startLine:endLine`.

Relay actually uses `uri:line:column` which aligns with with how locations are represented in stack traces. 

The values are also being interpreted as 0 indexed, when I believe they are actually 1 index in relay and stack traces

### DIFF
--- a/packages/graphql-language-service-server/README.md
+++ b/packages/graphql-language-service-server/README.md
@@ -180,7 +180,7 @@ module.exports = {
         );
         return { uri: path, range }; // range.start.line/range.end.character/etc, base 1
         // you can also return relay LSP style
-        // return '/path/to/file.py:20:23'; // (range: 20:1 )
+        // return '/path/to/file.py:20:1'; // (range: 20:1 20:1 )
         // return '/path/to/file.py'; // (range: 1:1 1:1)
       },
     },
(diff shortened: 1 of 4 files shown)


hayes:
cc @captbaritone

captbaritone:
This looks right to me. Thanks for opening the PR @hayes!

I opened https://github.com/graphql/graphiql/pull/3895 before I saw this one. I've closed it in favor of this one which also fixed the 1 based indexing, but I'll copy some of the description from mine:

> Our intention in Relay was for the response from the locate command to be treated as path:line:column to match the convention of path strings in CLI output used by most tools. For example, paths in stack traces or errors in compiler output.
> 
> This convention is a de facto standard understood by many terminal emulators. This means if a server implements a locate command it should be useful in GraphQL LSPs (like this one or Relay's) and also at the CLI, where a user could manually invoke the command and then cmd+click the result to jump to the correct location in their editor.

https://github.com/user-attachments/assets/53eb897e-76e1-41df-b8c6-4b08599d7a95

captbaritone:
I can also confirm that `path:line:column` with 1-based indexing is what Grats returns: https://grats.capt.dev/docs/getting-started/cli/#locate

jonathanawesome (approved):

PR #3895, opened 2025-04-25
[RFC] Treat locate command response as path:line:column

captbaritone:
Our intention in Relay was for the response from the locate command to be treated as `path:line:column` to match the convention of path strings in CLI output used by most tools. For example, paths in stack traces or errors in compiler output.

This convention is a de facto standard understood by many terminal emulators. This means if a server implements a locate command it should be useful in GraphQL LSPs (like this one or Relay's) and also at the CLI, where a user could manually invoke the command and then cmd+click the result to jump to the correct location in their editor.

I'm not sure what if any tests/docs etc will need to change here, but I figured this PR could start the conversation.

https://github.com/user-attachments/assets/53eb897e-76e1-41df-b8c6-4b08599d7a95

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -910,13 +910,11 @@ export class MessageProcessor {
         project,
       });
       if (typeof locateResult === 'string') {
-        const [uri, startLine = '1', endLine = '1'] = locateResult.split(':');
+        const [uri, line = '1', column = '1'] = locateResult.split(':');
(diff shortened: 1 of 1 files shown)


hayes:
I opened a similar PR here: https://github.com/graphql/graphiql/pull/3891

I think there is still a discrepancy here between 1 and 0 indexed line numbers

captbaritone:
Lol, I'm too slow. You even cced me. I'll close this in favor of yours.

jonathanawesome:
@captbaritone are you good with closing this PR in favor of #3891?

PR #3937, opened 2025-05-09
remove `Warning: useLayoutEffect does nothing on the server, because …` warnings on SSR

dimaMachina:


.changeset/flat-beers-help.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphiql/react': patch
+'graphiql': patch
+---
+
+dd
```
dimaMachina, line 6: remove `Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format` warnings on SSR

dimaMachina, line 6:
```suggestion
remove `Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format` warnings on SSR
```

PR #3940, opened 2025-05-09
feat(@graphiql/plugin-doc-explorer): migrate React context to zustand, replace `useExplorerContext` with `useDocExplorer` and `useDocExplorerActions` hooks

dimaMachina:


dimaMachina:
> @dimaMachina I think perhaps the versioning change on this PR was incorrect. It seems this was a breaking change as you renamed most of the exported code in packages/graphiql-plugin-doc-explorer/src/index.tsx which results in consumers of the code needing to make changes.

`@graphiql/react` still use 0.x. According to Semver is valid to introduce breaking changes in 0.x releases

mjfwebb:
> > @dimaMachina I think perhaps the versioning change on this PR was incorrect. It seems this was a breaking change as you renamed most of the exported code in packages/graphiql-plugin-doc-explorer/src/index.tsx which results in consumers of the code needing to make changes.
> 
> `@graphiql/react` still use 0.x. According to Semver is valid to introduce breaking changes in 0.x releases

Right! I didn't see that :+1:

PR #3943, opened 2025-05-11
feat(@graphiql/react): migrate React context to zustand, replace `useSchemaContext` with `useSchemaStore` hook

dimaMachina:


packages/graphiql-react/src/schema.ts
```diff
@@ -15,14 +15,183 @@ import {
...
+    | 'schemaDescription'
+    | 'fetcher'
+    | 'onSchemaChange'
+  >;
+
+export const schemaStore = createStore<SchemaStore>((set, get) => ({
+  inputValueDeprecation: null!,
+  introspectionQueryName: null!,
+  schemaDescription: null!,
+  fetcher: null!, // Explicitly set to null, as it is required
```
dimaMachina, line 39:
```suggestion
  fetcher: null!, // Explicitly set to null, as it's safe since we have TypeError thrown 
```

.changeset/old-beds-juggle.md
```diff
@@ -0,0 +1,9 @@
+---
+'@graphiql/plugin-doc-explorer': patch
+'@graphiql/plugin-explorer': patch
+'@graphiql/plugin-history': patch
```
dimaMachina, line 4:
```suggestion
```

PR #3945, opened 2025-05-12
feat(@graphiql/react): migrate React context to zustand, replace `usePluginContext` with `usePluginStore` hook

dimaMachina:


packages/graphiql-react/src/schema.ts
```diff
@@ -36,7 +36,7 @@ export const schemaStore = createStore<SchemaStore>((set, get) => ({
   inputValueDeprecation: null!,
   introspectionQueryName: null!,
   schemaDescription: null!,
-  fetcher: null!, // Explicitly set to null, as it's safe since we have TypeError thrown
+  fetcher: null!, // Explicitly set to null, as it is required
```
dimaMachina, line 39:
```suggestion
  fetcher: null!, // Explicitly set to null, as it's safe since we have TypeError thrown
```

PR #3964, opened 2025-05-22
fix: vs code extension setting `graphql-config.dotEnvPath` incorrect registration

pleb:
The `graphql-config.dotEnvPath` is not being registered against the `config.properties` section in the `package.json` file. 

This causes VS Code to report the setting as `Unknown` and causes `graphql-language-service-server` to not correctly load the env file when set.

Before this fix, when attempting to set `graphql-config.dotEnvPath` in VS Code

CleanShot 2025-05-22 at 12 23 52

After this fix, when attempting to set `graphql-config.dotEnvPath` in VS Code

CleanShot 2025-05-22 at 12 25 08

### DIFF
--- a/packages/vscode-graphql/package.json
+++ b/packages/vscode-graphql/package.json
@@ -149,14 +149,14 @@
           ],
           "description": "legacy mode for graphql config v2 config",
           "default": null
+        },
+        "graphql-config.dotEnvPath": {
+          "type": [
+            "string"
+          ],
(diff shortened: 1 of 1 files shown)


dimaMachina (approved):

pleb:
@dimaMachina I made a mistake with the changeset and incorrectly specified `graphql-language-service-server` when I should have specified `vscode-graphql`.

Can you advise on the best way to fix this? I'm unsure. Sorry!

PR #3974, opened 2025-05-31
chore: sort dependencies

benjie:
This just re-orders the dependencies to be alphabetical.


dimaMachina (commented):
In The Guild we used following prettier plugin https://github.com/the-guild-org/shared-config/blob/0425a4845aac576f2d573e905288d8c820174ca2/packages/prettier-config/package.json#L26 which automate it ;)

dimaMachina (approved):

Issue #3495, opened 2024-01-11
[graphiql] defaultQuery with empty string does not result in empty default query


incompl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently if you set `defaultQuery=""` you get the commented intro text in the editor by default.

### Expected Behavior

I expect the query editor to be empty in this case.

### Steps To Reproduce

* Add defaultQuery="" to your `GraphiQL` component
* Clear localstorage so you don't get your last query shown by default and refresh

### Environment

* GraphiQL Version: 3.1.0
* OS: OSX 13.6.1
* Browser: Chrome 120.0.6099.199
* Bundler: Webpack
* `react` Version: 18.2.0
* `graphql` Version: 16.6.0

### Anything else?

This worked as expected previously. I am upgrading from 1.x to 3.x.

dimaMachina:
will be fixed in upcoming graphiql 5

PR #4020, opened 2025-06-15
graphiql 5: run cypress tests in React strict mode

dimaMachina:
closes https://github.com/graphql/graphiql/issues/3495

### DIFF
--- a/packages/graphiql/cypress/e2e/tabs.cy.ts
+++ b/packages/graphiql/cypress/e2e/tabs.cy.ts
@@ -1,6 +1,9 @@
 describe('Tabs', () => {
   it('Should store editor contents when switching between tabs', () => {
     cy.visit('?defaultQuery=');
+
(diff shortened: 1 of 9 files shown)


packages/graphiql/cypress/e2e/tabs.cy.ts
```diff
@@ -1,15 +1,13 @@
 describe('Tabs', () => {
   it('Should store editor contents when switching between tabs', () => {
-    cy.visit('/?defaultQuery=&query=');
-
-    // Assert that tab visible when there's only one session
-    cy.get('.graphiql-tab-button').eq(0).should('exist');
-
+    cy.visit('?defaultQuery=');
```
dimaMachina, line 3:
```suggestion
    cy.visit('?defaultQuery=');

    // Assert that tab visible when there's only one session
    cy.get('.graphiql-tab-button').eq(0).should('exist');
```

PR #4022, opened 2025-06-16
Export GraphiQLProps

benjie:
Not sure if this is the right fix, but it feels like we should expose GraphiQLProps?


dimaMachina:
closing in favour of https://github.com/graphql/graphiql/pull/4026

you can also use `ComponentProps<typeof GraphiQL>`

Issue #4038, opened 2025-06-24
[graphiql] examples/graphql-vite dev command doesn't work


AndKiel:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

```shell
  VITE v6.3.5  ready in 870 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
✘ [ERROR] Cannot read file: /Users/andkiel/graphiql/examples/graphiql-vite/node_modules/.pnpm/monaco-graphql@1.7.1_graphql@16.11.0_monaco-editor@0.52.2_prettier@3.6.0/node_modules/monaco-graphql/esm/graphql.worker.js?worker

    node_modules/.pnpm/@graphiql+react@0.35.0_graphql@16.11.0_react-dom@19.1.0_react@19.1.0__react@19.1.0_use-_3492b0e3ed1a9fbbf2f5ae5f6cad1844/node_modules/@graphiql/react/dist/setup-workers/vite.js:3:26:
      3 │ import GraphQLWorker from "monaco-graphql/esm/graphql.worker.js?worker";
        ╵                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] Cannot read file: /Users/andkiel/graphiql/examples/graphiql-vite/node_modules/.pnpm/monaco-editor@0.52.2/node_modules/monaco-editor/esm/vs/editor/editor.worker.js?worker

    node_modules/.pnpm/@graphiql+react@0.35.0_graphql@16.11.0_react-dom@19.1.0_react@19.1.0__react@19.1.0_use-_3492b0e3ed1a9fbbf2f5ae5f6cad1844/node_modules/@graphiql/react/dist/setup-workers/vite.js:2:25:
      2 │ import EditorWorker from "monaco-editor/esm/vs/editor/editor.worker.js?worker";
        ╵                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] Cannot read file: /Users/andkiel/graphiql/examples/graphiql-vite/node_modules/.pnpm/monaco-editor@0.52.2/node_modules/monaco-editor/esm/vs/language/json/json.worker.js?worker

[...]

PR #4042, opened 2025-06-27
fix graphiql 5 with vite example

dimaMachina:
closes https://github.com/graphql/graphiql/issues/4038

### DIFF
--- a/docs/migration/graphiql-5.0.0.md
+++ b/docs/migration/graphiql-5.0.0.md
@@ -2,10 +2,38 @@
 
 Starting from GraphiQL 5, you need to set up Monaco workers in your project:
 
-- For **Vite** projects you must import:
+- For **Vite** projects you must:
 
-  ```js
-  import 'graphiql/setup-workers/vite';
+  1. Install `vite-plugin-monaco-editor` package:
+
+   ```bash
+   npm install vite-plugin-monaco-editor --save-dev
+   ```
+
+  2. Import and configure the plugin in your `vite.config.mjs` file:
+
+  ```diff
+  // vite.config.mjs
+  import { defineConfig } from 'vite'
+  import react from '@vitejs/plugin-react'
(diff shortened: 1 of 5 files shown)


eMerzh:
thanks @dimaMachina for the update 🙌 

example is now working 🚀 
now i still have to figure out why i have
"....can't access property "jsonDefaults"..." in my bigger project replicating the thing

dimaMachina:
> thanks @dimaMachina for the update 🙌
> 
> example is now working 🚀 now i still have to figure out why i have "....can't access property "jsonDefaults"..." in my bigger project replicating the thing

Hey @eMerzh feel free to create reproduction, and I'll take a look at your issue

eMerzh:
@dimaMachina , it seems that trying to use graphiql/react in some sort breaks...
like on the vite example, i tryied 

>  yarn add @graphiql/plugin-explorer

then 
```
import { GraphiQL } from 'graphiql';
import { explorerPlugin } from '@graphiql/plugin-explorer'
import 'graphiql/style.css';
import '@graphiql/plugin-explorer/style.css'

async function fetcher(graphQLParams) {
  const response = await fetch('https://graphql.earthdata.nasa.gov/api', {
    method: 'POST',
    headers: {
      Accept: 'application/json',
    },
  });
  return response.json();
}

function App() {
  return <GraphiQL fetcher={fetcher} plugins={[explorerPlugin()]} />;
}

export default App;
```

and that is breaking as well 🤔

dimaMachina:
@eMerzh 
your issue was fixed in https://github.com/graphql/graphiql/releases/tag/graphiql%405.0.3 thanks for feedback, lmk what we can improve more !

eMerzh:
@dimaMachina would it be possible to bump deps of @graphiql to require the correct version of @graphiql/react ? so that it's an intervention free install ? 🙏

Issue #4040, opened 2025-06-25
[graphiql]  The text cursor doesn't work well in the Playground page. It doesn't point correctly to end of the line or end of the text.


557029:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have problems with GraphQL Playground page. It stopped working yesterday.I have solved this problem like that:

https://github.com/graphql/graphiql/issues/4033#issuecomment-3000102372

However, the new page has text cursor issue. The cursor doesn't go to the end of line. 

<img width="392" alt="Image" src="https://github.com/user-attachments/assets/8e38b11d-014d-4cae-bebf-5b294e9789d8" />

### Expected Behavior

- Selection should cover all text.
- when user clicks "End" button, the cursor should go to the end of line. 

<img width="239" alt="Image" src="https://github.com/user-attachments/assets/31c81c13-cfcb-4f8c-9182-e2f92423a656" />

### Steps To Reproduce

1. Open GraphQL 5.0.0 Playground Page
2. Enter and longer than 20 characters text.
3. Check the cursor location.
4. Select all text with Cmd + A,
5. Check how text will be selected.

### Environment

* GraphiQL Version:
* OS: Windows
* Browser: Chrome, Edge
* Bundler:
* `react` Version: 19
* `graphql` Version: 5.0.0

557029:
Unfortunately, I'm not able to get video.
This is what I see when enter any text in the query window and trying to go to end of the line.
the highlighted text is what was selected from the first to the last characters in the line. 

<img width="664" alt="Image" src="https://github.com/user-attachments/assets/8f29e0bc-e168-4e51-be08-3d0b0ff424ad" />

dimaMachina:
thanks, I was able to reproduce your issue, happens only on Windows, will try to find a fix

dimaMachina:
@557029, @Mackey780, @OlaSkeie and @PhilipVLam

Caret issue on Windows was fixed in graphiql@5.0.2

Let me know what we can improve more, GraphiQL 5 was a huuuge release 🙂

Mackey780:
Thank you @dimaMachina! 

One bug I've found (I'm not convinced it isn't some other environmental/browser issue so not creating GH issue) is that tab line indicators can (in some situations) drift out of position as they are slightly too narrow. 
![Image](https://github.com/user-attachments/assets/d220494e-b412-4bc1-b422-60d671db5b8f)

Notes:
- Affects Edge but not Firefox
[...]

PR #4046, opened 2025-07-01
fix incorrect caret position on Windows

dimaMachina:
closes https://github.com/graphql/graphiql/issues/4040

### DIFF
--- a/.changeset/nervous-comics-happen.md
+++ b/.changeset/nervous-comics-happen.md
@@ -3,4 +3,4 @@
 'graphiql': patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/nervous-comics-happen.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphiql/react': patch
+'graphiql': patch
+---
+
+Enable font ligatures fix incorrect caret position on Windows
```
dimaMachina, line 6:
```suggestion
Enable font ligatures in monaco-editors fix incorrect caret position on Windows
```

PR #4069, opened 2025-07-12
graphiql 5: reduce bundle size, import `prettier` dynamically to avoid bundling Prettier

dimaMachina:
we don't need prettier on initial loading, it's only used by clicking on format query button/shortcut

diff from vite example

```diff
-dist/assets/index-BMgFrxsd.js             4,911.53 kB │ gzip: 1,339.77 kB
+dist/assets/index-BlpzusGL.js             4,221.28 kB │ gzip: 1,145.58 kB
```

cc @Benjie

### DIFF
--- a/packages/graphiql-react/src/constants.ts
+++ b/packages/graphiql-react/src/constants.ts
@@ -145,6 +145,7 @@ export const MONACO_GRAPHQL_API = initializeMode({
 
 export const DEFAULT_PRETTIFY_QUERY: EditorSlice['onPrettifyQuery'] =
   async query => {
+    // We don't need to load Prettier initially; it's only used when the 'Format Query' button or shortcut is triggered
     // @ts-expect-error -- wrong types
     const { printers } = await import('prettier/plugins/graphql');
     const { parsers } = await import('prettier/parser-graphql');
--- a/packages/graphiql-react/src/constants.ts
+++ b/packages/graphiql-react/src/constants.ts
@@ -1,9 +1,5 @@
 /* eslint-disable no-bitwise */
 import { initializeMode } from 'monaco-graphql/esm/lite.js';
-// @ts-expect-error -- wrong types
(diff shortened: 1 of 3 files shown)


packages/graphiql-react/src/utility/jsonc.ts
```diff
@@ -1,14 +1,16 @@
...
-import { parsers } from 'prettier/parser-babel';
 import {
   parse as jsoncParse,
   ParseError,
   printParseErrorCode,
 } from 'jsonc-parser';
 
-export function formatJSONC(content: string) {
+export async function formatJSONC(content: string): Promise<string> {
+  // Import dynamically to avoid bundling Prettier
```
dimaMachina, line 8:
```suggestion
  // We don't need to load Prettier initially; it's only used when the 'Format Query' button or shortcut is triggered
```

packages/graphiql-react/src/constants.ts
```diff
@@ -147,13 +143,20 @@ export const MONACO_GRAPHQL_API = initializeMode({
...
-    plugins: [
-      // Fix: Couldn't find plugin for AST format "graphql"
-      { printers },
-      // @ts-expect-error -- Fix: Couldn't resolve parser "graphql"
-      { parsers },
-    ],
-  });
+export const DEFAULT_PRETTIFY_QUERY: EditorSlice['onPrettifyQuery'] =
+  async query => {
+    // @ts-expect-error -- wrong types
```
dimaMachina, line 148:
```suggestion
    // We don't need to load Prettier initially; it's only used when the 'Format Query' button or shortcut is triggered
    // @ts-expect-error -- wrong types
```

benjie:
I'm not sure that GraphiQL itself should even embed prettier? Seems easy to add in a wrapper and something that people may or may not want (and may want to customize the options of).

Happy to see the bundle size reduced though!

PR #4081, opened 2025-07-19
fix: adjust command palette `width`, `border` and remove `box-shadow`

dimaMachina:
feat: add loader for initial loading of operation editor
feat: add short cut `Cmd/Ctrl + ,` for opening GraphiQL settings dialog


Mathieu-COSYNS:
Might also have caused #4093

PR #4087, opened 2025-07-30
Fix: JS grammar detects `gql` tags in comments

trevor-scheer:
Just a repro for now. `gql` tags inside of comments are detected by the grammar and should probably be ignored. Anyone motivated to fix this: please feel free to push to this branch (if you're able) or pick my repro commit to another PR and close this one out.

Had issues running `yarn vitest` and `yarn workspace vscode-graphql-syntax test` so I added `vitest` as a dev dependency (and matched the version the repo uses). Perhaps I'm missing the "right" way to run these tests.

Also: not obvious to me why git is detecting the entire file changes, but hiding whitespace while reviewing reveals the actual diff in the snapshot correctly.

### DIFF
--- a/.changeset/slimy-readers-heal.md
+++ b/.changeset/slimy-readers-heal.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql-syntax': patch
+---
+
+Prevent JS grammar from detecting gql tags in comments
--- a/packages/vscode-graphql-syntax/tests/__fixtures__/test.js
+++ b/packages/vscode-graphql-syntax/tests/__fixtures__/test.js
@@ -101,3 +101,7 @@ const queryWithLeadingAboveComment =
       }
     }
   `;
+
+// The comment below is its own standalone test:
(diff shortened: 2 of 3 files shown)


dimaMachina (approved):

packages/vscode-graphql-syntax/package.json
```diff
@@ -154,6 +154,7 @@
   "devDependencies": {
     "@vscode/vsce": "^2.22.1-2",
     "ovsx": "^0.3.0",
+    "vitest": "^2.1.9",
```
dimaMachina:
```suggestion
```
we already have `vitest` declared in root `package.json`?

dimaMachina (approved):

packages/vscode-graphql-syntax/package.json
trevor-scheer: Yeah, I don't remember why I added this. I mentioned in the description I had some difficulty running vitest for this package but not sure what the issue was.

PR #4096, opened 2025-08-20
Use headers from previous tabs, if available

hugo-vrijswijk:
## Issue

When opening a new tab and no `defaultHeaders` is configured, headers are cleared. Rather than keeping the headers entered from previous tabs. 

Headers are often used for authorization. Having to re-enter the headers for every tab is very inconvenient.

## Demo of the bug

https://github.com/user-attachments/assets/e34e6271-9d53-4dc2-aa57-d5f7c6f695b6

## Demo of the fix

https://github.com/user-attachments/assets/d0625c68-255d-449b-9587-098efd35c74b

### DIFF
--- a/.changeset/ninety-numbers-glow.md
+++ b/.changeset/ninety-numbers-glow.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Use headers from previous tabs, if available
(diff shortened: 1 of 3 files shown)


benjie (changes requested):
I think it should copy from the **active** tab, not the first tab with headers.

Further, I think this would need to be opt-in if we want it to go out in a minor; so maybe add a prop to control it?

PR #4106, opened 2025-09-19
Offer a desktop app alternative in the docs

luciomartinez:
Leave a link to the GraphiQL.app project so users can have a quick access to library with a simple installation process


trevor-scheer:
@luciomartinez last commit was 5 years ago, does it even still work / is it current? I kinda like the idea of offering it first party, but I'm not sure we want to point people to that repo.

PR #4120, opened 2025-10-27
docs: mention turbopack and fix file extension

notrab:
For those using Next.js 16 and above, Turbopack is the default which has the same `URL()` syntax as Webpack.

Also, the link to the Next.js example `page.ts` was wrong. This renames that file to use `.tsx` instead.


docs/migration/graphiql-5.0.0.md
```diff
@@ -3,7 +3,6 @@
 Starting from GraphiQL 5, you need to set up Monaco workers in your project:
 
 - For **Vite** projects you must:
```
dimaMachina:
```suggestion
- For **Vite** projects you must:

```

dimaMachina (commented):
> Also, the link to the Next.js example page.ts was wrong. This renames that file to use .tsx instead.

yep, link was wrong, but we don't need `tsx` extension since there is no jsx tags in this file, pushed fixes in your PR, LGTM

dimaMachina (approved):

PR #4123, opened 2025-10-29
Fix error encountered running in helix

forivall:
I manually applied this fix and the lsp works now. Single character fix to add a missing `?.`

### DIFF
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -213,7 +213,7 @@ export class MessageProcessor {
     // TODO: eventually we will instantiate an instance of this per workspace,
     // so rootDir should become that workspace's rootDir
     this._settings = { ...settings, ...vscodeSettings };
-    const rootDir = this._settings?.load?.rootDir.length
+    const rootDir = this._settings?.load?.rootDir?.length
       ? this._settings?.load?.rootDir
       : this._rootPath;
     if (settings?.dotEnvPath) {
--- a/packages/graphql-language-service-server/src/MessageProcessor.ts
+++ b/packages/graphql-language-service-server/src/MessageProcessor.ts
@@ -66,7 +66,12 @@ import {
   LoaderNoResultError,
   ProjectNotFoundError,
 } from 'graphql-config';
-import type { LoadConfigOptions, LocateCommand } from './types';
+import type {
+  GraphQLConfigSettings,
+  LoadConfigOptions,
+  LocateCommand,
+  VSCodeGraphQLSettings,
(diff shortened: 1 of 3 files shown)


trevor-scheer (commented):
@forivall change seems reasonable to me, but can you share a reproduction that demonstrates the issue?

forivall:
Here's my reproduction repository: https://github.com/forivall/gql-config-repro-workspace/tree/89204bd11c78655db950237db8264e205f624d90

setup: `npm install -g graphql-language-service-cli` and [helix editor](https://github.com/helix-editor/helix). helix has `graphql-lsp` support [out of the box](https://github.com/helix-editor/helix/blob/master/languages.toml#L61). 

to reproduce, i first watch the helix logs via `tail -f ~/.cache/helix/*.log | rg --max-columns 10000 graphql-language-service`

and then run `hx -vvv test-files/1.ts`

Among the log lines, i will see

```
2026-04-17T09:52:55.949 helix_lsp::transport [INFO] graphql-language-service <- {"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"TypeError: Cannot read properties of undefined (reading 'length')"}}
```

and the error will not appear.

The expected behaviour, which is how it works when i apply my fix, is to show this error:

ScreenShot 2026-04-17 at 9 55 18 AM

(the files for the above repo were adapted from the tests for [graphql-codegen](https://github.com/dotansimha/graphql-code-generator))

packages/graphql-language-service-server/src/MessageProcessor.ts
dimaMachina: Could you improve type for settings?

packages/graphql-language-service-server/src/MessageProcessor.ts
```diff
@@ -213,7 +213,7 @@ export class MessageProcessor {
     // TODO: eventually we will instantiate an instance of this per workspace,
     // so rootDir should become that workspace's rootDir
     this._settings = { ...settings, ...vscodeSettings };
-    const rootDir = this._settings?.load?.rootDir.length
+    const rootDir = this._settings?.load?.rootDir?.length
       ? this._settings?.load?.rootDir
```
dimaMachina:
```suggestion
      ? this._settings.load.rootDir
```

packages/graphql-language-service-server/src/MessageProcessor.ts
forivall: Done.

packages/graphql-language-service-server/src/MessageProcessor.ts
forivall: Done

PR #4162, opened 2026-01-30
Fix (graphql-language-service) Hover for first line

### LINKED ISSUE #3168: [language-service] getRange does not work for line 0
Seems there is a bug [here](https://github.com/graphql/graphiql/blob/06007498880528ed75dd4d705dcbcd7c9e775939/packages/graphql-language-service/src/interface/getDiagnostics.ts#LL185C27-L185C28
):

The call comes in from _doHover(..)_ if its called on the 1st line - it comes in as line 0. So the for loop never triggers and an invariant error is thrown.

```
export function getRange(location: SourceLocation, queryText: string): IRange {
  const parser = onlineParser();
  const state = parser.startState();
  const lines = queryText.split('\n');

  invariant(
    lines.length >= location.line,
    'Query text must have more lines than where the error happened',
  );

  let stream = null;

  for (let i = 0; i < location.line; i++) { // Likely should be <=
    stream = new CharacterStream(lines[i]);
    while (!stream.eol()) {
      const style = parser.token(stream, state);
      if (style === 'invalidchar') {
        break;
      }
    }
  }

  invariant(stream, 'Expected Parser stream to be available.');

```

### LINKED ISSUE #4157: [graphql-language-service] hover broken

### Current Behavior

If you hover at the beginning of the query editor, it shows an error 
`Error: Expected Parser stream to be available`

When I first filed the issue I thought this could be fixed in the GraphQLWorker file -> doHover function by changing to return early

```
const hover = this._languageService.getHover(uri, document, graphQLPosition);
if (!hover) return;
```

but a couple days after I filed this issue I realized it also occurs if I hover later in the first line and think it may be the same issue as #3168 from a few years ago.

I think it can be fixed by changing graphql-language-service -> interface -> getDiagnostics -> getRange 

```(let i = 0; i < location.line; i++) {```
to
```for (let i = 0; i <= location.line; i++) {```

### Expected Behavior

Should not throw `Error: Expected Parser stream to be available` in the UI console.

### Steps To Reproduce

To reproduce go to the swapi or netlify demos from the graphiql docs, open the UI console, and hover at position 0.

Image

Image

### Environment

* GraphiQL Version: 5.2.2
* OS: Mac Sequoia 15.7.3 (24G419)
* Browser: Chrome
* Bundler: tried with vite and esbuild and the demos
* `react` Version: 16.12
* `graphql` Version:  18.2

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by lesleydreyer**
fix hover for line 1 (0) in graphiql -> monaco-graphiql when hovering the first line

closes #4157 closes #3168

### DIFF
--- a/packages/graphql-language-service-server/src/__tests__/GraphQLLanguageService.test.ts
+++ b/packages/graphql-language-service-server/src/__tests__/GraphQLLanguageService.test.ts
@@ -12,7 +12,7 @@ import { join } from 'node:path';
 import { GraphQLConfig } from 'graphql-config';
 import { GraphQLLanguageService } from '../GraphQLLanguageService';
 import { SymbolKind } from 'vscode-languageserver-protocol';
-import { Position } from 'graphql-language-service';
+import { getRange, Position } from 'graphql-language-service';
 import { NoopLogger } from '../Logger';
 import { GraphQLEnumType } from 'graphql';
 
@@ -207,6 +207,7 @@ describe('GraphQLLanguageService', () => {
     );
     expect(definitionQueryResult?.definitions.length).toEqual(1);
   });
+  
   it('runs definition service on fragment spread', async () => {
     const definitionQueryResult = await languageService.getDefinition(
       'fragment TestFragment on Human { name }\nquery { ...TestFragment }',
@@ -234,6 +235,23 @@ describe('GraphQLLanguageService', () => {
       './queries/definitionQuery.graphql',
     );
     expect(definitionQueryResult?.definitions.length).toEqual(1);
+  });
+  
+  it('can get range on first line', async () => {
+    const range = await getRange(
+      { line: 0, column: 15},
+        `query myHero($id: ID) {
+          hero(id: $id) {
+            name
+            id
+          }
+        }
+    `);
+    
+    expect(range).toMatchObject({
+       end: { character: 23, line: -1 }, 
+       start: { character: 22, line: -1 }
+    });
   });
 
   it('runs hover service as expected', async () => {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/graphql-language-service-server/src/__tests__/GraphQLLanguageService.test.ts [hunk 1]
```diff
@@ -236,6 +237,24 @@ describe('GraphQLLanguageService', () => {
...
+            name
+            id
+          }
+        }
+    `,
+    );
+
+    expect(range).toMatchObject({
+      end: { character: 23, line: -1 },
+      start: { character: 22, line: -1 },
```
trevor-scheer, line 254: This smells funny, I suspect we need to normalize the off by 1 line number at the LSP/Monaco boundary where the line off by 1 is happening (presumably).

I can spend some more time investigating at some point but there's probably a more comprehensive solution that results in line numbers we'd actually expect here (presumably 0).

packages/graphql-language-service/src/interface/getDiagnostics.ts
```diff
@@ -181,7 +181,12 @@ export function getRange(location: SourceLocation, queryText: string): IRange {
 
   let stream = null;
 
-  for (let i = 0; i < location.line; i++) {
+  // LSP may count from 1, but monaco counts from 0
+  for (
+    let i = 0;
+    location.line === 0 ? i <= location.line : i < location.line;
```
trevor-scheer, line 187: Related to the comment above, this doesn't feel like the correct place for this fix.

trevor-scheer (commented):
Thanks for the PR @lesleydreyer! At some point I'll have a closer look at this, but I left a couple comments in case you feel like poking around a bit more to get this right. Hopefully I'll have some more helpful comments for you in the near future.

trevor-scheer:
This should be fixed in latest versions via https://github.com/graphql/graphiql/pull/4225

Please let me know if you see otherwise!

lesleydreyer:
Sorry, been busy. Thanks for fixing!

PR #4177, opened 2026-02-26
docs: fix incorrect casing for load.filepath setting

andres-bs12:
I've updated the documentation for the VS Code extension to use the correct casing for the graphql-config.load.filepath setting. The current README uses filePath, but the extension only recognizes the lowercase version. This change ensures developers can configure the file path correctly.

### DIFF
--- a/packages/vscode-graphql/README.md
+++ b/packages/vscode-graphql/README.md
@@ -147,7 +147,7 @@ You can search a folder for any of the matching config file names listed above:
 
 ```json
 "graphql-config.load.rootDir": "./config"
-"graphql-config.envfilePath": "./config/.dev.env"
+"graphql-config.dotEnvPath": "./config/.dev.env"
 ```
 
 Or a specific filepath:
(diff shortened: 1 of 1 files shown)


packages/vscode-graphql/README.md
```diff
@@ -147,7 +147,7 @@ You can search a folder for any of the matching config file names listed above:
 
 ```json
 "graphql-config.load.rootDir": "./config"
-"graphql-config.envFilePath": "./config/.dev.env"
+"graphql-config.envfilePath": "./config/.dev.env"
```
trevor-scheer: This one doesn't actually exist, I think it should be:
```suggestion
"graphql-config.dotEnvPath": "./config/.dev.env"
```

trevor-scheer (changes requested):
Thanks for the fix! One comment, then I'm happy to merge this once I've got maintainer perms sorted.

trevor-scheer (approved):
LGTM, thank you!

dimaMachina (approved):

PR #4181, opened 2026-03-27
fix: skip introspection when `IntrospectionQuery` is passed as `schema` prop

trevor-scheer:
The schema prop accepts `GraphQLSchema | IntrospectionQuery | null`, but `shouldIntrospect` only checked for a `GraphQLSchema` instance via `isSchema`. When an `IntrospectionQuery` was passed, it was ignored and a network introspection request was still triggered.

It now detects `IntrospectionQuery` data via `__schema` check, builds a `GraphQLSchema` from it, and correctly sets `shouldIntrospect` to `false`.

### DIFF
--- a/packages/graphiql-react/src/components/provider.tsx
+++ b/packages/graphiql-react/src/components/provider.tsx
@@ -45,9 +45,7 @@ import {
 } from '../constants';
 import { getDefaultTabState } from '../utility/tabs';
 
-function isIntrospectionData(
-  value: unknown,
-): value is IntrospectionQuery {
+function isIntrospectionData(value: unknown): value is IntrospectionQuery {
   return typeof value === 'object' && value !== null && '__schema' in value;
 }
 
--- a/packages/graphiql-react/src/components/provider.tsx
+++ b/packages/graphiql-react/src/components/provider.tsx
@@ -329,12 +329,11 @@ const InnerGraphiQLProvider: FC<GraphiQLProviderProps> = ({
    * Synchronize prop changes with state
    */
   useEffect(() => {
-    const newSchema = isSchema(schema)
-      ? schema
-      : isIntrospectionData(schema)
-        ? buildClientSchema(schema)
-        : schema == null
-          ? schema
+    const newSchema =
+      isSchema(schema) || schema == null
+        ? schema
+        : isIntrospectionData(schema)
+          ? buildClientSchema(schema)
           : undefined;
 
     const validationErrors =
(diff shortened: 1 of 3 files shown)


packages/graphiql-react/src/components/provider.tsx
```diff
@@ -336,7 +348,7 @@ const InnerGraphiQLProvider: FC<GraphiQLProviderProps> = ({
        */
       requestCounter: requestCounter + 1,
       schema: newSchema,
-      shouldIntrospect: !isSchema(schema) && schema !== null,
+      shouldIntrospect: !newSchema && schema !== null,
```
dimaMachina, line 351:
```suggestion
      shouldIntrospect: !newSchema,
```

@copilot this can be simplified?

packages/graphiql-react/src/components/provider.tsx
```diff
@@ -322,7 +328,13 @@ const InnerGraphiQLProvider: FC<GraphiQLProviderProps> = ({
...
    */
   useEffect(() => {
-    const newSchema = isSchema(schema) || schema == null ? schema : undefined;
+    const newSchema = isSchema(schema)
+      ? schema
+      : isIntrospectionData(schema)
+        ? buildClientSchema(schema)
+        : schema == null
+          ? schema
+          : undefined;
```
dimaMachina:
```suggestion
    const newSchema =
      isSchema(schema) || schema == null
        ? schema
        : isIntrospectionData(schema)
          ? buildClientSchema(schema)
          : undefined;
```
less ternary - better readability

packages/graphiql/src/GraphiQL.spec.tsx
```diff
@@ -157,6 +147,26 @@ describe('GraphiQL', () => {
...
+      await act(async () => {
+        render(
+          <GraphiQL
+            fetcher={trackingFetcher}
+            schema={simpleIntrospectionData}
+          />,
+        );
+      });
+
+      expect(fetcherCalled).toBe(false);
```
dimaMachina:
```suggestion
      const trackingFetcher = vi.fn().mockResolvedValue(simpleIntrospection);

      await act(() => {
        render(
          <GraphiQL
            fetcher={trackingFetcher}
            schema={simpleIntrospectionData}
          />,
        );
      });

      expect(trackingFetcher).not.toHaveBeenCalled();
```
let's use vitest mocks!

dimaMachina (approved):
left few comments

packages/graphiql-react/src/components/provider.tsx
trevor-scheer: We need this - if user passes explicit `null` we want `shouldIntrospect === false`.

If user passes explicit `null`, `!newSchema === true` so we need the secondary check.

PR #4183, opened 2026-04-15
Fix VS Code extension publishing

trevor-scheer:
## Summary

- Fix VS Code extension publishing to OpenVSX (and VS Code Marketplace) which has been broken since the Yarn 4 upgrade
- Add `--no-dependencies` to all `vsce` and `ovsx` commands, bypassing the incompatible `yarn list --prod --json` call that Yarn 4 removed
- Update `vscode-graphql-syntax`'s stale `ovsx` version and broken publish command

This attempts to fix #4178, will validate and close after a successful publish.

## Some additional context

`@vscode/vsce` [only supports](https://github.com/microsoft/vscode-vsce?tab=readme-ov-file#requirements) Yarn `>=1 <2`. When packaging an extension, it shells out to `yarn list --prod --json` to discover runtime dependencies. This doesn't exist in Yarn Berry/v4.

The graphql repo [upgraded to Yarn 4](https://github.com/graphql/graphiql/commit/4375aa7a) in June 2025, which seems to have broken all extension packaging. The root `release` script wraps `wsrun release` in `|| true` ([`package.json:55`](https://github.com/graphql/graphiql/blob/main/package.json#L55)), so these failures are silently swallowed and the workflow reports success. We also don't have logs from GH actions that long ago, so we're flying blind as far as past runs go.

There was a prior attempt to fix publishing in [#3883](https://github.com/graphql/graphiql/pull/3883) (which removed `|| true`), but the underlying packaging failure caused the release workflow to break, leading to a rollback in [#3900](https://github.com/graphql/graphiql/pull/3900). There are a [couple errors still visible](https://github.com/graphql/graphiql/actions/runs/14684686099/job/41329136728) that align with the yarn failures I suspect.

## Why `--no-dependencies`?

`--no-dependencies` tells `vsce`/`ovsx` to skip dependency discovery entirely. This is safe because all three extensions bundle their deps:

- **`vscode-graphql`**: esbuild bundles all deps into `out/extension.js` ([`esbuild.js:14`](https://github.com/graphql/graphiql/blob/main/packages/vscode-graphql/esbuild.js#L14): `bundle: true`)
- **`vscode-graphql-execution`**: same esbuild bundling ([`esbuild.js:10`](https://github.com/graphql/graphiql/blob/main/packages/vscode-graphql-execution/esbuild.js#L10): `bundle: true`)
- **`vscode-graphql-syntax`**: ships only grammar JSON files, no runtime code or dependencies

This is the same approach recommended in [microsoft/vscode-vsce#421](https://github.com/microsoft/vscode-vsce/issues/421) for projects using non-supported package managers.

## Changes

**All three extensions** (`vscode-graphql`, `vscode-graphql-execution`, `vscode-graphql-syntax`):
- Added `--no-dependencies` to `vsce:package`, `vsce:publish`, and `open-vsx:publish` scripts

**`vscode-graphql-syntax` only:**
- Updated `ovsx` from `^0.3.0` to `0.8.3` (matching the other two extensions)
- Fixed `open-vsx:publish` command — was `ovsx publish --yarn -i . --pat $OVSX_PAT` which incorrectly passes `.` (a directory) to `--packagePath` (expects `.vsix` files)

**`vscode-graphql` and `vscode-graphql-execution`:**
- Simplified `open-vsx:publish` — removed the `$(ls -1 *.vsix | sort -V | tail -n 1)` glob since `ovsx publish` (with no positional arg) packages from the working directory directly

## Test Plan

1. From `packages/vscode-graphql`, run `npx @vscode/vsce package --yarn --no-dependencies` — should produce a `.vsix` file successfully
2. Repeat for `packages/vscode-graphql-execution` (run `yarn compile` first)
3. Repeat for `packages/vscode-graphql-syntax`
4. Verify the full release flow works in CI once merged (the `|| true` in the root release script means a failure won't block other publishes, but check the workflow logs to confirm extension publishing succeeds)

### DIFF
--- a/.changeset/breezy-cows-do.md
+++ b/.changeset/breezy-cows-do.md
@@ -1,0 +1,7 @@
+---
+'vscode-graphql-execution': patch
+'vscode-graphql-syntax': patch
+'vscode-graphql': patch
+---
+
+Fix VS Code extension publishing scripts
--- a/packages/vscode-graphql-execution/package.json
+++ b/packages/vscode-graphql-execution/package.json
@@ -85,10 +85,10 @@
     "types:check": "tsc --noEmit",
     "compile": "node esbuild.js",
     "build-bundles": "yarn run compile",
-    "vsce:package": "yarn compile && vsce package --yarn",
+    "vsce:package": "yarn compile && vsce package --yarn --no-dependencies",
     "vsce:prepublish": "yarn run vsce:package",
-    "vsce:publish": "vsce publish --yarn",
-    "open-vsx:publish": "ovsx publish $(ls -1 *.vsix | sort -V | tail -n 1) --pat $OVSX_PAT",
+    "vsce:publish": "vsce publish --yarn --no-dependencies",
+    "open-vsx:publish": "ovsx publish --no-dependencies --pat $OVSX_PAT",
     "release": "yarn run compile && yarn run vsce:publish && yarn run open-vsx:publish"
   },
   "devDependencies": {
--- a/packages/vscode-graphql-syntax/package.json
+++ b/packages/vscode-graphql-syntax/package.json
@@ -153,17 +153,17 @@
   },
   "devDependencies": {
     "@vscode/vsce": "^2.22.1-2",
-    "ovsx": "^0.3.0",
+    "ovsx": "0.8.3",
     "vscode-oniguruma": "^1.7.0",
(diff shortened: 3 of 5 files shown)


dimaMachina (approved):

dwinegar-figma:
Thank you!

PR #4186, opened 2026-04-16
Bump node to 24, pin version in CI, `.nvmrc` -> `.node-version`

trevor-scheer:
* Update node version to LTS v24
* Update `@types/node` to v24
* Use repo's node version in CI runs
* Switch `.nvmrc` to the more tool-agnostic `.node-version` (this is what `graphql-js` currently has as well)


.github/workflows/main-test.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
+          node-version-file: '.node-version'
```
dimaMachina, line 20: I think it looks by default first in `.node-version` if version was not specified explicitly ?

trevor-scheer: @dimaMachina Appears to be a standing open feature request. I've subscribed to it!
https://github.com/actions/setup-node/issues/683

dimaMachina (approved):

PR #4188, opened 2026-04-16
Overhaul `release.yml` ready for OIDC publish

benjie:
I recommend walking this PR commit by commit as each of the changes are small and focused.


.github/workflows/release.yml
```diff
@@ -2,35 +2,48 @@ name: Release
 
 on:
   push:
-    branches: [main, graphiql-5]
+    branches:
+      - main
+      - graphiql-5
```
dimaMachina:
```suggestion
```
Can be removed

.github/workflows/release.yml
```diff
@@ -6,6 +6,10 @@ on:
       - main
       - graphiql-5
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  #cancel-in-progress: true
```
trevor-scheer: I'd say enable (or remove otherwise), but support enabling it

trevor-scheer (approved):

.github/workflows/release.yml
benjie: The previous job had a cancel in progress job, which is why I added it, but I'd be really worried about a release cancelling mid-release: 10 modules uploaded, 12 to go or whatever. So I'm on team "remove it" but figured it warranted discussion. It also indicates that it was known to exist and chosen to not be enabled, though I guess setting it as `: false` would achieve that better!

trevor-scheer: False with a comment? 👍

I don't see the downside. Another small upside is I can push consecutive commits and see the exact outcomes, kinda nice for a bugfix PR that starts with a failing test.

.github/workflows/release.yml
```diff
@@ -2,36 +2,48 @@ name: Release
...
 jobs:
   release:
-    permissions:
-      contents: write # for changesets/action to git push
-    environment: deploy
     name: Release
     runs-on: ubuntu-latest
+
+    environment: deploy
+
```
benjie:
```suggestion

```

.github/workflows/release.yml
```diff
@@ -2,36 +2,45 @@ name: Release
 
 on:
   push:
-    branches: [main, graphiql-5]
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  #cancel-in-progress: true
```
benjie:
```suggestion
  # Don't cancel a publish partway through!
  cancel-in-progress: false
```

.github/workflows/release.yml
```diff
@@ -2,36 +2,46 @@ name: Release
...
+  cancel-in-progress: false
+
 permissions: {}
+
 jobs:
   release:
-    permissions:
-      contents: write # for changesets/action to git push
-    environment: deploy
     name: Release
```
benjie, line 18: Per npm:

> The name of the [GitHub Actions environment](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) used for publishing. This is encouraged for projects with maintainers who should not have npm publishing access.

We don't have such maintainers currently, but let's future-proof ourselves. I'll add it back:

```suggestion
    environment: deploy
    name: Release
```

PR #4196, opened 2026-04-18
Drop `mock-fs` to unblock final `vitest` migration

trevor-scheer:
## Summary

`mock-fs` is the main blocker for migrating `graphql-language-service-server` to vitest (the last package still on Jest). It patches Node's `fs` at the C++ binding level, which is fundamentally incompatible with vitest's worker-based architecture. It's also the reason for the `// do not change to node:fs import, or it will break mock-fs` comment in `MessageProcessor.ts` and the gnarly hack in `MockProject` that preloads 17 real `node_modules` directories into the virtual filesystem so `cosmiconfig` can resolve its dynamic `require()` calls.

This PR replaces all of that with real tmpdirs via `fs.mkdtempSync`. Each `MockProject` instance gets a unique directory under `os.tmpdir()`, writes real files, and cleans up after itself. This matches how the LSP server actually behaves in production — it already writes generated schemas to `os.tmpdir()`.

## Changes

- Rewrite `MockProject` to use real tmpdirs with a `dispose()` method for cleanup
- Wire `dispose()` into `afterEach` hooks across both `MessageProcessor` test files
- Replace hardcoded `/tmp/test-default` assertions with dynamic `project.rootDir`-based values
- Drop `mock-fs` and `@types/mock-fs` from devDependencies

### DIFF
--- a/packages/graphql-language-service-server/package.json
+++ b/packages/graphql-language-service-server/package.json
@@ -68,11 +68,9 @@
   "devDependencies": {
     "@types/glob": "^8.1.0",
     "@types/mkdirp": "^1.0.1",
-    "@types/mock-fs": "^4.13.4",
     "@whatwg-node/fetch": "npm:^0.9.0",
     "cross-env": "^7.0.2",
     "debounce-promise": "^3.1.2",
-    "graphql": "^16.9.0",
-    "mock-fs": "^5.2.0"
+    "graphql": "^16.9.0"
   }
 }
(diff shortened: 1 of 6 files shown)


dimaMachina (approved):
Lgtm, ty for this refactoring 🙌

I remember I had an issue https://github.com/vitest-dev/vitest/issues/7082

It looks like it no longer exists? Feel free to merge 🚀

trevor-scheer:
@dimaMachina is that a separate issue with vitest migration? This + the other draft PR seem to be working ok, so maybe something's changed in vitest or changing the mocking strategy for fs and fetch work around the issue? Still seems unrelated to the IPC / messaging stuff, so if there's a concern there maybe you can help me understand a bit better.

PR #4197, opened 2026-04-18
Complete `jest` -> `vitest` migration

trevor-scheer:
## Summary

Migrates the last Jest package to vitest and removes all Jest infrastructure from the monorepo. The mechanical `jest.*` → `vi.*` port was straightforward, but `fetch-mock` couldn't follow — vitest's mock system doesn't intercept CJS `require()` calls from `node_modules`, so `jest.mock('@whatwg-node/fetch')` silently stops working. Replaced `fetch-mock` with MSW which intercepts at the network layer.

## Changes

- Port all 10 test files from `jest.*` to `vi.*` (imports, mocks, timers, type helpers)
- Add `vitest.config.mts` with `clearMocks`, `graphql` dedup alias, and `graphql-language-service` src alias for unbundled workspace resolution
- Replace `fetch-mock` v6 with `msw` v2 for HTTP mocking in `GraphQLCache.test` and `MessageProcessor.spec`
- Remove `@whatwg-node/fetch`, `cross-env` from server devDeps (only existed for the old mock setup)
- Remove `jest`, `ts-jest`, `@types/jest`, `@types/fetch-mock`, `fetch-mock`, `babel-plugin-macros`, `eslint-plugin-jest` from root
- Delete `jest.config.js`, `jest.config.base.js`, trim `babel.config.js` `env.test` block
- Simplify root scripts: `test` → `turbo run test`, `test:ci` → `yarn test`
- Replace `eslint-plugin-jest` with `eslint-plugin-vitest` for test lint coverage, including `vitest/no-disabled-tests`
- Switch `@testing-library/jest-dom` imports to the `/vitest` entrypoint and drop `@testing-library/jest-dom` from tsconfig `types` arrays
- Replace jest VS Code launch config with vitest equivalent (scoped to `packages/` via `--dir`)
- Fix `DEVELOPMENT.md` watch command: `yarn test -- --watch`
- Remove stale `!jest.config.js` exclusion from `codemirror-graphql` and `jest-haste-map` exclusion from `monaco-graphql` tsconfig

### `@ts-expect-error` on prettier subpath imports

Several files import prettier subpaths like `prettier/parser-graphql`. With `moduleResolution: "node"`, TypeScript can't resolve these because it doesn't read the `exports` field in `package.json`. Previously this worked because `@types/prettier@2.6.4` (a transitive dep of `jest-snapshot`) provided the type declarations. With jest removed, those types are gone. The proper fix is migrating to `moduleResolution: "node16"` or `"bundler"`, but that's out of scope here.

### `jsdom` added as an explicit dependency

Six packages set `environment: 'jsdom'` in their vitest configs. Vitest lists jsdom as an optional peer dep, so it must be installed explicitly. It was previously available as a transitive dependency of jest (`jest-environment-jsdom` → `jsdom`).

## Note

The `Jest Unit & Integration Tests` check will need to be marked as not required in order for this to pass once it's ready to merge.


package.json
```diff
@@ -58,12 +57,9 @@
     "repo:fix": "manypkg fix",
     "repo:resolve": "node scripts/set-resolution.js",
     "t": "yarn test",
-    "test": "yarn jest",
-    "test:ci": "yarn build && jest --coverage && yarn vitest",
-    "test:coverage": "yarn jest --coverage",
-    "test:watch": "yarn jest --watch",
+    "test": "turbo run test --filter='./packages/*'",
+    "test:ci": "yarn tsc && yarn test",
```
dimaMachina:
```suggestion
    "test:ci": "yarn test",
```
do we really need run this script before tests?

dimaMachina (approved):
amazing improvement!

trevor-scheer:
@dimaMachina there was quite a bit more cleanup since your last review. Should also be a lot easier to review by commit now!

package.json
```diff
@@ -60,12 +59,9 @@
     "repo:fix": "manypkg fix",
     "repo:resolve": "node scripts/set-resolution.mts",
     "t": "yarn test",
-    "test": "yarn jest",
-    "test:ci": "yarn build && jest --coverage && yarn vitest",
-    "test:coverage": "yarn jest --coverage",
-    "test:watch": "yarn jest --watch",
+    "test": "turbo run test --filter='./packages/*'",
+    "test:ci": "yarn test",
```
dimaMachina:
```suggestion
```
Can be just removed, only don't forget to rename github workflows

dimaMachina (approved):
Excellent migration

Just double check if there are no test name patterns that end on `-spec` or `-test` (I remember we had in the past), and if so, include them in vitest `include` option

PR #4199, opened 2026-04-18
Send spec-compliant Accept header in fetcher

### LINKED ISSUE #4189: [graphiql] Send `Accept: application/graphql-response+json, application/json;q=0.9`
Currently, [graphiql sends `Accept: application/json, multipart/mixed`. 
](https://github.com/graphql/graphiql/blob/5f2c3a45906c332e07200b0008d0e5bf329a8ffa/packages/graphiql-toolkit/src/create-fetcher/lib.ts#L160)
For compatibility with [GraphQL over HTTP](https://graphql.github.io/graphql-over-http/draft/#sel-GALHJNABAB5Br8e) and newer servers, could it send `Accept: application/graphql-response+json, application/json;q=0.9` instead?

trevor-scheer:
## Summary

- Send `application/graphql-response+json` as the preferred content type per the [GraphQL over HTTP spec](https://graphql.github.io/graphql-over-http/draft/#sel-GALHJNABAB5Br8e), falling back to `application/json;q=0.9`
- Applied to `createSimpleFetcher`, which previously sent no `Accept` header
- `createMultipartFetcher` is left unchanged — the correct `Accept` header for `multipart/mixed` responses is [still under discussion](https://github.com/graphql/graphql-over-http/issues/167)

Closes #4189

## Changes

- `createSimpleFetcher`: added `accept: 'application/graphql-response+json, application/json;q=0.9'`
- Both fetchers continue to allow user-provided headers (`options.headers`, `fetcherOpts.headers`) to override the default
- Added `acceptHeaders.spec.ts` with tests covering default headers and override behavior

## Validation Steps

1. `yarn dev:graphiql` to start the dev server
2. Open GraphiQL in the browser and run a query (e.g. `{ __typename }`)
3. In DevTools Network tab, inspect the fetch request to the GraphQL endpoint — the `Accept` request header should be `application/graphql-response+json, application/json;q=0.9`

### DIFF
--- a/packages/graphiql-toolkit/src/create-fetcher/__tests__/acceptHeaders.spec.ts
+++ b/packages/graphiql-toolkit/src/create-fetcher/__tests__/acceptHeaders.spec.ts
@@ -2,6 +2,8 @@ import { createSimpleFetcher, createMultipartFetcher } from '../lib';
 
 const SPEC_ACCEPT = 'application/graphql-response+json, application/json;q=0.9';
 const MULTIPART_ACCEPT = `${SPEC_ACCEPT}, multipart/mixed`;
+const QUERY = '{ __typename }';
+const BASE_URL = 'http://localhost';
 
 function mockFetch() {
   return vi.fn().mockResolvedValue({
@@ -13,8 +15,8 @@ function mockFetch() {
 describe('createSimpleFetcher', () => {
   it('sends spec-compliant accept header', async () => {
     const fetch = mockFetch();
-    const fetcher = createSimpleFetcher({ url: 'http://localhost' }, fetch);
-    await fetcher({ query: '{ __typename }' }, {});
+    const fetcher = createSimpleFetcher({ url: BASE_URL }, fetch);
+    await fetcher({ query: QUERY }, {});
 
     expect(fetch.mock.calls[0][1].headers.accept).toBe(SPEC_ACCEPT);
   });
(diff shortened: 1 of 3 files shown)


TallTed:
The currently specified `Accept:` header —
```
application/graphql-response+json, application/json;q=0.9, multipart/mixed
```
— persists an error. Leaving off the quality rating causes a media type to be treates as if `1.0` (maximum) quality, and that's what the old `Accept:` header had, and what the new header persists.

I'm fairly well certain that one of `application/graphql-response+json` or `multipart/mixed` is really preferred over the other, but I'm not sure which that is.

For sake of understanding, I offer two options. The first prefers `application/graphql-response+json` and the second prefers `multipart/mixed`.
```
application/graphql-response+json, application/json;q=0.9, multipart/mixed;q=0.95

application/graphql-response+json;q=0.95, application/json;q=0.9, multipart/mixed
```

Note — both of my offered options presume that `application/json` is the least-preferred. I hope that what I've written is clear enough that any reader of this comment can figure out how to make `application/json` more preferred than either of the other two.

packages/graphiql-toolkit/src/create-fetcher/__tests__/acceptHeaders.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { createSimpleFetcher, createMultipartFetcher } from '../lib';
+
+const SPEC_ACCEPT = 'application/graphql-response+json, application/json;q=0.9';
+const MULTIPART_ACCEPT = `${SPEC_ACCEPT}, multipart/mixed`;
+
```
dimaMachina, line 7:
```suggestion
const QUERY = '{ __typename }';
const BASE_URL = 'http://localhost';
```
I don't want to be annoying, but these vars are repeated 5 times each 😄

dimaMachina (approved):

trevor-scheer:
@TallTed I agree with what you're saying and I'm also unsure what the preference should be. For now I've backed out the multipart changes, we'll wait to see if this is ever prescribed by the spec. Some relevant conversation here: https://github.com/graphql/graphql-over-http/issues/167

Thanks for the review!

PR #4201, opened 2026-04-18
remove deprecated --frozen-lockfile from CI

trevor-scheer:
Flag is deprecated and redundant per the [migration guide](https://yarnpkg.com/migration/guide#migration-steps).


dimaMachina (approved):
Nice catch

PR #4202, opened 2026-04-18
Remove `install` job, use `setup-node`'s yarn cache

trevor-scheer:
## Context

We're already using the yarn cache in `setup-node`, so `yarn install` is cheap. Caching `node_modules` is extra work and possibly even slower.

## Changes

Remove `install` step that manually caches `node_modules`. Just run `yarn install` in jobs where it's needed.

### DIFF
--- a/.github/workflows/main-test.yml
+++ b/.github/workflows/main-test.yml
@@ -3,52 +3,22 @@ name: Main Tests
 on:
   push:
     branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  install:
-    name: Install
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Cancel Previous Runs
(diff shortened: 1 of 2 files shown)


.github/workflows/main-test.yml
```diff
@@ -3,52 +3,22 @@ name: Main Tests
...
-          path: |
-            **/node_modules
-          key: modules-${{ github.sha }}
-      - uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cache/Cypress
-          key: cypress-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
-      - run: yarn install --frozen-lockfile --immutable
+concurrency:
```
trevor-scheer, line 7: Replaces functionality of the removed `styfle/cancel-workflow-action`

dimaMachina: Nice! It’s out of box feature now 🎉

dimaMachina (approved):

PR #4205, opened 2026-04-19
Loosen example dependency versions

trevor-scheer:
## Context

Use ^major for >=1.0 deps and ^minor for 0.x deps consistently to reduce dependabot noise while keeping examples current for people referencing them.

`resolutions` shouldn't be necessary anymore.


dimaMachina (approved):
Needs rerun yarn install

PR #4207, opened 2026-04-20
Add `tsgo` for local development and informational CI

trevor-scheer:
Adds `@typescript/native-preview` (tsgo) alongside the existing `tsc` setup. The idea is to use tsgo locally for faster type-checking during development, while keeping `tsc` as the required CI check for now. The tsgo CI job runs but is allowed to fail (`continue-on-error: true`) so we can track compatibility without blocking PRs.

## Changes

- Install `@typescript/native-preview` and add a `check:tsgo` script
- Add an informational tsgo job to PR CI (non-blocking)
- Recommend the VS Code native-preview extension and configure `tsdk` to point at it

Once tsgo reaches parity for this repo, we'll cut over to it as the source of truth.


.github/workflows/pr.yml
dimaMachina, line 104: Should we remove this?

trevor-scheer: I kinda want to run both for a bit side by side at least in CI. Local dev is all tsgo (at least that's my plan).

package.json
```diff
@@ -36,6 +36,7 @@
     "watch-vscode": "yarn tsc && yarn workspace vscode-graphql compile",
     "watch-vscode-exec": "yarn workspace vscode-graphql-execution compile",
     "check": "yarn tsc --noEmit",
```
dimaMachina:
```suggestion
    "check": "yarn tsc",
```

I found an issue in this command `--noEmit` is attached to end of `tsc` script `"tsc": "tsc --build && node resources/patch-monaco-editor-type.mjs",`

dimaMachina (approved):

package.json
trevor-scheer: Oh good catch. I agree with your change, but seems like the real issue is having a script called `tsc` is kinda confusing. I renamed the script so `yarn tsc ...` does what you'd expect.
[83a6d74](https://github.com/graphql/graphiql/pull/4207/commits/83a6d74be8170dfd1f0cb0b14466a465b912e254)

dimaMachina: great 💪

trevor-scheer: I misunderstood originally, `yarn tsc` didn't behave how I expected. Fixed separately.
https://github.com/graphql/graphiql/pull/4214

dimaMachina (approved):

PR #4211, opened 2026-04-22
 fix: add `*.css` to `sideEffects` in all CSS-exporting graphiql packages

davidjb:
When attempting to set up `graphiql` in my Webpack 5 environment, and follow the example code https://github.com/graphql/graphiql/tree/main/packages/graphiql#using-as-package, I noticed that the line:

```js
import 'graphiql/style.css';
```

was having no effect. The line was being processed, but Webpack was not emitting CSS. The resulting page looked like this, a working UI, but no CSS styles:

image

The cause is that the `graphiql` package has a `sideEffects` property in its `package.json`, which is set up to allow side effects but only for `dist/setup-workers/*`. Because `graphiql/style.css` (aka `dist/style.css`) falls outside of this path, Webpack ignores the imported CSS.

This is explained at https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free, down the bottom of this section:

> Note that any imported file is subject to tree shaking. This means if you use something like css-loader in your project and import a CSS file, it needs to be added to the side effect list so it will not be unintentionally dropped in production mode:
>
> ```js
> {
>  "name": "your-project",
>  "sideEffects": ["./src/some-side-effectful-file.js", "*.css"]
> }
> ```

This PR follows that directive and adds the required `sideEffects` entry to all packages - especially `graphiql` - that have CSS exports in their `package.json` files. Testing locally by patching this in to my build pipeline sees the JS `import` line now working.

----

Ahead of this PR being accepted, a Webpack configuration can manually override `sideEffects` like so:

```js
const config = {
  module: {
    rules: [
      ...,
      {
        test: /graphiql/,
        sideEffects: true,
      }
    ]
  }
}
```

That forces `sideEffects` to be on for all files, but in Webpack, it's only able to be a Boolean.

Another workaround is to use Sass/SCSS or another CSS preprocessor and `@use 'graphiql/style.css';` instead -- using another tool bypasses Webpack's checking of `sideEffects`. This is actually how I found the issue in the first place -- importing within Sass worked fine, but when using `import` in JS, nothing loaded.

### DIFF
--- a/.changeset/quick-cars-bathe.md
+++ b/.changeset/quick-cars-bathe.md
@@ -1,0 +1,10 @@
+---
+"@graphiql/plugin-code-exporter": minor
+"@graphiql/plugin-doc-explorer": minor
+"@graphiql/plugin-explorer": minor
+"@graphiql/plugin-history": minor
+"@graphiql/react": minor
+"graphiql": minor
+---
+
+Add \*.css to sideEffects to allow import of CSS in Webpack Javascript
--- a/.changeset/quick-cars-bathe.md
+++ b/.changeset/quick-cars-bathe.md
@@ -1,10 +1,10 @@
 ---
-"@graphiql/plugin-code-exporter": minor
-"@graphiql/plugin-doc-explorer": minor
-"@graphiql/plugin-explorer": minor
-"@graphiql/plugin-history": minor
-"@graphiql/react": minor
-"graphiql": minor
+'@graphiql/plugin-code-exporter': patch
+'@graphiql/plugin-doc-explorer': patch
+'@graphiql/plugin-explorer': patch
+'@graphiql/plugin-history': patch
+'@graphiql/react': patch
+'graphiql': patch
 ---
 
 Add \*.css to sideEffects to allow import of CSS in Webpack Javascript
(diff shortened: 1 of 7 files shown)


.changeset/quick-cars-bathe.md
```diff
@@ -0,0 +1,10 @@
+---
+"@graphiql/plugin-code-exporter": minor
+"@graphiql/plugin-doc-explorer": minor
+"@graphiql/plugin-explorer": minor
+"@graphiql/plugin-history": minor
+"@graphiql/react": minor
+"graphiql": minor
```
trevor-scheer: Seems like a patch/bugfix to me, @dimaMachina ?

trevor-scheer (approved):
This looks great and thanks for the fantastic PR description. Going to validate the change on my side when I get a chance but LGTM!

.changeset/quick-cars-bathe.md
davidjb: I've just updated these `patch` (and also formatted it to pass CI); I'd intended this originally but the `yarn changeset` CLI puzzled me on first use. 

Anyway, having CSS output where it's intended to exist but isn't currently working feels like a fix, but please let me know if this needs further adjusting.

dimaMachina (approved):
Lgtm, strange that nobody reported it before, looks like it's very strict in webpack

Issue #3951, opened 2025-05-15
Playground feature parity


martinbonnin:
This is a follow up from https://github.com/graphql/graphql-wg/issues/1648. The current situation of having 2 different IDEs is confusing. Playground hasn't had much activity lately and it'd be grand to consolidate everything in GraphiQL.

That being said, as a Playground user, there are several aspects that I like better there and I'm curious what the team thoughts are about them. 

The full context is in https://github.com/graphql/graphql-wg/issues/1648 but the most important ones are:

* [ ] Copy as curl
* [ ] SDL/JSON download
* [ ] Editable endpoint url
* [ ] Multi column schema doc
* [ ] Search the docs
* [ ] Maybe schema polling? (I would have expected this to be available but didn't find it)
* [ ] Settings editor

This is all from a cursory look of comparing the sample index.html ([playground](https://github.com/graphql/graphql-playground/blob/main/packages/graphql-playground-html/minimal.html), [graphiql](https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html)). It may very well be that GraphiQL has all those features but that they require more configuration, it which case, some better defaults would be very welcome.

Any thoughts?

jonathanawesome:
- Controls 
  - Copy as curl
    - We should be able to get something done here quickly if a Toolbar button is acceptable. I suppose we don't have _all_ of the header information from the fetcher, but we could work around that.
- Schema  
  - SDL/JSON download
    - We don't have a screen that displays the schema (like Playground) so this control would be out-of-context unless we build additional UI around it. Still, should be _as possible_ as the `copy as curl` request.
  - Editable endpoint url
    - This doesn't quite fit into the current layout. We'd have to wrap the entire app, as is, with another set of tabs that connect to different endpoints. Genuinely curious how often you use this feature in Playground. Is it primarily for dev/prod/staging comparisons, or are you connected to unrelated endpoints in the same Playground session?
  - Multi column schema doc
    - As above, this won't fit well into the current layout very well. In Playground, the docs are _above_ the IDE such that they're allowed to expand when necessary. In GraphiQL they're limited to the left column. Sure, there's a handle for changing the size, but we wouldn't want to do that automatically like Playground does.
  - Maybe schema polling? (I would have expected this to be available but didn't find it)
[...]

PR #4220, opened 2026-05-01
feat: Add url details and copy to cURL button

jerelmiller:
Part of https://github.com/graphql/graphiql/issues/3951

Adds a new URL bar that shows the GraphQL server URL along with a "Copy cURL" button.

Since this is my first PR, design suggestions welcome! I've added review comments to point out various details about this PR.

**Screenshot**

Screenshot 2026-05-07 at 12 04 41 AM


packages/graphiql-toolkit/src/create-fetcher/createFetcher.ts
```diff
@@ -56,4 +57,6 @@ export function createGraphiQLFetcher(options: CreateFetcherOptions): Fetcher {
     }
     return httpFetcher(graphQLParams, fetcherOpts);
   };
+
+  return Object.assign(fetcher, { [FETCHER_OPTIONS_SYMBOL]: options });
```
jerelmiller, line 61: We don't have a great way to access options provided to the fetcher since all the `GraphiQL` component sees is a function. I went with a symbol approach which lets us read those values in the component without the need for additional props.

That said, are there situations where this function isn't used? I hide the new UI when we can't detect these options on the fetcher function, but I'm wondering if we need to have some sort of prop on `GraphiQL` itself as replacement in case custom implementations provide fetchers that don't use this function.

I'm open to other ideas for sharing this information, but this was the best I could think of without introducing a breaking change.

packages/graphiql-react/src/components/url-details/index.css
```diff
@@ -0,0 +1,26 @@
+.url-details {
```
jerelmiller, line 1: Design suggestions welcome!

packages/graphiql/src/GraphiQL.tsx
```diff
@@ -343,6 +355,57 @@ export const GraphiQLInterface: FC<GraphiQLInterfaceProps> = ({
...
+  };
+
+  const handleCurlClick: ButtonHandler = () => {
+    const url = getFetcherUrl();
+    const query = queryEditor?.getValue();
+    const variables = safeParse<Record<string, unknown>>(
+      variableEditor?.getValue(),
+    );
+
+    const headers: Record<string, string> = {
```
jerelmiller: GraphQL Playground adds a [whole lot more](https://github.com/graphql/graphql-playground/blob/5d737c25b993d7b5be835a84a2f4286b922cbee6/packages/graphql-playground-react/src/components/Playground/TopBar/TopBar.tsx#L135-L144) header values than I've included here, but I chose to keep this minimal for now. Let me know if we should include more.

packages/graphiql/src/GraphiQL.tsx
```diff
@@ -468,61 +533,69 @@ export const GraphiQLInterface: FC<GraphiQLInterfaceProps> = ({
...
-                  </Tab>
-                ))}
-              </Tabs>
-              <Tooltip label={LABEL.newTab}>
-                <UnStyledButton
-                  type="button"
-                  className="graphiql-tab-add"
-                  onClick={addTab}
-                  aria-label={LABEL.newTab}
+          <div className="graphiql-main-container">
```
jerelmiller, line 474: The "Hide whitespace" setting is recommended while reviewing this component.

packages/graphiql-toolkit/src/create-fetcher/createFetcher.ts
trevor-scheer: This seems reasonable to me. I would say maybe just leave as-is, and if there's a good place to document the limitation that'd be good.

jerelmiller: Great! That should be enough for me to write a changeset and add additional documentation. Thanks!

trevor-scheer:
Not a strong opinion here (and I think your placement is reasonable given playground's), but I'd imagined this as a button alongside the others in the editor (near prettify, etc.). Any preference between the two options?

jerelmiller:
I thought about that too, though I struggled to figure out a good icon to use that would convey what it does, so I took the easy way out 😅. I didn't want it confused with the copy query button. Any ideas on what icon makes sense I moved it with the others?

packages/graphiql/src/GraphiQL.tsx
```diff
@@ -30,16 +30,20 @@ import {
...
   useMonaco,
+  Button,
 } from '@graphiql/react';
 import { HistoryStore, HISTORY_PLUGIN } from '@graphiql/plugin-history';
 import {
   DocExplorerStore,
   DOC_EXPLORER_PLUGIN,
 } from '@graphiql/plugin-doc-explorer';
 import { GraphiQLLogo, GraphiQLToolbar, GraphiQLFooter, Sidebar } from './ui';
+import { FETCHER_OPTIONS_SYMBOL } from '@graphiql/toolkit';
```
trevor-scheer: Build is failing because toolkit is listed as a dev dependency which is actually pretty surprising to me. I'm curious if that was an intentional decision. Maybe @dimaMachina knows more.

trevor-scheer:
@jerelmiller how about a terminal icon?
https://www.svgrepo.com/svg/511163/terminal

jerelmiller:
That works for me! I'll try and push an update tonight to try moving that button over. Thanks for the suggestion!

trevor-scheer:
FYI this'll need a rebase against `main` due to #4203 and making the new/renamed `Lint` check required.

jerelmiller:
@trevor-scheer got those updates in place! I've added some new comments as followup questions as well that I'd love some input on. Thanks for the reviews thus far!

packages/graphiql-react/src/stores/editor.ts
```diff
@@ -146,6 +152,11 @@ export interface EditorSlice extends TabsState {
    * @remarks from graphiql 5
    */
   operations?: OperationFacts['operations'];
+
+  /**
+   * The GraphQL server url
+   */
+  url?: string;
```
jerelmiller, line 159: Should this be called `fetcherUrl`, or is `url` enough?

packages/graphiql/src/ui/toolbar.tsx
```diff
@@ -60,5 +71,15 @@ export const GraphiQLToolbar: FC<{
     </ToolbarButton>
   );
 
-  return children({ prettify, copy, merge });
+  const copyAsCurl = (
+    <ToolbarButton
+      label="Copy as cURL"
+      disabled={!url}
```
jerelmiller, line 77: I opted to disable the button when the `url` is not available. Should we hide the component altogether instead?

packages/graphiql-react/src/stores/editor.ts
```diff
@@ -532,9 +551,62 @@ export const createEditorSlice: CreateEditorSlice = initial => (set, get) => {
...
+        operationName = definition?.name?.value;
+      }
+
+      const data = JSON.stringify({
+        query,
+        variables,
+        operationName,
+      });
+
+      const command = `curl '${url}' ${headersString} --data-binary '${data}'`;
```
trevor-scheer, line 590: I think we need to escape single quotes here. I'd extract this bit of the function:
`toCurl(url, headers, data): string` so we can easily unit test it.

packages/graphiql-react/src/stores/editor.ts
```diff
@@ -532,9 +551,62 @@ export const createEditorSlice: CreateEditorSlice = initial => (set, get) => {
...
+        .map(key => {
+          const value = headers[key];
+          return `-H '${key}: ${value}'`;
+        })
+        .join(' ');
+
+      let operationName: string | undefined;
+
+      if (query) {
+        const definition = parse(query).definitions.find(
```
trevor-scheer, line 577: This can throw, and I think we have a couple options...
1. on query update enable/disable the copy as curl button
2. show something in the UI
3. just log to console
4. do nothing

Thoughts? 1 seems nice but maybe a bit heavy.

packages/graphiql/src/ui/toolbar.tsx
trevor-scheer: I like disabling. It'd be nice to explain why in the tooltip but might be a bit verbose, so I'm on the fence.

I'd say leave as is (and if you come up with some very concise copy for the disabled state, open to that).

packages/graphiql-react/src/stores/editor.ts
trevor-scheer: No strong feelings, both seem reasonable. I haven't thought of another relevant URL that this needs to be disambiguated from, but I also haven't thought too hard about it :smiling:

trevor-scheer (changes requested):
Looks pretty good, a couple important comments to address within. Still needs a changeset as well. Thanks!

PR #4225, opened 2026-05-06
Fix `monaco-graphql` hover crashing on the first line

### LINKED ISSUE #3168: [language-service] getRange does not work for line 0

### LINKED ISSUE #4157: [graphql-language-service] hover broken

trevor-scheer:
Hover is broken in `monaco-graphql`. On the first line it throws and renders nothing; on later lines it doesn't crash but highlights the wrong token.

`GraphQLWorker.doHover` passes a 0-indexed position to `getRange`, which actually wants a 1-indexed `SourceLocation` (per the GraphQL spec, and what its other callers pass via `error.locations[0]`). On line 0 the loop never runs and `Expected Parser stream to be available` fires. On later lines `getRange` parses to end-of-line and ignores `column`, so the range comes back pointing at the previous line's last token.

`getRange` was the wrong tool here anyway. It was built for diagnostics, not for finding the token under a cursor. `getTokenAtPosition` is what we want, and it's what the rest of the language service already uses.

Fixes #4157
Fixes #3168

Supersedes #4162, which tried to special-case `line === 0` inside `getRange` itself. That left `getRange` returning a `Position` with `line: -1` (which becomes `0` after `toMonacoRange`'s `+1`, invalid in Monaco's 1-indexed API), and didn't address the bigger problem that `getRange` doesn't compute a token range to begin with.

## Validation

```bash
yarn install
yarn build
yarn workspace example-monaco-graphql-react-vite dev
```

Open the URL Vite prints, with browser devtools open.

1. Hover the word `query` on the first line. Before: console logs `Error: Expected Parser stream to be available` and no popup appears. After: popup renders and `query` is highlighted.
2. Press Enter to push the query down a line, then hover a field name. Before: popup appears but the highlight is on the previous line's last token. After: the field under the cursor is highlighted.

### DIFF
--- a/.changeset/fix-monaco-hover-line-zero.md
+++ b/.changeset/fix-monaco-hover-line-zero.md
@@ -1,0 +1,7 @@
+---
+'monaco-graphql': patch
+---
+
+Fix hover crashing on the first line of a query
+
+`GraphQLWorker.doHover` was passing 0-indexed positions to `getRange`, which expects a 1-indexed `SourceLocation` (per the GraphQL spec). On the first line this caused `Expected Parser stream to be available` to be logged and hover to return `null`. On other lines it returned the range of the previous line's last token rather than the token under the cursor. Use `getTokenAtPosition` to compute the actual token range instead.
--- a/packages/monaco-graphql/__tests__/GraphQLWorker.test.ts
+++ b/packages/monaco-graphql/__tests__/GraphQLWorker.test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect } from 'vitest';
+import { buildSchema } from 'graphql';
+import { GraphQLWorker } from '../src/GraphQLWorker';
+
+const schema = buildSchema(`
+  type Query { hero: Hero }
+  type Hero { name: String }
+`);
+
+function createWorker(document: string) {
+  const uri = 'inmemory://model/1';
+  const ctx = {
+    getMirrorModels: () => [
+      {
+        uri: { toString: () => uri },
+        getValue: () => document,
+      },
+    ],
+  } as any;
+  const worker = new GraphQLWorker(ctx, {
+    languageId: 'graphql',
+    languageConfig: {
+      schemas: [{ uri: 'schema.graphql', schema }],
+    },
+    diagnosticSettings: {},
+  });
+  return { worker, uri };
+}
+
+describe('GraphQLWorker.doHover', () => {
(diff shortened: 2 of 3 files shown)


dimaMachina (approved):
Nice, did you create failling test first?

trevor-scheer:
@dimaMachina test was after but I did test it against `main` for failure!

PR #4249, opened 2026-05-11
chore: Fix import path for React in graphiql-cdn example

Nick-MX:


trevor-scheer (approved):
Thanks, lgtm just need the CLA signature.

Nick-MX:
Signed.

dimaMachina (commented):
@trevor-scheer it looks like this file is generated by some script now? And script should be fixed instead 👀

Nick-MX:
I updated the generation script. Please let me know if there's anything else that needs to be updated.

dimaMachina (approved):
Thank you for contribution!

trevor-scheer:
@dimaMachina good catch, thanks!

PR #4317, opened 2026-05-20
v6: History `⌥`-click opens a diff against the current query

trevor-scheer:
## Summary

- Hold `⌥` (or `Alt` on non-Mac) and click a history row to open a side-by-side Monaco diff of the current query against the row's query.
- `Apply` loads the row into the editors (matching the existing plain-click behavior). `Cancel` or `ESC` dismisses with no changes.
- New `OperationDiffEditor` component and `diffOverlay` slice in `@graphiql/react` are generic: any caller can prime them to render a diff in the operation editor area.
- History panel subtitle updated to advertise the gesture.

## Test plan

- [ ] Run a query so a history row exists. Edit the query.
- [ ] `⌥`-click the row. Verify the side-by-side diff appears with "Compare with <row name>" in the header.
- [ ] Click `Apply`. Editor swaps to the row's query.
- [ ] Repeat the setup, `⌥`-click again, then `Cancel` (or `ESC`). Editor stays as you left it.

Refs: graphql/graphiql#4219

### DIFF
--- a/.changeset/history-alt-click-diff.md
+++ b/.changeset/history-alt-click-diff.md
@@ -1,0 +1,7 @@
+---
+'@graphiql/react': patch
+'@graphiql/plugin-history': patch
+'graphiql': patch
+---
+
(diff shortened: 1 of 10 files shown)


trevor-scheer:
Not loving this, might rethink and come back to it.

PR #4329, opened 2026-06-04
fix(graphql-language-service): allow optional field blocks in SDL type definitions

### LINKED ISSUE #4206: `schema-kitchen-sink.graphql` contains spec-invalid GraphQL
We're ignoring it in `oxfmt` config for now, but we should update the test (and fix the parser, because it doesn't handle this properly).

For reference, this was corrected in `graphql-js`'s version of the test which this copied.
https://github.com/graphql/graphql-js/commit/74d5bf54d1e75b6ef224cb#diff-fd3ee0ca4f0deb83ed6f82e5e204b6a3b4677870e858d4f67538ca01c463bef8

vishwakt:
## Summary

The online parser used by `codemirror-graphql` required a field block on object, interface, input, and enum type definitions. Per the GraphQL spec those blocks are optional, so valid SDL like `type Bar @baz` or `extend type Foo @onType` (directives only, no body) was highlighted as an error (`invalidchar`).

This makes the field/value block optional, mirroring how `SelectionSet` is already optional on fields.

Closes #4206.

## Changes

- `graphql-language-service`: field/value blocks are now optional for object, interface, input, and enum type definitions (and their extensions) in the online parser grammar.
- Corrected the `schema-kitchen-sink.graphql` test fixture, which contained spec-invalid empty bodies (`extend type Foo @onType {}` and `type NoFields {}`), to match graphql-js, and removed it from the oxfmt ignore list so it is formatted again.

## How to verify

This affects the CodeMirror 5 `graphql` mode in `codemirror-graphql`. The GraphiQL netlify preview will not exercise it, since GraphiQL 5 uses Monaco rather than this parser.

To reproduce in a CodeMirror 5 editor:

```js
import CodeMirror from 'codemirror';
import 'codemirror-graphql/esm/mode';

CodeMirror(document.getElementById('editor'), {
  value: 'type Bar @baz',
  mode: 'graphql',
});
```

On `main`, the `type Bar @baz` tokens render with the `cm-invalidchar` class (red). On this branch they render normally. The same applies to bodyless `interface`, `enum`, and `input` definitions and their extensions.

Before / after screenshots of that editor are attached in a comment below.

### DIFF
--- a/.changeset/schema-kitchen-sink-optional-fields.md
+++ b/.changeset/schema-kitchen-sink-optional-fields.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service': patch
+---
+
+Make the fields block optional when parsing object, interface, input, and enum type definitions (and their extensions) in the online parser. Per the GraphQL spec these blocks are optional, so spec-valid SDL such as `extend type Foo @onType` or `type Foo @onType` (directives only, no body) no longer reports `invalidchar` during syntax highlighting.
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -6,11 +6,7 @@
   "proseWrap": "never",
   "printWidth": 80,
   "sortPackageJson": false,
-  "ignorePatterns": [
-    "packages/codemirror-graphql/src/__tests__/schema-kitchen-sink.graphql",
-    "**/CHANGELOG.md",
-    "working-group/agendas/**"
-  ],
+  "ignorePatterns": ["**/CHANGELOG.md", "working-group/agendas/**"],
   "overrides": [
     {
       "files": ["*.md", "*.mdx"],
--- a/packages/codemirror-graphql/src/__tests__/mode.test.ts
+++ b/packages/codemirror-graphql/src/__tests__/mode.test.ts
@@ -91,6 +91,24 @@ describe('graphql-mode', () => {
     });
   });
 
+  it('parses type system definitions without a body without invalidchar', () => {
+    const schema = `
+      type EmptyType @onType
+      interface EmptyInterface @onInterface
+      enum EmptyEnum @onEnum
+      input EmptyInput @onInputObject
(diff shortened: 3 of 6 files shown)


vishwakt:
Heads up for reviewers: the failing `Types Check` here is **not introduced by this PR**. It is a pre-existing latent error in `graphql-language-service-server` (a `Logger` import that only resolves through a fragile re-export chain), normally masked by Turborepo's cache. This PR changes `graphql-language-service`, which invalidates that cache and forces the real re-run, surfacing the error.

Full diagnosis and proposed fix: #4330 (fix in #4331)

This PR's own packages type-check and test cleanly; the red check will clear once #4331 lands and this branch is rebased.

trevor-scheer (commented):
Thanks for fixing! I'll take a closer look at this at some point.

The entire verification section of your description is unhelpful to a reviewer. It would be much more helpful to provide a list of steps a reviewer can take to manually verify the fix. The netlify preview is usually a great place to start.

Fixes are always welcome, but please take the time to make your AI-generated descriptions easily digestible and coherent.

vishwakt:
Before:
Screenshot 2026-06-04 at 10 25 31

After:
Screenshot 2026-06-04 at 10 26 29

vishwakt:
@trevor-scheer thanks, that's fair. I've rewritten the description to drop the automated-results dump and added manual steps with before/after behavior.

Also added screenshots of the verification.

trevor-scheer:
Updating with merge after landing your fix, expecting type check to pass now.

trevor-scheer (changes requested):
The fix LGTM. It looks like the `OnlineParser` tests only update the existing tests? Can we add new tests for the cases we're allowing now (bare types/interfaces/etc with no selection set)?

trevor-scheer (approved):
Thanks!

PR #4331, opened 2026-06-04
fix(graphql-language-service-server): import Logger from vscode-jsonrpc

### LINKED ISSUE #4330: [lsp-server] 🐞 types:check fails on Logger import, masked by Turborepo cache

### Summary

`graphql-language-service-server` imports `Logger` from `vscode-languageserver`:

```ts
// packages/graphql-language-service-server/src/GraphQLLanguageService.ts:56
import type { Logger } from 'vscode-languageserver';
```

On CI (Linux), this fails type checking:

```
src/GraphQLLanguageService.ts(56,15): error TS2305:
Module '"vscode-languageserver"' has no exported member 'Logger'.
```

`Logger` is defined in `vscode-jsonrpc` and only reaches `vscode-languageserver` through a re-export chain. `vscode-languageserver`'s entry does:

```ts
export * from 'vscode-languageserver-protocol/'; // note the trailing slash
```

`tsgo` (the TypeScript native preview the repo migrated to in #4274) appears to resolve that trailing-slash re-export differently across platforms, so `Logger` is not visible from `vscode-languageserver` on CI.

### Why it has gone unnoticed

The error is latent and masked by Turborepo's cache. `graphql-language-service-server#types:check` is a cache hit for any PR that does not invalidate it, so the stale green result is reused. Only a PR that changes `graphql-language-service` (a dependency) busts the cache and forces a real re-run, which is when the error surfaces.

This is currently visible on #4329, whose `graphql-language-service` change invalidates the cache (`cache miss, executing bf0aa89f182b2312`). Other open PRs still pass `types:check` because their cached result is reused.

### Evidence

- Locally (macOS), `yarn types:check` produces the same Turbo hash CI computes (`bf0aa89f182b2312`) and passes (24/24), so this is not a source-logic issue, it is environment/resolution specific.
- The lockfile pins `vscode-languageserver@8.0.1 -> vscode-languageserver-protocol@3.17.1`, so the resolved versions are identical on CI and locally. The difference is in how `tsgo` resolves the trailing-slash re-export per platform.

### Proposed fix

Import `Logger` directly from `vscode-jsonrpc`, which defines and exports it, and which `graphql-language-service-server` already lists as a direct dependency (`vscode-jsonrpc: ^8.0.1`):

```ts
import type { Logger } from 'vscode-jsonrpc';
```

This avoids relying on the fragile transitive re-export chain. (Same dependency-hygiene class as #4324.)

### Environment

* LSP Server Version: `graphql-language-service-server` on `main`
* CI: GitHub Actions, Linux, `tsgo` (`@typescript/native-preview` 7.0.0-dev.20260418.1)
* Not reproducible on macOS with the same lockfile/tsgo, so CI is the verification path

vishwakt:
## Summary

`graphql-language-service-server` imported the `Logger` type from `vscode-languageserver`, but `Logger` is defined in `vscode-jsonrpc` and only reaches `vscode-languageserver` through a transitive re-export (`export * from 'vscode-languageserver-protocol/'`). On CI this breaks `types:check`:

```
src/GraphQLLanguageService.ts(56,15): error TS2305:
Module '"vscode-languageserver"' has no exported member 'Logger'.
```

This imports `Logger` from `vscode-jsonrpc` instead, which owns the type and is already a direct dependency of the package (`vscode-jsonrpc: ^8.0.1`), avoiding the fragile re-export chain.

Fixes #4330.

## Two separate things are going on

These answer different questions and are both true:

**Why it surfaced now (Turborepo cache).** `graphql-language-service-server#types:check` is a cache hit for any PR that does not invalidate it, so the stale green result is reused. Only a PR that changes `graphql-language-service` (a dependency) busts that cache and forces a real re-run. That is when the latent error becomes visible (it surfaced on #4329, a `graphql-language-service` change).

**Where the type error manifests (platform).** When `types:check` actually runs, it fails on CI (Linux) but passes locally (macOS). The lockfile pins identical dependency versions in both places (`vscode-languageserver@8.0.1 -> vscode-languageserver-protocol@3.17.1`), so resolution is the only variable. `tsgo` is `@typescript/native-preview`, which ships a separate compiled native binary per OS/arch (`native-preview-linux-x64` on CI vs `native-preview-darwin-arm64` locally), and the pre-release binaries appear to resolve the trailing-slash re-export differently.

I am confident about the cache behavior and the fix. The platform mechanism is my best inference, since I can only test on macOS and cannot directly inspect the Linux `tsgo`'s resolution. Either way the fix is robust because it imports `Logger` from the package that owns it.

## Verification

- `tsgo --noEmit` and `oxlint` pass locally on the changed package.
- This PR's CI is the real verification: changing this file invalidates the cached `types:check` and forces a fresh run on Linux, and `Types Check` (along with all other checks) passes.

### DIFF
--- a/.changeset/lsp-server-logger-import.md
+++ b/.changeset/lsp-server-logger-import.md
@@ -1,0 +1,5 @@
+---
+'graphql-language-service-server': patch
+---
+
+Import `Logger` from `vscode-jsonrpc` instead of `vscode-languageserver`. `Logger` is defined in `vscode-jsonrpc` (a direct dependency) and only reached `vscode-languageserver` through a transitive re-export, which `tsgo` failed to resolve on CI (`Module '"vscode-languageserver"' has no exported member 'Logger'`). Importing from the package that owns the type avoids relying on that fragile re-export chain.
--- a/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
+++ b/packages/graphql-language-service-server/src/GraphQLLanguageService.ts
@@ -53,7 +53,7 @@ import type { GraphQLCache } from './GraphQLCache';
 
 import { GraphQLConfig, GraphQLProjectConfig } from 'graphql-config';
 
-import type { Logger } from 'vscode-languageserver';
+import type { Logger } from 'vscode-jsonrpc';
 import {
   Hover,
   SymbolInformation,


trevor-scheer (approved):
Fix LGTM but is the PR description correct?

My understanding is that this is a turbo cache issue, but the description also says not reproducible on Mac / platform-specific. Are both things true? Platform-specific would be surprising to me.

vishwakt:
@trevor-scheer Both are true, and they answer two different questions:

- **Why it surfaced on a PR now (Turborepo):** `graphql-language-service-server#types:check` is a cache hit for any PR that doesn't invalidate it, so the stale green result gets reused. Only a PR touching `graphql-language-service` (a dependency) busts that cache and forces a real re-run, which is what exposed the error.
- **Where the type error actually manifests (platform):** when `types:check` does run for real, it fails on CI (Linux) but passes on my machine (macOS). I agree platform-specificity is surprising for a type error, but `tsgo` is `@typescript/native-preview`, which ships a separate compiled native binary per OS/arch (`native-preview-linux-x64` on CI vs `native-preview-darwin-arm64` locally). A pre-release native compiler resolving the unusual `export * from 'vscode-languageserver-protocol/'` (trailing slash) differently between platform builds is believable, and the lockfile pins identical dependency versions in both places, so resolution is the only variable left.

To be fully transparent: I'm confident about the Turbo cache behavior and the fix. The platform mechanism is my best inference, since I only have macOS to test on and can't directly inspect the Linux `tsgo`'s resolution. Either way the fix is robust, since it imports `Logger` from the package that actually owns it (`vscode-jsonrpc`, already a direct dependency) rather than relying on the transitive re-export.

Happy to tighten the PR description to separate these two points more clearly if that helps.

PR #4332, opened 2026-06-05
fix(graphiql-react): preserve host when previewing image URLs

### LINKED ISSUE #4101: [GraphiQL] 5.2.0 – Image preview not loaded correctly

### Current Behavior

Image

After migrating from version 3 to 5, we are experiencing a problem where the playground attempts to fetch images via a relative path instead of fetching them directly from the provided URL.

### Expected Behavior

If the link starts with http/https, fetch the image directly from the provided URL.
If not, continue with the current behavior of fetching the image from the absolute path.

### Steps To Reproduce

_No response_

### Environment

* GraphiQL Version: 5.2.0
* OS: Windows 11
* Browser: Edge latest version
* Bundler: Vite
* `react` Version: -
* `graphql` Version: -

### Anything else?

_No response_

vishwakt:
## Summary

In the response viewer, hovering an image URL shows an inline preview. The preview gets the hovered word from Monaco's `getWordAtPosition`, and Monaco's JSON word pattern treats `:` as a separator. So a full URL like `https://example.com/img.png` arrives as the protocol-relative `//example.com/img.png`.

`pathToURL` then stripped the leading character (`path.slice(1)`), turning `//example.com/img.png` into the host-relative `/example.com/img.png`, so the image was fetched from the current origin (`https://<current-host>/example.com/img.png`) instead of the real host.

The fix stops stripping that character. `//host/path` now resolves to the original host, and relative paths such as `/images/foo.png` still resolve against the current origin.

Closes #4101.

## How to verify

1. Run a query whose response contains a field with a full `https://` image URL (for example a logo or avatar hosted on another domain).
2. In the response pane, hover over the path part of that URL to trigger the image preview.

- Before: the preview is broken, because it requests the image from the current host.
- After: the preview loads, because it requests the image from the original host.

A unit test (`image-preview.spec.ts`) also covers this: `pathToURL('//example.com/img.png')` now resolves to host `example.com` rather than the current origin.

## Credit

Thanks to @nastoychev for the report and for pinpointing `pathToURL` and the expected behavior.

### DIFF
--- a/.changeset/image-preview-absolute-url.md
+++ b/.changeset/image-preview-absolute-url.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Fix image previews in the response viewer fetching from the wrong host. Monaco splits the hovered word on `:`, so a full URL like `https://example.com/img.png` reaches the preview as the protocol-relative `//example.com/img.png`. The preview stripped the leading character, turning that into the host-relative `/example.com/img.png` and fetching it from the current origin. The original host is now preserved.
--- a/packages/graphiql-react/src/components/image-preview.spec.ts
+++ b/packages/graphiql-react/src/components/image-preview.spec.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from 'vitest';
+import { ImagePreview, pathToURL } from './image-preview';
+
+describe('pathToURL', () => {
+  it('keeps the original host for full URLs', () => {
+    // Monaco splits the word on `:`, so a full `https://example.com/img.png`
+    // arrives here as the protocol-relative `//example.com/img.png`.
+    const url = pathToURL('//example.com/img.png');
+    expect(url?.host).toBe('example.com');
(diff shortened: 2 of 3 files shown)


trevor-scheer:
Do you happen to have a live endpoint / query I can run that returns an image? If not I can spin one up but if you already tested and have one handy that'd be great. Thanks!

vishwakt:
> Do you happen to have a live endpoint / query I can run that returns an image? If not I can spin one up but if you already tested and have one handy that'd be great. Thanks!

@trevor-scheer, Yes! The public Rick and Morty API works well for this:

**Endpoint:** `https://rickandmortyapi.com/graphql`
```graphql
{
  characters(page: 1) {
    results { name image }
  }
}
```

The `image` field returns a full cross-origin URL

**Before** (image requested from the current host, so it fails to load):
graphiql-4332-before


**After** (loads from the original host):
graphiql-4332-after

trevor-scheer:
@vishwakt perfect, thanks!

PR #4356, opened 2026-06-16
fix(graphql-language-service): wrap list input values in brackets on autocomplete

### LINKED ISSUE #587: Wrap auto-completed list input values with square brackets if required
In Gatsby's connections, you can add a sort input like the following:

```graphql
allMarkdownRemark(sort: { fields: [frontmatter___title], order: DESC}) {
  edges {
    node {
      id
    }
  }
}
```

But in graphiql, it doesn't enforce that fields must be an array so people regularly complain sorting is broken as they just add a field in the UI and then when they try to copy that into a project it won't work.

Is there a reason graphiql is like this? It seems to be casting the single values to an array which is why it seems to work.

DukeDeSouth:
Completing an enum or boolean value at a list-typed position (like `[Episode]`) inserts a bare `JEDI`, so you end up with `values: JEDI` instead of `values: [JEDI]` — issue #587. The online parser already unwraps one list level once the caret is inside the brackets, so I count the remaining list depth and add that many brackets; nested lists like `[[Episode]]` work too. Values typed inside an existing `[ ]`, single non-list values, and variables are left as-is. Added tests across the different positions and nullability combos, plus a changeset.

Closes #587

### DIFF
--- a/.changeset/wrap-list-input-values.md
+++ b/.changeset/wrap-list-input-values.md
@@ -1,0 +1,7 @@
+---
+'graphql-language-service': patch
+---
+
+Wrap autocompleted list input values in square brackets
+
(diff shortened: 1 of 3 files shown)


trevor-scheer (approved):
Thanks!

PR #4391, opened 2026-07-09
v6: Stop unstyled buttons from leaking the native `buttonface` background

trevor-scheer:
Two small theming fixes:

- **Unstyled buttons no longer leak the native `buttonface` background.** `.graphiql-un-styled` relies on `all: unset`, but that leaves the native button appearance, so these buttons painted a UA `buttonface` background that ignored the active theme whenever the OS color scheme differed from it. Setting `background: none` clears it.
- **Single writer for `data-theme`.** `useGraphiQLSettings` already writes `data-theme` to the GraphiQL container, which is where dialogs, dropdowns, and tooltips portal (via #4381); `stores/theme.ts` was redundantly writing it to `<html>` too. Dropping that write leaves one source of truth.

Refs: graphql/graphiql#4219

### DIFF
--- a/.changeset/quiet-plums-relax.md
+++ b/.changeset/quiet-plums-relax.md
@@ -2,4 +2,4 @@
 '@graphiql/react': patch
 ---
 
-Make `useGraphiQLSettings` the single writer of the `data-theme` attribute (scoped to the GraphiQL container) by dropping a redundant write to `<html>` from the theme store.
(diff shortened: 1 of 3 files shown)


trevor-scheer:
Superseded by #4381 — overlays already portal into the themed container and follow the active theme, so this `data-theme` single-writer cleanup is redundant. Closing.

PR #4412, opened 2026-07-11
v6: Query builder inline fragment extraction and editing

trevor-scheer:
## Summary

Reworks how the query builder handles fragments end to end.

**Extraction is now an inline row action.** Fragment extraction used to be a single "Create fragment from selection" button pinned to the bottom of the panel, acting on wherever the editor's text cursor happened to sit. It now lives on the field rows: expand a composite field, tick the children you want, and an "Extract to fragment" action appears on that row. Triggering it lifts the selection into a named fragment and replaces it with a spread (`user { ...UserFields }`). The generated name (`UserFields`, de-duped when taken) stays editable in place from the Fragments list.

**Extracted rows stay editable.** An extracted field no longer collapses into a locked `...FragmentName` badge. It stays a normal, expandable composite: the spread renders as a `...FragmentName` reference row among the field's children, and ticking more fields adds them to the **base query** alongside the spread. A field's only tie to its fragment is the real spread in the document — it doesn't "remember" being extracted.

**Fragments are editable in place.** Because a fragment's contents are shared across every field that spreads it, editing them needs its own surface. The active editing target now **follows the editor cursor**: move the cursor into a `fragment … { … }` block and the builder switches to editing that fragment's tree, rooted at its type condition. You can also jump in by clicking a `...FragmentName` reference row or a name in the Fragments list; "Back to query" returns to the operation. Fragments are flat peers — no nesting or breadcrumbs.

**The fragment reuse suggestion buttons are gone.** The short-lived "Use ...FragmentName" reuse action (and its dead helpers) is removed — it crowded the row, and multiple matching fragments on one field overflowed the small row.

The row actions are plain keyboard-focusable buttons. A dedicated shortcut and keycap hint are left for the app-wide accessibility pass.

### Implementation notes

The document mutators and readers were generalized from an operation-only `operationName` to a `DefinitionTarget` (`operation | fragment`), so the same field tree and handlers drive operation and fragment editing alike. Emptying a fragment is a no-op (a fragment can't be emptied without orphaning its spreads); promote-arg-to-variable stays operation-only.

## Test plan

- [x] Open the query builder against a schema. Expand a composite field (e.g. `person`), tick a scalar child (e.g. `name`), and confirm an "Extract to fragment" action appears on that row.
- [x] Trigger it. Confirm the query reads `person { ...PersonFields }` with a `fragment PersonFields on Person { name }` block, and the `person` row now shows a `...PersonFields` reference among its children.
- [x] With the field still expanded, tick another child (e.g. `age`). Confirm it lands in the **base query** — `person { ...PersonFields age }` — and the fragment is unchanged.
- [x] Click the `...PersonFields` reference row. Confirm a focused fragment editor opens (header `PersonFields on Person`, "Back to query" control), rooted at `Person`.
- [x] Tick a field in the fragment editor. Confirm it edits the fragment (`fragment PersonFields on Person { name age }`) and the operation still just spreads it.
- [x] Move the editor cursor into the `fragment PersonFields …` block. Confirm the builder switches to editing that fragment. Move it back into the operation; confirm it switches back.
- [x] Click a fragment's name in the Fragments list. Confirm it focuses that fragment and is highlighted as active.
- [x] In the Fragments list, rename `PersonFields`. Confirm the definition and every `...PersonFields` reference update together.
- [x] Keyboard-only: Tab to the extract action / reference row / fragment name, confirm each takes focus and activates with Enter/Space.
- [x] Unit and `query-builder.cy.ts` suites pass; Storybook a11y check clean.

Refs: graphql/graphiql#4219


packages/graphiql-plugin-query-builder/src/style.css
```diff
@@ -798,39 +798,303 @@ select.graphiql-qb-arg-select:focus {
   list-style: none;
   display: flex;
   flex-direction: column;
-  gap: 2px;
+  gap: 1px;
 }
 
+/* Roomier rows echoing the operation headers: a bold name and its type
```
trevor-scheer:
```suggestion
```

PR #4415, opened 2026-07-12
fix(vscode-graphql-syntax): don't highlight .graphql() member calls as embedded GraphQL

### LINKED ISSUE #3810: [graphiql] Syntax highlighting breaks with .graphql as a function

### Current Behavior

When using amplify's graphql api capabilities, the syntax highlighting for anything past the graphql function stops working properly. See screenshot. You can see where the if after .graphql() is yellow instead of the purple like previous as well as the const following that if is white instead of navy blue.

image

### Expected Behavior

This should affect graphql syntax, but not affect other highlighting after using a .graphql() function.

### Steps To Reproduce

_No response_

### Environment

* GraphiQL Version: N/A
* OS: Windows 11
* Browser: N/A
* Bundler: N/A
* `react` Version: 18.2.0
* `graphql` Version: N/A
* `aws-amplify` Version: 6.6.7

### Anything else?

_No response_

mvanhorn:
## Summary

The `vscode-graphql-syntax` TextMate grammar treated a bare `graphql` / `gql` identifier as an embedded-GraphQL trigger even when it appeared as a member access, so `client.graphql<Response>(...)` and `foo.graphql()` were highlighted as embedded GraphQL and broke downstream JS/TS scoping (regression from #3540). This restricts the trigger to genuine tagged-template / helper calls.

## Why

The injection regex now requires the keyword not to be preceded by a member/identifier boundary, so member calls (`.graphql()`, private `#graphql()`) are excluded, and it preserves quoted angle brackets inside generic arguments (`graphql<Result<User>>(...)`). Cross-line member access is handled as far as a single-line TextMate grammar can. The bulk of the diff is regenerated tokenizer snapshots reflecting the corrected scoping.

## Testing

`vitest run tests/js-grammar.spec.ts` in `packages/vscode-graphql-syntax` -- passes, with fixtures added for member calls, private-member calls, and generic arguments.

Closes #3810

### DIFF
--- a/.changeset/fix-graphql-method-highlight.md
+++ b/.changeset/fix-graphql-method-highlight.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql-syntax': patch
+---
+
+Prevent GraphQL syntax highlighting from leaking into JavaScript and TypeScript after `.graphql()` member calls.
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -6,7 +6,11 @@
   "proseWrap": "never",
   "printWidth": 80,
   "sortPackageJson": false,
-  "ignorePatterns": ["**/CHANGELOG.md", "working-group/agendas/**"],
+  "ignorePatterns": [
+    "**/CHANGELOG.md",
+    "working-group/agendas/**",
+    "packages/vscode-graphql-syntax/tests/__fixtures__/**"
+  ],
   "overrides": [
     {
       "files": ["*.md", "*.mdx"],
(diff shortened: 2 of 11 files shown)


mvanhorn:
Fixed the CSpell and Format checks in e38d5c8. CSpell was flagging `alnum` from the POSIX class `[_$[:alnum:]]` in the grammar fixture, so it's in the custom word list now. For Format: oxfmt wanted to rewrite the test fixtures, but that would erase exactly what they test (the multiline member split, the `client. graphql` space, the comment between the dot and the method name), so I excluded `tests/__fixtures__/**` in .oxfmtrc.json instead — grammar fixtures need their exact source text preserved. The Canary Builds failure looks like a registry hiccup (Cloudflare error page in the log); this push re-runs it.

PR #4426, opened 2026-07-13
v6: Ship Inter and JetBrains Mono as the default fonts

trevor-scheer:
## Summary

- v6 now ships Inter and JetBrains Mono by default, matching the original design handoff. `tokens.css` already declared `--font-family: 'Inter'` and `--font-family-mono: 'JetBrains Mono'`, but nothing actually loaded those fonts, and a legacy block in `root.css` redeclared both vars with the old Roboto/Fira Code values and won the cascade. So the app had actually been rendering Roboto and Fira Code this whole time regardless of what the tokens said.
- Self-hosted `font/inter.css` and `font/jetbrains-mono.css` (base64, Latin + Latin Extended, regular and italic) load by default now, following the same pattern as the existing `font/roboto.css` / `font/fira-code.css`. Roboto and Fira Code stay in the repo and stay importable for anyone who wants the v5 look.
- Removed the `--font-family`, `--font-family-mono`, and `--font-size-body` redeclarations from `root.css`'s `.graphiql-container` block so `tokens.css` actually wins. That also fixes a second bug: `--font-size-body` was pinned to a fixed 15px by the same shadowing, so the Settings > Font Size preset never touched body text. It scales correctly now.
- Monaco's editor font is now JetBrains Mono instead of Fira Code, matching the token.
- Bundle impact: the two self-hosted font files add about 160 KB to `graphiql`'s built `style.css` (710 KB → 875 KB) compared to today's Roboto + Fira Code. That's the cost of matching the handoff exactly; happy to trim further (e.g. drop italic, which is most of the added weight) if the size isn't worth it.

## Test plan

- [ ] Open the app. The UI (panel headers, buttons, sidebar) should render in Inter, and the query/response editors should render in JetBrains Mono, not Roboto/Fira Code.
- [ ] Open Settings and change Font Size to Large or Extra Large. Body text throughout the app should visibly grow (it previously stayed fixed regardless of this setting).
- [ ] Set Font Size back to Default and confirm everything returns to its original size.
- [ ] Check both Dark and Light themes look right at a couple of font sizes.
- [ ] Type a query with a `#` comment and confirm it renders in italic in the editor.

Refs: #4228


trevor-scheer:
Deciding against.

PR #4432, opened 2026-07-15
v6: fix broken migration guide examples and transport docs

trevor-scheer:
## Summary

A docs review cross-checked the migration guide and package READMEs against the actual `graphiql-6` source and found several that describe APIs that no longer exist. The migration guide's `useDocExplorer`/`useHistory` "after" examples pass a selector argument and destructure fields (`.navStack`, `.items`) that don't exist — both hooks take zero arguments and return the value directly, so copying the examples as written is a compile error. The `graphiql` and `graphiql-react` READMEs describe `transport` as "a function," which is the old `Fetcher` shape; it's an object with a `send()` method. The `createTransport` link in `graphiql/README.md` pointed at `createFetcher.ts`, and neither README mentioned that `fetcher` still works. The "Editor Theme" section documented a CodeMirror API that hasn't existed since the Monaco rewrite. The toolkit's own README never mentioned `createTransport`, its biggest new export. `create-fetcher.md`'s deprecation banner linked to a `create-transport.md` file that doesn't exist, and (found while fixing that line) the migration-guide link right next to it was also pointing one directory too shallow.

Also added the `ExecuteButton` removal to the migration guide's breaking-changes notes, reworded the `.browserslistrc` line (the file still exists, only its contents changed to `defaults`), and added the v6 migration guide to `graphiql/README.md`'s getting-started list, which already linked to it later in the doc but didn't list it up top.

Refs: graphql/graphiql#4219


docs/migration/graphiql-6.0.0.md
```diff
@@ -616,6 +616,7 @@ Nothing below is happening in v6. These are informal signals about where the pro
...
 - **Monaco editor theme registration.** The built-in Monaco themes (`graphiql-DARK` and `graphiql-LIGHT`) were rewritten to use the v6 accent palette — every GraphQL token type (keywords, type names, field identifiers, variables, annotations, strings, numbers, comments) and the surrounding editor chrome (suggest widget, hover widget, quick input) now pull from the same tokens as the rest of the UI. The `editorTheme` prop on `<GraphiQL>` still takes the same `{ dark, light }` pair of Monaco theme names or definitions it always did, so a custom theme registered through that prop should continue to work unchanged. If you were depending on the specific color values baked into the previous built-in `graphiql-DARK` / `graphiql-LIGHT` themes (for a screenshot test, for example), expect those values to have shifted.
 - **`cn` removed from `@graphiql/react`.** The `cn` helper was just a re-export of `clsx`. It's gone now, so import `clsx` directly from the `clsx` package instead.
+- **`ExecuteButton` removed from `@graphiql/react`.** The Run button and its execution logic moved into the top bar; there's no longer a standalone `ExecuteButton` export to compose into a custom toolbar.
```
trevor-scheer: Fixing this, we'll keep exporting this button.

PR #4442, opened 2026-07-26
fix: navigate docs for input object fields

### LINKED ISSUE #4441: [graphiql] Nested input object fields do not open in Documentation Explorer
## Current Behavior

Cmd/Ctrl-clicking a field inside a GraphQL input object literal in the Monaco operation editor opens the Documentation Explorer, but the explorer does not navigate to the clicked input field.

For example, given a schema with nested input object types and this mutation:

```graphql
mutation {
  submit(
    order: {
      lineItem: {
        quantity: 1
      }
    }
  )
}
```

Cmd/Ctrl-clicking `lineItem` or `quantity` leaves the Documentation Explorer at its previous location. Equivalent Cmd/Ctrl-click navigation for query selection fields works.

The Monaco definition provider already recognizes these tokens:

```ts
kind === 'ObjectField' && step === 0 && typeInfo.fieldDef
```

However, `packages/graphiql-plugin-doc-explorer/src/schema-reference.ts` only resolves `Field` and `AliasedField` field references. `ObjectField` is ignored. In addition, `typeInfo.parentType` still points to the enclosing output type, so the input object type that owns the clicked input field must be resolved from the schema.

## Expected Behavior

Cmd/Ctrl-clicking a nested input object field should navigate the Documentation Explorer to that input field under its owning `GraphQLInputObjectType`, just as clicking a query selection field navigates to its field documentation.

## Steps To Reproduce

1. Render GraphiQL with the default Documentation Explorer and a schema containing a mutation argument whose type is a nested input object.
2. Enter a mutation using nested input object literals.
3. Cmd-click on macOS, or Ctrl-click on other platforms, an input object key.
4. Observe that the Documentation Explorer becomes visible but does not navigate to the clicked field.
5. Cmd/Ctrl-click a query selection field and observe that navigation works.

## Environment

- **GraphiQL**: 5.2.2 (also present on current `main`)
- **@graphiql/react**: 0.37.3 / current `main`
- **@graphiql/plugin-doc-explorer**: 0.4.1 / current `main`
- **OS**: macOS
- **Browser**: Chromium
- **React**: 19
- **GraphQL**: 16

## Proposed Fix

Teach the Documentation Explorer's schema-reference resolver to handle `ObjectField`. For that reference kind, locate the `GraphQLInputObjectType` whose field map owns `typeInfo.fieldDef`, then create the same field reference shape used for ordinary fields.

A regression test should cover a nested mutation input and verify both the owning input type and clicked field are resolved.

pur3miish:
## Summary

- handle `ObjectField` schema references in the Documentation Explorer
- preserve the owning input-object type as `inputObjectType` in `AllTypeInfo`
- resolve input fields directly from parser-provided type information instead of scanning the schema
- add regression coverage for nested mutation input fields
- declare release intent for `graphql-language-service` and `@graphiql/plugin-doc-explorer`

## Root cause

The operation editor already emits an `ObjectField` schema-reference event for command/control-clicks inside input object literals. The Documentation Explorer did not handle that event, and `getTypeInfo` retained only the input object's field map—not the owning input-object type needed to build the documentation navigation path.

## Implementation

- add `inputObjectType` to `AllTypeInfo` and populate it while processing `OBJECT_VALUE`
- have the Documentation Explorer read the owning type from `typeInfo.inputObjectType`
- remove the schema parameter and reverse `getTypeMap()` lookup from `getSchemaReference`
- intentionally require the updated type-information contract without a compatibility fallback
- update the changeset with:
  - `graphql-language-service`: minor
  - `@graphiql/plugin-doc-explorer`: patch

## Testing

- `graphql-language-service`: 277 tests passed
- `@graphiql/plugin-doc-explorer`: 21 tests passed
- TypeScript checks passed for both affected packages
- formatting and lint checks passed for the changed files
- Changesets release-plan validation completed

Fixes #4441


trevor-scheer:
Thanks for the PR!

The part that's giving me pause is threading `schema` into `getSchemaReference` so it can scan `getTypeMap()` for the input object that owns `fieldDef`. I think we're fixing this a layer too late. Could it go in the parser instead?

I'm imagining something around the `OBJECT_VALUE` case in `getTypeInfo.ts`, which already resolves the owning type and then keeps only its field map:

```ts
case RuleKinds.OBJECT_VALUE:
  const objectType = getNamedType(inputType!);
  objectFieldDefs = objectType instanceof GraphQLInputObjectType
    ? objectType.getFields()
    : null;
```

If we held onto `objectType` as well (a new field on `AllTypeInfo`?), the doc explorer could read it directly from `typeInfo` and the `schema` param disappears, + some other cleanup.

Let me know what you think!

pur3miish:
> Let me know what you think!

That makes sense and avoids the reverse schema lookup. One concern is that AllTypeInfo is exported from graphql-language-service, while the doc explorer accepts older @graphiql/react versions that won’t populate the new field. Would you prefer that I make inputObjectType optional and retain a compatibility fallback, or update the relevant package versions/peer range as part of this PR?

trevor-scheer:
@pur3miish good observation. I'd be happy with the latter (no back-compat with proper versioning bumps). I'm pretty sure this just means adding a `major` entry to your changeset for `@graphiql/react` (and the doc explorer should automatically receive a patch bump for it...I can confirm it's correct before we cut the release).

So just make sure that entry is added to the changeset as well and I'll make sure the versioning is correct before release. Thanks!

I believe [this changesets PR](https://github.com/changesets/changesets/pull/2090/changes) is relevant to our scenario.

pur3miish:
@trevor-scheer Thanks for the guidance! I’ve updated the PR based on your recommendation:

- added `inputObjectType` to `AllTypeInfo` and populated it in `getTypeInfo`
- updated the Documentation Explorer to read the owning input-object type directly from `typeInfo`
- removed the schema-wide reverse lookup and compatibility fallback
- added parser and Documentation Explorer regression coverage
- updated the changeset with a major bump for `@graphiql/react`, a minor bump for `graphql-language-service`, and a patch bump for `@graphiql/plugin-doc-explorer`

The affected tests, type checks, formatting, and lint checks pass.

After updating the branch, the Changesets bot currently calculates major bumps for the Documentation Explorer and the other `@graphiql/react` peer dependents. Since you offered to confirm the versioning before release, could you take a look and let me know whether any further changes are needed? Thanks!

pur3miish:
Hi @trevor-scheer — friendly follow-up. The parser-level change and requested `@graphiql/react` major changeset are now included, all CI checks pass, and the PR is mergeable. The only remaining question appears to be the Changesets propagation to peer dependents. Could you confirm whether the current versioning is acceptable and, if everything looks good, review/merge the PR? Thanks!

trevor-scheer:
Hey @pur3miish, thank you for your patience. I'm solo parenting for the week and pretty preoccupied, I'll try to get to this as soon as I'm able to but it probably won't be until next week.

.changeset/bright-pens-smile.md
```diff
@@ -0,0 +1,7 @@
+---
+'@graphiql/react': major
```
trevor-scheer: My mistake, I'm actually unsure why I suggested this but I think I meant the `graphql-language-service` package (which already had a correct pre-v0 minor, so no change needed there).
```suggestion
```

trevor-scheer (approved):

.changeset/bright-pens-smile.md
pur3miish: Done in 6c9ec24 — I removed the `@graphiql/react` major entry and left the existing `graphql-language-service` minor and `@graphiql/plugin-doc-explorer` patch entries. I also updated the PR description to match. Thanks for clarifying!

PR #4448, opened 2026-07-27
feat(graphiql-react): expose customScalarSchemas for the variable editor

### LINKED ISSUE #4447: [graphiql] Expose customScalarSchemas so custom scalars aren't flagged as invalid in the variable editor
### Problem

The variable editor's live JSON Schema linter (`graphql-language-service`'s `getVariablesJSONSchema`, consumed via `monaco-graphql`) assumes every custom scalar only accepts primitives:

```js
definition.type = ['string', 'number', 'boolean', 'integer'];
```

This is correct for a scalar like `DateTime` serialized as a string, but wrong for any custom scalar that legitimately holds an object or array, e.g. a `JSON` scalar, or in our case a `GeoJSON` scalar (RFC 7946) used by [`@graphile/postgis`](https://github.com/graphile/postgis). Passing a valid value like:

```json
{ "type": "Point", "coordinates": [9.5, 51.3] }
```

...for a `GeoJSON!` variable gets flagged in the editor as:

> Incorrect type. Expected one of string, number, boolean, integer.

The query still executes successfully server-side. This is an editor-side false positive, confusing when using the tool, since it looks like the request will fail.

### Existing (unexposed) solution

Both underlying libraries already support fixing this per-scalar:

- `graphql-language-service`'s `getVariablesJSONSchema` accepts a `scalarSchemas` option (added in [graphql/graphiql#3376](https://github.com/graphql/graphiql/pull/3376)).
- `monaco-graphql`'s `SchemaConfig` already threads this through as `customScalarSchemas` (`LanguageService.getVariablesJSONSchema` forwards `schema.customScalarSchemas` into `scalarSchemas`).

`@graphiql/react`'s `operation-editor.tsx` seems to be a gap: it calls `monacoGraphQL.setSchemaConfig([{ uri, schema }])` with a hardcoded two-field object, with no prop anywhere in `GraphiQLProps`/`SchemaProps` to supply a third `customScalarSchemas` field. So the capability exists two layers down, but the top-level component never surfaces it.

### Why this isn't solvable via the plugin API

I want to address this per the template's guidance to explore the plugin/sidebar API first: this isn't a sidebar UI concern, it's the Monaco *language service* schema registration that happens in `operation-editor.tsx` on mount/schema-change. There's no hook today for a consumer to intercept or extend that `setSchemaConfig` call, and I don't think there's a way to build this as a sidebar plugin as it needs to affect the shared schema config an existing core component registers.

### Proposed fix

Add an optional `customScalarSchemas?: Record<string, JSONSchema6>` prop to `SchemaProps` (alongside `schema`, `schemaDescription`, etc.), threaded through the schema store slice and included in `operation-editor.tsx`'s `setSchemaConfig` call. Purely additive/opt-in, no behavior change for anyone not using it.

I have a working patch for this (~35 lines) and I'm happy to open a PR! Wanted to check in here first before doing so. Any preference on API shape, e.g. `customScalarSchemas` (matching `monaco-graphql`'s own naming) vs. something else?

dargmuesli:
## Summary

Adds an optional `customScalarSchemas` prop to `@graphiql/react`, so custom GraphQL scalars that legitimately hold objects/arrays (a `JSON` scalar, a `GeoJSON` scalar, etc.) aren't flagged by the variable editor's live JSON Schema linter with a false-positive "Incorrect type. Expected one of string, number, boolean, integer." error.

Both `graphql-language-service` (`scalarSchemas` option, added in #3376) and `monaco-graphql` (`SchemaConfig.customScalarSchemas`) already support this per-scalar override. `@graphiql/react`'s `operation-editor.tsx` was the missing link: it registers the schema with Monaco via a hardcoded `{ uri, schema }`, with no way for a consumer to add the third field.

Fixes #4447. See that issue for the full problem description.

## Changes

- `stores/schema.ts`: added `customScalarSchemas?: Record<string, JSONSchema6>` to `SchemaProps`, threaded through `SchemaSlice` and `CreateSchemaSlice`.
- `components/provider.tsx`: destructures the new prop, passes it into `createSchemaSlice(...)`.
- `components/operation-editor.tsx`: reads it from the store and includes it in the `monacoGraphQL.setSchemaConfig(...)` call (and its effect's dependency array).

Existing behavior for anyone not passing the prop is unchanged.

## Usage

```tsx
<GraphiQL
  fetcher={fetcher}
  customScalarSchemas={{
    GeoJSON: {}, // accept any JSON value
    DateTime: { type: 'string', format: 'date-time' },
  }}
/>
```

### DIFF
--- a/.changeset/expose-custom-scalar-schemas.md
+++ b/.changeset/expose-custom-scalar-schemas.md
@@ -1,0 +1,6 @@
+---
+'@graphiql/react': minor
+'graphiql': minor
+---
+
+Add `customScalarSchemas` prop to `GraphiQLProvider`/`GraphiQL`, forwarded to `monaco-graphql`'s per-schema `customScalarSchemas` config. Without it, the variable editor's live JSON Schema linter assumes every custom scalar only accepts primitives (string, number, boolean, integer) and reports a spurious "Incorrect type" error for scalars that legitimately accept objects or arrays (e.g. a `JSON` or `GeoJSON` scalar). `graphql-language-service` and `monaco-graphql` already supported this per-scalar override; `@graphiql/react` just never exposed it.
--- a/packages/graphiql-react/src/components/operation-editor.tsx
+++ b/packages/graphiql-react/src/components/operation-editor.tsx
@@ -59,6 +59,7 @@ export const OperationEditor: FC<OperationEditorProps> = ({
   const {
     initialQuery,
     schema,
+    customScalarSchemas,
     referencePlugin,
     operations,
     operationName,
@@ -70,6 +71,7 @@ export const OperationEditor: FC<OperationEditorProps> = ({
     pick(
       'initialQuery',
       'schema',
+      'customScalarSchemas',
       'referencePlugin',
       'operations',
       'operationName',
@@ -273,7 +275,11 @@ export const OperationEditor: FC<OperationEditorProps> = ({
       return;
     }
     monacoGraphQL.setSchemaConfig([
-      { uri: `${uriInstanceId}${URI_NAME.schema}`, schema },
+      {
+        uri: `${uriInstanceId}${URI_NAME.schema}`,
+        schema,
+        customScalarSchemas,
+      },
     ]);
     monacoGraphQL.setExternalFragmentDefinitions([
       ...externalFragments.values(),
@@ -338,6 +344,7 @@ export const OperationEditor: FC<OperationEditorProps> = ({
     return cleanupDisposables(disposables);
   }, [
     schema,
+    customScalarSchemas,
     referencePlugin,
     setSchemaReference,
     setVisiblePlugin,
--- a/packages/graphiql-react/src/components/provider.tsx
+++ b/packages/graphiql-react/src/components/provider.tsx
@@ -151,6 +151,7 @@ const InnerGraphiQLProvider: FC<GraphiQLProviderProps> = ({
   onCopyQuery,
   onPrettifyQuery = DEFAULT_PRETTIFY_QUERY,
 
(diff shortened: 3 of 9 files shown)


dargmuesli:
I'll sign the CLA once there is positive feedback on my proposal. Btw, the CLA document template looks off: the fields for me to fill are randomly placed on some page in the second step.

.changeset/expose-custom-scalar-schemas.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphiql/react': minor
```
trevor-scheer, line 2: Changesets _will_ bump `graphiql` without this due to its dependency on `@graphiql/react`, but this will ensure your description below lands in the `graphiql` changelog which is arguably the more interesting place for users to see it.
```suggestion
'@graphiql/react': minor
'graphiql': minor
```

trevor-scheer (commented):
Thanks, lgtm. Could you add a custom scalar usage to the demo app and a cypress test?

`packages/graphiql/src/e2e.ts` powers the demo and our cypress tests, and the schema behind it is `packages/graphiql/test/schema.js`. Then add a test in `packages/graphiql/cypress/e2e/lint.cy.ts` along the existing variable type tests.

trevor-scheer (approved):
Thanks!

dargmuesli:
I've update the cspell words.

dargmuesli:
... and the unit test, good god I was certain that I ran everything locally

trevor-scheer:
Sorry I can't perma-approve the CI runs, that certainly doesn't help. Thanks for your patience!

PR #4463, opened 2026-08-22
fix(monaco-graphql): support monaco-editor >= 0.56

### LINKED ISSUE #4382: Support monaco-editor@0.53+
Hi!

I see that across the repo `monaco-editor` is pinned to `0.52.2`. 
I know that they changed how worker are loaded in `0.53.x` and onwards, and I wanted to know if you plan to support it.

Thank you

lukasbash:
Fixes #4382.

monaco-editor 0.53+ rewrote how custom workers are created and dropped
the `exports` field escape hatch monaco-graphql relied on for its deep
imports, breaking every monaco-graphql install on modern monaco-editor.

## Two independent breaks bundled together

1. **`editor.createWebWorker()` no longer loads a worker by `moduleId`
   or delivers `createData`.** The type is now
   `{ worker: Worker | Promise<Worker> }` - the caller is expected to
   construct the worker itself. `WorkerManager` now builds the worker via
   `MonacoEnvironment.getWorker()` and configures it through a new
   `GraphQLWorker#initialize()` RPC call once the worker is ready,
   instead of via constructor args that are no longer transported.

2. **`graphql.worker.ts`'s bootstrap wrapped `initialize()` inside an
   extra `onmessage` handler.** monaco-editor's client now sends a
   throwaway ping message immediately followed by the real
   `$initialize` handshake; the extra wrapper consumed the ping and
   only called `initialize()` on the next message - the real handshake -
   one message too late for its reply to reach the client, hanging
   every worker RPC call forever.
   `monaco-editor/editor/editor.worker`'s own bootstrap calls
   `initialize` unwrapped for the same reason; mirrored that here.

Also updates the now-relocated `monaco-editor/esm/vs/...` deep imports
(the `exports` field dropped the old paths), the deprecated
`languages.json.jsonDefaults` reference, and bumps the `monaco-editor`
peer range to `>= 0.56.0`.

This is a **breaking change** for consumers on monaco-editor < 0.53 -
no attempt is made to support both API shapes at once, matching the
scope implied by #4382.

## Test plan

- [x] Full monorepo `tsgo --build` passes clean
- [x] Verified end-to-end against a real app (React + Vite, native
      `MonacoEnvironment.getWorker` worker setup, no
      `vite-plugin-monaco-editor`) on monaco-editor 0.56.0: GraphQL
      autocomplete, hover, and validation all round-trip through the
      worker correctly - confirmed via instrumented worker calls that
      the full `WorkerManager -> createWebWorker -> initialize() ->
      doValidation()` chain resolves and returns real diagnostics
      (previously threw `Missing requestHandler or method: doValidation`
      / `doComplete` / `doHover` on every call)
- [ ] Package's own test suite (`vitest run`) - not run in this pass,
      happy to run if maintainers want it before review

### DIFF
--- a/packages/monaco-graphql/package.json
+++ b/packages/monaco-graphql/package.json
@@ -78,13 +78,13 @@
   "devDependencies": {
     "execa": "^7.1.1",
     "graphql": "^16.9.0",
-    "monaco-editor": "0.52.2",
+    "monaco-editor": "^0.56.0",
     "prettier": "3.3.2",
     "vscode-languageserver-types": "^3.17.1"
   },
   "peerDependencies": {
     "graphql": "^15.5.0 || ^16.0.0 || ^17.0.0",
-    "monaco-editor": ">= 0.20.0 < 0.53",
+    "monaco-editor": ">= 0.56.0",
     "prettier": "^2.8.0 || ^3.0.0"
   }
 }
--- a/packages/monaco-graphql/src/GraphQLWorker.ts
+++ b/packages/monaco-graphql/src/GraphQLWorker.ts
@@ -22,8 +22,21 @@ export class GraphQLWorker {
   private _languageService: LanguageService;
   private _formattingOptions: FormattingOptions | undefined;
 
-  constructor(ctx: monaco.worker.IWorkerContext, createData: ICreateData) {
+  constructor(ctx: monaco.worker.IWorkerContext, createData?: ICreateData) {
     this._ctx = ctx;
+    this._languageService = new LanguageService(
+      createData?.languageConfig ?? {},
+    );
+    this._formattingOptions = createData?.formattingOptions;
+  }
+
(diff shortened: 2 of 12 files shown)


trevor-scheer:
Thanks @lukasbash! I'd like to try to land this in our upcoming major release (on the `graphiql-6` branch) but it needs a fair amount of work. Would you mind if I push changes to your branch/PR?

lukasbash:
> Thanks @lukasbash! I'd like to try to land this in our upcoming major release (on the `graphiql-6` branch) but it needs a fair amount of work. Would you mind if I push changes to your branch/PR?

@trevor-scheer I would appreciate it. I wanted to invest more time but honestly I am not willing to sign a multi-page CLI form for putting effort in a migration to new dependencies.
Would be still super awesome if this makes it into core.

PR #4472, opened 2026-08-29
v6: don't create an empty tab on reload when no editor state was stored

trevor-scheer:
Load GraphiQL with empty storage, don't touch anything, and reload: a second `<untitled>` tab appears, empty and active, next to the default tab. Every fresh-visit-then-reload session starts with two tabs instead of one.

The cause is in `getDefaultTabState`. The `query`/`variables`/`headers` storage keys are only written on edit, so after a first visit that never touched the editors they're all still `null`. On reload the function hashes those `null`s, finds no stored tab with a matching hash (the default tab hashes its `defaultQuery` contents), and takes the "preserve the stored editor state in a new tab" branch, pushing a tab made entirely of `null`s. That branch exists to carry editor content forward from pre-tabs storage; when all three keys are empty there's nothing to carry.

The fix skips the push when none of the three keys hold a value and storage already has at least one tab, keeping the stored `activeTabIndex` (clamped to the tab range, since that index is no longer unconditionally rewritten on this path).

Reproduction is the first commit (the new test fails on `graphiql-6`: it gets 2 tabs where it expects 1), fix is the second.

I could see this predating v6 in principle, but I haven't verified against v5; the hash-match-then-push logic is unchanged from `main`, so if v5's CodeMirror editors also never wrote the storage keys on mount it has the same behavior.

### DIFF
--- a/.changeset/fix-phantom-tab-on-reload.md
+++ b/.changeset/fix-phantom-tab-on-reload.md
@@ -1,0 +1,5 @@
+---
+'@graphiql/react': patch
+---
+
+Fix an extra empty tab being created on reload. The `query`/`variables`/`headers` storage keys are only written on edit, so reloading a session that never touched the editors found no stored editor state matching any tab and pushed a new empty one.
--- a/packages/graphiql-react/src/utility/tabs.spec.ts
+++ b/packages/graphiql-react/src/utility/tabs.spec.ts
@@ -143,6 +143,32 @@ describe('getDefaultTabState', () => {
     });
   });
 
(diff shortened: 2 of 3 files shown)


trevor-scheer:
Reproduced this on `graphiql@5.3.0` as well (clean storage → load → reload gives two tabs), so it's a longstanding `main` bug rather than a v6 regression. Opened #4476 with the same test+fix against `main`; if that lands and merges into `graphiql-6`, this PR can close in its favor.

trevor-scheer:
Superseded by #4476

PR #4385, opened 2026-07-05
Highlight GraphQL string values as strings, not comments

glen-84:
## Summary

- GraphQL string values — single-line and block (triple-quoted) — in directive and field arguments, list items, and default values are now highlighted as strings (`string.quoted.double.graphql` / `string.quoted.triple.graphql`) instead of comments. The grammar's comment rule previously matched any `"…"`, so argument and default string values were scoped as comments.
- Descriptions are scoped as documentation (`comment.block.documentation.graphql` / `comment.line.documentation.graphql`) and only in valid description positions per the GraphQL spec — type-system definitions and executable operation, fragment, and variable descriptions — so themes can distinguish them while still defaulting to comment coloring.
- Field and input default values are now tokenized (the `=` and value on a field definition were previously left unscoped).

## Test plan

- `yarn workspace vscode-graphql-syntax test` passes.
- Added a `descriptions-and-values.graphql` fixture and snapshot covering descriptions in every valid position and string values (single- and triple-quoted) across arguments, lists, and default values; updated the embedded-GraphQL fixture to distinguish an operation description from a `#` comment.

Closes #2358

_Drafted by Claude (Anthropic AI assistant)._

PR #4491, opened 2026-09-08
fix(vscode-graphql-syntax): correct consecutive description highlighting

rrbe:
## Summary

Fixes a regression introduced in #4385.

Field and argument definitions did not end before a following description. Without a separating comma, descriptions were tokenized inside the previous definition. Commas within description text could then shift quote pairing and cause subsequent code to be highlighted as documentation.

End these definitions before description strings while preserving string-value highlighting.

## Reproduction

Open this schema with GraphQL: Syntax Highlighting 1.3.12:

```graphql
type Query {
  users(
    "First argument"
    first: Int
    "Second argument, with punctuation"
    second: Int
  ): String

  """Next field"""
  next: String
}
```

Expected: descriptions are highlighted as documentation and field definitions retain their normal highlighting.

Actual: the second description is misclassified, and subsequent code can be highlighted as documentation.

While writing this PR, I found that GitHub's own graphql code rendering can directly reproduce this issue. Maybe GitHub uses this repository?

image

## Validation

- Added a regression fixture, scope assertions, and snapshot coverage.
- Confirmed the regression test fails before the fix.
- Syntax extension tests: 17 passed.
- Root `yarn build` and `yarn test` passed.
- Formatting and lint checks passed for the changed grammar/test files.
- Packaged and installed a local VSIX; manually verified the fix in VS Code.

### DIFF
--- a/.changeset/fix-consecutive-descriptions.md
+++ b/.changeset/fix-consecutive-descriptions.md
@@ -1,0 +1,5 @@
+---
+'vscode-graphql-syntax': patch
+---
+
+Fix highlighting of consecutive field and argument descriptions without separating commas, preventing description punctuation from causing subsequent code to be highlighted as documentation.
--- a/packages/vscode-graphql-syntax/grammars/graphql.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.json
@@ -388,7 +388,6 @@
       },
       "patterns": [
         { "include": "#graphql-comment" },
-        { "include": "#graphql-block-string-value" },
         { "include": "#graphql-description-singleline" },
         { "include": "#graphql-arguments" },
         { "include": "#literal-quasi-embedded" },
--- a/packages/vscode-graphql-syntax/grammars/graphql.json
+++ b/packages/vscode-graphql-syntax/grammars/graphql.json
@@ -131,7 +131,7 @@
     "graphql-type-definition": {
       "comment": "key (optionalArgs): Type",
       "begin": "\\s*([_A-Za-z][_0-9A-Za-z]*)(?=\\s*\\(|:)",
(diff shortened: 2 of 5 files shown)


F-OBrien:
I hit this issue also. This PR fixes the cascade completely on a large schema (descriptions on most fields).

One case remains after this PR: a `"""` block description that follows a field
carrying a **directive**.

```graphql
type T {
  "single-line description"
  a: String @index
  """
  block description
  """
  b: String
}
```

- `"single-line description"` → `comment.line.documentation.graphql` ✅
- `"""block description"""` → `string.quoted.triple.graphql` ❌
  (expected `comment.block.documentation.graphql`)

The directive is the trigger: `@index` and `@index(unique: false)` both reproduce it,
and without a directive the same block is scoped correctly. 1.3.10 rendered this
shape correctly, so it looks like part of the same #4385 window.

It's conspicuous in the default dark themes — `comment.*` resolves to `#6A9955` and
`string.quoted.triple` to `#ce9178`, so one description in a type renders a different
colour from its neighbours with nothing in the source to explain it.

**Cause:** `graphql-directive` uses `"end": "(?=.)"` with `"applyEndPatternLast": 1`,
so its sub-patterns are tried before the end pattern, and `#graphql-block-string-value`
claims the `"""` before the directive rule can close and return to the object body.
A single `"` matches none of its sub-patterns, which is why the single-line case works.

**Possible fix** — that include looks redundant:

```diff
 "graphql-directive": {
   "patterns": [
     { "include": "#graphql-comment" },
-    { "include": "#graphql-block-string-value" },
```

Block strings used as directive *argument values* still highlight correctly, since
`#graphql-arguments` → `#graphql-value` already includes it:
`a: String @foo(bar: """v""")` stays `string.quoted.triple.graphql`.

rrbe:
@F-OBrien I have reproduced the case you described, so I've made an update

trevor-scheer (approved):
Thanks for the fix(es)!

PR #4563, opened 2026-09-21
Document loading dotenv files relative to the config file

unrevised6419:
Follow-up to #2951, full write-up in #4562.

The VS Code extensions call graphql-config's `loadConfig({ rootDir })` with the workspace folder, but `rootDir` only locates `graphql.config.*`. The config module itself runs with the extension host's working directory, which is wherever VS Code was launched from (`/` when started from the Dock). So `import 'dotenv/config'` or a bare `dotenv.config()` in a config file only finds `.env` when VS Code happens to start in the project directory.

This PR updates the docs to resolve the path from the config file:

```ts
import path from 'path';
import dotenv from 'dotenv';

dotenv.config({ path: path.resolve(import.meta.dirname, '.env') });
```

- `packages/vscode-graphql/README.md`: fix the "Advanced Example", explain why the path must be explicit, show the CommonJS `__dirname` form, and point YAML/JSON configs to `graphql-config.dotEnvPath` (language server only, not `vscode-graphql-execution`).
- `packages/graphql-language-service-server/README.md`: replace the "import `dotenv()`" note with the same guidance.
- `packages/vscode-graphql/package.json`: update the `graphql-config.dotEnvPath` setting description.
- Changeset: `vscode-graphql` patch, since the setting description ships with the extension.

Notes:
- `import.meta.dirname` works in `.ts`/ESM configs on older Node because jiti (used by graphql-config) rewrites it to the config file's directory at load time.
- The examples use `'path'` rather than `'node:path'` and don't suggest `process.loadEnvFile`, because the extension supports VS Code `^1.63.0` (Node 14.16).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


unrevised6419:
Wanted to sign the CLA, but because I'm using email aliases I've got blocked (temporary_email_block), and can't get out of this state even if I clear the cookies.

image

